PR #14, opened 2019-04-09
panic/crashbackoffloop on misconfigured prometheus

### LINKED ISSUE #11: Stop Panic-ing when Prometheus misconfigured
See Issue #10 

We should instead return a sensible error in the response and write to the container logs.

AjayTripathy:
Fixes #11 

The server will panic when prometheus is misconfigured/doesn't have a valid endpoint. While it doesn't return a nice error to the frontend, it's ok to fail fast in this situation.

### DIFF
--- a/kubernetes/deployment.yaml
+++ b/kubernetes/deployment.yaml
@@ -27,7 +27,7 @@ spec:
               memory: "55M"
           env:
             - name: PROMETHEUS_SERVER_ENDPOINT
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Also added a broken, but nonempty prometheus endpoint to test. conveniently conforms to the mustache template spec to allow easy templating for later integration tests.

PR #17, opened 2019-04-10
Add a healthz endpoint to enable readinessProbe

mdaniel:


main.go
```diff
@@ -91,10 +91,16 @@ func (a *Accesses) CostDataModelRange(w http.ResponseWriter, r *http.Request, ps
 	w.Write(wrapData(data, err))
 }
 
+func (a *Accesses) Healthz(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
AjayTripathy: nit: accesses stores the data required to hit prometheus/kubernetes api, etc. It's probably not necessary to make Healthz a method of the Accesses type since this doesn't require anything on it.

PR #22, opened 2019-04-10
Discover the ClusterID from EC2 instance tags

mdaniel:


cloud/awsprovider.go
```diff
@@ -316,14 +320,75 @@ func (aws *AWS) NodePricing(key string) (*Node, error) {
...
+			tagKey := *tag.Key
+			for matchNum, group := range clusterIdRx.FindStringSubmatch(tagKey) {
+				if matchNum != 1 {
+					continue
+				}
+				return makeStructure(group)
+			}
+		}
+	}
+	return nil, fmt.Errorf("Unable to sniff out cluster ID, perhaps set $%s to force one", ClusterIdEnvVar)
```
AjayTripathy: Because this is "best effort" to find a clustername, I'd prefer to fall back to some sort of hardcoded string here (and still return the provider) rather than an error which will be rolled up to a 500 at the top. Can this be a log statement instead, and we preserve the original behavior of returning "AWS Cluster #1" and reserving the error argument for exceptional cases like the k8's api failing to return the nodelist?

PR #30, opened 2019-04-12
klog migration

AjayTripathy:
Migration from log -> klog. Setup/Error related warnings get v1. v2/3 for everything else.

### DIFF
--- a/cloud/provider.go
+++ b/cloud/provider.go
@@ -163,19 +163,19 @@ func NewProvider(clientset *kubernetes.Clientset, apiKey string) (Provider, erro
 	}
 	provider := strings.ToLower(nodes.Items[0].Spec.ProviderID)
 	if strings.HasPrefix(provider, "aws") {
-		klog.V(3).Info("Found ProviderID starting with \"aws\", using AWS Provider")
+		klog.V(2).Info("Found ProviderID starting with \"aws\", using AWS Provider")
 		return &AWS{
 			Clientset: clientset,
 		}, nil
 	} else if strings.HasPrefix(provider, "azure") {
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -100,18 +102,18 @@ func Healthz(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
 func (a *Accesses) recordPrices() {
 	go func() {
 		for {
-			log.Print("Recording prices...")
+			klog.V(3).Info("Recording prices...")
 			data, err := costModel.ComputeCostData(a.PrometheusClient, a.KubeClientSet, a.Cloud, "1h")
 			if err != nil {
-				log.Printf("Error in price recording: " + err.Error())
+				klog.V(3).Info("Error in price recording: " + err.Error())
```
dwbrown2: does V3 feel right here? seems higher priority.

dwbrown2: does V3 feel right here? seems higher priority.

cloud/provider.go
```diff
@@ -162,19 +163,19 @@ func NewProvider(clientset *kubernetes.Clientset, apiKey string) (Provider, erro
...
-		log.Print("Found ProviderID starting with \"azure\", using Azure Provider")
+		klog.V(3).Info("Found ProviderID starting with \"azure\", using Azure Provider")
 		return &Azure{
 			CustomProvider: &CustomProvider{
 				Clientset: clientset,
 			},
 		}, nil
 	} else {
-		log.Print("Unsupported provider, falling back to default")
+		klog.V(3).Info("Unsupported provider, falling back to default")
```
dwbrown2: Nit: it seems like these could be V2 to me

PR #56, opened 2019-04-24
Embed the git commit and emit it on startup

mdaniel:


main.go
```diff
@@ -160,6 +165,7 @@ func main() {
 	klog.InitFlags(nil)
 	flag.Set("v", "3")
 	flag.Parse()
+	klog.V(1).Info("Starting cost-model (git commit \"%s\")", gitCommit)
```
AjayTripathy: Guessing you meant Infof here?

PR #67, opened 2019-05-01
Ajay tripathy fix historical

### LINKED ISSUE #62: Containers that are no longer running may be missing node prices
Certain containers that were deleted or are no longer running for another reason seem to be missing pricing data.

AjayTripathy:
Fixes #62 -- mostly. If the node has disappeared, you would need to examine the pricing data from the emitted /metrics endpoint.

Also adds better GPU support.

### DIFF
--- a/cloud/awsprovider.go
+++ b/cloud/awsprovider.go
@@ -77,7 +77,7 @@ type AWSProductAttributes struct {
 	OperatingSystem string `json:"operatingSystem"`
 	PreInstalledSw  string `json:"preInstalledSw"`
 	InstanceFamily  string `json:"instanceFamily"`
-	GPU             string `json:gpu`
+	GPU             string `json:"gpu"` // GPU represents the number of GPU on the instance
 }
 
 // AWSPricingTerms are how you pay for the node: OnDemand, Reserved, or (TODO) Spot
@@ -111,7 +111,7 @@ type AWSProductTerms struct {
 	Memory   string        `json:"memory"`
 	Storage  string        `json:"storage"`
 	VCpu     string        `json:"vcpu"`
-	GPU      string        `json:"gpu"`
+	GPU      string        `json:"gpu"` // GPU represents the number of GPU on the instance
 }
 
 // ClusterIdEnvVar is the environment variable in which one can manually set the ClusterId
@@ -468,7 +468,7 @@ func (aws *AWS) createNode(terms *AWSProductTerms, usageType string, k Key) (*No
 			UsageType:    usageType,
 		}, nil
 	}
-	c, ok  := terms.OnDemand.PriceDimensions[terms.Sku+OnDemandRateCode+HourlyRateCode]
+	c, ok := terms.OnDemand.PriceDimensions[terms.Sku+OnDemandRateCode+HourlyRateCode]
 	if !ok {
 		return nil, fmt.Errorf("Could not fetch data for \"%s\"", k.ID())
 	}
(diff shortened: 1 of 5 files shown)


cloud/gcpprovider.go
```diff
@@ -252,60 +261,128 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]bool) (map[string]*G
...
-						if hourlyPrice == 0 {
-							continue
-						} else if strings.Contains(strings.ToUpper(product.Description), "RAM") {
-							if instanceType == "custom" {
-								klog.V(2).Infof("RAM custom sku is: " + product.Name)
+
+						for k, key := range inputKeys {
+							if key.GPUType() == gpuType {
+								if region == strings.Split(k, ",")[0] {
+									klog.V(3).Infof("MATCHED GPU TO NODE in region " + region)
```
mdaniel: `.Infof("...", region)`

main.go
```diff
@@ -123,6 +123,29 @@ func (p *Accesses) GetAllNodePricing(w http.ResponseWriter, r *http.Request, ps
...
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+	data, err := p.Cloud.UpdateConfig(r.Body)
+	if err != nil {
+		w.Write(wrapData(data, err))
+		return
+	}
+	w.Write(wrapData(data, err))
+	err = p.Cloud.DownloadPricingData()
+	if err != nil {
+		klog.V(1).Infof("Error redownloading data on config update")
```
mdaniel: Please don't swallow errors -- the first thing anyone who sees this message will want to know is _what_ was the error while redownloading ...

cloud/gcpprovider.go
```diff
@@ -252,60 +261,128 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]bool) (map[string]*G
...
 					partialCPU = 0.2
 				} else if strings.ToLower(instanceType) == "g1small" {
 					partialCPU = 0.5
 				}
 
+				var gpuType string
+				provIdRx := regexp.MustCompile("(Nvidia Tesla [^ ]+) ")
+				for matchnum, group := range provIdRx.FindStringSubmatch(product.Description) {
+					if matchnum == 1 {
+						gpuType = strings.ToLower(strings.Join(strings.Split(group, " "), "-"))
```
mdaniel: for your consideration, it's quite strange to use a regex, with a grouping construct, and then use some split-join magic to extract the thing one actually cares about from that string

cloud/gcpprovider.go
```diff
@@ -435,6 +520,11 @@ func (gcp *gcpKey) Features() string {
 	} else {
 		usageType = "ondemand"
 	}
+
+	if _, ok := gcp.Labels["cloud.google.com/gke-accelerator"]; ok {
```
mdaniel: Making this a constant would likely go a long way toward avoiding updated-in-one-place-but-not-the-other bugs

cloud/awsprovider.go
```diff
@@ -76,6 +76,8 @@ type AWSProductAttributes struct {
 	UsageType       string `json:"usagetype"`
 	OperatingSystem string `json:"operatingSystem"`
 	PreInstalledSw  string `json:"preInstalledSw"`
+	InstanceFamily  string `json:"instanceFamily"`
+	GPU             string `json:gpu`
```
mdaniel: It would be outstanding if this carried some godoc, since `GPU` of type `string` could be **anything**

Same thing fro its `GPU string` friend in the GCP version below

main.go
AjayTripathy: Done

cloud/gcpprovider.go
AjayTripathy: Done

cloud/gcpprovider.go
AjayTripathy: Done

cloud/awsprovider.go
AjayTripathy: Done

PR #70, opened 2019-05-02
emit GPU price-per-node metrics

AjayTripathy:


main.go
```diff
@@ -173,11 +174,15 @@ func (a *Accesses) recordPrices() {
 				cpu, _ := strconv.ParseFloat(node.VCPU, 64)
 				ramCost, _ := strconv.ParseFloat(node.RAMCost, 64)
 				ram, _ := strconv.ParseFloat(node.RAMBytes, 64)
+				gpu, _ := strconv.ParseFloat(node.GPU, 64)
+				gpuCost, _ := strconv.ParseFloat(node.GPUCost, 64)
```
dwbrown2: We're emitting 0 today when there is no GPU present. I believe standard Prometheus practice is to emit nothing in this situation.

PR #105, opened 2019-06-07
run image as non-root user to comply with PSP

RGPosadas:
We are using the cost-model deployment along with our own custom PodSecurityPolicies that require images to run as non-root.


dwbrown2:
Awesome, we actually had an open issue to do this here: https://github.com/kubecost/cost-analyzer-helm-chart/issues/64. We'll get this tested today. Thanks so much for the contribution!

PR #107, opened 2019-06-10
Add AWS PV Pricing

mda590:
* Pull in cost data from AWS cost json for EBS volume types
* Added a `CostPerIO` field in the `PV` object to handle `io1` volumes which have a per GB price + a per provisioned IO price

Tested in live AWS EKS environment and everything looked good. I am noticing certain cases where the `CostPerIO` field for `io1` volume types is not being filled out - let me know if you have any ideas on that.

### DIFF
--- a/cloud/awsprovider.go
+++ b/cloud/awsprovider.go
@@ -140,8 +140,7 @@ const HourlyRateCode = ".6YS6EN2CT7"
 var volTypes = map[string]string{
 	"EBS:VolumeUsage.gp2":     "gp2",
 	"EBS:VolumeUsage":         "standard",
-	"EBS:VolumeUsage.sc1":     "sc1", 
-	"EBS:VolumeP-IOPS.piops":  "io1",
+	"EBS:VolumeUsage.sc1":     "sc1",
 	"EBS:VolumeUsage.st1":     "st1",
 	"EBS:VolumeUsage.piops":   "io1",
 	"gp2":                     "EBS:VolumeUsage.gp2",
@@ -517,9 +516,14 @@ func (aws *AWS) DownloadPricingData() error {
 					aws.ValidPricingKeys[spotKey] = true
 				} else if strings.Contains(product.Attributes.UsageType, "EBS:Volume") {
 					key := locationToRegion[product.Attributes.Location] + "," + product.Attributes.UsageType
+					// UsageTypes may be prefixed with a region code - we're removing this when using
+					// volTypes to keep lookups generic
+					usageTypeRegx := regexp.MustCompile(".*(-|^)(EBS.+)")
+					usageTypeMatch := usageTypeRegx.FindStringSubmatch(product.Attributes.UsageType)
+					usageTypeNoRegion := usageTypeMatch[len(usageTypeMatch)-1]
 					spotKey := key + ",preemptible"
 					pv := &PV{
-						Class:      volTypes[product.Attributes.UsageType],
+						Class:      volTypes[usageTypeNoRegion],
 						Region:     locationToRegion[product.Attributes.Location],
 					}
 					productTerms := &AWSProductTerms{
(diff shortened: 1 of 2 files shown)


cloud/awsprovider.go
```diff
@@ -437,6 +478,20 @@ func (aws *AWS) DownloadPricingData() error {
 					}
 					aws.ValidPricingKeys[key] = true
 					aws.ValidPricingKeys[spotKey] = true
+				} else if strings.HasPrefix(product.Attributes.UsageType, "EBS:Volume") {
```
AjayTripathy: You probably want contains here, as only us-east-1 is prefixed in this fashion.
For example, here's a us-west2 usagetype:
 "usagetype" : "USW2-EBS:VolumeUsage.gp2",

This may be related to your observation that not all costPerIOOperation is being set, though I am still investigating that.

mda590: Addressed this by dynamically adding the correct region prefix to UsageTypes in regions which are not us-east-1. Included a map which maps region identifiers (us-east-2) into these UsageType region codes (USE2).

cloud/awsprovider.go
```diff
@@ -489,7 +515,7 @@ func (aws *AWS) DownloadPricingData() error {
 					}
 					aws.ValidPricingKeys[key] = true
 					aws.ValidPricingKeys[spotKey] = true
-				} else if strings.HasPrefix(product.Attributes.UsageType, "EBS:Volume") {
+				} else if strings.Contains(product.Attributes.UsageType, "EBS:Volume") {
 					key := locationToRegion[product.Attributes.Location] + "," + product.Attributes.UsageType
```
AjayTripathy: Usagetype would also need to be stripped here because the key mapping above doesn't have region, otherwise you would end up with a key like: us-east-2,USE2-EBS:VolumeUsage.sc1 and a lookup key is something like us-east-2,EBS:VolumeUsage.sc1

cloud/awsprovider.go
```diff
@@ -546,15 +572,23 @@ func (aws *AWS) DownloadPricingData() error {
...
 								cost := offerTerm.PriceDimensions[sku.(string)+OnDemandRateCode+HourlyRateCode].PricePerUnit.USD
-								cost_F, _ := strconv.ParseFloat(cost, 64)
-								piops_key := curr_region + ",EBS:VolumeUsage.piops"
+								// UsageType in most regions starts with a region identifier
+								// Here we get that region identifier and then use it as part of
+								// the key for looking up and storing the per IO cost with the io1 price
+								billing_region_code := regionToBillingRegionCode[curr_region]
+								piops_key := ""
+								if billing_region_code != "" {
+									piops_key = curr_region + "," + billing_region_code + "-EBS:VolumeUsage.piops"
```
AjayTripathy: nit: Go convention is to use camelCase vs underscores-- linters complain about this in general.

AjayTripathy: (ditto for other underscores)

mda590: fixed these - I noticed I also did this in my previous Glue PR, so I'll submit a change for that as well.

cloud/awsprovider.go
mda590: Pushed a commit for this - let me know if you are good with the change I made to handle this or if you were thinking of something else.

AjayTripathy: I'd suggest setting the key on L519 here to the stripped value, as the lookup further down on L354 looks up based on pvk.Features(), which returns a key of the format "region,usage type without region". This seems like a cleaner approach, as then there's no need to reconstruct a regional P-IOPS key when you reconstruct the keys on L581. 

Also, as is, L596 on to get non P-IOPS data isn't correct, as the key hasn't been made regional.

AjayTripathy: @mda590 as discussed I'll go ahead and merge this and follow it with a fix to support non us-east1 regions. Thank you for this!

PR #113, opened 2019-06-20
Support Azure managed identity

omerlh:
Azure support [managed identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview), which let you easily consume Azure resources without credentials - based on machine credentials (or pod). I took the relevant code from [AAD pod identity](https://github.com/Azure/aad-pod-identity#list-vms-with-seamless-authorization), but I need to test it.  I opened the PR to get feedback about adding this before testing.

### DIFF
--- a/cloud/azureprovider.go
+++ b/cloud/azureprovider.go
@@ -278,6 +278,14 @@ func (az *Azure) DownloadPricingData() error {
 		}
 		authorizer = a
 	}
+	else {
+		a, err := auth.NewAuthorizerFromEnvironment()
+		if err != nil {
+		    return err
+		}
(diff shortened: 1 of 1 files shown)


AjayTripathy:
This looks like a good approach to me. Thanks for this contribution! Let me know when it's tested and I will merge.

omerlh:
Sorry, it's my mistake for not reading all the code :( the support for `NewAuthorizerFromEnvironment` is already in place, no need for this PR. When we'll test this project, we will try using it with MSI and update if we'll encounter issues.

PR #139, opened 2019-08-20
only charge running pods

### LINKED ISSUE #136: Should we be billing on completed jobs?
Right now we still charge an allocation to completed jobs if they have resource requests. This was originally intended to encourage teams to clean up old jobs. Should we be doing this? 

Some questions to consider:

Do they block downscaling on some cluster autoscalers?
Does the k8s scheduler have to schedule around them?

AjayTripathy:
Fixes #136

### DIFF
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -352,6 +352,9 @@ func ComputeCostData(cli prometheusClient.Client, clientset kubernetes.Interface
 	}
 	currentContainers := make(map[string]v1.Pod)
 	for _, pod := range podlist.Items {
+		if pod.Status.Phase != "Running" {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


costmodel/costmodel.go
```diff
@@ -352,6 +352,9 @@ func ComputeCostData(cli prometheusClient.Client, clientset kubernetes.Interface
 	}
 	currentContainers := make(map[string]v1.Pod)
 	for _, pod := range podlist.Items {
+		if pod.Status.Phase != "Running" {
```
dwbrown2: Is there a chance that status is undefined here, i.e. unsafe access?

dwbrown2: Is there potentially a version of k8s that may have a phase that doesn't equal EXACTLY "Running"?

PR #142, opened 2019-09-04
#141 unregistr gauge 'container_uptime_seconds'

PetzJohannes:
Hi @dwbrown2 ,
today I had an hour to fix the last cached gauge.
Here is my pull request.

It completes this one:
https://github.com/kubecost/cost-model/pull/141/files#diff-7ddfb3e035b42cd70649cc33393fe32c

And we discussed the issue here:
https://github.com/kubecost/cost-model/issues/140

Cheers
Johannes


AjayTripathy:
Looks good to me! Thanks for this.

PR #160, opened 2019-09-25
Fix for "thrift used for two different module paths" issue.

mbolt35:
This seems to have resolved the build issue. Fingers crossed it corrects CI as well.


AjayTripathy:
Thanks, LGTM.

PR #161, opened 2019-09-26
Add Network Costs to Model

mbolt35:


costmodel/networkcosts.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+		var vectors []*Vector
+		for _, value := range values {
+			dataPoint, ok := value.([]interface{})
+			if !ok || len(dataPoint) != 2 {
+				return nil, fmt.Errorf("Improperly formatted datapoint from Prometheus")
+			}
+
+			strVal := dataPoint[1].(string)
+			v, _ := strconv.ParseFloat(strVal, 64)
```
AjayTripathy: It would be very weird if this was not parseable as a float, but let's not swallow errors all the same.

costmodel/costmodel.go
```diff
@@ -1236,6 +1295,12 @@ func (cm *CostModel) ComputeCostDataRange(cli prometheusClient.Client, clientset
 		}
 	}
 
+	networkUsageMap, err := GetNetworkUsageData(resultNetZoneRequests, resultNetRegionRequests, resultNetInternetRequests, true)
+	if err != nil {
+		klog.V(1).Infof("Unable to get Network Cost Data: %s", err.Error())
+		return nil, err
```
AjayTripathy: Remember that returning the error here blocks costmodel execution. We probably don't want an error in the networking data fetch to halt the execution of the program, so you may just want to log here and set networkUsageMap := make(map...)

mbolt35: ah, good catch. Thanks

PR #165, opened 2019-09-27
Kubernetes Caching Watchers Implementation

mbolt35:
Use caching watchers to maintain up to date cached cluster resources


costmodel/watchcontroller.go
```diff
@@ -2,34 +2,104 @@ package costmodel
...
+
+	// GetAll returns all of the resources
+	GetAll() []interface{}
+
+	// SetUpdateHandler sets a specific handler for adding/updating individual resources
+	SetUpdateHandler(WatchHandler) WatchController
+
+	// SetRemovedHandler sets a specific handler for removing individual resources
+	SetRemovedHandler(WatchHandler) WatchController
+}
```
mbolt35: Added the above APIs in order to specifically watch a resource, but these aren't exposed via the `CacheCluster` interface yet. We'll want to hook these in when we start listening to pod updates.

costmodel/clustercache.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (kcc *KubernetesClusterCache) Run(stopCh chan struct{}) {
+	go kcc.namespaceWatch.Run(1, stopCh)
+	go kcc.nodeWatch.Run(1, stopCh)
+	go kcc.podWatch.Run(1, stopCh)
+	go kcc.serviceWatch.Run(1, stopCh)
+	go kcc.deploymentsWatch.Run(1, stopCh)
+	go kcc.pvWatch.Run(1, stopCh)
+	go kcc.storageClassWatch.Run(1, stopCh)
+}
```
mbolt35: I'm pretty sure it's ok to use the same `stopCh` here as long as we actually stop using `close(stopCh)`

costmodel/costmodel.go
```diff
@@ -44,52 +42,20 @@ const (
...
-	cm.ContainerWatcher(podListWatcher)
-	return cm
+	stop chan struct{}
 }
 
-func (cm *CostModel) ContainerWatcher(podListWatcher cache.ListerWatcher) {
-
-	// create the workqueue
-	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+func NewCostModel(client kubernetes.Interface) *CostModel {
```
AjayTripathy: The underlying controllers should provide a HasSynced() function for an "initial sync" of your underlying caches. I'd prefer if this waited to return until all the controllers have synced. No need to do anything fancy, I think a while loop in KubernetesClusterCache should do the trick.

costmodel/watchcontroller.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+// Type alias for a receiver func
+type WatchHandler = func(interface{})
+
+// WatchController defines a contract for an object which watches a specific resource set for
+// add, updates, and removals
+type WatchController interface {
+	// Initializes the cache
+	WarmUp(chan struct{})
```
mbolt35: Split the Run() into WarmUp() and Run() where WarmUp waits for the HasSync on the informer.

costmodel/clustercache.go
```diff
@@ -0,0 +1,165 @@
...
+
+	go initializeCache(kcc.namespaceWatch, &wg, cancel)
+	go initializeCache(kcc.nodeWatch, &wg, cancel)
+	go initializeCache(kcc.podWatch, &wg, cancel)
+	go initializeCache(kcc.serviceWatch, &wg, cancel)
+	go initializeCache(kcc.deploymentsWatch, &wg, cancel)
+	go initializeCache(kcc.pvWatch, &wg, cancel)
+	go initializeCache(kcc.storageClassWatch, &wg, cancel)
+
+	wg.Wait()
```
mbolt35: Executes blocking WarmUp() in a go func for each of the watchers. WarmUp waits for HasSync on the informer, so this should return the cluster cache populated.

PR #167, opened 2019-10-01
add pv cost

AjayTripathy:


main.go
```diff
@@ -684,6 +687,10 @@ func main() {
 		Name: "container_gpu_allocation",
 		Help: "container_gpu_allocation GPU used",
 	}, []string{"namespace", "pod", "container", "instance", "node"})
+	PVAllocation := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "pod_pvc_allocation",
```
dwbrown2: Should we be prefixing all our new metrics with `_kubecost`? That's what we're doing for recording rules now.

AjayTripathy: I did this for consistency with the others-- we should rename them all at once if we're going to do that.

dwbrown2: Let's leave as is then for now.

Issue #173, opened 2019-10-07
Cost model caching


dwbrown2:
Let's start caching data so that backend response times are faster. My proposal is to start this with the aggregation api with the following goals:

* [P0] let's try to be really careful about not causing regressions. My suggestion is to keep v1 of this implementation simple and err heavily on the side of missing a cache hit vs returning wrong data
* Make cache short-lived, i.e. 2-5 minutes, but this is ideally configurable later
* Can keep a single cache when parameters match (e.g. byNamespace, timeWindow, offset) or cache for multiple aggregations (e.g. namespace, label, etc.)
* I think just starting with namespace caching could be easiest

AjayTripathy:
We can start with a simple cache that just uses the url parameters on the [aggregateCostModel](https://github.com/kubecost/cost-model/blob/master/main.go#L205) endpoint as keys 
and invalidates based on a TTL. [go-cache](https://github.com/patrickmn/go-cache) seems to support something like this out of the box

I'd also suggest a parameter to disable the cache and a parameter to clear the cache for debugging.

Once that's working, as a later step we can think about what timeWindows we might want to precompute/how to warm up the cache.

dwbrown2:
Marking this as closed, lots of great work here :)

PR #176, opened 2019-10-08
Issue #173 implement caching for aggregatedCostModel endpoint

nikovacevic:
Done
- Imported in [go-cache](https://github.com/patrickmn/go-cache) for TTL caching
- In `/aggregatedCostModel` handler, implemented parameter-keyed cache with default expiry of 2 minutes. Pass `clearCache=true` to clear the cache and `disableCache=true` to force a one-time cache miss.
- Added helper method for identifying cache hits and misses in HTTP responses
- Manual testing

Questions
- Should we attempt a second layer of caching of the `ComputeCostDataRange` results?
- This solution is tightly-coupled to the `AggregateCostModel` handler. Should we try to factor this out into a more general-purpose component, or save that for when we move caching out of the `cost-model`?
- Should we start by warming the cache with a call to `aggregate=namespace` and `window=1d` or save that for an improvement?

To Do
- Unit and/or integration tests
- Remove helper methods and TODO comments

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,5 +40,3 @@ require (
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 	k8s.io/klog v0.4.0
 )
-
-go 1.13
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/julienschmidt/httprouter v1.2.0
 	github.com/lib/pq v1.2.0
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
+	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pkg/errors v0.8.1 // indirect
 	github.com/prometheus/client_golang v1.0.0
 	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
@@ -39,3 +40,5 @@ require (
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 	k8s.io/klog v0.4.0
 )
+
+go 1.13
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ import (
 	"github.com/julienschmidt/httprouter"
 	costAnalyzerCloud "github.com/kubecost/cost-model/cloud"
 	costModel "github.com/kubecost/cost-model/costmodel"
+	"github.com/patrickmn/go-cache"
 	prometheusClient "github.com/prometheus/client_golang/api"
 	prometheusAPI "github.com/prometheus/client_golang/api/prometheus/v1"
 	v1 "k8s.io/api/core/v1"
@@ -61,6 +62,7 @@ type Accesses struct {
 	ServiceSelectorRecorder       *prometheus.GaugeVec
 	DeploymentSelectorRecorder    *prometheus.GaugeVec
 	Model                         *costModel.CostModel
+	Cache                         *cache.Cache
 }
 
 type DataEnvelope struct {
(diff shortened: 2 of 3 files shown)


nikovacevic:
Tomorrow I'm planning to write integration tests for at least the following:
- two requests in rapid succession produce identical responses
- one request, followed by a `disableCache=true` request produce values within 0.1% of each other, but not necessarily identical
- one request, followed by a `clearCache=true` request produce values within 0.1% of each other, but not necessarily identical

It would be nice to make the expiration configurable so that we don't have to sleep for 2 minutes in the middle of the test to assert natural expiry, but something testing that is also on my radar.

main.go
```diff
@@ -202,12 +230,25 @@ func (a *Accesses) ClusterCostsOverTime(w http.ResponseWriter, r *http.Request,
...
-	aggregation := r.URL.Query().Get("aggregation")
 	namespace := r.URL.Query().Get("namespace")
+	aggregationField := r.URL.Query().Get("aggregation")
 	aggregationSubField := r.URL.Query().Get("aggregationSubfield")
 
+	disableCache := r.URL.Query().Get("disableCache") == "true"
+	clearCache := r.URL.Query().Get("clearCache") == "true"
+
+	// TODO nikovacevic-caching this should be required, right? Can we return 400 Bad Request if not set?
+	if aggregationField == "" {
```
dwbrown2: Yep, seems fine to me. This is a required field. Nit: I may provide a more specific error message, maybe "Missing aggregationField parameter"?

main.go
```diff
@@ -219,6 +260,8 @@ func (a *Accesses) AggregateCostModel(w http.ResponseWriter, r *http.Request, ps
 		endTime = endTime.Add(-1 * o)
 	}
 
+	// if window is defined in terms of days, convert it to hours
```
dwbrown2: thanks for these comments! :)

main.go
```diff
@@ -832,6 +910,13 @@ func main() {
 	router.GET("/containerUptimes", a.ContainerUptimes)
 	router.GET("/aggregatedCostModel", a.AggregateCostModel)
 
+	router.GET("/image", func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Header().Set("Access-Control-Allow-Origin", "*")
+		w.Write([]byte("nikovacevic/kubecost-cost-model:caching-4"))
```
dwbrown2: do we want to check this reference in?

go.mod
```diff
@@ -39,3 +40,5 @@ require (
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 	k8s.io/klog v0.4.0
 )
+
+go 1.13
```
AjayTripathy: we're still supporting 1.12 for now, what does this line do?

main.go
AjayTripathy: SGTM

main.go
```diff
@@ -239,23 +284,52 @@ func (a *Accesses) AggregateCostModel(w http.ResponseWriter, r *http.Request, ps
...
+
+	// TODO nikovacevic-caching cache raw cost data and unmarshal it
+	// rawKey := fmt.Sprintf("raw:%s:%s:%s:%s", start, end, "1h", namespace)
+	// data, found := a.Cache.Get(rawKey)
+	// if !found || disableCache ...
+
 	data, err := a.Model.ComputeCostDataRange(a.PrometheusClient, a.KubeClientSet, a.Cloud, start, end, "1h", namespace)
 	if err != nil {
-		w.Write(wrapData(nil, err))
+		w.Write(wrapDataWithMessage(nil, err, fmt.Sprintf("cache miss")))
```
AjayTripathy: could we see aggKey in the returned message?

AjayTripathy:
Re: questions

* Should we attempt a second layer of caching of the ComputeCostDataRange results?
- Let's hold on this to see if this gets us most of the way to speeding things up.
*This solution is tightly-coupled to the AggregateCostModel handler. Should we try to factor this out into a more general-purpose component, or save that for when we move caching out of the cost-model?
- I think it's okay to tightly couple for now.
* Should we start by warming the cache with a call to aggregate=namespace and window=1d or save that for an improvement?
- Let's hold on this also for now until we think it's necessary to get the performance we need.

main.go
nikovacevic: Nope, that whole endpoint is just for development purposes. Removing now.

go.mod
nikovacevic: That's the line my VS Code Go extension keeps adding. Until I find and disable the setting that controls it, I'll be careful to delete manually.

main.go
nikovacevic: Agreed, changing now.

main.go
nikovacevic: Sure thing!

AjayTripathy:
@nikovacevic as discussed, had to get out #177 for a user issue before this, this LGTM after conflicts fixed.

PR #182, opened 2019-10-11
Aggregation caching improvements

nikovacevic:
- enable custom pricing overrides to cost model data
- filtering by cluster name using `cluster=...`
- enable time series data with `timeSeries=true` URL param
- update clusterID to reflect frontend name


cloud/provider.go
```diff
@@ -57,6 +55,11 @@ type Node struct {
 	GPUCost          string `json:"gpuCost"`
 }
 
+// IsSpot determines whether or not a Node uses spot by usage type
+func (n *Node) IsSpot() bool {
+	return strings.Contains(n.UsageType, "spot") || strings.Contains(n.UsageType, "emptible")
```
nikovacevic: @dwbrown2 I pulled this logic from the FE. Does that seem right here?

costmodel/aggregations.go
```diff
@@ -166,36 +186,68 @@ func mergeVectors(costDatum *CostData, aggregation *Aggregation, discount float6
...
-func getPriceVectors(costDatum *CostData, discount float64, idleCoefficient float64) ([]*Vector, []*Vector, []*Vector, [][]*Vector) {
+func getPriceVectors(cp cloud.Provider, costDatum *CostData, discount float64, idleCoefficient float64) ([]*Vector, []*Vector, []*Vector, [][]*Vector) {
+	cpuCostStr := costDatum.NodeData.VCPUCost
+	ramCostStr := costDatum.NodeData.RAMCost
+	gpuCostStr := costDatum.NodeData.GPUCost
+	pvCostStr := costDatum.NodeData.StorageCost
+
+	// If custom pricing is enabled and can be retrieved, replace
+	// default cost values with custom values
+	customPricing, err := cp.GetConfig()
```
AjayTripathy: Don't swallow the error here, since it's entirely possible we lost touch with the filesystem somehow. I'd say log it.

nikovacevic: Ok logging sounds good. Think it's ok to move forward like this, just skipping the custom pricing?

AjayTripathy: I'd say yeah, I think it's okay to just use the cloud pricing if you can't fetch custom pricing. We can revisit though, but this should be pretty rare.

cloud/provider.go
dwbrown2: @AjayTripathy isn't there one more test that can be done on the backend? Check spot_label in modelConfigs?

PR #185, opened 2019-10-15
Thanos Remote Endpoint

mbolt35:


costmodel/router.go
```diff
@@ -411,7 +420,16 @@ func (a *Accesses) CostDataModelRange(w http.ResponseWriter, r *http.Request, ps
...
 	}
-	data, err := a.Model.ComputeCostDataRange(a.PrometheusClient, a.KubeClientSet, a.Cloud, start, end, window, namespace, cluster, remoteEnabled)
+
+	// Use Thanos Client if it exists and remote != false
+	var pClient prometheusClient.Client
+	if remote != "false" && a.ThanosClient != nil {
+		pClient = a.ThanosClient
+	} else {
+		pClient = a.PrometheusClient
+	}
```
mbolt35: This happens after the check for `remoteEnabled` for Postgres. Right now, it seems like Postgres will take priority if it is enabled at the same time as thanos.

PR #186, opened 2019-10-15
Add rate data to aggregated cost model

nikovacevic:
- add rate date to aggregateCostModel functionality
- allow aggregateCostModel endpoint to optionally be parametrized by start and end rather than window and offset
- allow querying by multiple subfields with OR composition

### DIFF
--- a/costmodel/aggregations.go
+++ b/costmodel/aggregations.go
@@ -12,24 +12,25 @@ import (
 )
 
 type Aggregation struct {
-	Aggregator    string    `json:"aggregation"`
-	Subfields     []string  `json:"subfields"`
-	Environment   string    `json:"environment"`
-	Cluster       string    `json:"cluster"`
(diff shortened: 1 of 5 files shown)


nikovacevic:
Aggregation API progress update:
- Cumulative costs are matching precisely across-the-board when identical windows are used
  - Note that this changes the definition of window=1d from 24 data points (e.g. 10/02 0:00:00 to 10/01 01:00:00, inclusive) to 25 data points (e.g. 10/02 0:00:00 to 10/01 0:00:00, inclusive) so the "window=1d" numbers are expected to rise by a factor of 1.0417
- Rate data is complete and looks correct on `niko` cluster
- Network costs are integrated and look correct on `niko` cluster

Moving on to writing tests now, but if this gets the sign-off and we want to cut a release before that is done, then I'm comfortable with that.

PR #193, opened 2019-10-17
Fix ClusterCosts regression

nikovacevic:
Remove offset templating in the handler, which is now being done in the ClusterCosts function itself


AjayTripathy:
LGTM, merging

PR #196, opened 2019-10-17
run integration tests in cluster

AjayTripathy:
Step towards automation-- the cost-model integration tests should run in cluster or locally.

Bonus: fix multicluster setup for idle allocation

### DIFF
--- a/costmodel/cluster.go
+++ b/costmodel/cluster.go
@@ -2,6 +2,7 @@ package costmodel
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	costAnalyzerCloud "github.com/kubecost/cost-model/cloud"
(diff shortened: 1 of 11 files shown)


nikovacevic:
Multicluster idle allocation fixes look good to me. When this goes in, I'll go back and make sure that [PR 197](https://github.com/kubecost/cost-model/pull/197) works with these changes.

costmodel/cluster.go
```diff
@@ -99,22 +102,106 @@ func resultToTotal(qr interface{}) ([][]string, error) {
...
-	if !ok || len(dataPoint) != 2 {
-		return nil, fmt.Errorf("Improperly formatted datapoint from Prometheus")
+	return toReturn, nil
+}
+
+// ClusterCostsForAllClusters gives the cluster costs averaged over a window of time for all clusters.
+func ClusterCostsForAllClusters(cli prometheusClient.Client, cloud costAnalyzerCloud.Provider, windowString, offset string) (map[string]*Totals, error) {
+
+	if offset != "" {
+		offset = fmt.Sprintf("offset 3h") // Set offset to 3h for block sync
```
nikovacevic: Whether or not this is the right thing to do, this looks like it could cause some confusion, e.g. calling `ClusterCostsForAllClusters(..., offset = "1h", ...)` will result in `offset = "offset 3h"`. Is that the desired behavior?

PR #207, opened 2019-10-23
Caching out-of-cluster cost requests

nikovacevic:
Branched off of [PR 204](https://github.com/kubecost/cost-model/pull/204), so that is a prerequisite to this for all intents and purposes.

### DIFF
--- a/costmodel/cluster.go
+++ b/costmodel/cluster.go
@@ -142,7 +142,7 @@ func resultToTotal(qr interface{}) (map[string][][]string, error) {
 }
 
 // ClusterCostsForAllClusters gives the cluster costs averaged over a window of time for all clusters.
-func ClusterCostsForAllClusters(cli prometheusClient.Client, cloud costAnalyzerCloud.Provider, windowString, offset string) (map[string]*Totals, error) {
+func ClusterCostsForAllClusters(cli prometheusClient.Client, windowString, offset string) (map[string]*Totals, error) {
 
 	if offset != "" {
 		offset = fmt.Sprintf("offset %s", offset)
--- a/costmodel/cluster.go
+++ b/costmodel/cluster.go
@@ -148,9 +148,17 @@ func ClusterCostsForAllClusters(cli prometheusClient.Client, cloud costAnalyzerC
 		offset = fmt.Sprintf("offset %s", offset)
 	}
 
+	localStorageQuery, err := cloud.GetLocalStorageQuery(offset)
+	if err != nil {
+		return nil, err
+	}
+	if localStorageQuery != "" {
+		localStorageQuery = fmt.Sprintf("+ %s", localStorageQuery)
+	}
+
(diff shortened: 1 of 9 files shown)


dwbrown2:
Just wanted to clarify... are you removing the ability to pass start and end date to the aggregatedCostModel? That's fine if so, just wanted to make sure I understand.

nikovacevic:
> Just wanted to clarify... are you removing the ability to pass start and end date to the aggregatedCostModel? That's fine if so, just wanted to make sure I understand.

@dwbrown2 Correct. At least until we come up with a sophisticated cache that can pull chunks out of a cached timeline, this makes the logic of when to cache and how more simpler and thus better IMHO.

Any objections? (e.g. does this screw up the frontend's custom date range querying?) It wouldn't be hard to put it back and set a rule to avoid the cache if the time range is defined by start/end rather than window/offset.

dwbrown2:
Nope, this shouldn't be a problem. Just wanted to know so I can communicate
correctly to new users.

Webb
Kubecost, Co-founder

ᐧ

costmodel/cluster.go
```diff
@@ -142,7 +142,7 @@ func resultToTotal(qr interface{}) (map[string][][]string, error) {
 }
 
 // ClusterCostsForAllClusters gives the cluster costs averaged over a window of time for all clusters.
-func ClusterCostsForAllClusters(cli prometheusClient.Client, cloud costAnalyzerCloud.Provider, windowString, offset string) (map[string]*Totals, error) {
+func ClusterCostsForAllClusters(cli prometheusClient.Client, windowString, offset string) (map[string]*Totals, error) {
```
AjayTripathy: We actually do need this, see: https://github.com/kubecost/cost-model/pull/210/files#diff-58d9ebd2d60cdef59d19220180531c6fR151

costmodel/router.go
```diff
@@ -260,169 +326,100 @@ func (a *Accesses) AggregateCostModel(w http.ResponseWriter, r *http.Request, ps
...
 
+	// enable remote if it is available and not disabled
+	remoteAvailable := os.Getenv(remoteEnabled) == "true"
+	remoteEnabled := remote && remoteAvailable
+	klog.Infof("REMOTE ENABLED: %t", remoteEnabled)
+
+	// default to given Prometheus client, but use Thanos Client if it
+	// exists and remote is not disabled
+	pClient := a.PrometheusClient
+	if remoteEnabled && a.ThanosClient != nil {
```
AjayTripathy: just use a.ThanosClient != nil here. The remoteEnabled flag is for postgres. ThanosClient will only be instantiated if Thanos is available.

nikovacevic: That would be a change from master (see https://github.com/kubecost/cost-model/blob/master/costmodel/router.go#L414). Is that still the intended logic?

costmodel/cluster.go
nikovacevic: Ok I'll put that back

PR #208, opened 2019-10-23
Ajay tripathy fix thanos

AjayTripathy:


costmodel/costmodel.go
```diff
@@ -165,7 +165,7 @@ type PrometheusMetadata struct {
 
 // ValidatePrometheus tells the model what data prometheus has on it.
 func ValidatePrometheus(cli prometheusClient.Client) (*PrometheusMetadata, error) {
-	data, err := Query(cli, "up")
+	data, err := Query(cli, "up offset 3h")
```
mbolt35: We use this same query to validate the non-thanos Prometheus client as well right? Still ok to offset here?

AjayTripathy: No.. good catch.

mbolt35:
LGTM 👍

PR #212, opened 2019-10-26
Deployments and Services Considered Multicluster and for Deleted Pods

mbolt35:


costmodel/promparsers.go
```diff
@@ -0,0 +1,350 @@
...
+
+		result[label] = value
+	}
+
+	return result
+}
+
+// NewQueryResults accepts the raw prometheus query result and returns an array of
+// PromQueryResult objects
+func NewQueryResults(queryResult interface{}) ([]*PromQueryResult, error) {
```
mbolt35: @AjayTripathy This is my proposal for query results. It handles both `value` and `values` based on metrics, does careful checking for each property and propagates errors correctly. Helper function for retrieving labels from the metrics as well as getting specific string values. The query parsing in this file uses this, and I switched Network parsing over to this as well. It seems to reduce the noise of the parsing methods a good bit.

PR #215, opened 2019-10-28
Cost Model caching, Out Of Cluster caching, Allocation in Aggregation API

nikovacevic:


costmodel/router.go
```diff
@@ -646,6 +682,56 @@ func (a *Accesses) OutofClusterCosts(w http.ResponseWriter, r *http.Request, ps
...
+	// clear cache prior to checking the cache so that a clearCache=true
+	// request always returns a freshly computed value
+	if clearCache {
+		a.OutOfClusterCache.Flush()
+	}
+
+	// attempt to retrieve cost data from cache
+	key := fmt.Sprintf(`%s:%s:%s`, start, end, aggregation)
+	if value, found := a.OutOfClusterCache.Get(key); found && !disableCache {
+		klog.V(1).Infof("out of cluser cache hit: %s", key)
```
AjayTripathy: This will probably be a bit noisy on server. Watch for log levels of v(1) to V3), those are reserved for things that would be useful for end user debugging. This should probably be shipped as V(4)

costmodel/router.go
```diff
@@ -646,6 +682,56 @@ func (a *Accesses) OutofClusterCosts(w http.ResponseWriter, r *http.Request, ps
...
+		if data, ok := value.([]*costAnalyzerCloud.OutOfClusterAllocation); ok {
+			w.Write(wrapDataWithMessage(data, nil, fmt.Sprintf("out of cluser cache hit: %s", key)))
+			return
+		}
+		klog.Errorf("caching error: failed to type cast data: %s", key)
+	}
+	klog.V(1).Infof("out of cluster cache miss: %s", key)
```
AjayTripathy: Same as above

costmodel/router.go
nikovacevic: Good call. Meant to just remove those, anyways, as we're providing cache hit info in the response.

PR #229, opened 2019-11-04
fix idle resolutions

AjayTripathy:


costmodel/router.go
```diff
@@ -440,16 +441,11 @@ func (a *Accesses) AggregateCostModel(w http.ResponseWriter, r *http.Request, ps
 		return
 	}
 
-	// if custom pricing has changed, then clear the cache and recompute data
-	if a.CustomPricingHasChanged() {
```
dwbrown2: did you mean to remove this?

costmodel/router.go
```diff
@@ -533,35 +532,40 @@ func (a *Accesses) AggregateCostModel(w http.ResponseWriter, r *http.Request, ps
...
 	}
 
-	c, err := a.Cloud.GetConfig()
+	c, err := A.Cloud.GetConfig()
 	if err != nil {
 		w.Write(WrapData(nil, err))
 		return
 	}
-	discount, err := parsePercentString(c.Discount)
+	discount, err := strconv.ParseFloat(c.Discount[:len(c.Discount)-1], 64)
```
dwbrown2: niko just added this. why are we removing?

costmodel/router.go
```diff
@@ -441,6 +441,10 @@ func (a *Accesses) AggregateCostModel(w http.ResponseWriter, r *http.Request, ps
 		return
 	}
 
+	if a.CustomPricingHasChanged() {
```
dwbrown2: why still lower case?

costmodel/router.go
AjayTripathy: Fixed.

costmodel/router.go
AjayTripathy: Fixed, uses ParsePercentString now.

costmodel/router.go
AjayTripathy: Fixed.

PR #230, opened 2019-11-05
Performance Improvements

mbolt35:
* Remove `findDeletedPodInfo` from ComputeCostDataRange as we are already cleanly processing pod labels, service labels, and deployment labels for "missing" pods. 
* For ComputeCostDataRange, add a singleflight implementation which allows any requests made with similar inputs to rely on a single call rather than multiple concurrent calls.

### DIFF
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -1335,83 +1335,127 @@ func (cm *CostModel) ComputeCostDataRange(cli prometheusClient.Client, clientset
 	var promErr error
 	var resultRAMRequests interface{}
 	go func() {
-		resultRAMRequests, promErr = QueryRange(cli, queryRAMRequests, start, end, window)
+		defer measureTime(time.Now(), queryRAMRequests)
 		defer wg.Done()
+
+		resultRAMRequests, promErr = QueryRange(cli, queryRAMRequests, start, end, window)
 	}()
 	var resultRAMUsage interface{}
 	go func() {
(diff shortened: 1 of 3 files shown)


costmodel/costmodel.go
```diff
@@ -1289,7 +1297,47 @@ func costDataPassesFilters(costs *CostData, namespace string, cluster string) bo
...
+
+	klog.V(3).Infof("ComputeCostDataRange with Key: %s", key)
+
+	// If there is already a request out that uses the same data, wait for it to return to share the results.
+	// Otherwise, start executing.
+	result, err, _ := cm.RequestGroup.Do(key, func() (interface{}, error) {
+		return cm.costDataRange(cli, clientset, cp, startString, endString, windowString, filterNamespace, filterCluster, remoteEnabled)
+	})
+
+	data := result.(map[string]*CostData)
```
nikovacevic: Is there no chance of this type cast failing? (I was curious when doing similar operations in the caching work and checked `ok` just in case, but seems like it would never fail.)

AjayTripathy: Even if the chance is none, let's get in the habit of Never Panicking by catching these and logging the error.

mbolt35: It's pretty weird that go as a language has overloads based on return types. Even if you just do 
```go
data, _ := result.(map[string]*CostData)
```

That will avoid the panic. However, if you leave off the `, _` it panics. I'll fix

PR #231, opened 2019-11-07
Refresh Provider data every week

mbolt35:
This is somewhat over architected for what I ended up with, but due to some inconsistencies with lack of etags on GCP responses, I've punted for now and just added this directly to the run loop. 

* Does it make sense to add a private reference to Accesses so we can call `Reset()` if the data is force reloaded?
* Should we expose a configuration value?

### DIFF
--- a/costmodel/router.go
+++ b/costmodel/router.go
@@ -920,7 +920,23 @@ func (a *Accesses) recordPrices() {
 			return strings.Split(key, ",")
 		}
 
+		// Create job to refresh provider prices every week
+		// TODO: Make configurable?
+		refreshProviderCosts := NewTimedJob(7*24*time.Hour, func() error {
+			return a.Cloud.DownloadPricingData()
+		})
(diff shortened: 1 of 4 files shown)


costmodel/router.go
```diff
@@ -920,7 +920,23 @@ func (a *Accesses) recordPrices() {
 			return strings.Split(key, ",")
 		}
 
+		// Create job to refresh provider prices every week
+		// TODO: Make configurable?
+		refreshProviderCosts := NewTimedJob(7*24*time.Hour, func() error {
```
dwbrown2: So if this fails, retry currently happens in a week? Seems fine for an initial implementation but just wanted to ask

mbolt35: Yeah - it just doesn't feel like something we want to retry. I'm not sure what type of "failure" can happen outside of failing to download the sku list.

PR #242, opened 2019-11-25
ClusterCache Refactor to Provider + Reserved Instances GCP

mbolt35:
* Moved ClusterCache utility into a separate package to prevent cross dependency between costmodel and provider packages.
* ClusterCache is now created and initialized after Kubernetes client initialization and passed into costmodel as well as provider
* Pulling in GCP committed use, applying to costs _after_ node VCPU and RAM is set. 

----
Note: I will remove aggressive logging of Reserved Instance data before merge.

### DIFF
--- a/cloud/provider.go
+++ b/cloud/provider.go
@@ -15,10 +15,9 @@ import (
 	"k8s.io/klog"
 
 	"cloud.google.com/go/compute/metadata"
+	"github.com/kubecost/cost-model/clustercache"
 
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 )
 
 const clusterIDKey = "CLUSTER_ID"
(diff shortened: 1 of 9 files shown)


mbolt35:
Note that I will revise and trim down the debug logging that currently exists. Leaving for the time being while testing :)

costmodel/costmodel.go
```diff
@@ -974,7 +968,7 @@ func GetPVCost(pv *costAnalyzerCloud.PV, kpv *v1.PersistentVolume, cp costAnalyz
 	return nil
 }
 
-func getNodeCost(cache ClusterCache, cp costAnalyzerCloud.Provider) (map[string]*costAnalyzerCloud.Node, error) {
+func getNodeCost(cache clustercache.ClusterCache, cp costAnalyzerCloud.Provider) (map[string]*costAnalyzerCloud.Node, error) {
```
mbolt35: Perhaps we should refactor this entire method into provider as something like: 
```go
func GetNodeCost(cp Provider) (map[string]*Node, error)
```

It does appear to create an "instanced" version of provider `Node`s. Not sure of the best approach here with regards to the `ApplyReservedInstancePricing` method call below.

costmodel/costmodel.go
```diff
@@ -1110,10 +1104,12 @@ func getNodeCost(cache ClusterCache, cp costAnalyzerCloud.Provider) (map[string]
 		nodes[name] = &newCnode
 	}
 
+	cp.ApplyReservedInstancePricing(nodes)
```
mbolt35: There were a few options here regarding how this should work. `ApplyReservedInstancePricing` requires VCPU and RAM values to be set for the `Node` for GCP (AWS will use machine type, but can still be applied here). See my notes above about the parent method `getNodeCost`. It may be better to change `ApplyReservedInstancePricing` to something more generic like `ApplyNodePricing` where each `Provider` could apply (or not) any special pricing schemes (including reserved instances).

cloud/provider.go
```diff
@@ -156,6 +165,7 @@ type Provider interface {
 	GetManagementPlatform() (string, error)
 	GetLocalStorageQuery(offset string) (string, error)
 	ExternalAllocations(string, string, string) ([]*OutOfClusterAllocation, error)
+	ApplyReservedInstancePricing(map[string]*Node)
```
mbolt35: See notes in costmodel about this. It may be better to rename this to something more generic, like `ApplyNodePricing` where each provider can implement any custom node pricing. It also seems like moving the `getNodeCost` method to `provider.go` is justifiable.

cloud/provider.go
```diff
@@ -34,25 +33,35 @@ var createTableStatements = []string{
...
+	StorageCost      string                `json:"storageHourlyCost"`
+	UsesBaseCPUPrice bool                  `json:"usesDefaultPrice"`
+	BaseCPUPrice     string                `json:"baseCPUPrice"` // Used to compute an implicit RAM GB/Hr price when RAM pricing is not provided.
+	BaseRAMPrice     string                `json:"baseRAMPrice"` // Used to compute an implicit RAM GB/Hr price when RAM pricing is not provided.
+	BaseGPUPrice     string                `json:"baseGPUPrice"`
+	UsageType        string                `json:"usageType"`
+	GPU              string                `json:"gpu"` // GPU represents the number of GPU on the instance
+	GPUName          string                `json:"gpuName"`
+	GPUCost          string                `json:"gpuCost"`
+	Reserved         *ReservedInstanceData `json:"reserved,omitempty"`
```
mbolt35: This property is sort of a rough draft of the protocol. AWS will be a bit different, but I think I can adapt that implementation to the same protocol.

cloud/provider.go
nikovacevic: I like the `ApplyNodePricing` idea

AjayTripathy:
LGTM, resolve conflicts and merge

PR #244, opened 2019-11-26
Caching updates

nikovacevic:


util/semaphore.go
```diff
@@ -0,0 +1,25 @@
+package util
+
+// Semaphore implements a non-weighted semaphore for restricting
+// concurrent access to a limited number of processes.
+type Semaphore struct {
```
nikovacevic: @AjayTripathy @mbolt35 We're not using this in this repo yet, but Bolt requested it go here so that we can use it in both places. I'm open to putting it anywhere, but here seemed fine.

mbolt35: You could be hyper optimal and use the quantum entangled data type `struct{}` for the channel. Never have I witnessed such an impossibly small data type. 💯 💯 💯 💯 💯

PR #249, opened 2019-12-02
Ajay tripathy fix n2price

AjayTripathy:


dwbrown2:
do you have a test cluster running for this?

dwbrown2:
are there other machine types that we are missing, e.g. m1, m2, and c2?

dwbrown2:
was this not supposed to fall back to default pricing in this case?

PR #270, opened 2020-01-08
add configmap init and watcher

AjayTripathy:


dwbrown2:
Would you be able to add a test to this?

costmodel/router.go
```diff
@@ -822,6 +823,24 @@ func Initialize() {
...
+	watchConfigFunc := func(c interface{}) {
+		conf := c.(*v1.ConfigMap)
+		if conf.GetName() == "pricing-configs" {
+			_, err := cloudProvider.UpdateConfigFromConfigMap(conf.Data)
+			if err != nil {
+				klog.Infof("ERROR UPDATING CONFIG: %s", err.Error())
+			}
+		}
+	}
+	// We need an initial invocation because the init of the cache has happened before we had access to the provider.
```
mbolt35: 😢

costmodel/router.go
```diff
@@ -822,6 +823,24 @@ func Initialize() {
...
+	configs, err := kubeClientset.CoreV1().ConfigMaps("kubecost").Get("pricing-configs", metav1.GetOptions{})
```
mbolt35: Saw in the clustercache we used `os.Getenv("KUBECOST_NAMESPACE")` - should that be the same here? I'm guilty of this in a few places I think.

AjayTripathy: Yes thank you good catch.

PR #276, opened 2020-01-15
NaN Checks

mbolt35:
`NaNNaNNaNNaNNaNNaNNaN Watman!`

### DIFF
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -605,8 +605,8 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, clientset kube
 					NamespaceLabels: nsLabels,
 					ClusterID:       clusterID,
 				}
-				costs.CPUAllocation = getContainerAllocation(costs.CPUReq, costs.CPUUsed)
-				costs.RAMAllocation = getContainerAllocation(costs.RAMReq, costs.RAMUsed)
+				costs.CPUAllocation = getContainerAllocation(costs.CPUReq, costs.CPUUsed, "CPU")
+				costs.RAMAllocation = getContainerAllocation(costs.RAMReq, costs.RAMUsed, "RAM")
 				if filterNamespace == "" {
 					containerNameCost[newKey] = costs
 				} else if costs.Namespace == filterNamespace {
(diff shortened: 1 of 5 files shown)


costmodel/costmodel.go
```diff
@@ -843,11 +843,22 @@ func findDeletedNodeInfo(cli prometheusClient.Client, missingNodes map[string]*c
...
 
-func getContainerAllocation(req []*Vector, used []*Vector) []*Vector {
+func getContainerAllocation(req []*Vector, used []*Vector, allocationType string) []*Vector {
 	// The result of the normalize operation will be a new []*Vector to replace the requests
 	allocationOp := func(r *Vector, x *float64, y *float64) bool {
 		if x != nil && y != nil {
-			r.Value = math.Max(*x, *y)
+			x1 := *x
+			if math.IsNaN(x1) {
+				klog.V(1).Infof("Warning: NaN value found during %s allocation calculation for requests.", allocationType)
```
AjayTripathy: I liked your convention of "[Reserved] blah blah blah" on logs, can we do that for [Warning]?

costmodel/promparsers.go
```diff
@@ -148,6 +148,11 @@ func parseDataPoint(dataPoint interface{}) (*Vector, error) {
 		return nil, err
 	}
 
+	if math.IsNaN(v) {
+		klog.V(1).Infof("Warning: Found NaN value parsing vector data point.")
```
AjayTripathy: Any chance we could see labels here? I know you'd need to pass them in from somewhere but it would be useful to know what's NaN. otherwise there will be thousands of these statements with no real context.

costmodel/promparsers.go
```diff
@@ -148,6 +148,11 @@ func parseDataPoint(dataPoint interface{}) (*Vector, error) {
...
+		v = 0.0
```
AjayTripathy: thought you wanted to not return 0? change your mind?

costmodel/costmodel.go
mbolt35: Yeah

costmodel/promparsers.go
mbolt35: Yeah, we can do that...

costmodel/promparsers.go
mbolt35: Well, I started looking later in the path and noticed we do a lot of checking for 0 when it can be dangerous. Since, this happens early on, I think it's safe to rely on our 0 checking later.

PR #277, opened 2020-01-16
More profiling

nikovacevic:


costmodel/costmodel.go
```diff
@@ -1378,6 +1378,12 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 	}
 	clusterID := os.Getenv(clusterIDKey)
 
+	durHrs := end.Sub(start).Hours() + 1
+	durStr := fmt.Sprintf("%fh", durHrs)
+	if durHrs > 24.0 {
```
AjayTripathy: it's ok to drop this and just print hours, we can convert. less code is better here esp since fmt.Sprintf(...) can have errors and we're trying to turn this around quickly.

costmodel/costmodel.go
```diff
@@ -1378,6 +1378,12 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 	}
 	clusterID := os.Getenv(clusterIDKey)
 
+	durHrs := end.Sub(start).Hours() + 1
+	durStr := fmt.Sprintf("%fh", durHrs)
```
AjayTripathy: can we just use durHrs everywhere and divide ourself when we read logs, for simplicity?

nikovacevic: Yep, on it

AjayTripathy:
LGTM

PR #282, opened 2020-01-20
NaN Logging in GetNodeCost

mbolt35:
Added aggressive NaN checks for GetNodeCost to weed out any potential 0 / 0 scenarios.

### DIFF
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -1036,7 +1036,7 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 
 		cnode, err := cp.NodePricing(cp.GetKey(nodeLabels))
 		if err != nil {
-			klog.V(1).Infof("Error getting node. Error: " + err.Error())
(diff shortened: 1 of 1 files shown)


costmodel/costmodel.go
```diff
@@ -1142,8 +1215,17 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 					return nil, err
 				}
 			}
+			if math.IsNaN(nodePrice) {
+				klog.V(1).Infof("[Warning] nodePrice parsed as NaN. Setting to 0.")
+				nodePrice = 0
+			}
 
 			ramPrice := (nodePrice / ramMultiple)
```
AjayTripathy: Should we be dividing here and accepting the nan if this could be 0?

mbolt35: So yeah... I thought about this, but it also seemed worthwhile to note each possible problem case rather than skipping some cases knowing they could now be 0 due to our "fixups" Hopefully this will shed some light on the issue.

PR #288, opened 2020-01-22
Resolution-related improvements

nikovacevic:
Supports https://github.com/kubecost/kubecost-cost-model/pull/52

Changes here aim to fix accuracy problems related to decreased-resolution Prometheus queries:
* Reverted the scalar modification of the normalization query in favor of the following.
* RAM and CPU allocation queries now return the total (resource*hour) amounts over the given windows, rather than normalized hourly rate data. This vastly improves the accuracy of cumulative results.
* Correcting for scrape-misses has become slightly more complicated, but seems to do a good job within a reasonable percentage of scrape misses. If anything, this is the part that could use review and improvement.

### DIFF
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -338,7 +338,6 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, clientset kube
 	queryNetZoneRequests := fmt.Sprintf(queryZoneNetworkUsage, window, "")
 	queryNetRegionRequests := fmt.Sprintf(queryRegionNetworkUsage, window, "")
 	queryNetInternetRequests := fmt.Sprintf(queryInternetNetworkUsage, window, "")
-	//normalization := fmt.Sprintf(normalizationStr, window, offset, 1.0)
 	normalization := fmt.Sprintf(normalizationStr, window, offset)
 
 	// Cluster ID is specific to the source cluster
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -1837,8 +1837,8 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 
 	normalizationValue, err := getNormalizations(normalizationResults)
 	if err != nil {
-		return nil, fmt.Errorf("error computing normalization for start=%s, end=%s, window=%s: %s",
-			start, end, window, err.Error())
+		return nil, fmt.Errorf("error computing normalization for start=%s, end=%s, window=%s, res=%f: %s",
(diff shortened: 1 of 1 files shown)


costmodel/costmodel.go
```diff
@@ -1948,51 +1985,58 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
...
 	}
 	for key := range CPUReqMap {
 		containers[key] = true
 	}
-	CPUUsedMap, err := GetContainerMetricVectors(resultCPUUsage, false, normalizationValue, clusterID) // No need to normalize here, as this comes from a counter
+
+	// No need to normalize here, as this comes from a counter
+	// CPUUsedMap, err := GetContainerMetricVectors(resultCPUUsage, clusterID)
+	// TODO Is that ^ true? We were normalizing before (in spite of the comment) so I'm keeping it.
+	CPUUsedMap, err := GetNormalizedContainerMetricVectors(resultCPUUsage, normalizationValue, clusterID)
```
mbolt35: So, before, we were passing the `normalizationValue`, but the second parameter to `GetContainerMetricVectors` is `normalize bool`, which we were passing `false`. The old line was:

```
                                          This is tells the method not to normalize
                                                              |
                                                              V
CPUUsedMap, err := GetContainerMetricVectors(resultCPUUsage, false, normalizationValue, clusterID)
```

The actual logic in the method:
```
               if normalize && normalizationValue != 0 {
			for _, v := range val.Values {
				v.Value = v.Value / normalizationValue
			}
		}
```

mbolt35: So, I realized I made a mistake. The `GetContainerMetricVector` for non-ranged queries actually honors the `normalize bool` parameter while the `GetContainerMetricVectors` method didn't. I'm not sure if this was intended or not, and may have actually been a bug I introduced when moving to the `PromQueryResult` refactor. Going to investigate a bit more here.

mbolt35: Tracked down my refactor changeset, and it doesn't look like we ever honored the `normalize bool` parameter in `GetContainerMetrics()` Here's the version from early November:

https://github.com/kubecost/cost-model/blob/95d335b75ebbc283efcd4d152c73bebbc3c1a5a1/costmodel/costmodel.go#L2563-L2615

However, if you go back before the existence of `vector.go`, there is a version where the logic was similar to `GetContainerMetricVector`:

https://github.com/kubecost/cost-model/blob/f7eb99baa6872f071081334059ac82477116c2d4/costmodel/costmodel.go#L1704-L1758

AjayTripathy:
LGTM

PR #289, opened 2020-01-22
Ajay tripathy update fixes

AjayTripathy:


costmodel/router.go
```diff
@@ -868,8 +868,9 @@ func Initialize() {
 	configs, err := kubeClientset.CoreV1().ConfigMaps(kubecostNamespace).Get("pricing-configs", metav1.GetOptions{})
 	if err != nil {
 		klog.Infof("ERROR FETCHING configmap: %s", err.Error())
+	} else {
+		watchConfigFunc(configs)
```
mbolt35: I saw this the other day, but it still seemed fine given that the `ConfigMaps(kubecostNamespace).Get()` will always return a valid pointer. Probably more clear this way :)

PR #295, opened 2020-01-24
more locking safety, getcustompricing safety

AjayTripathy:


cloud/provider.go
```diff
@@ -233,36 +251,27 @@ func GetDefaultPricingData(fname string) (*CustomPricing, error) {
...
 		cj, err := json.Marshal(c)
 		if err != nil {
-			return nil, err
+			return c, err
 		}
 
 		err = ioutil.WriteFile(path, cj, 0644)
 		if err != nil {
+			klog.Infof("Could not write Custom Pricing file to path '%s'", path)
 			return nil, err
```
mbolt35: do we want to return `c` here as well?

mbolt35:
One other important note is that this will also require an update to `kubecost-cost-model` gcp provider uses `GetDefaultPricing` instead of `GetCustomPricing`

PR #305, opened 2020-01-31
nodeprice fixes

AjayTripathy:
Remove some unnecessary code.

The cost model building loop should not use the k8s api at all. Data is placed in Prometheus by the emission loop, and is read from prometheus/thanos via the consumption loop. 

Not that "findMissingNode" is misnamed-- it now finds all nodes. Consequently, we should remove the regex and just get all the node prices. That is TODO, wanted to get an initial review and let this bake because it is a big change.

Additionally, Azure GPU prices are baked into nodes, and this has code to support that.

### DIFF
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -2065,7 +2065,6 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 		if _, ok := containerNameCost[key]; ok {
 			continue // because ordering is important for the allocation model (all PV's applied to the first), just dedupe if it's already been added.
 		}
-		// The container has been deleted, or is from a different clusterID
 		c, _ := NewContainerMetricFromKey(key)
 		RAMReqV, ok := RAMReqMap[key]
 		if !ok {
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -2071,7 +2071,6 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 			continue // because ordering is important for the allocation model (all PV's applied to the first), just dedupe if it's already been added.
 		}
 		// The container has been deleted, or is from a different clusterID
-		// Not all information is sent to prometheus via ksm, so fill out what we can without k8s api
 		c, _ := NewContainerMetricFromKey(key)
 		RAMReqV, ok := RAMReqMap[key]
 		if !ok {
--- a/costmodel/costmodel.go
+++ b/costmodel/costmodel.go
@@ -2072,7 +2072,6 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 		}
 		// The container has been deleted, or is from a different clusterID
 		// Not all information is sent to prometheus via ksm, so fill out what we can without k8s api
-		klog.V(4).Info("The container " + key + " has been deleted. Calculating allocation but resulting object will be missing data.")
 		c, _ := NewContainerMetricFromKey(key)
 		RAMReqV, ok := RAMReqMap[key]
 		if !ok {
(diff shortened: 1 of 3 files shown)


costmodel/costmodel.go
```diff
@@ -2023,6 +2044,10 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 	}
 	for key := range GPUReqMap {
 		containers[key] = true
+		klog.Infof("GPUMAP: %s", key)
```
mbolt35: Log level for this?

costmodel/costmodel.go
```diff
@@ -2023,6 +2044,10 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
...
+		for _, vec := range GPUReqMap[key] {
+			klog.Infof("%f: %f", vec.Timestamp, vec.Value)
```
mbolt35: Log Level?

costmodel/costmodel.go
AjayTripathy: Oops. Accidentally left these in for debug. Thanks

costmodel/costmodel.go
```diff
@@ -174,6 +174,7 @@ const (
 					count_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!=""}[%s] %s) 
 					*  
 					avg_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!=""}[%s] %s)
+					* %f
```
nikovacevic: This will be subject to the same startup accuracy issues that we recently solved for CPU and RAM, but it's a fine solution for steady-state. Not saying we need to change it now, but putting it on everyone's radar.

costmodel/costmodel.go
```diff
@@ -1075,6 +1076,28 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
...
+		q, ok := n.Status.Capacity["nvidia.com/gpu"]
+		if ok {
+			gpuCount := q.Value()
+			if gpuCount != 0 {
+				newCnode.GPU = fmt.Sprintf("%d", q.Value())
+				gpuc = float64(gpuCount)
+			}
+		} else {
+			gpuc, err = strconv.ParseFloat(newCnode.GPU, 64)
+			if err != nil {
```
nikovacevic: Do you think we want a log here?

costmodel/costmodel.go
```diff
@@ -2060,296 +2065,147 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
...
-			}
-
-			nsKey := c.Namespace + "," + c.ClusterID
-			podKey := c.Namespace + "," + c.PodName + "," + c.ClusterID
-
-			namespaceLabels, ok := namespaceLabelsMapping[nsKey]
-			if !ok {
-				klog.V(3).Infof("Missing data for namespace %s", c.Namespace)
-			}
+		// The container has been deleted, or is from a different clusterID
```
nikovacevic: This comment no longer applies, correct?

costmodel/costmodel.go
AjayTripathy: I know. we need to migrate to the gpu_allocation stat after this bakes for a while (that stat was recently added)

costmodel/costmodel.go
AjayTripathy: Pretty high volume place to put a log, considering that parsefloat is often "", which returns an error.

AjayTripathy:
Closing in favor of #306, this was initially intended to be a patch but will go out in the full release.

PR #309, opened 2020-02-04
fix nil pointer, add daemon and statefulsets

AjayTripathy:


costmodel/promparsers.go
```diff
@@ -380,6 +380,41 @@ func GetStatefulsetMatchLabelsMetrics(queryResult interface{}, defaultClusterID
...
 }
 
+func GetPodDaemonsetsWithMetrics(queryResult interface{}, defaultClusterID string) (map[string]string, error) {
+	toReturn := make(map[string]string)
+	result, err := NewQueryResults(queryResult)
+	if err != nil {
+		return toReturn, err
+	}
+	for _, val := range result {
+		// We want Deployment, Namespace and ClusterID for key generation purposes
```
nikovacevic: This looks like an un-updated copy/paste comment

costmodel/promparsers.go
```diff
@@ -380,6 +380,41 @@ func GetStatefulsetMatchLabelsMetrics(queryResult interface{}, defaultClusterID
...
+			return toReturn, err
+		}
+
+		clusterID, err := val.GetString("cluster_id")
+		if clusterID == "" {
+			clusterID = defaultClusterID
+		}
+
+		pod, err := val.GetString("pod")
+		if clusterID == "" {
```
nikovacevic: Should this be `pod == ""`?

costmodel/costmodel.go
```diff
@@ -2144,6 +2166,16 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 			}
 		}
 
+		var podStatefulSets []string
+		if _, ok := podStatefulsetsMapping[nsKey]; ok {
```
nikovacevic: If `nsKey := c.Namespace + "," + c.ClusterID` how does this match with the `podStatefulsetsMapping` keys, which are `nsKey := ns + "," + ss + "," + clusterID`?

AjayTripathy: podStatefulsetsMapping should be a map[string]map[string][]string where the key is two levels, [namespace,clusterid][podname]

nikovacevic: Yes, but I'm pointing out that there might be in issue here: https://github.com/kubecost/cost-model/blob/93f04c52cf43e32111747f30703a3e23769ea51f/costmodel/promparsers.go#L376

The type is still `map[string]map[string][]string`, but I think the top-level keys will not match.

nikovacevic: Perhaps that `nsKey` in the promparsers for StatefulSets should just be `ns,clusterId`, no?

nikovacevic: Or perhaps the function `getPodDeploymentsWithMetrics` is translating between the two. Looks like that might be the case. It's pretty confusing in there, but if it seems to be working then no need to clarify.

PR #318, opened 2020-02-11
add e2 support

AjayTripathy:
* e2highmem/cpu are billed at the same rate per resource as e2standard
* e2small/medium/micro are billed for fractional CPU

### DIFF
--- a/cloud/gcpprovider.go
+++ b/cloud/gcpprovider.go
@@ -463,6 +463,13 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]Key, pvKeys map[stri
 					instanceType = "n2standard"
 				}
 
+				if (instanceType == "ram" || instanceType == "cpu") && strings.Contains(strings.ToUpper(product.Description), "E2 INSTANCE") {
+					instanceType = "e2"
+				}
+				partialCPUMap := make(map[string]float64)
+				partialCPUMap["e2micro"] = 0.25
+				partialCPUMap["e2small"] = 0.5
+				partialCPUMap["e2medium"] = 1
 				/*
 					var partialCPU float64
 					if strings.ToLower(instanceType) == "f1micro" {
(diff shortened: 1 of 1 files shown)


cloud/gcpprovider.go
```diff
@@ -480,11 +487,23 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]Key, pvKeys map[stri
 					}
 				}
 
-				for _, sr := range product.ServiceRegions {
-					region := sr
-					candidateKey := region + "," + instanceType + "," + usageType
-					candidateKeyGPU := candidateKey + ",gpu"
+				candidateKeys := []string{}
+				for _, region := range product.ServiceRegions {
+					if instanceType == "e2" { // this needs ot be done to handle a partial cpu mapping
```
nikovacevic: Typo "ot" instead of "to"

cloud/gcpprovider.go
```diff
@@ -560,11 +578,10 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]Key, pvKeys map[stri
...
 										RAMCost: strconv.FormatFloat(hourlyPrice, 'f', -1, 64),
 									}
-									/*
-										if partialCPU != 0 {
-											product.Node.VCPU = fmt.Sprintf("%f", partialCPU)
-										}
-									*/
+									partialCPU, pcok := partialCPUMap[instanceType]
+									if pcok {
+										product.Node.VCPU = fmt.Sprintf("%f", partialCPU)
```
nikovacevic: It's pretty tough to tell what's going on here for such a big and deeply-nested function, so I just want to check: this block appears to be setting RAM cost for the given node. Do we also want to be setting VCPU here? Doesn't seem harmful, I'm just confused.

AjayTripathy: Agreed its confusing, but the pattern is (or should be): anytime you add a new object to the block, set its VCPU, since we know how many virtualCPUs there are as soon as we know the key.

PR #326, opened 2020-02-13
add version info to /clusterinfo

AjayTripathy:


costmodel/router.go
```diff
@@ -579,6 +579,18 @@ func (p *Accesses) ClusterInfo(w http.ResponseWriter, r *http.Request, ps httpro
 	w.Header().Set("Access-Control-Allow-Origin", "*")
 
 	data, err := p.Cloud.ClusterInfo()
+
```
nikovacevic: @AjayTripathy would it be clearer to write the error and return early if `err != nil` here?

PR #327, opened 2020-02-14
Cluster Manager Implementation

mbolt35:
### ClusterManager
Cluster manager is a very basic key/value storage. The out of box implementation is provided by boltdb (etcd flavor), and will fall back to a memory storage on failure to load the database from disk.

### ClusterDefinition
The `ClusterDefinition` object is used by the server and client to represent a cluster entry. When the client wishes to modify an existing `ClusterDefinition`, it should include the `id` as part of the request. The server will always respond with the `id` present. The `details` field is completely optional and is provided to attach custom metadata from the client to the cluster definition.

### Cluster Definition JSON
```json
    {
        "id": "<guid>",
        "name": "ClusterName",
        "address": "http://cluster-addr:9090",
        "details": {}
    }
```

### Cluster Manager Endpoints
```
    GET /model/clusters
    Returns an array of all ClusterDefinitions

    PUT /model/clusters
    Accepts JSON for a ClusterDefinition to add or update in the manager. If an 'id' is included, then 
    entry corresponding to that id is updated. Otherwise, a new cluster definition is added.

    DELETE /model/clusters/:id
    Deletes a ClusterDefinition from management. Must specifically provide an 'id' in the url. 
```

### DIFF
--- a/pkg/clustermanager/boltdbstorage.go
+++ b/pkg/clustermanager/boltdbstorage.go
@@ -1,0 +1,89 @@
+package clustermanager
+
+import (
+	bolt "github.com/etcd-io/bbolt"
+	_ "k8s.io/klog"
+)
+
+type BoltDBClusterStorage struct {
+	bucket []byte
+	db     *bolt.DB
+}
+
+func NewBoltDBClusterStorage(bucket string, db *bolt.DB) (ClusterStorage, error) {
+	bucketKey := []byte(bucket)
+
+	err := db.Update(func(tx *bolt.Tx) error {
+		_, err := tx.CreateBucketIfNotExists(bucketKey)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &BoltDBClusterStorage{
+		bucket: bucketKey,
+		db:     db,
+	}, nil
+}
+
+// Adds the entry if the key does not exist
+func (cs *BoltDBClusterStorage) AddIfNotExists(key string, cluster []byte) error {
+	return cs.db.Update(func(tx *bolt.Tx) error {
+		k := []byte(key)
+		bucket := tx.Bucket(cs.bucket)
+
+		if bucket.Get(k) != nil {
+			return nil
+		}
+		return bucket.Put(k, cluster)
+	})
+}
+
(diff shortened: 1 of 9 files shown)


pkg/clustermanager/boltdbstorage.go
```diff
@@ -0,0 +1,89 @@
...
+func (cs *BoltDBClusterStorage) Remove(key string) error {
+	return cs.db.Update(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(cs.bucket)
+
+		return bucket.Delete([]byte(key))
+	})
+}
+
+// Retrieve all the keys stored
+func (cs *BoltDBClusterStorage) Each(handler func(string, []byte)) error {
```
nikovacevic: Two pieces of feedback here:

1. Shouldn't the `handler` func return `error`, which should then be checked and returned by `ForEach`? We do check for errors coming back from a call to `Each` in `ClusterManager.GetAll()` but this function can never return an error because it's hard-coded to always `return nil`, as opposed to `return handler(...)`. There may be a reason for the signature as-is, but wanted to check.

(One example of how I'm imagining we change this can be found here: https://godoc.org/go.etcd.io/bbolt#Bucket.ForEach)

2. This is a nit, but we may want the comment to mention that this function executes a handler func for each key/pair in the bucket. You're using it to retrieve the keys, but it could be used for any manner of handler funcs.

mbolt35: 1. I thought about this for a while during implementation, and I really didn't want a failed handler to halt the transaction, and imagined the handler as being a receiver, and nothing more. That is, the `Each()` does not care about feedback from the handler itself. This would be a problem if you were searching and wanted to end your search at a specific result, but that didn't support anything we were currently doing so it just seemed unnecessary. It's funny that you provided an example of _exactly_ the implementation I am using. 

2. Yeah, the comment was just a stub -- I can definitely update.

mbolt35: > Everything except the `Each` implementation looks good to me.
> 
> Conceptually, does this mean that each cluster will have its own cluster manager? In multi-cluster, will there be a "master" cluster tracking all the ClusterDefinitions, or will each cluster have to track a ClusterDefinition for each other cluster?

For right now, yes, we're keeping this per cluster. The first phase shifts the cluster management from the frontend to the backend. Phase 2 will be more about centralized management of multiple clusters.

PR #329, opened 2020-02-15
Ajay tripathy change default filter

AjayTripathy:
* Support filtering on out-of-cluster tags in gke
* Change the AWS default filter so that empty selectors do not match everything.


AjayTripathy:
closed in favor of 337

PR #337, opened 2020-02-21
Ajay tripathy multilabel

AjayTripathy:
Support multiple labeltypes to match on
Support multiple cloud providers as sources of out-of-cluster cost.


cloud/provider.go
```diff
@@ -203,6 +204,19 @@ func CustomPricesEnabled(p Provider) bool {
 	return config.CustomPricesEnabled == "true"
 }
 
+func NewCrossClusterProvider(ctype string, overrideConfigPath string, cache clustercache.ClusterCache) (Provider, error) {
+	if ctype == "aws" {
```
nikovacevic: I'm confused about how we're using this bidirectionally (AWS CCP for GCP Provider and GCP CCP for AWS Provider) but `ctype` is always `"aws"`. Shouldn't there be an option for a `ctype == "gcp"`?

cloud/awsprovider.go
```diff
@@ -1169,17 +1183,36 @@ func (a *AWS) ExternalAllocations(start string, end string, aggregator string, f
...
 				oocAllocs = append(oocAllocs, ooc)
 			}
 		} else {
-			klog.V(1).Infof("No results available for %s at database %s between %s and %s", aggregator_column_name, customPricing.AthenaTable, start, end)
+			klog.V(1).Infof("No results available for %s at database %s between %s and %s", strings.Join(formattedAggregators, ","), customPricing.AthenaTable, start, end)
+		}
+	}
+
+	if customPricing.BillingDataDataset != "" { // There is GCP data, meaning someone has tried to configure a GCP out-of-cluster allocation
+		gcp, err := NewCrossClusterProvider("aws", "aws.json", a.Clientset)
```
nikovacevic: Perhaps it makes no difference, but confused about why this is `ctype = "aws"`

cloud/gcpprovider.go
```diff
@@ -211,28 +261,131 @@ func (gcp *GCP) UpdateConfig(r io.Reader, updateType string) (*CustomPricing, er
...
+    service.description as service,
+    TO_JSON_STRING(labels) as gl,
+    SUM(cost)
+  FROM  `guestbook-227502.billing_data.gcp_billing_export_v1_01AC9F_74CF1D_5565A2`
+  WHERE
+    EXISTS (SELECT * FROM UNNEST(labels) AS l WHERE l.key = "kubernetes_namespace" AND l.value = "kubecost")
+    AND EXISTS(SELECT * FROM UNNEST(labels) AS l2 WHERE l2.key = "kubernetes_label_app")
+    AND usage_start_time >= "2020-02-10" AND usage_start_time < "2020-02-13"
+	GROUP BY  service,gl)
+*/
```
nikovacevic: I'm assuming this is just left over from the work-in-progress and should be deleted?

cloud/awsprovider.go
```diff
@@ -1169,17 +1183,36 @@ func (a *AWS) ExternalAllocations(start string, end string, aggregator string, f
...
 				}
 				oocAllocs = append(oocAllocs, ooc)
 			}
 		} else {
-			klog.V(1).Infof("No results available for %s at database %s between %s and %s", aggregator_column_name, customPricing.AthenaTable, start, end)
+			klog.V(1).Infof("No results available for %s at database %s between %s and %s", strings.Join(formattedAggregators, ","), customPricing.AthenaTable, start, end)
+		}
+	}
+
+	if customPricing.BillingDataDataset != "" { // There is GCP data, meaning someone has tried to configure a GCP out-of-cluster allocation
```
nikovacevic: I would just note that once this field is set, it cannot be deleted. (I have tried to delete my service key, etc., for testing purposes and it does not work.) If we're going to pull GCP data on the basis of that field being preset, I think we should allow it to be deleted.

cloud/awsprovider.go
AjayTripathy: Ah good catch thanks.

cloud/provider.go
AjayTripathy: Yes, added.

PR #348, opened 2020-02-25
Ajay tripathy fix dev

AjayTripathy:


pkg/cloud/awsprovider.go
```diff
@@ -1205,12 +1205,12 @@ func (a *AWS) ExternalAllocations(start string, end string, aggregators []string
 		}
 	}
 
-	if customPricing.BillingDataDataset != "" { // There is GCP data, meaning someone has tried to configure a GCP out-of-cluster allocation.
-		gcp, err := NewCrossClusterProvider("gcp", "gcp.json", a.Clientset)
+	if customPricing.BillingDataDataset != "" && !crossCluster { // There is GCP data, meaning someone has tried to configure a GCP out-of-cluster allocation.
+		gcp, err := NewCrossClusterProvider("gcp", "aws.json", a.Clientset)
```
nikovacevic: I trust that this solved something, but it looks a bit confusing to create a `gcp` cross cluster provider with the `aws.json` config path.

AjayTripathy: Configs are just listed as aws.json when you're on AWS-- vice versa for GCP. it's the same config file

PR #354, opened 2020-02-27
ClusterCosts improvements

nikovacevic:
* Fix ClusterCosts being too high when nodes come up and down (problem with `fmtQueryDataCount`)
* Apply discount to ClusterCosts on back-end
* Include percentage breakdown in ClusterCosts


AjayTripathy:
This LGTM, is it still a work in progress?

nikovacevic:
@AjayTripathy Yes, PV breakdown is not complete, and implications for multi-cluster are unclear. Should wrap up those two this morning.

PR #356, opened 2020-02-27
Fix data count query for ComputeClusterCosts

nikovacevic:
Corrects the "minutes per window" query to account for nodes coming up and down.


AjayTripathy:
Lgtm

PR #358, opened 2020-02-27
Fix Azure PV Price Units

mbolt35:


pkg/cloud/azureprovider.go
```diff
@@ -408,7 +408,9 @@ func (az *Azure) DownloadPricingData() error {
 						for _, rate := range v.MeterRates {
 							priceInUsd += *rate
 						}
-						priceStr := fmt.Sprintf("%f", priceInUsd)
+						// rate is in GB per month, resolve to GB per hour
+						pricePerHour := priceInUsd / 720.0
```
AjayTripathy: there are ~730 hours per month

PR #399, opened 2020-04-17
Expose YAML definition for Details in the ClusterEntry

mbolt35:
### Summary
Expose the YAML definition for "details" in the cluster entries defined in configmap. This opens up the frontend to implement it's own protocol for cluster entries.

If there is a configmap provided, which contains cluster entry definitions, they can now properly set the details field.


mbolt35:
This PR did not satisfy requirements well enough. Closing, and will make a new PR with auth additions.

PR #403, opened 2020-04-22
Ajay tripathy allow direct

AjayTripathy:


pkg/costmodel/router.go
```diff
@@ -897,18 +897,19 @@ func Initialize(additionalConfigWatchers ...ConfigWatchers) {
...
-		klog.Fatalf("No valid prometheus config file at %s. Error: %s . Troubleshooting help available at: %s", address, err.Error(), prometheusTroubleshootingEp)
-	}
-	klog.V(1).Info("Success: retrieved a prometheus config file from: " + address)
-
-	_, err = ValidatePrometheus(promCli, false)
-	if err != nil {
-		klog.Fatalf("Failed to query prometheus at %s. Error: %s . Troubleshooting help available at: %s", address, err.Error(), prometheusTroubleshootingEp)
+	m, err := ValidatePrometheus(promCli, false)
+	if err != nil || m.Running == false {
+		klog.Errorf("Failed to query prometheus at %s. Error: %s . Troubleshooting help available at: %s", address, err.Error(), prometheusTroubleshootingEp)
```
rvolykh, line 902: in case `err = nil` and `m.Running = false` you will get panic at `err.Error()`

AjayTripathy: Thanks so much for this review and sorry for the late response-- I think we may end up going a different direction on support for direct thanos connection.

PR #410, opened 2020-04-29
add where clause

AjayTripathy:
Add a simple clause to reduce amount of data returned to the server. Regression tested with AWS external cost tests.

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -1274,6 +1274,8 @@ func (a *AWS) ExternalAllocations(start string, end string, aggregators []string
 		formattedAggregators = append(formattedAggregators, aggregator_column_name)
 	}
 	aggregatorNames := strings.Join(formattedAggregators, ",")
+	aggregatorOr := strings.Join(formattedAggregators, " <> '' OR ")
+	aggregatorOr = aggregatorOr + " <> ''"
 
 	filter_column_name := "resource_tags_user_" + filterType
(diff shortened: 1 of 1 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -1290,8 +1290,8 @@ func (a *AWS) ExternalAllocations(start string, end string, aggregators []string
 			%s,
 			SUM(line_item_blended_cost) as blended_cost
 		FROM %s as cost_data
-		WHERE (%s='%s') AND line_item_usage_start_date BETWEEN date '%s' AND date '%s'
-		GROUP BY %s`, aggregatorNames, filter_column_name, customPricing.AthenaTable, filter_column_name, filterValue, start, end, groupby)
+		WHERE (%s='%s') AND line_item_usage_start_date BETWEEN date '%s' AND date '%s AND %s <> ''
```
nikovacevic, line 1293: Are the quotes correct here? Seems like it should be `... AND date '%s' AND...` (with a close quote on the %s) but perhaps I'm just reading this wrong.

nikovacevic: And why only use `formattedAggregators[0]`?

AjayTripathy: Fixed, thanks. It turns out I was testing the wrong if clause...

nikovacevic: Ah that makes sense. This one looks better :+1:

PR #417, opened 2020-05-06
Prometheus helpers package

nikovacevic:
## Changes
- Group existing Prometheus utilities (e.g. promparsers) into new package that can be neatly imported elsewhere w/o the entire costmodel package
- Implement clean interface for async Prometheus queries

### Example
From [here](https://github.com/kubecost/kubecost-cost-model/pull/131/files#diff-01c9ea57a2c6da08768f763742664ce4R85)
```go
resChs := ctx.QueryAll(
	fmtQueryCPUAllocation(p, window, true),
	fmtQueryCPUAllocation(p, window, false),
	fmtQueryRAMAllocation(p, window, true),
	fmtQueryRAMAllocation(p, window, false),
	fmtQueryMaxPodCPU(p, window),
	fmtQueryMaxPodRAM(p, window),
)

...

resultsDaemonSetCPU := resChs[0].Await()
resultsStaticCPU := resChs[1].Await()
```

### DIFF
--- a/pkg/prom/query.go
+++ b/pkg/prom/query.go
@@ -1,0 +1,112 @@
+package prom
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/kubecost/cost-model/pkg/util"
+	prometheus "github.com/prometheus/client_golang/api"
+	"k8s.io/klog"
+)
+
+const (
+	apiPrefix = "/api/v1"
+	epQuery   = apiPrefix + "/query"
+)
+
+// Context wraps a Prometheus client and provides methods for querying and
+// parsing query responses and errors.
+type Context struct {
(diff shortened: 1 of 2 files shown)


mbolt35:
From your example, can you await all via:

```go
resChs.Await()
```
?

This appeared to be available, but maybe I misunderstood.

nikovacevic:
@mbolt35 Ah, no, but I like the idea.

So we'd need a new type, something like:
```go
type QueryResultsChans []QueryResultsChan
```
And the signature would be something like
```go
func (qrcs QueryResultsChans) Await() [][]*QueryResults
```

I could see a use for that, but I ultimately just call `Await()` on each individual channel as I need to use their results. Is there a particular use-case you had in mind where an `AwaitAll()` type call makes things a lot simpler?

PR #418, opened 2020-05-11
Contextualize errors for better debugging

nikovacevic:
## Changes
- Wrap errors with a contextual label
- In some cases, don't return both a value and an error (would like confirmation, ultimately, but that makes a response that's confusing to interpret)

## Testing
- Swapped images on kc-niko cluster and confirmed things work
- Added a fake error and confirmed that the label is printed in logs


AjayTripathy:
lgtm

AjayTripathy:
Can we also go directly to master btw

PR #430, opened 2020-05-21
ComputeClusterCosts: stabilize startup

nikovacevic:
## Changes
- Query for data count with equal resolution to cost data, so that mismatch in granularity stops causing oscillating rate data
- Remove logs

## Testing
Installed fresh from master, confirmed costs jumping. Then installed this change, confirmed costs are more stable.

### Before
2m
Screenshot from 2020-05-21 10-15-05
3m
Screenshot from 2020-05-21 10-17-49
4m
Screenshot from 2020-05-21 10-18-39
5m (correct)
Screenshot from 2020-05-21 10-19-46
6m (back up)
Screenshot from 2020-05-21 10-20-10

### After
8m
Screenshot from 2020-05-21 10-23-18

9m
Screenshot from 2020-05-21 10-25-49

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/kubecost/cost-model/pkg/cloud"
-	"github.com/kubecost/cost-model/pkg/log"
 	"github.com/kubecost/cost-model/pkg/prom"
 	"github.com/kubecost/cost-model/pkg/util"
 	prometheus "github.com/prometheus/client_golang/api"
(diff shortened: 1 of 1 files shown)


pkg/costmodel/cluster.go
```diff
@@ -146,7 +145,7 @@ func ComputeClusterCosts(client prometheus.Client, provider cloud.Provider, wind
 	hourlyToCumulative := float64(minsPerResolution) * (1.0 / 60.0)
 
 	const fmtQueryDataCount = `
-		count_over_time(sum(kube_node_status_capacity_cpu_cores) by (cluster_id)[%s:1m]%s)
+		count_over_time(sum(kube_node_status_capacity_cpu_cores) by (cluster_id)[%s:%dm]%s) * %d
```
nikovacevic, line 148: Previously, at startup this would return 1, 2, ..., 5, 6, 7, .... as the minutes pass. Now it will be less accurate, but will match the scaling of `hourlyToCumulative`, returning 5, 5, ..., 5, 10, 10, ... as the first minutes pass.

That query is used to create `dataMinsPerCluster`, which feeds the function `NewClusterCostsFromCumulative` to compute monthly costs.

PR #431, opened 2020-05-23
Cost Model Error Reporting

mbolt35:


pkg/errors/panic.go
```diff
@@ -0,0 +1,121 @@
...
+// PanicHandler is a func that receives a Panic and returns a bool representing whether or not
+// the panic should recover or not.
+type PanicHandler = func(p Panic) bool
+
+var (
+	enabled    = false
+	dispatcher = make(chan Panic)
+)
+
+// SetPanicHandler sets the handler that is executed when
```
nikovacevic, line 46: cliffhanger

pkg/costmodel/cluster.go
```diff
@@ -40,7 +41,7 @@ const (
 // TODO move this to a package-accessible helper
 type PromQueryContext struct {
 	Client         prometheus.Client
-	ErrorCollector *util.ErrorCollector
+	ErrorCollector *errors.ErrorCollector
```
nikovacevic, line 44: Complete nit, but might be nice if this is just called `errors.Collector` to avoid naming stutter.

nikovacevic:
@mbolt35 a nice artifact for this PR might be a quick summary of how this was tested

pkg/errors/panic.go
mbolt35: Yikes! Thanks :)

PR #435, opened 2020-05-25
The version of the Deployment spec

hvandenb:
This version has changed with K8s and also now requires a selector.


AjayTripathy:
Thank you for this PR-- LGTM!

hvandenb:
Ah didn’t realize there was the template as well. Good to know.

PR #455, opened 2020-06-25
fix: removing duplicated node name labels

caarlos0:
some metrics are exposed with an `instance` and a `node` label, and both have the same values.

as far as I was able to look at, the places where `instance` is used, `node` could be used instead.

this patch (although I'm not yet 100% sure is correct) would reduce metric cardinality (specially in clusters that autoscale a lot and/or with spot/preemptible nodes) and improve query performance by dropping some `label_replace`s.

### DIFF
--- a/PROMETHEUS.md
+++ b/PROMETHEUS.md
@@ -14,7 +14,7 @@ After deploying the Kubecost model (see [README](README.md) for more info on ins
   static_configs:
   - targets:
     - < address of cost-model service> # example: <service-name>.<namespace>:<port>
-``` 
+```
 
 ## Example queries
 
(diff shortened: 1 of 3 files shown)


AjayTripathy:
Hi @caarlos0 , thank you for this PR. We're currently doing it this way because the name of the label changed and in in a version of kube-state-metrics.

In general, I'm supportive of having prometheus do less work (no label replace, smaller vectors), although because these values are equal, it isn't changing the cardinality of these series. 

The additional work we'd have to do to support this would be making sure none of our users are using KSM  < 1.9, or place this behind a feature flag.

caarlos0:
> The additional work we'd have to do to support this would be making sure none of our users are using KSM < 1.9, or place this behind a feature flag.

wouldn't add a relabel config fix that? Or maybe query prometheus to check which label is available an choose appropriately? 🤔

dwbrown2:
@AjayTripathy let's discuss options here further

@caarlos0 In the meantime, I actually suspect these two PRs are going to have a much bigger impact on reducing your Thanos memory consumption:

https://github.com/kubecost/cost-analyzer-helm-chart/pull/536
https://github.com/kubecost/cost-model/pull/462

PR #458, opened 2020-06-29
Spot Refresh Updates

mbolt35:
* Added logging when a spot instance ID is added into the Spot price mapping
* Reduce refresh time from 1h to 15m

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -44,6 +44,9 @@ const supportedSpotFeedVersion = "1"
 const SpotInfoUpdateType = "spotinfo"
 const AthenaInfoUpdateType = "athenainfo"
 
+// How often spot data is refreshed
+const SpotRefreshDuration = 15 * time.Minute
+
 const defaultConfigPath = "/var/configs/"
 
 var awsRegions = []string{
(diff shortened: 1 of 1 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -1896,6 +1899,7 @@ func parseSpotData(bucket string, prefix string, projectID string, region string
 		klog.V(5).Infof("ListObjects \"s3://%s/%s\" produced no keys", *ls2.Bucket, *ls2.Prefix)
 	}
 
+	// TODO: Worth it to use LastModifiedDate to determine if we should reparse the spot data?
```
mbolt35, line 1902: We'd need to keep track of a LastModifiedDate in the AWS struct and check against that for each file in the set. We could avoid downloading and parsing the files in this case, which may let us be more aggressive in our refresh times. Either way, we're still subject to the delay from AWS -> S3

pkg/cloud/awsprovider.go
```diff
@@ -1980,7 +1984,7 @@ func parseSpotData(bucket string, prefix string, projectID string, region string
 				continue
 			}
 
-			klog.V(4).Infof("Found spot info %+v", spot)
+			klog.V(1).Infof("Found spot info for: %s", spot.InstanceID)
```
dwbrown2, line 1987: Is this log potentially noisy?

mbolt35: If it's only refreshing 15 minutes, not sure why it would be?

PR #462, opened 2020-07-06
Rate Limited Prometheus Client

mbolt35:
* Add a rate limiting prometheus client implementation
* Remove query context rate limiting
* Allow max concurrency to be passed via env var

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -24,6 +24,7 @@ import (
 	"github.com/kubecost/cost-model/pkg/clustercache"
 	cm "github.com/kubecost/cost-model/pkg/clustermanager"
 	"github.com/kubecost/cost-model/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/costmodel/router.go
```diff
@@ -163,6 +165,22 @@ func normalizeTimeParam(param string) (string, error) {
...
+// Parses the max query concurrency environment variable
+func maxQueryConcurrency() int {
+	v := os.Getenv(maxQueryConcurrencyEnvVar)
+	if v == "" {
+		return 5
+	}
+
+	result, err := strconv.Atoi(v)
+	if err != nil {
+		klog.Infof("[Warning] Failed to parse MAX_QUERY_CONCURRENCY. Defaulting to 5 - %s", err)
```
nikovacevic, line 177: Just a reminder that we have `log.Warningf(...)` for this stuff. A nit, but it standardizes the labeling conventions.

PR #471, opened 2020-07-14
[WIP] initial proposal for log limiting

AjayTripathy:
@nikovacevic @mbolt35 initial proposal for limiting logs. I think we'd want this logic to be applied to Errors, Warnings, and Info but not debug or profile. What do you think?

### DIFF
--- a/pkg/cloud/csvprovider.go
+++ b/pkg/cloud/csvprovider.go
@@ -139,7 +139,7 @@ func (c *CSVProvider) DownloadPricingData() error {
 		c.Pricing = pricing
 		c.PricingPV = pvpricing
 	} else {
-		log.DedupedWarningf("No data received from csv at %s", 5, c.CSVLocation)
+		log.DedupedWarningf(5, "No data received from csv at %s", c.CSVLocation)
 	}
 	time.AfterFunc(refreshMinutes*time.Minute, func() { c.DownloadPricingData() })
 	return nil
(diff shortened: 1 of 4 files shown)


pkg/cloud/csvprovider.go
```diff
@@ -138,7 +139,7 @@ func (c *CSVProvider) DownloadPricingData() error {
 		c.Pricing = pricing
 		c.PricingPV = pvpricing
 	} else {
-		klog.Infof("[WARNING] No data received from csv")
+		log.DedupedWarningf("No data received from csv at %s", 5, c.CSVLocation)
```
nikovacevic, line 142: This syntax is confusing because it looks like you're trying to template in `5` as a `%d`. Can we change the order of the params so that that count comes first?
```go
log.DedupedWarningf(5, "No data received from csv at %s", c.CSVLocation)
```

pkg/log/log.go
```diff
@@ -7,18 +7,68 @@ import (
 	"k8s.io/klog"
 )
 
+var seen = make(map[string]int)
+
 func Errorf(format string, a ...interface{}) {
 	klog.Errorf(fmt.Sprintf("[Error] %s", format), a...)
 }
 
+func DedupedErrorf(format string, logTypeLimit int, a ...interface{}) {
```
nikovacevic, line 16: RE the previous comment, this would become:
```go
func DedupedErrorf(logTypeLimit int, format string, a ...interface{}) {
```

PR #477, opened 2020-07-20
Environment Variable Management API + Thanos Offset

mbolt35:
* Add more convenient environment variable API and convention to add new environment variables. 
* Add a `thanos` package which contains specific configuration data related to thanos.

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -47,12 +47,6 @@ const (
 	epCleanTombstones = apiPrefix + "/admin/tsdb/clean_tombstones"
 	epConfig          = apiPrefix + "/status/config"
 	epFlags           = apiPrefix + "/status/flags"
-
-	clusterIDKey   = "CLUSTER_ID"
-	remoteEnabled  = "REMOTE_WRITE_ENABLED"
-	thanosEnabled  = "THANOS_ENABLED"
-	thanosQueryUrl = "THANOS_QUERY_URL"
-	thanosOffset   = "THANOS_QUERY_OFFSET"
 )
 
 type CostModel struct {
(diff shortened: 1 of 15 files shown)


pkg/costmodel/costmodel.go
```diff
@@ -51,6 +52,7 @@ const (
 	remoteEnabled  = "REMOTE_WRITE_ENABLED"
 	thanosEnabled  = "THANOS_ENABLED"
 	thanosQueryUrl = "THANOS_QUERY_URL"
+	thanosOffset   = "THANOS_QUERY_OFFSET"
```
nikovacevic, line 55: Is this being used?

pkg/env/costmodelenv.go
```diff
@@ -0,0 +1,163 @@
...
+
+// GetConfigPath returns the environment variable value for ConfigPathEnvVar which represents the cost
+// model configuration path
+func GetConfigPath() string {
+	return Get(ConfigPathEnvVar, "")
+}
+
+// GetConfigPath returns the environment variable value for ConfigPathEnvVar which represents the cost
+// model configuration path
+func GetConfigPathWithDefault(defaultValue string) string {
```
nikovacevic, line 111: Do you think we could get away with one function called `GetConfigPath()` that defaulted to some constant in the package whose value is `"/models/"`? The reasoning being that the only places where `GetConfigPath()` is currently used are commented-out, and everywhere this function is used, the same default is given.

(I don't feel strongly about this, just playing devil's advocate.)

pkg/costmodel/costmodel.go
mbolt35: Ah, crap - looks like all of these got added back.

pkg/env/costmodelenv.go
mbolt35: You're probably right, but was just trying to set a convention for any env vars in the future that would need a reasonable default passed in.

PR #478, opened 2020-07-21
refresh pricing on new node types without pod restart

AjayTripathy:
Tested by:
* Add a new type of node
* Note these warnings
I0720 21:29:12.634973       1 costmodel.go:1143] [Warning] Error getting node pricing. Error: Warning: no pricing data found for &{map[beta.kubernetes.io/arch:amd64 beta.kubernetes.io/fluentd-ds-ready:true beta.kubernetes.io/instance-type:n1-standard-1 beta.kubernetes.io/os:linux cloud.google.com/gke-accelerator:nvidia-tesla-k80 cloud.google.com/gke-nodepool:mpool cloud.google.com/gke-os-distribution:cos failure-domain.beta.kubernetes.io/region:us-central1 failure-domain.beta.kubernetes.io/zone:us-central1-a kubernetes.io/arch:amd64 kubernetes.io/hostname:gke-standard-cluster-1-mpool-b754c721-fhfm kubernetes.io/os:linux providerID:gce://guestbook-227502/us-central1-a/gke-standard-cluster-1-mpool-b754c721-fhfm]}
*Delete new type of node
* Add new build
* Add new node
* Note no warnings

### DIFF
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -1353,30 +1353,37 @@ func (gcp *GCP) AllNodePricing() (interface{}, error) {
 	return gcp.Pricing, nil
 }
 
+func (gcp *GCP) getPricing(key Key) (*GCPPricing, bool) {
+	gcp.DownloadPricingDataLock.RLock()
+	defer gcp.DownloadPricingDataLock.RUnlock()
+	n, ok := gcp.Pricing[key.Features()]
+	return n, ok
+}
+func (gcp *GCP) isValidPricingKey(key Key) bool {
+	gcp.DownloadPricingDataLock.RLock()
+	defer gcp.DownloadPricingDataLock.RUnlock()
+	_, ok := gcp.ValidPricingKeys[key.Features()]
+	return ok
+}
+
 // NodePricing returns GCP pricing data for a single node
 func (gcp *GCP) NodePricing(key Key) (*Node, error) {
-	gcp.DownloadPricingDataLock.RLock()
-	if n, ok := gcp.Pricing[key.Features()]; ok {
+	if n, ok := gcp.getPricing(key); ok {
 		klog.V(4).Infof("Returning pricing for node %s: %+v from SKU %s", key, n.Node, n.Name)
 		n.Node.BaseCPUPrice = gcp.BaseCPUPrice
-		gcp.DownloadPricingDataLock.RUnlock()
 		return n.Node, nil
-	} else if _, ok := gcp.ValidPricingKeys[key.Features()]; ok {
-		gcp.DownloadPricingDataLock.RUnlock()
+	} else if ok := gcp.isValidPricingKey(key); ok {
 		err := gcp.DownloadPricingData()
 		if err != nil {
 			return nil, fmt.Errorf("Download pricing data failed: %s", err.Error())
 		}
-		if n, ok := gcp.Pricing[key.Features()]; ok {
+		if n, ok := gcp.getPricing(key); ok {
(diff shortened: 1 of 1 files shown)


pkg/cloud/gcpprovider.go
```diff
@@ -1349,12 +1356,27 @@ func (gcp *GCP) AllNodePricing() (interface{}, error) {
...
 	gcp.DownloadPricingDataLock.RLock()
-	defer gcp.DownloadPricingDataLock.RUnlock()
 	if n, ok := gcp.Pricing[key.Features()]; ok {
 		klog.V(4).Infof("Returning pricing for node %s: %+v from SKU %s", key, n.Node, n.Name)
 		n.Node.BaseCPUPrice = gcp.BaseCPUPrice
+		gcp.DownloadPricingDataLock.RUnlock()
 		return n.Node, nil
+	} else if _, ok := gcp.ValidPricingKeys[key.Features()]; ok {
+		gcp.DownloadPricingDataLock.RUnlock()
+		err := gcp.DownloadPricingData()
```
mbolt35, line 1366: This is the biggest problem with go RWMutexes is there's not API support to upgrade the lock from read to readwrite, so this "gap" where we unlock the read and then lock read/write leaves room for something else to potentially take the lock before executing `DownloadPricingData()`. 

Given the circumstances and likely low contention on that specific lock, this bit of code is likely ok, as I don't think that anything stealing the lock between these calls will create any unhealthy scenarios in `DownloadPricingData()`.

pkg/cloud/gcpprovider.go
```diff
@@ -1349,12 +1356,27 @@ func (gcp *GCP) AllNodePricing() (interface{}, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("Download pricing data failed: %s", err.Error())
+		}
+		if n, ok := gcp.Pricing[key.Features()]; ok {
+			klog.V(4).Infof("Returning pricing for node %s: %+v from SKU %s", key, n.Node, n.Name)
+			n.Node.BaseCPUPrice = gcp.BaseCPUPrice
+			return n.Node, nil
+		}
```
mbolt35, line 1374: This code should be lock protected by a read lock, but since you already have acquired a write lock, it'd be optimal to just hold onto it beyond the `DownloadPricingData()`. 

I think maybe what you should here is actually split out the non-lock code from `DownloadPricingData()` into a new method called `downloadPricing()` or something similar. Have `DownloadPricingData()` perform the locking and call `downloadPricing()`. Then above these lines, instead of calling `DownloadPricingData()`, acquire the readwrite lock via `gcp.DownloadPricingLock.Lock()`, call the `downloadPricing()` function, then call `gcp.DownloadPricingLock.Unlock()` before the `return` in each of these conditionals.

pkg/cloud/gcpprovider.go
```diff
@@ -1349,12 +1356,27 @@ func (gcp *GCP) AllNodePricing() (interface{}, error) {
...
+		klog.V(1).Infof("[Warning] no pricing data found for %s: %s", key.Features(), key)
+		gcp.DownloadPricingDataLock.RUnlock()
```
mbolt35, line 1376: This seems unnecessary maybe. At this point, we would've already release the read lock. However, if you do something like the suggestion above, this line would need to be `gcp.DownloadPricingDataLock.Unlock()`

The other possibility for ALL of this code is create a lock-less `downloadPricing()` defined in the way described above, and then just acquire the read/write lock and defer it at the top of the method. It doesn't leverage the most optimal RW locking, but the constant lock switching probably creates more performance impact than just a single read/write lock on the entire method.

AjayTripathy:
@mbolt35 refactored to write two threadsafe helpers instead-- what do you think about this?

PR #486, opened 2020-07-29
Add ProviderID to node metrics

nikovacevic:
Required for https://github.com/kubecost/kubecost-cost-model/pull/184

## Changes
- Adds `provider_id` to four node metrics: 
  - node_cpu_hourly_cost
  - node_ram_hourly_cost
  - node_gpu_hourly_cost
  - node_total_hourly_cost

## Testing
Manually tested on GCP querying `node_total_hourly_cost`
```
node_total_hourly_cost{instance="gke-niko-n1-standard-2-wljla-8df8e58a-hfy7",instance_type="n1-standard-2",job="kubecost",node="gke-niko-n1-standard-2-wljla-8df8e58a-hfy7",provider_id="gce://guestbook-227502/us-central1-a/gke-niko-n1-standard-2-wljla-8df8e58a-hfy7",region="us-central1"} | 0.0941597579689026
```

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -1158,6 +1158,7 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 		if newCnode.Region == "" {
 			newCnode.Region = n.Labels[v1.LabelZoneRegion]
 		}
+		newCnode.ProviderID = n.Spec.ProviderID
 
 		var cpu float64
 		if newCnode.VCPU == "" {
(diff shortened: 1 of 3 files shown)


pkg/costmodel/costmodel.go
```diff
@@ -1158,6 +1158,7 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 		if newCnode.Region == "" {
 			newCnode.Region = n.Labels[v1.LabelZoneRegion]
 		}
+		newCnode.ProviderID = n.Spec.ProviderID
```
dwbrown2, line 1161: On-prem clusters, and other custom provisioners, may not have this field set. Just want to make sure that scenario wouldn't cause problems for you here.

nikovacevic: Shouldn't be, that field would just be an empty string, which seems appropriate.

AjayTripathy:
LGTM!

PR #491, opened 2020-08-02
Ajay tripathy aws keyinfo

AjayTripathy:


AjayTripathy:
On AWS with a few checks: 
Screen Shot 2020-08-03 at 1 26 04 PM

On GCP, no checks implemented: 
Screen Shot 2020-08-03 at 1 24 53 PM

pkg/costmodel/costmodel.go
```diff
@@ -1136,7 +1136,7 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 
 		cnode, err := cp.NodePricing(cp.GetKey(nodeLabels, n))
 		if err != nil {
-			log.DedupedWarningf(10, "Error getting node pricing. Error: %s", err.Error())
+			klog.Infof("Error getting node pricing. Error: %s", err.Error())
```
nikovacevic, line 1139: Was this for testing, or do we want to un-dedupe this log in production?

AjayTripathy: I think we will want to undedupe this in prod, unless there are major concerns. node pricing issues have been our top support questions so far.

PR #494, opened 2020-08-06
Promo instances and more disks for Azure

Subreptivus:


dwbrown2:
Thanks for the great PR, Yurii. Were you able to test all of these new cases or are there specific areas that we can help with? Thanks again!

Subreptivus:
@dwbrown2 Those cases were my solely purpose of changing the code. Yes, I've tested all these new cases.
Among others I'm using `NC6_Promo` GPU instances in Azures `Europe North` region and all of my clusters are using Azure-File StorageClasses for PVCs (unfortunately there is no easy way of getting their costs, so I've decided to set the price to "0" for better unification).
And all the Disks cases (HDD, SSD, SSD Premium) have been tested on real environment as well. I haven't added support of Ultra disks, because I'm not going to use them and have no time to really test it.

dwbrown2:
Super helpful, thank you again!

AjayTripathy:
@Subreptivus thanks so much for the PR-- sorry it took so long to review and test. Merging now!

PR #497, opened 2020-08-07
add cluster management costs for GKE/EKS

AjayTripathy:
Adds cluster management costs to EKS/GKE. TODO is to hook this up to the assets pipeline and add AKS costs.

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -73,6 +73,7 @@ type Accesses struct {
 	CPUAllocationRecorder         *prometheus.GaugeVec
 	GPUAllocationRecorder         *prometheus.GaugeVec
 	PVAllocationRecorder          *prometheus.GaugeVec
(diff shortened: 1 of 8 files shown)


pkg/costmodel/router.go
```diff
@@ -1129,6 +1136,10 @@ func Initialize(additionalConfigWatchers ...ConfigWatchers) {
 		Name: "kubecost_network_internet_egress_cost",
 		Help: "kubecost_network_internet_egress_cost Total cost per GB of internet egress.",
 	})
+	ClusterManagementCostRecorder := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "kubecost_cluster_management_cost",
```
nikovacevic, line 1140: Might be nice to call this `kubecost_cluster_management_hourly_cost` to be more explicit, and to match the `node_cpu_hourly_cost`, etc.

PR #506, opened 2020-08-11
Cost Model Metric Recording Fix and Refactor

mbolt35:
* Fix issue where `seen` maps were resetting the deletion status creating a false error log.
* Move metric emit loop to `metrics.go`, add Start/Stop methods
* Organize metrics.go by Collector/Metric implementations. 

Note: It would probably be better if we moved the `metrics.go` source to a new package (enables separation of metric emission loop code from cost-model source); however, this change does not include that move. Possibly a next refactor step.


pkg/costmodel/metrics.go
```diff
@@ -228,11 +283,323 @@ func (s ServiceMetric) Write(m *dto.Metric) error {
...
+					a.PVAllocationRecorder.DeleteLabelValues(labels...)
+					delete(pvcSeen, labelString)
+				} else {
+					pvcSeen[labelString] = false
+				}
+			}
+
+			select {
+			case <-time.After(time.Minute):
+			case <-recordingStop:
```
nikovacevic, line 561: Does the `close(recordingStop)` somehow send something on this channel? I'm confused about how the stop mechanism works. This `select` for combo timer/escape mechanism is cool.

mbolt35: `close(recordingStop)` will signal anything currently receiving the channel and anything that will receive the channel in the future (ie: it auto resolves). This is especially nice if you you have 1 channel and many consumers and you need all consumer goroutines to "complete" at once.

mbolt35: I believe that receiving on a channel actually returns the `T` value as well as `bool` for whether or not the receive was successful or closed. 
```go
    ch := make(chan string)

    go func() {
        result, ok := <- ch
        if ok {
            fmt.Printf("Successful Receive: %s\n", result)
        } else {
            fmt.Println("Channel Closed")
        }
    }()
    time.Sleep(2 * time.Second)
    close(ch)
    time.Sleep(1 * time.Second)
```

I haven't ever used the `_, ok` on receive before but I think this is how `range` over a channel is able to tell when to stop iterating.

mbolt35: While I'm on this tangent, I recognized that it would probably be "better" to use a `chan struct{}` but honestly, this grosses me out.

mbolt35:
> Definitely support these changes. Curious how you tested the start/stop mechanism? Is there a use-case for that today, or just building it in advance of an application?

This is a pattern I've used a few times already (I'll send you a link to network-costs source in slack), but here's a similar use-case in cost-model:
[clustercache.go](https://github.com/kubecost/cost-model/blob/43fa0ed72cc0eb2797961c167c6cede9704831ac/pkg/clustercache/clustercache.go#L153-L160) which wires into [watchcontroller.go](https://github.com/kubecost/cost-model/blob/43fa0ed72cc0eb2797961c167c6cede9704831ac/pkg/clustercache/watchcontroller.go#L198)

Kubernetes `wait.Until` is basically a glorified `select` that uses `case <-stopCh` similar to the emission loop. 

 Admittedly, this case is a bit more complicated due to the potential of stopping while the emission loop is in-progress. For this scenario, we use that `recordingStopping bool` to denote that we've executed a `close()` on the channel to prevent calling `close()` again (panic). This allows the current iteration of the loop to complete before receiving the stop signal.

After a bit more inspection, I'm fairly sure that there is a race related to `select`. Since go doesn't allow prioritization on the `select` cases, I believe it's possible for the `stop` to be closed at the same time as a minute elapses. If go "chooses" the minute case, then you'd just execute an additional iteration similar to if you stopped mid emission loop.

You can "protect" against this behavior by adding something like this at the start of the loop, but I don't think it's really that much of an issue in this case:

```go
select {
case <-recordingStop:
    recordingLock.Lock()
    recordingStopping = false
    recordingStop = nil
    recordingLock.Unlock()
    return
default:
}
```

The main way I test the general signal flow is using a dummy app and writing in `time.Sleep(...)` to represent work. Here's my scratch pad for this particular setup:

```go
package main

import (
	"fmt"
	"sync"
	"time"
)

var (
	recordingLock     sync.Mutex
	recordingStop     chan bool
	recordingStopping bool
)

// Checks to see if there is a metric recording stop channel. If it exists, a new
// channel is not created and false is returned. If it doesn't exist, a new channel
// is created and true is returned.
func checkOrCreateRecordingChan() bool {
	recordingLock.Lock()
	defer recordingLock.Unlock()

	if recordingStop != nil {
		return false
	}

	recordingStop = make(chan bool, 1)
	return true
}

func run() bool {
	if !checkOrCreateRecordingChan() {
		fmt.Println("[Error]: Attempted to start cost model metric recording when it's already running.")
		return false
	}

	fmt.Println("STARTING")

	go func() {
		for {
			// Work
			time.Sleep(2 * time.Second)

			select {
			case <-time.After(10 * time.Second):
			case <-recordingStop:
				recordingLock.Lock()
				recordingStopping = false
				recordingStop = nil
				fmt.Println("STOPPED")
				recordingLock.Unlock()
				return
			}
		}
	}()

	return true
}

func stop() {
	recordingLock.Lock()
	defer recordingLock.Unlock()

	if !recordingStopping && recordingStop != nil {
		recordingStopping = true
		close(recordingStop)
	}
}

func main() {
	run()
	run()

	time.Sleep(5 * time.Second)
	stop()
	stop()
	stop()
	stop()

	for !run() {
		time.Sleep(20 * time.Millisecond)
	}

	time.Sleep(5 * time.Second)

	stop()
	stop()
	stop()

	time.Sleep(5 * time.Second)
}
```

PR #507, opened 2020-08-13
support running cost-model locally

almonteb:
expose the env var KUBECONFIG_PATH in order to run cost-model outside of a kubernetes pod.


almonteb:
@AjayTripathy thoughts on this?

AjayTripathy:
Hi @almonteb -- sorry for the late response. I'm happy to merge this-- just wanted to let you know that because we rely pretty heavily on prometheus, it's tough to run kubecost outside of its full environment. That being said, this definitely looks like a step in the right direction!

PR #510, opened 2020-08-18
Prevent nil *QueryResults error

nikovacevic:
## Changes
- Always return a non-nil QueryResults. Now that it includes error, we need to deprecate the extra error and eliminate the `return nil, err` idiom.

## Testing
- Injected random errors, noted that nothing panicked. Queries just retried as expected.

### DIFF
--- a/pkg/prom/result.go
+++ b/pkg/prom/result.go
@@ -63,32 +63,40 @@ type QueryResult struct {
 
 // NewQueryResults accepts the raw prometheus query result and returns an array of
 // QueryResult objects
-func NewQueryResults(query string, queryResult interface{}) (*QueryResults, error) {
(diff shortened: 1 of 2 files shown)


pkg/prom/result.go
```diff
@@ -166,7 +181,7 @@ func NewQueryResults(query string, queryResult interface{}) (*QueryResults, erro
 	return &QueryResults{
```
mbolt35, line 181: Return `qrs` here! 👍

nikovacevic: Oh yep

mbolt35: False alarm I think. Looks like you return qrs every error, although might be possible to just attach results to `qrs` and return.

PR #524, opened 2020-08-27
Prometheus Client Profiling

mbolt35:
* Add profiling queries to `query.go`
* Add request counts to `RateLimitedPrometheusClient` and a function to log the current request counts


dwbrown2:
Diig it.

PR #529, opened 2020-09-02
support for cost model currency conversion

itscalvinwang:
Tag: add currency conversion support to the cost model (work in progress)

Description: new code calls gcp.GetConfig() and then appends the CurrencyCode to the url, GET request on the url returns json body, after which `nanos` is automatically returned as the new units (e.g., nanos could be 100 in USD, but returns as 135 if `&currencyCode=CAD` is appended to the url), after which it is fed to the client (intended behavior)

Testing: built and pushed to Docker, ran by port forwarding docker url to local. queried Prometheus for `node_total_hourly_cost` to see no value changed upon selecting a different currency in the settings dropdown.

Problems: ran into a bug where json unmarshalling was failing because of PricingInfo struct `CurrencyConversionRate` variable typing, fixed it and now it's back to the original behavior (selecting a different currency in the settings dropdown and saving only changes the currency character in the client, but doesn't convert the numbers)

Questions:
1. What is the code path that calls `DownloadPricingData()` when the currency setting is changed?
2. Does the aforementioned function even need to be called again?
3. Are any of my assumptions in the Description incorrect? How does the back end send the price to the client?

### DIFF
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -166,6 +166,9 @@ func (gcp *GCP) GetConfig() (*CustomPricing, error) {
 	if c.NegotiatedDiscount == "" {
 		c.NegotiatedDiscount = "0%"
 	}
+	if c.CurrencyCode == "" {
+		c.CurrencyCode = "USD"
+	}
 	return c, nil
 }
 
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -575,7 +575,7 @@ type GCPPricing struct {
 type PricingInfo struct {
 	Summary                string             `json:"summary"`
 	PricingExpression      *PricingExpression `json:"pricingExpression"`
-	CurrencyConversionRate int                `json:"currencyConversionRate"`
+	CurrencyConversionRate float64            `json:"currencyConversionRate"`
 	EffectiveTime          string             `json:""`
 }
 
@@ -874,7 +874,11 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]Key, pvKeys map[stri
 
 func (gcp *GCP) parsePages(inputKeys map[string]Key, pvKeys map[string]PVKey) (map[string]*GCPPricing, error) {
 	var pages []map[string]*GCPPricing
-	url := "https://cloudbilling.googleapis.com/v1/services/6F81-5844-456A/skus?key=" + gcp.APIKey
+	c, err := gcp.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	url := "https://cloudbilling.googleapis.com/v1/services/6F81-5844-456A/skus?key=" + gcp.APIKey + "&currencyCode=" + c.CurrencyCode
 	klog.V(2).Infof("Fetch GCP Billing Data from URL: %s", url)
 	var parsePagesHelper func(string) error
 	parsePagesHelper = func(pageToken string) error {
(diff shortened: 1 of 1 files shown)


pkg/cloud/gcpprovider.go
```diff
@@ -874,7 +874,11 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]Key, pvKeys map[stri
 
 func (gcp *GCP) parsePages(inputKeys map[string]Key, pvKeys map[string]PVKey) (map[string]*GCPPricing, error) {
 	var pages []map[string]*GCPPricing
-	url := "https://cloudbilling.googleapis.com/v1/services/6F81-5844-456A/skus?key=" + gcp.APIKey
+	c, err := gcp.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	url := "https://cloudbilling.googleapis.com/v1/services/6F81-5844-456A/skus?key=" + gcp.APIKey + "&currencyCode=" + c.CurrencyCode
```
AjayTripathy, line 881: If no currencyCode is supplied, we should explicitly default to USD

AjayTripathy (changes requested):

dwbrown2:
I think it's ok to no re download pricing data on this for v1. That should happen on pod restart and most users will be setting this from values.yaml instead of UI.

dwbrown2:
What happens if you pass CAD or EUR in values and then install/upgrade Kubecost? You can do that here:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml#L479

Does this update the prices that you receive?

AjayTripathy:
I don't think it's a big deal that we don't change currencies immediately when the dropdown changes. We can tell users to set it on install from our helm chart, under productConfigs:
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L405

As an introduction to helm, I'd suggest that you try playing with the currencyCode value in that block of helm, installing from a local values.yaml file, and noting that the price does change in prometheus when the new values.yaml is applied

itscalvinwang:
I've sent the currencyCost values from the values.yaml file, but it doesn't update the prices on the client, nor in Prometheus. Any ideas on how to fix this?

dwbrown2:
Have you looked at the API request that you are making and the data being returned? Does it match what you would expect to see, based on the currency being set?

dwbrown2:
Btw, you can see GCP prices for different currencies + assets here: https://cloud.google.com/skus/?currency=CAD&filter=Compute+Engine

itscalvinwang:
tested again in accordance to the comments via the following process

1. edit values.yaml
2. upgrade release
3. edit deployment to point to correct docker tag
4. port forward both cost-model and prometheus
5. dump logs to text file

the correct URL is queried on the currencyCode flag change, and logs show the correct VCPUCost change across different currencies. Prometheus also reflects the conversions in `node_total_hourly_cost`, so the back end is fetching the URL and getting these values. However, the client isn't receiving these values somehow.

Screen Shot 2020-09-03 at 12 47 39 PM

0.06166 / 0.04270 ≈ 1.44, which is the conversion of CAD/CHF

dwbrown2:
Nice!! Sounds like good validation to me... you likely just need to have ETL with this new data. We can test that part next week.

dwbrown2:
Still one minor comment from Ajay here: https://github.com/kubecost/cost-model/pull/529/files#r482313600

itscalvinwang:
Okay I'm quite certain that it's working actually. One thing to note was that I misunderstood the expected behavior; I thought the values would update and convert immediately (e.g., monthly total cost would go from CAD $10 to 15 CHF as soon as the setting is updated), but @nikovacevic (and the hover tooltip) revealed that it is a 24h average that slowly updates over the next 24h after the setting change.

PR #532, opened 2020-09-04
Ajay tripathy fix ooc masterpayer

AjayTripathy:
This query was not properly collecting the masterpayer credentials for athena CUR.


nikovacevic (approved):
LGTM

PR #535, opened 2020-09-08
fix e2 issues

AjayTripathy:
Running on 34.71.168.144

http://34.71.168.144/assets.html has the correct prices-- compare 1d * 30 of 23.7 vs 24.46 on https://cloud.google.com/compute/vm-instance-pricing#e2_predefined (50c difference due to memory accesible to k8s)

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -346,6 +346,10 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset 
 		nodeMap[key].CPUCost += cpuCost
 		nodeMap[key].NodeType = nodeType
 	}
+	partialCPUMap := make(map[string]float64)
+	partialCPUMap["e2-micro"] = 0.25
+	partialCPUMap["e2-small"] = 0.5
+	partialCPUMap["e2-medium"] = 1.0
 
 	for _, result := range resNodeCPUCores {
 		cluster, err := result.GetString("cluster_id")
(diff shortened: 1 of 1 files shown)


pkg/costmodel/cluster.go
```diff
@@ -346,6 +346,10 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset
 		nodeMap[key].CPUCost += cpuCost
 		nodeMap[key].NodeType = nodeType
 	}
+	partialCPUMap := make(map[string]float64)
+	partialCPUMap["e2-micro"] = 0.25
```
dwbrown2, line 350: comment might be helpful?

pkg/costmodel/cluster.go
```diff
@@ -346,6 +346,10 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset
...
+	partialCPUMap["e2-small"] = 0.5
+	partialCPUMap["e2-medium"] = 1.0
```
nikovacevic, line 352: @AjayTripathy looks like you decided against the f1- and g1- entries, so we're handling that above?

nikovacevic (approved):
If f1 and g1 are being handled correctly, then this looks good :+1:

pkg/costmodel/cluster.go
AjayTripathy: I believe we are on the price emission side. I don't want to blow up scope. e2 is very high priority given that new clusters are provisioned with e2. I have added a sprint task to test this fully with f1/g1

PR #543, opened 2020-09-15
Ajay tripathy pv reconciliation

AjayTripathy:
Reconciliation for PV and network costs
Screen Shot 2020-09-27 at 9 26 14 PM

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -2435,7 +2435,7 @@ func (aws *AWS) ParseID(id string) string {
 }
 
 func (aws *AWS) ParsePVID(id string) string {
-	rx := regexp.MustCompile("aws:/[^/]*/[^/]*/([^/]+)")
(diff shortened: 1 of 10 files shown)


AjayTripathy:
@nikovacevic let me know what you think. I don't think this is possible without introducing a "secondary" provider ID() to match cluster on, but not merge with assets.

nikovacevic:
@AjayTripathy we should discuss briefly today. Unless I'm missing something, I think we will not need two provider IDs. Can we just match by the pair (type, providerID), such that (type1 == type2 && providerID1 == providerID2) means a full match and (type1 != type2 && providerID1 == providerID2) means different aspects of the same resource (e.g. local disk or network costs for some node)? Also, I have to imagine we can do better than requiring two separate functions for parsing IDs (i.e. `ParseID` and `ParsePVID` should be combined).

pkg/cloud/awsprovider.go
```diff
@@ -2427,3 +2433,16 @@ func (aws *AWS) ParseID(id string) string {
 
 	return match[1]
 }
+
+func (aws *AWS) ParsePVID(id string) string {
+	rx := regexp.MustCompile("aws:/[^/]*/[^/]*/([^/]+)")
```
nikovacevic, line 2438: Can you provide and example provider ID here?

PR #545, opened 2020-09-16
Cw 407 load balancer costs

itscalvinwang:
Description: writing load balancer cost to prometheus
Issue: https://github.com/kubecost/cost-model/issues/407
Testing: build, push, and edit docker deployment

Steps & Goals in Feature Development (**step I'm currently on**):
1. create a new Prometheus metric called `load_balancer_cost` that detects the ELB on my cluster
2. _start dynamically fetching cloud provider data from GCP based on region_ (skip for now)
3. _extend cloud functionality to handle internal load balancers as well as other providers (Azure, AWS, etc.)_ (skip for now)
4. add metric to assets pipeline
5. **test, debug, release**

What I've Changed:
1. `router.go` -- added a LBCostRecorder as a gauge vector, called `prometheus.MustRegister()` on it, and added it to the `Accesses` struct
2. `provider.go` -- created a `LoadBalancer` struct to store dynamically fetched cloud provider pricing rates, as well as the cost of the load balancer (mimicking the `Node` struct before it)
3. `costmodel.go` -- added a `getLBCost()` function that identifies which services are load balancers, and subsequently fetches and populates a `LoadBalancer` struct with cloud pricing info (currently hardcoded values, but will dynamically fetch after I get Step 1 working)
4. `metrics.go` -- `StartCostModelMetricRecording()` calls `getLBCost()` and loops through the returned map of `LoadBalancer` structs

Questions:
1. What else am I missing from my understanding of how Prometheus metrics are recorded?
2. Why is `getNodeCost()` called in both `metrics.go:StartCostModelMetricRecording()` as well as `costmodel.go:ComputeCostData()`, which `StartCostModelMetricRecording()` also calls again in the same for loop?
3. _See comments on the diff._

Feel free to point out anything that warrants discussion or correct any assumptions or misunderstandings that I have regarding the code.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -559,6 +559,7 @@ k8s.io/apimachinery v0.0.0-20190913075812-e119e5e154b6 h1:tGU1C/vMoUV2ZakSH6wQq2
 k8s.io/apimachinery v0.0.0-20190913075812-e119e5e154b6/go.mod h1:nL6pwRT8NgfF8TT68DBI8uEePRt89cSvoXUVqbkWHq4=
 k8s.io/apimachinery v0.18.6 h1:RtFHnfGNfd1N0LeSrKCUznz5xtUP1elRGvHJbL3Ntag=
 k8s.io/apimachinery v0.19.0 h1:gjKnAda/HZp5k4xQYjL0K/Yb66IvNqjthCb03QlKpaQ=
+k8s.io/apimachinery v0.19.1 h1:cwsxZazM/LA9aUsBaL4bRS5ygoM6bYp8dFk22DSYQa4=
 k8s.io/client-go v0.0.0-20190404172613-2e1a3ed22ac5 h1:BwY2C//EoWktJi74O6R2REBonrhsfhRI0qfVwOjOPp8=
 k8s.io/client-go v0.0.0-20190404172613-2e1a3ed22ac5/go.mod h1:bIEHXHbykaOlj+pgLllzLJ2RPGdzkjtqdk0Il07KPEM=
 k8s.io/client-go v0.0.0-20190620085101-78d2af792bab h1:E8Fecph0qbNsAbijJJQryKu4Oi9QTp5cVpjTE+nqg6g=
--- a/go.sum
+++ b/go.sum
@@ -560,6 +560,7 @@ k8s.io/apimachinery v0.0.0-20190913075812-e119e5e154b6/go.mod h1:nL6pwRT8NgfF8TT
 k8s.io/apimachinery v0.18.6 h1:RtFHnfGNfd1N0LeSrKCUznz5xtUP1elRGvHJbL3Ntag=
 k8s.io/apimachinery v0.19.0 h1:gjKnAda/HZp5k4xQYjL0K/Yb66IvNqjthCb03QlKpaQ=
 k8s.io/apimachinery v0.19.1 h1:cwsxZazM/LA9aUsBaL4bRS5ygoM6bYp8dFk22DSYQa4=
+k8s.io/apimachinery v0.19.2 h1:5Gy9vQpAGTKHPVOh5c4plE274X8D/6cuEiTO2zve7tc=
 k8s.io/client-go v0.0.0-20190404172613-2e1a3ed22ac5 h1:BwY2C//EoWktJi74O6R2REBonrhsfhRI0qfVwOjOPp8=
 k8s.io/client-go v0.0.0-20190404172613-2e1a3ed22ac5/go.mod h1:bIEHXHbykaOlj+pgLllzLJ2RPGdzkjtqdk0Il07KPEM=
 k8s.io/client-go v0.0.0-20190620085101-78d2af792bab h1:E8Fecph0qbNsAbijJJQryKu4Oi9QTp5cVpjTE+nqg6g=
--- a/pkg/cloud/provider.go
+++ b/pkg/cloud/provider.go
@@ -73,14 +73,14 @@ func (n *Node) IsSpot() bool {
 // LoadBalancer is the interface by which the provider and cost model communicate LoadBalancer prices.
 // The provider will best-effort try to fill out this struct.
 type LoadBalancer struct {
-	IngressIPAddresses           []string `json:"IngressIPAddresses"`
-	Cost                         float64  `json:"hourlyCost"` // TODO: find out if cloud providers return these values as strings or floats
-	FirstFiveForwardingRulesCost float64  `json:"firstFiveForwardingRulesCost"`
-	AdditionalForwardingRuleCost float64  `json:"additionalForwardingRuleCost"`
-	IngressDataCostPerGB         float64  `json:"ingressDataCostPerGB"`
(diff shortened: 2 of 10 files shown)


go.sum
```diff
@@ -559,6 +559,7 @@ k8s.io/apimachinery v0.0.0-20190913075812-e119e5e154b6 h1:tGU1C/vMoUV2ZakSH6wQq2
 k8s.io/apimachinery v0.0.0-20190913075812-e119e5e154b6/go.mod h1:nL6pwRT8NgfF8TT68DBI8uEePRt89cSvoXUVqbkWHq4=
 k8s.io/apimachinery v0.18.6 h1:RtFHnfGNfd1N0LeSrKCUznz5xtUP1elRGvHJbL3Ntag=
 k8s.io/apimachinery v0.19.0 h1:gjKnAda/HZp5k4xQYjL0K/Yb66IvNqjthCb03QlKpaQ=
+k8s.io/apimachinery v0.19.1 h1:cwsxZazM/LA9aUsBaL4bRS5ygoM6bYp8dFk22DSYQa4=
```
itscalvinwang, line 562: not really sure what this does

pkg/costmodel/metrics.go
```diff
@@ -511,6 +512,18 @@ func StartCostModelMetricRecording(a *Accesses) bool {
...
+				// TODO: parse (if necessary) and calculate cost associated with loadBalancer based on dynamic cloud prices fetched into each lb struct on GetLBCost() call
+				keyParts := getLabelStringsFromKey(lbKey)
+				namespace := keyParts[0]
+				serviceName := keyParts[1]
+				a.LBCostRecorder.WithLabelValues(namespace, serviceName).Set(lb.Cost)
+
+				labelKey := getKeyFromLabelStrings(namespace, serviceName)
+				loadBalancerSeen[labelKey] = true
+			}
+
```
itscalvinwang, line 526: I just dumb copied this without understanding its purpose because `nodes` and `pvs` also do it.

I assume it's to prevent gauge vector double counting?

pkg/costmodel/metrics.go
```diff
@@ -623,6 +636,14 @@ func StartCostModelMetricRecording(a *Accesses) bool {
...
 				}
 			}
+			for labelString, seen := range loadBalancerSeen {
+				if !seen {
+					labels := getLabelStringsFromKey(labelString)
+					a.LBCostRecorder.DeleteLabelValues(labels...)
+				} else {
+					loadBalancerSeen[labelString] = false
+				}
+			}
```
itscalvinwang, line 646: I just dumb copied this without understanding its purpose because `nodes` and `pvs` also do it.

I assume it's to prevent gauge vector double counting?

pkg/costmodel/router.go
```diff
@@ -887,6 +888,10 @@ func Initialize(additionalConfigWatchers ...ConfigWatchers) {
 		Name: "kubecost_cluster_management_cost",
 		Help: "kubecost_cluster_management_cost Hourly cost paid as a cluster management fee.",
 	}, []string{"provisioner_name"})
+	LBCostRecorder := prometheus.NewGaugeVec(prometheus.GaugeOpts{ // don't know if necessary to differentiate ELB vs. ALB cost
+		Name: "load_balancer_cost",
```
dwbrown2, line 892: Nice PR!

We didn't prefix any of our original metrics with `kubecost_` but we've started doing with recording rules, etc. What do you guys think we should do here? @itscalvinwang @mbolt35 @AjayTripathy @nikovacevic

dwbrown2 (approved):

pkg/costmodel/metrics.go
itscalvinwang, line 526: I haven't made any changes to `ComputeCostData()`. Do I need to?

It looks to me that `metrics.go:StartCostModelMetricRecording()` calls `getNodeCost()` to fetch cloud provider data, but actually does the total cost calculation in `StartCostModelMetricRecording()` (`metrics.go:500`). Is this the pattern we want to repeat going forward?

AjayTripathy: ComputeCostData is building the allocation model, not the assets. Since the goal of this PR is to build up LoadBalancer assets in prometheus and not necessarily to perfectly allocate LB costs, this seems fine.

pkg/costmodel/costmodel.go
```diff
@@ -1183,6 +1190,48 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
...
+				AdditionalForwardingRuleCost: 0.010,
+				IngressDataCostPerGB:         0.008,
+			}
+			newLoadBalancer := *loadBalancer
+			if len(service.Status.LoadBalancer.Ingress) > 0 { // should actually check if LoadBalancer.Ingress exists
+				for _, loadBalancerIngress := range service.Status.LoadBalancer.Ingress {
+					newLoadBalancer.IngressIPAddresses = append(newLoadBalancer.IngressIPAddresses, loadBalancerIngress.IP)
+				}
+			}
+			newLoadBalancer.Cost = 88.88
```
AjayTripathy, line 1228: Flat costs should be made hourly (730 hours per month)

pkg/cloud/provider.go
```diff
@@ -70,6 +70,18 @@ func (n *Node) IsSpot() bool {
 	}
 }
 
+// LoadBalancer is the interface by which the provider and cost model communicate LoadBalancer prices.
+// The provider will best-effort try to fill out this struct.
+type LoadBalancer struct {
+	IngressIPAddresses           []string `json:"IngressIPAddresses"`
```
AjayTripathy, line 76: The interface should probably be a bit more general, with "cost" representing an hourly cost that takes into account the specifics of GCP loadbalancers. A function for computing cost probably needs to be provider specific, ie in gcpprovider.go. So in provider.go you may want a constructor-like function LoadBalancer) that takes in a service and returns a LoadBalancer with a "Cost" field calculated.

pkg/costmodel/costmodel.go
```diff
@@ -1183,6 +1190,48 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
...
+	for _, service := range servicesList {
+		namespace := service.GetObjectMeta().GetNamespace()
+		name := service.GetObjectMeta().GetName()
+		// TODO: add clusterID to key? possible if called in ComputeCostData(), but what about metrics.go?
+		key := namespace + "," + name
+
+		// Does this identify ELB vs. ILB? Need to test the /api/allServices call with an ALB. Current work is on ELBs.
+		if service.Spec.Type == "LoadBalancer" {
+			// TODO: dynamically fetch based on cloud provider and region. Currently using hard-coded GCP us-central1 values.
+			loadBalancer := &costAnalyzerCloud.LoadBalancer{
```
AjayTripathy, line 1217: See above. I suggest you want a provider.LoadBalancer() that's a bit more generic than doing this here.

AjayTripathy (changes requested):

PR #546, opened 2020-09-16
Getter/Setter/Mapper Abstraction

mbolt35:
* Abstract Get/Set logic from the `env` package into `util/mapper` package that can leverage a `Map` interface.
* Add EnvVar implementation and move current `env` package to use new mapper. 
* Add QueryParams implementation.


nikovacevic (approved):
I dig LGTM :metal:

PR #553, opened 2020-09-24
FIFO Concurrency Policy for Prometheus Queries

mbolt35:
Replace unfair semaphore throttling with fair blocking queue implementation for prometheus queries.


AjayTripathy:
LGTM!

PR #554, opened 2020-09-25
ClusterDisks: fix missing initialization

nikovacevic:
## Changes
- Initialize a `Breakdown` instance in the case where it is missing (the other three cases are covered)


caarlos0 (approved):

AjayTripathy:
LGTM!

PR #558, opened 2020-09-30
Feature/cost data model range metadata

ameyajoshi99:
Added metadata for Nodes & PVC in costmodeldatarange api

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -1895,6 +1895,8 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 
 	profileStart = time.Now()
 
+	nodes, _ := cm.GetNodeCost(cp)
+
 	missingNodes := make(map[string]*costAnalyzerCloud.Node)
 	missingContainers := make(map[string]*CostData)
 	for key := range containers {
@@ -1939,10 +1941,10 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 		}
 
 		var node *costAnalyzerCloud.Node
-		if n, ok := missingNodes[c.NodeName]; ok {
-			node = n
-		} else {
-			node = &costAnalyzerCloud.Node{}
+		if node, ok = missingNodes[c.NodeName]; !ok {
+			if node, ok = nodes[c.NodeName]; !ok {
+				node = &costAnalyzerCloud.Node{}
+			}
 			missingNodes[c.NodeName] = node
 		}
 
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -1720,6 +1720,17 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 	if pvAllocationMapping != nil {
 		addMetricPVData(pvAllocationMapping, pvCostMapping, cp)
 		for k, v := range pvAllocationMapping {
+			for _,pvcData := range v {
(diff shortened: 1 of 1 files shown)


dwbrown2:
Thanks for this contribution, @ameyajoshi99!

@mbolt35 would you be able to take review?

ameyajoshi99:
@dwbrown2 / @mbolt35 Did you get chance to look at this?

mbolt35:
@ameyajoshi99 Trying to understand the motivation behind this change. If you could provide some more context, that would be great.

pkg/costmodel/costmodel.go
```diff
@@ -1895,6 +1906,8 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, clientset kubern
 
 	profileStart = time.Now()
 
+	nodes, _ := cm.GetNodeCost(cp)
```
mbolt35, line 1909: This could create bugs due to what costDataRange queries. The context of these queries is over a specific time range and can include multiple clusters. GetNodeCost() only returns data concerning the current set of nodes from kubernetes. This is fine for `ComputeCostData` which operates in scope of the local cluster, but it would definitely cause issues if we added this to our range queries over multiple clusters

ameyajoshi99:
> stand the motivation behind this change. If you could provide some more context, that would be

The costdata api is returning nodes metadata such as provider Id region. But the Range api doesnot. So all we want is add metadata of the node in range api.

pkg/costmodel/costmodel.go
ameyajoshi99: As in the other comment I mentioned, we are looking for the metadata of the node. GetNodeCost is handling kubernates api calls right? So I didn't interfere with the missing node logic. I think that is where costs are getting added right?

mbolt35: The reason the `costDataRange` doesn't use `GetNodeCost` data is because we are technically unable to attach node data (consistently) for historical records. The results of `costDataRange` are historical, and the results from `GetNodeCost()` are current. If we applied metadata for the _current_ state, then it's possible that it conflicts with the state N hours/days ago.  We can't just simply apply current state to all the historical records for that node. Does that make sense?

In your specific case, your nodes aren't changing, so I totally understand why this seems feasible. However, the way we're modeling range data wouldn't adequately allow us to show changes over time to metadata.

To correctly implement what you're asking for, we'd need to determine a consistent way of modeling changes over time for metadata. 
* Find this data in KSM, or write it into `metrics.go`
* Validate that the scale of this data is manageable
* Add queries that fetch this data historically  

>  I think that is where costs are getting added right?

You are correct that for the local cluster, that is where the costs are calculated for a Node, but it is the _current_ cost, and it does not include costs of nodes from other clusters. `ComputeCostData` is mainly driven by our metric emission loop which emits metrics on the current state of the local cluster. This is why we're able to use the kubernetes API to lookup and attach the metadata. `ComputeCostDataRange` is designed to pull historical/summary records that includes (potentially) multicluster data, and which is not available from the kubernetes API. 

Hopefully this makes sense, we can certainly do a better job with documentation and code clarity with regards to this, so thank you for the feedback.

ameyajoshi99: @mbolt35 Thanks a lot for detailed explanation. This does makes sense.

michaelmdresser:
I believe this PR is obsolete due to changes in Kubecost's core APIs since this PR was opened. Check out our [official documentation](https://guide.kubecost.com/hc/en-us) to learn more. If you still don't have the functionality desired, please feel free to open a new issue or PR!

PR #564, opened 2020-10-08
Cluster Info Map and Prometheus/Thanos Client API Rework

mbolt35:
# Changes

### Prometheus/Thanos Client Identification
The `prom` package now has the ability to identify the specific `prometheus.Client` via:
```go
var promCli prometheus.Client = // ... 

prom.IsPrometheus(promCli) // true
prom.IsThanos(promCli) // false
```
This feature is superficial and takes advantage of a string identifier that is set during client creation. It doesn't do any expensive examination of the client outside of the ID check. 

### Prometheus/Thanos Client "Factory" Methods
In order to better identify the client, the instantiation of the prom and thanos clients were moved to their own respective packages:
```go
prom.NewPrometheusClient(address string, timeout, keepAlive time.Duration, queryConcurrency int, queryLogFile string) (prometheus.Client, error)

thanos.NewThanosClient(address string, timeout, keepAlive time.Duration, queryConcurrency int, queryLogFile string) (prometheus.Client, error)
```
These specific methods streamlined the client authentication mechanisms a bit, and abstracted away the noisy bootstrapping happening in `router.go`

### Prometheus/Thanos Validation
The `costmodel.ValidatePrometheus` method was moved from `costmodel.go` to `validate.go` in the `prom` package. It was also renamed to `Validate` and no longer requires the `isThanos bool` to be set due to identification changes mentioned above:
```go
var (
	prometheusValidateQuery string = "up"
	thanosValidateQuery     string = fmt.Sprintf("up offset %s", env.GetThanosOffset())
)

func Validate(cli prometheus.Client) (*PrometheusMetadata, error) {
	if IsThanos(cli) {
		return validate(cli, thanosValidateQuery)
	}

	return validate(cli, prometheusValidateQuery)
}
```
### ClusterMap
Added a new interface `ClusterMap` in the `pkg/costmodel/clusters` package. The default implementation accepts a `prometheus.Client` implementation, which it uses to self-populate with the `kubecost_cluster_info` metric. For single clusters, this means passing a client targeting Prometheus. For multicluster, this means passing a client targeting Thanos. 

The refresh interval is also configurable at instantiation: 
```go
// Initialize ClusterMap for maintaining ClusterInfo by ClusterID
var clusterMap clusters.ClusterMap
if thanosClient != nil {
    clusterMap = clusters.NewClusterMap(thanosClient, 10*time.Minute)
} else {
    clusterMap = clusters.NewClusterMap(promCli, 5*time.Minute)
}
```

There is also an internal retry mechanism that will continually retry to load query results on failure. There is some randomized backoff that prevents high frequency retries, and a hard limit of 6 attempts:
```go
time.Sleep(delay)

// add some random backoff
jitter := time.Duration(rand.Int63n(int64(delay)))
delay = delay + jitter/2
```
The purpose of the retry is to try to ensure consistency with respect to clusterID -> clusterInfo mappings. Additionally, because we load this data at startup, it's possible prometheus and/or thanos won't immediately be available.

### ClusterID -> ClusterName/ClusterID Mapping
The reason that this changeset is classified as a WIP is because of the execution of this specific change. Originally, there was speculation that we could just update the clusterID to a `<clusterName>/<clusterID>` schema on the `CostData`. The current implementation does this, and also takes into account filtering such that a cluster filter will look at both the name and id. 

Everything seems to be working correctly with this change, but something doesn't quite feel right about it. We use the `cluster_id` label from Thanos metrics constantly as keys into maps. Any relationships which require matching on clusterID could immediately break due to this change. For instance, `ComputeClusterCosts` keys the total cluster costs by cluster id. In order to map `CostData` breakdowns into the total cluster costs, you have to be aware that `costData.ClusterID` is actually a `name/id` formatted field. This feels bad.

However, the current efforts were immediately viewable on the frontend without any frontend changes.

##### Proposed Solution
I think the best solution in this case is to add a field to `CostData`, namely `ClusterName` that will be set to the name of the cluster identified by cluster id. This allows existing relationship connections to be made on ClusterID, and any consuming our API will now have a ClusterName field to display. 

Any other ideas?


pkg/util/blockingqueue.go
```diff
@@ -63,6 +63,9 @@ func (q *blockingSliceQueue) Dequeue() interface{} {
 	}
 
 	e := q.q[0]
+
+	// nil 0 index to prevent leak
+	q.q[0] = nil
```
nikovacevic, line 68: :eyes:

nikovacevic (approved):
Okay, definitely some Thanos stuff here I'm not an expert on, but no red flags from me, so LGTM!

PR #567, opened 2020-10-12
add array length check before accessing lb.IngressIPAddresses

itscalvinwang:
Description: Add length check before array access to get ingress IP addresses.

Potential diagnosis: Application load balancer gets treated like ELB, despite not having ingress IPs.

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -518,7 +518,12 @@ func StartCostModelMetricRecording(a *Accesses) bool {
 				keyParts := getLabelStringsFromKey(lbKey)
 				namespace := keyParts[0]
 				serviceName := keyParts[1]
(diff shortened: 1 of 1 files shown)


pkg/costmodel/metrics.go
```diff
@@ -518,7 +518,12 @@ func StartCostModelMetricRecording(a *Accesses) bool {
 				keyParts := getLabelStringsFromKey(lbKey)
 				namespace := keyParts[0]
 				serviceName := keyParts[1]
-				ingressIP := lb.IngressIPAddresses[0] // assumes one ingress IP per load balancer
+				ingressIP := ""
+				if len(lb.IngressIPAddresses) > 0 {
+					ingressIP = lb.IngressIPAddresses[0] // assumes one ingress IP per load balancer
+				} else {
```
AjayTripathy, line 524: I'd just drop the "else" here and go leave it as an empty string, that's sufficient for determining nonexistance.

PR #570, opened 2020-10-15
Simplify CPU mode query & propagate query in errors

nikovacevic:
## Changes
- Remove `group_left` in favor of a simpler query for totals only. Then, compute breakdown post-hoc.
- Propagate `query` in more types of Prometheus query errors. (Previously, some kinds of errors would propagate, some would not.)

### Resource utilization
- Tested manually on gcp-niko. @AjayTripathy will take further testing there
### Query propagation
- Tested manually on gcp-niko with randomized manual errors and log checking
#### Before: no query
```
log.go:13] [Error] ComputeAssets: error: invalid parameter 'query': 1:5: parse error: unexpected identifier "query"
```
#### After: includes 'bad query', which was the query we tried to run
```
log.go:13] [Error] ComputeAssets: error: 'invalid parameter 'query': 1:5: parse error: unexpected identifier "query"' parsing query 'bad query'
```


dwbrown2:
Nice!! This LGTM, but would be helpful to get Ajay's GO input.

AjayTripathy:
LGTM! I tested by dropping node_exporter and multicluster on bolt-1

AjayTripathy (approved):

PR #572, opened 2020-10-16
Proposal for required and optional queries

nikovacevic:


AjayTripathy:
I'm supportive of this! Does this still have the whole failing query logged as warnings?

nikovacevic:
> I'm supportive of this! Does this still have the whole failing query logged as warnings?

Yes, it should! And we'll actually log them all, instead of just the one that we return.

AjayTripathy:
LGTM!

AjayTripathy (approved):

PR #573, opened 2020-10-16
Optimization: Apply Operation to Multiple Vectors

mbolt35:
### Summary
`util.ApplyVectorOp` is the top allocator running cost-model over time. This method is in the critical path aggregating query results. We've made a few iterations to improve memory consumption and total number of allocations by providing various map pooling implementations. Even with these improvements, we're still seeing it at the top of pprof heap profiles. This is likely due to the way have to chain call `ApplyVectorOp` to aggregate totals over multiple vectors. For instance
```go
sumVectors := func (....) { }

var a []*Vector
var b []*Vector
var c []*Vector
var d []*Vector

ab := util.ApplyVectorOp(sumVectors, a, b)
abc := util.ApplyVectorOp(sumVectors, ab, c) 
abcd := util.ApplyVectorOp(sumVectors, abc, d)
```
The total number of allocations increases with the total vector results we aggregate. 

### Proposal
We can perform a similar operation in a more performant way with a more flexible API. This still needs more testing, but the godocs for the new utility provide good insight into it's implementation. The idea would be to add this API, convert associative operations, then move `ApplyVectorOp` to use `ApplyMultiVectorOp`.

### Details
`ApplyMultiVectorOp` accepts M vectors, synchronizes timestamps (N), and executes an operation
on each vector's values for the collection timestamps. Providing M=5 []Vectors, assuming N=8 unique
timestamps across all []Vectors, `ApplyMultiVectorOp` will construct a `MxN (5x8)` matrix, and call the
provided `MultiVectorJoinOp` for each column (timestamp) passing the values at that timestamp for each
of the 5 Vectors.
```
            ++-------------------------------------------------------++
            ||                    TimeStamps                         ||
 ++---------++-----++-----++-----++-----++-----++-----++-----++------++
 || Vectors || 599 || 659 || 719 || 779 || 839 || 899 || 959 || 1019 ||
 ++---------++-----++-----++-----++-----++-----++-----++-----++------++
 ||    A    || 2.1 || nil || nil || 3.2 || nil || 1.4 || nil || nil  ||
 ||    B    || nil || 1.1 || 5.6 || 2.1 || nil || nil || 1.1 || 3.2  ||
 ||    C    || nil || 3.5 || nil || nil || nil || 8.3 || 5.2 || nil  ||
 ||    D    || 6.6 || nil || nil || 1.0 || 9.7 || nil || nil || nil  ||
 ||    E    || nil || 8.2 || 7.2 || 4.4 || 5.0 || 3.9 || nil || nil  ||
 ++---------++-----++-----++-----++-----++-----++-----++-----++------++
```

For each timestamp (starting with the earliest), the `MultiVectorJoinOp` will get called passing in a
result `*Vector` with the timestamp set, and the in-order slice for that column (representing values at
that timestamp for `Vectors A-E`). For example, the first call to `MultiVectorJoinOp` will contain the following:

```go
op(&Vector{ timestamp: 599 }, []*float64{ 2.1, nil, nil, 6.6, nil })
```

The `MultiVectorJoinOp`'s job will be to set the `*Vector`'s Value field to the result of an operation on those values.

It's important to note that it is completely up to the developer of the `MultiVectorJoinOp` to intelligently
implement the operation aware of the semantics of this function. Associative operations are safe (+, *) where
the ordering doesn't matter. Associativity is not a hard limitation, but one must use caution where ordering
matters. For instance, you could have a normalization operation which adds values at indices 0 through `len(values)-2`,
then divide the result by the value at index `len(values)-1`. However, this operation would require passing the `[]Vector`s
into `ApplyMultiVectorOp()` in the order required for achieve the expected result. These "special" rules with ordering
should be formalized using a utility/helper method that enforces the rules. ie:

```go
func NormalizeMulti(normalizeBy []*Vector, joinVecs ...[]*Vector) []*Vector {
     // Sums all the values except for the last, divides by last index
     normalizeOp := func(result *util.Vector, values []*float64) bool {
         var result *float64
         for i, v := range values {
             if i == len(values)-1 {
                 if result != nil && v != nil {
                     sum := *result
                     normalizeBy := *v
                     if sum == 0 || normalizeBy == 0 {
                         result.Value = 0
                     } else {
                         result.Value = sum / normalizeBy
                     }
                 }
                 continue
             }
             if v != nil {
                 if result != nil {
                     *result += *v
                 } else {
                     xv := *v
                     result = &xv
                 }
             }
         }
         return true
     }

     // create a single input based on the concrete op
     joinVecs = append(joinVecs, normalizeBy)
     return ApplyMultiVectorOp(normalizeOp, joinVecs)
 }
```

 The above function could also be implemented by an associative `addVectors()` `MultiVectorOp`
 followed by a binary `NormalizeVectorByVector()`

### Results
The hypothesis that `ApplyMultiVectorOp` would out-perform `ApplyVectorOp` was incorrect in a few scenarios. In fact, the more common scenario in cost-model (low vector count 5-15 and 1400+ data points per vector) performs nearly the same. `ApplyMultiVectorOp` is way more efficient (and fast) as the total number of vectors to apply at once grows. We saw excellent speedup and memory usage going from 5 to 100 vectors. Justification for the design is there, but we unfortunately don't have any direct use-cases just yet. The discussions internally revolve around getting the API committed along with the unit tests and benchmarks, and we will at least be equipped to deal with the specific scenarios that support `ApplyMultiVectorOp` down the road.


mbolt35:
Benchmarking was yielding somewhat abysmal results. Bottleneck on floatpool locking (with insertion and sort). After updating the implementation to use a linked list, we got a substantial improvement. After updating the search algorithms to allow passing in a hint to where to start the search, performance gains were a bit better. I'm still not sure that this change is justifiable - still need to think about whether or not the parallelism is actually a meaningful metric (relative to cost-model). While total allocations are still down, it's not as dramatic as expected.

Here are the benchmarks after all performance improvements:
```
BenchmarkApplyMultiVectorOp                        51439             21204 ns/op            4856 B/op        277 allocs/op
BenchmarkApplyMultiVectorOp-2                      80676             14344 ns/op            4856 B/op        277 allocs/op
BenchmarkApplyMultiVectorOp-4                      83032             14143 ns/op            4856 B/op        277 allocs/op
BenchmarkApplyMultiVectorOp-8                      82536             14384 ns/op            4856 B/op        277 allocs/op
BenchmarkParallelApplyMultiVectorOp                57801             21273 ns/op            4856 B/op        277 allocs/op
BenchmarkParallelApplyMultiVectorOp-2              66045             17595 ns/op            4856 B/op        277 allocs/op
BenchmarkParallelApplyMultiVectorOp-4              61599             19392 ns/op            4856 B/op        277 allocs/op
BenchmarkParallelApplyMultiVectorOp-8              54474             22046 ns/op            4856 B/op        277 allocs/op
BenchmarkApplyVectorOp                             55174             22326 ns/op            5984 B/op        388 allocs/op
BenchmarkApplyVectorOp-2                           68998             17067 ns/op            5984 B/op        388 allocs/op
BenchmarkApplyVectorOp-4                           70052             17073 ns/op            5984 B/op        388 allocs/op
BenchmarkApplyVectorOp-8                           66832             17208 ns/op            5984 B/op        388 allocs/op
BenchmarkParallelApplyVectorOp                     56398             21812 ns/op            5984 B/op        388 allocs/op
BenchmarkParallelApplyVectorOp-2                   82910             13899 ns/op            5984 B/op        388 allocs/op
BenchmarkParallelApplyVectorOp-4                  142105              8077 ns/op            6353 B/op        390 allocs/op
BenchmarkParallelApplyVectorOp-8                  177829              6237 ns/op            7795 B/op        395 allocs/op
```

mbolt35:
### Benchmarking
Wishing that I could definitively put this changeset to rest, I had to examine the scenarios which put `MultiVectorOp` at an advantage to the binary `VectorOp`. Due to the existence of the `AllocationList` and inevitable allocations representing used space, in order for `MultiVectorOp` bring value, the allocations had to be fairly large (with respect to slice length). Otherwise, the total volume of small allocations would add to search time when fitting new allocations, which lead to longer blocking time while running in parallel. For example, if we were to add 1000 Vectors containing 10 data points, then `MultiVectorOp` would be much more optimal than `VectorOp`:

```
BenchmarkApplyMultiVectorOp                 2590            449971 ns/op           81496 B/op      10051 allocs/op
BenchmarkApplyMultiVectorOp-2               4616            252501 ns/op           81419 B/op      10049 allocs/op
BenchmarkApplyMultiVectorOp-4               9326            126865 ns/op           81369 B/op      10047 allocs/op
BenchmarkApplyMultiVectorOp-8              12602             93350 ns/op           81358 B/op      10047 allocs/op

BenchmarkApplyVectorOp                       580           2122270 ns/op          600189 B/op      35989 allocs/op
BenchmarkApplyVectorOp-2                     792           1614792 ns/op          599978 B/op      35982 allocs/op
BenchmarkApplyVectorOp-4                    1096           1052915 ns/op          656002 B/op      36412 allocs/op
BenchmarkApplyVectorOp-8                    1028           1173477 ns/op          762099 B/op      37077 allocs/op
```

However, this scenario is not in the current set of use-cases with respect to cost-model. In fact, it's the opposite: We're more likely to pull high data points with less total vectors. In this case, the `MultiVectorOp` performs much worse by allocating more (`AllocationList` bloat), and just being slower all-around (`AllocationList` search times are much higher). For example, if we were to add 5 vectors with 1000 data points each:

```
BenchmarkApplyMultiVectorOp                  928           1327393 ns/op          160600 B/op       9017 allocs/op
BenchmarkApplyMultiVectorOp-2                702           1714984 ns/op          160664 B/op       9019 allocs/op
BenchmarkApplyMultiVectorOp-4                370           3149489 ns/op          160909 B/op       9025 allocs/op
BenchmarkApplyMultiVectorOp-8                189           6413026 ns/op          164973 B/op       9040 allocs/op

BenchmarkApplyVectorOp                      1239            968808 ns/op          194608 B/op      12084 allocs/op
BenchmarkApplyVectorOp-2                    2234            535702 ns/op          194606 B/op      12084 allocs/op
BenchmarkApplyVectorOp-4                    3854            324077 ns/op          199205 B/op      12087 allocs/op
BenchmarkApplyVectorOp-8                    4370            250835 ns/op          236116 B/op      12113 allocs/op
```

In the above results, our `MultiVectorOp` managed to keep allocations/op below `VectorOp` but ran significantly slower.

Finally, the final benchmarks remove the `AllocationList` from the equation and just straight allocate `[]*float64` dynamically (5 vectors with 1000 data points each):
```
BenchmarkApplyMultiVectorOp                 1746            613797 ns/op          128510 B/op       8014 allocs/op
BenchmarkApplyMultiVectorOp-2               3759            317847 ns/op          128455 B/op       8013 allocs/op
BenchmarkApplyMultiVectorOp-4               6691            157633 ns/op          128435 B/op       8012 allocs/op
BenchmarkApplyMultiVectorOp-8               8971            116132 ns/op          132227 B/op       8013 allocs/op

BenchmarkApplyVectorOp                      1268            969224 ns/op          194349 B/op      12080 allocs/op
BenchmarkApplyVectorOp-2                    2254            515192 ns/op          194255 B/op      12076 allocs/op
BenchmarkApplyVectorOp-4                    4347            269468 ns/op          197216 B/op      12084 allocs/op
BenchmarkApplyVectorOp-8                    5108            210289 ns/op          246103 B/op      12120 allocs/op
```

Surprisingly, this out performed `VectorOp` and had 25% less allocations/op (significantly less bytes/op as well).

Now, flipping this back to 1000 Vectors with 10 data points:
```
BenchmarkApplyMultiVectorOp                 2457            500399 ns/op          162626 B/op      10042 allocs/op
BenchmarkApplyMultiVectorOp-2               4496            261693 ns/op          162542 B/op      10039 allocs/op
BenchmarkApplyMultiVectorOp-4               7646            144143 ns/op          162501 B/op      10037 allocs/op
BenchmarkApplyMultiVectorOp-8               9745            111834 ns/op          162490 B/op      10037 allocs/op

BenchmarkApplyVectorOp                       577           2107929 ns/op          600193 B/op      35990 allocs/op
BenchmarkApplyVectorOp-2                     736           1631009 ns/op          600022 B/op      35984 allocs/op
BenchmarkApplyVectorOp-4                    1110           1059998 ns/op          654357 B/op      36401 allocs/op
BenchmarkApplyVectorOp-8                    1021           1206727 ns/op          753693 B/op      37020 allocs/op
```

We can see `MultiVectorOp` is still better in speed and allocations, but isn't quite as performant as when the `AllocationList` is added:

```
-- With Allocation List --
BenchmarkApplyMultiVectorOp                 2590            449971 ns/op           81496 B/op      10051 allocs/op
BenchmarkApplyMultiVectorOp-2               4616            252501 ns/op           81419 B/op      10049 allocs/op
BenchmarkApplyMultiVectorOp-4               9326            126865 ns/op           81369 B/op      10047 allocs/op
BenchmarkApplyMultiVectorOp-8              12602             93350 ns/op           81358 B/op      10047 allocs/op

-- Dynamic Allocations --
BenchmarkApplyMultiVectorOp                 2457            500399 ns/op          162626 B/op      10042 allocs/op
BenchmarkApplyMultiVectorOp-2               4496            261693 ns/op          162542 B/op      10039 allocs/op
BenchmarkApplyMultiVectorOp-4               7646            144143 ns/op          162501 B/op      10037 allocs/op
BenchmarkApplyMultiVectorOp-8               9745            111834 ns/op          162490 B/op      10037 allocs/op
```
The `AllocationList` version was slightly more allocations/op but the bytes per operation were about 50% of the dynamic allocations, and was a good order of magnitude faster.

### Summary 
The following were the best in both speed (total iterations and ns/op) as well as allocations:

5 Vectors, 1000 DataPoints per Vector: **MultiVectorOp - No AllocationList**
```
BenchmarkApplyMultiVectorOp-8               8971            116132 ns/op          132227 B/op       8013 allocs/op
```
1000 Vectors, 10 DataPoints per Vector: **MultiVectorOp - AllocationList**
```
BenchmarkApplyMultiVectorOp-8              12602             93350 ns/op           81358 B/op      10047 allocs/op
```

The more frequent use-case in cost-model is definitely less Vectors with higher data points (24h - 1440 data points). However, there may be cases in specific environments where Vector counts may be increased. The problem currently is that `VectorOp` performs average across both scenarios. It starts to become a problem at higher Vector counts, but doesn't tip over quite as quickly as `MultiVectorOp + AllocationList` at lower vector counts. 

The next set of questions are related to constraints, and if we're even able to determine which algorithm to apply based on the incoming data set.

AjayTripathy:
I want to ask a more fundamental question about this: should we be doing this math on pointers at all? It seems like that's blowing up our GC time and the real reason behind these allocations?

PR #578, opened 2020-10-20
Threadsafe logging

nikovacevic:
## Changes
- In deduped logging, turn the `seen` map into a threadsafe counter

## Testing
- Reproduced concurrent map write panics
```
$ go test -v ./pkg/log/
=== RUN   TestCounter_Ops
--- PASS: TestCounter_Ops (0.00s)
=== RUN   TestCounter_Threadsafety
fatal error: concurrent map writes

goroutine 23 [running]:
runtime.throw(0x55c456, 0x15)
	/usr/local/go/src/runtime/panic.go:774 +0x72 fp=0xc000059eb0 sp=0xc000059e80 pc=0x42eca2
runtime.mapassign_faststr(0x529f40, 0xc0000981e0, 0x55c993, 0x16, 0xc0000ee088)
	/usr/local/go/src/runtime/map_faststr.go:211 +0x417 fp=0xc000059f18 sp=0xc000059eb0 pc=0x413667
github.com/kubecost/cost-model/pkg/log.(*counter).increment(...)
	/home/niko/code/kubecost/model/cost-model/pkg/log/counter.go:29
github.com/kubecost/cost-model/pkg/log.DedupedInfof(0xa, 0x55c993, 0x16, 0xc000059fb8, 0x1, 0x1)
	/home/niko/code/kubecost/model/cost-model/pkg/log/log.go:54 +0x5a fp=0xc000059f80 sp=0xc000059f18 pc=0x50561a
github.com/kubecost/cost-model/pkg/log.TestCounter_Threadsafety.func1(0xc0000a00d0)
	/home/niko/code/kubecost/model/cost-model/pkg/log/counter_test.go:54 +0x92 fp=0xc000059fd8 sp=0xc000059f80 pc=0x505f42
runtime.goexit()
```

- With changes, test are passing
```
$ go test -v ./pkg/log/
=== RUN   TestCounter_Ops
--- PASS: TestCounter_Ops (0.00s)
=== RUN   TestCounter_Threadsafety
--- PASS: TestCounter_Threadsafety (1.58s)
=== RUN   TestDeduping
--- PASS: TestDeduping (0.00s)
PASS
ok  	github.com/kubecost/cost-model/pkg/log	1.579s
```

### DIFF
--- a/pkg/log/counter.go
+++ b/pkg/log/counter.go
@@ -12,8 +12,8 @@ func newCounter() *counter {
 }
 
 func (ctr *counter) count(key string) int {
-	ctr.mu.Lock()
-	defer ctr.mu.Unlock()
+	ctr.mu.RLock()
+	defer ctr.mu.RUnlock()
 	return ctr.seen[key]
 }
 
--- a/pkg/log/counter.go
+++ b/pkg/log/counter.go
@@ -1,0 +1,31 @@
+package log
+
+import "sync"
+
+type counter struct {
+	mu   sync.RWMutex
+	seen map[string]int
+}
+
+func newCounter() *counter {
+	return &counter{seen: map[string]int{}}
+}
+
+func (ctr *counter) count(key string) int {
+	ctr.mu.Lock()
+	defer ctr.mu.Unlock()
+	return ctr.seen[key]
+}
+
+func (ctr *counter) delete(key string) {
+	ctr.mu.Lock()
+	delete(ctr.seen, key)
+	ctr.mu.Unlock()
+}
+
+func (ctr *counter) increment(key string) int {
+	ctr.mu.Lock()
+	defer ctr.mu.Unlock()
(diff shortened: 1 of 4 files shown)


mbolt35:
The only "improvements" [I don't think we have a lot of contention here] would be to use a `RWMutex` and an atomic counter. `Lock()` on `delete`, `RLock()` on `count`, and `RLock()` on increment and use `atomic.AddInt32`. Unfortunately, I think go prevents taking the address of map space, so you'd have to use [AtomicInt32](https://github.com/kubecost/cost-model/blob/develop/pkg/util/atomic.go). This would allow multiple readers and the atomic would handle concurrent increments. 

I'm not advocating for this change, but using it as sort of a thought experiment. I think just using a `RWMutex` and write locking on `delete` and `increment` is fine :)

AjayTripathy (approved):

PR #579, opened 2020-10-20
added length check to keyParts before indexing

itscalvinwang:
Description: checks length of keyParts before accessing


AjayTripathy:
@itscalvinwang this doesn't work at all. variables need declaration outside the loop. CircleCI compiles this for you as well.

PR #584, opened 2020-10-25
Ajay tripathy return aws errors

AjayTripathy:
* Error handling code-- save and propagate athena and spot errors to the frontend.
* Quick update to spot code-- if you detect an instance as "spot" from AWS, just force it's type to be preemptible. Also clean up a bit of the hardcoded strings with a constant.
* Support over 1000 rows in the out-of-cluster queries


pkg/cloud/awsprovider.go
```diff
@@ -1514,6 +1556,82 @@ func generateAWSGroupBy(lastIdx int) string {
 	return strings.Join(sequence, ",")
 }
 
+func (a *AWS) QueryAthenaPaginated(query string) (*athena.GetQueryResultsInput, *athena.Athena, error) {
+	customPricing, err := a.GetConfig()
+	if err != nil {
+		return nil, nil, err
+	}
+	if customPricing.ServiceKeyName != "" {
```
mbolt35, line 1564: This needs new auth.

mbolt35: `a.ConfigureAuthWith(customPricing)`

AjayTripathy: Thanks good catch :)

PR #587, opened 2020-10-28
Mapper GetList/SetList

mbolt35:
`GetList` and `SetList` added to the mapper utility for getting/setting delimited strings to/from slices.

### DIFF
--- a/pkg/util/mapper/mapper.go
+++ b/pkg/util/mapper/mapper.go
@@ -2,6 +2,7 @@ package mapper
 
 import (
 	"strconv"
+	"strings"
 )
 
 //--------------------------------------------------------------------------
@@ -83,6 +84,11 @@ type PrimitiveMapReader interface {
 	// GetBool parses a bool from the map key parameter. If the value
 	// is empty or fails to parse, the defaultValue parameter is returned.
 	GetBool(key string, defaultValue bool) bool
+
+	// GetList returns a string list which contains the value set by key split using the
+	// provided delimiter with each entry trimmed of space. If the value doesn't exist,
+	// nil is returned
+	GetList(key string, delimiter string) []string
 }
 
 // PrimitiveMapWriter is an implementation contract for an object capable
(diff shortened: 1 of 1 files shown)


pkg/util/mapper/mapper.go
```diff
@@ -83,6 +84,11 @@ type PrimitiveMapReader interface {
 	// GetBool parses a bool from the map key parameter. If the value
 	// is empty or fails to parse, the defaultValue parameter is returned.
 	GetBool(key string, defaultValue bool) bool
+
+	// GetList returns a string list which contains the value set by key split using the
+	// provided delimiter with each entry trimmed of space. If the value doesn't exist,
+	// nil is returned
+	GetList(key string, delimiter string) []string
```
nikovacevic, line 91: Should we call this `GetStringList` in case we want to leave room for `GetIntList` or something like that?

pkg/util/mapper/mapper.go
```diff
@@ -373,6 +383,27 @@ func (rom *readOnlyMapper) GetBool(key string, defaultValue bool) bool {
...
+// nil is returned
+func (rom *readOnlyMapper) GetList(key string, delimiter string) []string {
+	value := rom.Get(key, "")
+	if value == "" {
+		return nil
+	}
+
+	split := strings.Split(value, delimiter)
+
+	// reuse slice created for split
```
nikovacevic, line 397: Fancy

nikovacevic (approved):
LGTM

pkg/util/mapper/mapper.go
mbolt35: Just trying to avoid the `strings.Split` allocation churn our key tuples were generating months ago -- it can be somewhat disorienting to read at first, but in general, we avoid allocating a new backing storage every time we create a slice.

pkg/util/mapper/mapper.go
mbolt35: Yeah... Thought about this as well, multiple times, changed my mind, then changed it again. The general resolution I landed on was that the underlying data provider is a string map `Get(key string) string`, and the `Mapper` `Get` implementation is: `Get(key string, defaultValue string) string`, so the specialization adds the qualifiers, and `string` is always the default "value" type.

PR #593, opened 2020-11-02
Ajay tripathy use regional pricing

AjayTripathy:


dwbrown2 (commented):
What's the outcome if node regions change, i.e. a second region is added at a later time? Not sure I've seen this before, but just want to understand the behavior if it does.

nikovacevic (approved):
LGTM

AjayTripathy:
> What's the outcome if node regions change, i.e. a second region is added at a later time? Not sure I've seen this before, but just want to understand the behavior if it does.

I have not tested this exact case, but it would be similar to a new node class joining the fleet. When we're missing a key for a node, DownloadPricingData gets called one time. We then detect multiple regions and download the full 1gb file. Nodes are keyed by region, class, and whether or not they are spot, so we should be missing a key.

nikovacevic (approved):
LGTM

PR #594, opened 2020-11-03
Ajay tripathy fix duplication issue

AjayTripathy:
Fixes a prometheus query that would sometimes fail if the node KSM was on was removed from the fleet/terminated.

Regression Testing:
* Over consistent intervals, where the node duplication does not exist
Screen Shot 2020-11-03 at 3 22 10 PM
 the two queries are identical
* Over consistent intervals where the node duplicatin does exist, the old query fails:
Screen Shot 2020-11-03 at 3 21 23 PM
* The added row for the changing node looks reasonable, and the existing rows don't change much, when a time range expands to have the duplication (added node highlighted)
Screen Shot 2020-11-03 at 3 25 53 PM


nikovacevic (approved):
LGTM

PR #595, opened 2020-11-06
Ajay tripathy downsampling safe queries

AjayTripathy:


nikovacevic (approved):
Okay, I haven't manually tested each individual query, but the methodology looks sound.

One thing I would say we should remove is `kubecostMinsPerHour` and associated queries. Seems we don't use that any more, so we can save an entire query (`queryKubecostUpMinsPerHour`) and some parsing.

Also, the comments describing the queries that have changed now do not apply. I can come back around and update those, but at least want to plant the flag there.

PR #598, opened 2020-11-08
refine savings plan estimates

AjayTripathy:
For time periods where savings plans expired, ran out of allocated credits, or had partial usages covered, we were underestimating the hourly cost. My proposal is we instead charge the savings plan rate to each node. This is still likely an underestimate, in the sense that the real value is the partial coverage in the "effective cost" column + the regular line item usage. 

For perfect accuracy you need to wait for reconciliation data anyway-- my suggestion is just to improve the estimate here for now, instead of trying to perfectly calculate an hourly cost w/ all things like amortized initial savings plan fee and amortized monthly recurring savings plan fee etc.

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -1710,25 +1710,29 @@ func (a *AWS) GetSavingsPlanDataFromAthena() error {
 	tOneDayAgo := tNow.Add(time.Duration(-25) * time.Hour) // Also get files from one day ago to avoid boundary conditions
 	start := tOneDayAgo.Format("2006-01-02")
 	end := tNow.Format("2006-01-02")
+	// Use Savings Plan Effective Rate as an estimation for cost, assuming the 1h most recent period got a fully loaded savings plan.
+	//
 	q := `SELECT   
 		line_item_usage_start_date,
 		savings_plan_savings_plan_a_r_n,
 		line_item_resource_id,
-		savings_plan_savings_plan_effective_cost
+		savings_plan_savings_plan_rate 
 	FROM %s as cost_data
 	WHERE line_item_usage_start_date BETWEEN date '%s' AND date '%s'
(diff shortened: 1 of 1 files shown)


dwbrown2:
Can you share the difference between `savings_plan_savings_plan_effective_cost` and `savings_plan_savings_plan_rate`? I would expect the latter to be a better option and so it's not clear why this PR is an improvement.

AjayTripathy:
"savings_plan_savings_plan_rate": The hourly rate of the node under the savings plan.
"savings_plan_effective_cost": The amount actually billed against the savings plan. 

In cases where you exceed your savings plan, savings_plan_effective cost is low, as there is a "remainder" line item that appears elsewhere in the CUR. We could try to search the CUR for this line when effective cost < rate, but given that we do this in the price reconciliation pipeline I think this is overkill here.

dwbrown2:
SGTM, thanks for sharing

Webb
Kubecost, Co-founder

ᐧ

pkg/cloud/awsprovider.go
```diff
@@ -1752,8 +1756,20 @@ func (a *AWS) GetSavingsPlanDataFromAthena() error {
...
 		}
 		a.SavingsPlanDataLock.Unlock()
-	} else {
-		klog.Infof("No savings plan applied instance data found")
+		return true
+	}
+
+	query := fmt.Sprintf(q, cfg.AthenaTable, start, end)
+
+	klog.V(3).Infof("Running Query: %s", query)
```
nikovacevic, line 1764: Noisy log?

nikovacevic (approved):
LGTM

pkg/cloud/awsprovider.go
AjayTripathy: It doesn't happen that often and has been useful in debugging so far.

PR #602, opened 2020-11-12
Prometheus Recording Rules API

mbolt35:
### Summary
Basic API support for loading/caching the configured prometheus recording rules. Currently is limited to `LoadRecordingRules()` and `HasRecordingRule(name string)`. This implementation can be queried and cached early, and can act as a diagnostic for failed queries.

### DIFF
--- a/pkg/prom/rules.go
+++ b/pkg/prom/rules.go
@@ -1,0 +1,139 @@
+package prom
+
+import (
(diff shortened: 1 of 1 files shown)


mbolt35:
This code can be replaced by the use of `Prometheus.API.Rules` implementation.

PR #604, opened 2020-11-13
nan shield on total of 0

AjayTripathy:
I don't think we have full control over the data coming back from Prometheus, and this seems like the place a NaN could sneak in.


nikovacevic (approved):
LGTM

PR #608, opened 2020-11-18
set spot gpu values

AjayTripathy:
sets a default spot gpu value if it doesn't exist, and uses it when a spot label is applied.


nikovacevic (approved):
LGTM

PR #609, opened 2020-11-18
pull scrape interval from prometheus

AjayTripathy:


pkg/costmodel/promparsers.go
```diff
@@ -8,8 +8,30 @@ import (
...
+}
+
+func GetPrometheusConfig(pcfg string) (PromCfg, error) {
+	var promCfg PromCfg
+	err := yaml.Unmarshal([]byte(pcfg), &promCfg)
+	return promCfg, err
+}
+
+func GetKubecostJobName() string {
+	return DEFAULT_KUBECOST_JOB_NAME // TODO: look this up from a prometheus variable?
```
nikovacevic, line 32: So, until we do this, someone with a different job name would get the default of "1m", right?

nikovacevic (approved):
Makes sense to me. Just that one quick question about behavior when the job name isn't the default job name.

nikovacevic:
Also, do we have an env in which to test weird scrape intervals?

pkg/costmodel/promparsers.go
AjayTripathy: Yeah. That seem right to you?

PR #616, opened 2020-12-02
Ajay tripathy fix offset

AjayTripathy:
Certain direct queries will pass "" as the offset, which would have broken the old data.


nikovacevic (approved):
LGTM!

PR #619, opened 2020-12-04
Feature/costdatarange porviderids

ameyajoshi99:
@dwbrown2 / @mbolt35 
Hi,
This is change is based off your suggestion on https://github.com/kubecost/cost-model/pull/558

Here, I've used Kube state metrics to fetch the provider id for node and volumenid for PVC. This is now able to fetch the historical data.
FYI, for PVC, KSM has added support in their alpha release.
https://github.com/kubernetes/kube-state-metrics/blob/v2.0.0-alpha/docs/persistentvolume-metrics.md

I have no experience in other providers than aws, so not returning the provider id's for Azure, CSV, Custom & GCP


ameyajoshi99:
@mbolt35 / @dwbrown2  Did you get a chance to look at this?

mbolt35:
@ameyajoshi99 I don't see `providerID` on the KSM persistentvolume metrics. https://github.com/kubernetes/kube-state-metrics/blob/master/docs/persistentvolume-metrics.md

Just wanted to provide some insight here to give you an idea of our current goals with the open source code base. Currently, the CostDataRange taking on huge responsibility (way too much). We've just performed a major overhaul to error handling, and we're going to continue to iterate in order to remove the bottleneck scenarios that occur in that function. Many of existing patterns in this function are going to be refactored or separated in order to reduce unnecessary dependencies as well as hopefully setup clear extensibility patterns for adding new functionality (neither of which are in place). 

With that being said, our hesitation with this PR is mostly due to the additions of more queries to this function. However, we are definitely interested in what your change adds functionally so we can keep it in mind as we make previously mentioned improvements. If you could help me understand a bit more about the changeset: It looks like there's an expectation for the `providerId` field to exist on the metrics; I wasn't able to find that specific prom label on the KSM metrics for PVs, PVCs, or Storage Classes. For the node info query, I did see the `provider_id` that exists on `kube_node_info`.  

I seem to recall that the end goal is to supply better information for the PVClaimData (ie: the storageclass) as well as the specific Volume data. If you could let me know about the `providerId` and if I understand the goal accurately, I can be sure to create a ticket that's trackable for providing this functionality when we refactor this code in the future. 

Apologies for the delayed response -- busy time of year! Let me know if you have anymore questions or comments.

ameyajoshi99:
> @ameyajoshi99 I don't see `providerID` on the KSM persistentvolume metrics. https://github.com/kubernetes/kube-state-metrics/blob/master/docs/persistentvolume-metrics.md
> 
> Just wanted to provide some insight here to give you an idea of our current goals with the open source code base. Currently, the CostDataRange taking on huge responsibility (way too much). We've just performed a major overhaul to error handling, and we're going to continue to iterate in order to remove the bottleneck scenarios that occur in that function. Many of existing patterns in this function are going to be refactored or separated in order to reduce unnecessary dependencies as well as hopefully setup clear extensibility patterns for adding new functionality (neither of which are in place).
> 
> With that being said, our hesitation with this PR is mostly due to the additions of more queries to this function. However, we are definitely interested in what your change adds functionally so we can keep it in mind as we make previously mentioned improvements. If you could help me understand a bit more about the changeset: It looks like there's an expectation for the `providerId` field to exist on the metrics; I wasn't able to find that specific prom label on the KSM metrics for PVs, PVCs, or Storage Classes. For the node info query, I did see the `provider_id` that exists on `kube_node_info`.
> 
> I seem to recall that the end goal is to supply better information for the PVClaimData (ie: the storageclass) as well as the specific Volume data. If you could let me know about the `providerId` and if I understand the goal accurately, I can be sure to create a ticket that's trackable for providing this functionality when we refactor this code in the future.
> 
> Apologies for the delayed response -- busy time of year! Let me know if you have anymore questions or comments.

No worries @mbolt35 ....Thanks for the reply.  
As you mentioned the KSM metric `kube_node_info` do have provider_id for instances. However, for PVC it does not  have the field under same label. This is in their pipeline, currently released in beta version: 
https://github.com/kubernetes/kube-state-metrics/blob/v2.0.0-beta/docs/persistentvolume-metrics.md
Under `kube_persistentvolume_info` you would get persistentvolume(name), storageclass & ebs_volume_id (in case of aws provider, for rest they have other fields like gce_persistent_disk_name).
In this PR I've replaced `ebs_volume_id` in aws provider with label `providerId` to keep it consistent with the code, that's why you don't see this field in KSM.
Once this is released, we can straight away start using this, and even if the field is missing in current version, prom query will not break but just field providerId would be missing in the query result which can be handled in the code.

michaelmdresser:
I think this PR is obsolete due to many improvements to Kubecost's core APIs since it was opened. Check out our [official documentation](https://guide.kubecost.com/hc/en-us) to learn more. If you still don't have the functionality desired, please feel free to open a new issue or PR!

PR #623, opened 2020-12-10
Ajay tripathy fix e2custom

AjayTripathy:
Support e2 custom prices, using the same pricing tier as e2 standard but charging for the full CPU/RAM (no partial CPU usage)


nikovacevic (approved):
LGTM

PR #625, opened 2020-12-10
do not pass in a 'cost' value for aws spot nodes

AjayTripathy:
Testing done: Force all nodes to use this case of "createNode" and check emitted values:
Before:
Screen Shot 2020-12-10 at 3 32 11 PM
After:
Screen Shot 2020-12-10 at 3 26 10 PM

Note that a todo is to correctly assign spot GPU costs. I want to patch/hotfix 1.70.0 with this, so have made the most minimal set of changes. 1.71.0 can better support spot GPU nodes.


nikovacevic (approved):
LGTM

For posterity, we're following up with more fixes to this in a near-future release.

PR #626, opened 2020-12-14
Ajay tripathy handle parsefails

AjayTripathy:
Stop blowing up the cost-model on invalid configs
Fix unit tests to all be in the same package
Add unit test + bad config


dwbrown2:
Can we add a test or two?

mbolt35 (approved):
LGTM

PR #630, opened 2020-12-15
Ajay tripathy handle spotgpu

AjayTripathy:
WIP, needs a unit test. 

You need to not pass any of RAM, CPU, GPU "cost" to the cloud provider in order for our split calculation to take effect. Pass just 1, and whatever cost you have written down gets passed through to the metrics emitter.


nikovacevic (approved):
TL;DR: LGTM (I think.)

Here's my breakdown of what this change affects:
`ComputeCostData` (costmodel.go) and `costModelMetricsEmitter.Start()` (metrics.go) both call `GetNodeCost`:
- `costmodel.GetNodeCost`
  - `provider.NodePricing`
    - `aws.NodePricing`
      - if `aws.Pricing` exists for the given key (i.e. node)
        - `aws.createNode(terms, usageType, k)`
      - else
        - `DownloadPricingData`, then if that goes without errors
          - `aws.createNode(terms, usageType, k)`

Within `aws.createNode`, we have:
- `aws.createNode(terms, usageType, key)`
  - if spot and spotInfo is available
    - return Node with spot data
  - else if spot
    - return Node with default values (i.e. no spot cost) !!! this is the proposed change to remove node.RAMCost !!!
  - else if savings plan
    - return Node with savings plan cost
  - else if reserved instance
    - return Node with reserved instance cost
  - else
    - return Node with on demand cost

Q: What happens in the case of spot, but no spot data? What should happen?
A: The Node from `createNode` gets returned through `aws.NodePricing` (i.e. `provider.NodePricing`) straight to `costmodel.GetNodeCost`. From there, this is the flow:

- `costmodel.GetNodeCost`
  - get all the nodes in the cluster, then for each one:
    - cnode, err := cp.NodePricing(cp.GetKey(nodeLabels, n))
      - if err and cnode
        - set and continue !!! this is weird !!!
      - if err and nil cnode
        - start with Node that only has default CPU/RAM cost
      - else
        - start with what was returned  !!! this is the path in question (err == nil) !!!
    - set instance type, region, size, etc. if not set and the data is available
    - if GPU != "" && GPUCost == "", set CPU, RAM, GPU cost after "fixing" CPU, RAM
    - else if RAMCost == "", set RAM after "fixing" CPU
  - apply reserved instance pricing to everything (even though it's now after the 4th else block of `createNode`?)

Are we sure that these two cases cover everything? That is, do (have GPU, but no GPUCost) and (no RAMCost) cover the whole space of possibilities?

Seems like, in either case, we're relying on VCPUCost to be set, so removing that would actually break things and make GetNodeCost return (nil, err). Right? If we agree on that, then this change seems reasonable.

In terms of how to test this, I think we're probably talking about re-designing this set of functionality. It's extremely difficult to reason about how small changes in specific implementations will affect generalized logic; e.g. if using the interface requires hours of study of each individual implementation, then the interface isn't doing anything to reduce complexity and increase testability. The best I can offer without a rewrite would be to simulate as many of these permutations as possible in an integration-test-type scenario, which I would not recommend doing because there are too many permutations and the test would probably be very fragile.

Issue #637, opened 2020-12-17
Support same instance type matching in custom CSV


rcallphin:
When a node instanceID does not match in the custom CSV, having support for the process to infer a match based on other instanceID's in the csv that share the same instance type and region or vmss in azure.

dwbrown2:
Following up after our discussion on Thursday, the current plan is to use existing data points in the CSV pipeline to infer new node prices. Initial algorithm will follow this logic:

* First search for an exact match in CSV pipeline
* If exact match not available, search for an existing CSV data point that matches region, instanceType, and AssetClass 
* If neither available, fallback to pricing estimates

dwbrown2:
Initial implementation PR now out: https://github.com/kubecost/cost-model/pull/641

AjayTripathy:
merged and released  in v1.71.1!

PR #641, opened 2020-12-21
add csv fallback

AjayTripathy:
See #637

Goal will be to use existing data points in the CSV pipeline to infer new node prices. Initial algorithm will follow this logic:
* First search for an exact match in CSV pipeline
* If exact match not available, search for an existing CSV data point that matches region, instanceType, and AssetClass 
* If neither available, fallback to pricing estimates

### DIFF
--- a/test/cloud_test.go
+++ b/test/cloud_test.go
@@ -1,6 +1,8 @@
 package test
 
 import (
+	"fmt"
+	"math"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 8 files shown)


test/cloud_test.go
```diff
@@ -310,3 +314,48 @@ func TestNodePriceFromCSVWithCase(t *testing.T) {
 	}
 
 }
+
+func TestNodePriceFromCSVByClass(t *testing.T) {
+	n := &v1.Node{}
+	n.Spec.ProviderID = "fakeproviderid"
+	n.Labels = make(map[string]string)
+	n.Labels[v1.LabelZoneRegion] = "eastus2"
```
dwbrown2, line 322: Are we matching on region or region/Zone set?

AjayTripathy: Region. v1.LabelZoneRegion is either topology.kubernetes.io/region or failure-domain.beta.kubernetes.io/region depending on k8s version. v1.LabelZoneFailureDomain would be the zone.

mbolt35 (approved):
LGTM

PR #642, opened 2020-12-21
Annotation Emission

mbolt35:
Check and register annotations collectors if enabled.


mbolt35:
Discussed with @AjayTripathy - Merging.

PR #644, opened 2020-12-22
Ajay tripathy pvc error fix

AjayTripathy:
empty volumenames could result in many-to-many match errors, and this is potentially possible for claims that never get fulfilled.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/aws/aws-sdk-go v1.28.9
 	github.com/dimchansky/utfbom v1.1.0 // indirect
 	github.com/getsentry/sentry-go v0.6.1
-	github.com/google/martian v2.1.0+incompatible // indirect
+	github.com/google/martian v2.1.0+incompatible
 	github.com/google/uuid v1.1.1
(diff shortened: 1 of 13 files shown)


dwbrown2 (commented):
Could persistentvolumeclaim also be empty? If so, would that cause issues?

AjayTripathy:
> Could persistentvolumeclaim also be empty? If so, would that cause issues?

Seems unlikely given that we're joining kube_persistentvolumeclaim_info and kube_persistentvolumeclaim_resource_requests_storage_bytes

nikovacevic (commented):
I understand the query change (filtering out metrics with an empty field) but the rest I don't exactly understand. Is this a combination of two different PRs, or am I just not seeing how they relate?

Query changes LGTM.

AjayTripathy:
> I understand the query change (filtering out metrics with an empty field) but the rest I don't exactly understand. Is this a combination of two different PRs, or am I just not seeing how they relate?
> 
> Query changes LGTM.

This branched off the wrong place, super sorry. I'll update to branch from develop

PR #657, opened 2021-01-12
Sean/agg on annotations

Sean-Holcomb:
Tested on bolt-1 using /model/aggregatedCostModel?window=1d&offset=1m&aggregation=annotation&allocateIdle=true&timeSeries=true&aggregationSubfield=test_kubecost_annotation_name.

Change aggregation subfield to different annotation labels to for different annotations. Given limited annotation data points all calls to actual annotation labels will result in aggregation on label and __unallocated__.


nikovacevic (commented):
Seems like this contains all changes from https://github.com/kubecost/cost-model/pull/654 and some more. So it'll suffice to just reference the other review: https://github.com/kubecost/cost-model/pull/654#pullrequestreview-569726865

Sean-Holcomb:
On aggregation on annotations, there can be a difference in numbers when the etl flag is on and off. This is due to the fact that if a pod has an annotation on one day within the window, then ETL applies that annotation to that pod on all days in that window even if not present

nikovacevic (approved):
LGTM

PR #662, opened 2021-01-13
Aggregation by label

SheepGotoHeaven:
Updated AssetSet and AssetSetRange to aggregate by a new property aggStrings []string.
The props []AssetProperty value is still maintained on a given AssetSet, but rather than use this for aggregation,
we now use aggStrings, which can include values other than enumerated props. Specifically, strings prefixed with "label:"
are interpreted as labels and can be grouped on.

Strings in aggStrings which match the enumerated AssetProperty strings are stored in AssetSet.props, as before.

Also updated the relevant asset_test tests to call the Aggregation funcs with []string rather than []AssetProperty

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -63,8 +63,7 @@ type Asset interface {
 // values will key by only those values.
 // Valid values of `aggregateBy` elements are strings which are an `AssetProperty`, and strings prefixed
 // with `"label:"`.
-func key(a Asset, aggregateBy []string) (string, []error) {
-	var errors []error
+func key(a Asset, aggregateBy []string) (string, error) {
 	keys := []string{}
 
 	if aggregateBy == nil {
@@ -82,44 +81,46 @@ func key(a Asset, aggregateBy []string) (string, []error) {
 	}
 
 	for _, s := range aggregateBy {
+		key := ""
 		switch true {
-		case s == string(AssetProviderProp) && a.Properties().Provider != "":
-			keys = append(keys, a.Properties().Provider)
-		case s == string(AssetAccountProp) && a.Properties().Account != "":
-			keys = append(keys, a.Properties().Account)
-		case s == string(AssetProjectProp) && a.Properties().Project != "":
-			keys = append(keys, a.Properties().Project)
-		case s == string(AssetClusterProp) && a.Properties().Cluster != "":
-			keys = append(keys, a.Properties().Cluster)
-		case s == string(AssetCategoryProp) && a.Properties().Category != "":
-			keys = append(keys, a.Properties().Category)
-		case s == string(AssetTypeProp) && a.Type().String() != "":
-			keys = append(keys, a.Type().String())
-		case s == string(AssetServiceProp) && a.Properties().Service != "":
-			keys = append(keys, a.Properties().Service)
-		case s == string(AssetProviderIDProp) && a.Properties().ProviderID != "":
-			keys = append(keys, a.Properties().ProviderID)
-		case s == string(AssetNameProp) && a.Properties().Name != "":
-			keys = append(keys, a.Properties().Name)
+		case s == string(AssetProviderProp):
+			key = a.Properties().Provider
+		case s == string(AssetAccountProp):
+			key = a.Properties().Account
+		case s == string(AssetProjectProp):
+			key = a.Properties().Project
+		case s == string(AssetClusterProp):
+			key = a.Properties().Cluster
+		case s == string(AssetCategoryProp):
+			key = a.Properties().Category
+		case s == string(AssetTypeProp):
+			key = a.Type().String()
+		case s == string(AssetServiceProp):
+			key = a.Properties().Service
+		case s == string(AssetProviderIDProp):
+			key = a.Properties().ProviderID
+		case s == string(AssetNameProp):
+			key = a.Properties().Name
 		case strings.HasPrefix(s, "label:"):
-			if label := a.Labels()[s[6:]]; label != "" {
-				keys = append(keys, label)
+			if labelKey := strings.TrimPrefix(s, "label:"); labelKey != "" {
+				key = a.Labels()[labelKey]
 			} else {
-				keys = append(keys, "__unallocated__")
+				// Don't allow aggregating on label ""
+				log.Errorf("Attempted to aggregate on empty label.")
+				return "", fmt.Errorf("Attempted to aggregate on invalid key: %s", s)
 			}
 		default:
-			_, err := ParseAssetProperty(s)
-			if err != nil {
-				if errors == nil {
-					errors = []error{}
-				}
-				errors = append(errors, fmt.Errorf("Invalid asset aggregation key: %s", s))
-			} else {
-				keys = append(keys, "__unallocated__")
-			}
+			log.Errorf("Attempted to aggregate on invalid key: %s", s)
+			return "", fmt.Errorf("Attempted to aggregate on invalid key: %s", s)
+		}
+
+		if key != "" {
+			keys = append(keys, key)
+		} else {
+			keys = append(keys, "__undefined__")
 		}
 	}
-	return strings.Join(keys, "/"), errors
+	return strings.Join(keys, "/"), nil
 }
 
 func toString(a Asset) string {
@@ -2462,9 +2463,9 @@ func (as *AssetSet) FindMatch(query Asset, aggregateBy []string) (Asset, error) 
 	as.RLock()
 	defer as.RUnlock()
 
-	matchKey, errors := key(query, aggregateBy)
-	if errors != nil {
-		return nil, fmt.Errorf("Bad key(s) given to FindMatch: %v", errors)
+	matchKey, err := key(query, aggregateBy)
+	if err != nil {
+		return nil, err
 	}
 	for _, asset := range as.assets {
 		if k, _ := key(asset, aggregateBy); k == matchKey {
@@ -2477,7 +2478,7 @@ func (as *AssetSet) FindMatch(query Asset, aggregateBy []string) (Asset, error) 
 
 // ReconciliationMatch attempts to find an exact match in the AssetSet on
 // (Category, ProviderID). If a match is found, it returns the Asset with the
-// intent to adjut it. If no match exists, it attempts to find one on only
+// intent to adjust it. If no match exists, it attempts to find one on only
 // (ProviderID). If that match is found, it returns the Asset with the intent
 // to insert the associated Cloud cost.
 func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
@@ -2486,17 +2487,23 @@ func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
 
 	// Full match means matching on (Category, ProviderID)
 	fullMatchProps := []string{string(AssetCategoryProp), string(AssetProviderIDProp)}
-	fullMatchKey, fullMatchErrors := key(query, fullMatchProps)
+	fullMatchKey, err := key(query, fullMatchProps)
 
 	// This should never happen because we are using enumerated properties,
 	// but the check is here in case that changes
-	if fullMatchErrors != nil {
-		return nil, false, fmt.Errorf("ReconciliationMatch: Invalid key(s) passed: %v", fullMatchErrors)
+	if err != nil {
+		return nil, false, err
 	}
 
 	// Partial match means matching only on (ProviderID)
 	providerIDMatchProps := []string{string(AssetProviderIDProp)}
-	providerIDMatchKey, _ := key(query, providerIDMatchProps)
+	providerIDMatchKey, err := key(query, providerIDMatchProps)
+
+	// This should never happen because we are using enumerated properties,
+	// but the check is here in case that changes
+	if err != nil {
+		return nil, false, err
+	}
 
 	var providerIDMatch Asset
 	for _, asset := range as.assets {
@@ -2545,9 +2552,9 @@ func (as *AssetSet) Insert(asset Asset) error {
 	defer as.Unlock()
 
 	// Determine key into which to Insert the Asset.
-	k, keyErrors := key(asset, as.aggregateBy)
-	if keyErrors != nil {
-		return fmt.Errorf("Attempted Insert on an `AssetSet` with invalid `aggregateBy` values: %v", keyErrors)
+	k, err := key(asset, as.aggregateBy)
+	if err != nil {
+		return err
 	}
 
 	// Add the given Asset to the existing entry, if there is one;
@@ -2615,9 +2622,9 @@ func (as *AssetSet) Set(asset Asset, aggregateBy []string) error {
 
 	// Expand the window to match the AssetSet, then set it
 	asset.ExpandWindow(as.Window)
-	k, keyErrors := key(asset, aggregateBy)
-	if keyErrors != nil {
-		return fmt.Errorf("Attempted to Set() Asset into AssetSet with invalid aggregateBy: %v", keyErrors)
+	k, err := key(asset, aggregateBy)
+	if err != nil {
+		return err
 	}
 	as.assets[k] = asset
 	return nil
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -109,9 +109,10 @@ func key(a Asset, aggregateBy []string) (string, error) {
 			if labelKey := strings.TrimPrefix(s, "label:"); labelKey != "" {
 				labelVal, ok := a.Labels()[labelKey]
 				if !ok {
-					labelVal = "__undefined__"
+					key = "__undefined__"
+				} else {
+					key = fmt.Sprintf("%s=%s", labelKey, labelVal)
 				}
-				key = fmt.Sprintf("%s=%s", labelKey, labelVal)
 			} else {
 				// Don't allow aggregating on label ""
 				return "", fmt.Errorf("Attempted to aggregate on invalid key: %s", s)
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -15,6 +15,10 @@ import (
 
 const timeFmt = "2006-01-02T15:04:05-0700"
 
+// UndefinedKey is used in composing Asset group keys if the group does not have that property defined.
+// E.g. if aggregating on Cluster, Assets in the AssetSet where Asset has no cluster will be grouped under key "__undefined__"
+const UndefinedKey = "__undefined__"
+
 // Asset defines an entity within a cluster that has a defined cost over a
 // given period of time.
 type Asset interface {
@@ -106,18 +110,16 @@ func key(a Asset, aggregateBy []string) (string, error) {
 				key = a.Labels()[labelKey]
 			} else {
 				// Don't allow aggregating on label ""
-				log.Errorf("Attempted to aggregate on empty label.")
 				return "", fmt.Errorf("Attempted to aggregate on invalid key: %s", s)
 			}
 		default:
-			log.Errorf("Attempted to aggregate on invalid key: %s", s)
 			return "", fmt.Errorf("Attempted to aggregate on invalid key: %s", s)
 		}
 
 		if key != "" {
 			keys = append(keys, key)
 		} else {
-			keys = append(keys, "__undefined__")
+			keys = append(keys, UndefinedKey)
 		}
 	}
 	return strings.Join(keys, "/"), nil
@@ -2468,7 +2470,9 @@ func (as *AssetSet) FindMatch(query Asset, aggregateBy []string) (Asset, error) 
 		return nil, err
 	}
 	for _, asset := range as.assets {
-		if k, _ := key(asset, aggregateBy); k == matchKey {
+		if k, err := key(asset, aggregateBy); err != nil {
+			return nil, err
+		} else if k == matchKey {
 			return asset, nil
 		}
 	}
@@ -2507,11 +2511,15 @@ func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
 
 	var providerIDMatch Asset
 	for _, asset := range as.assets {
-		if k, _ := key(asset, fullMatchProps); k == fullMatchKey {
+		if k, err := key(asset, fullMatchProps); err != nil {
+			return nil, false, err
+		} else if k == fullMatchKey {
 			log.DedupedInfof(10, "Asset ETL: Reconciliation[rcnw]: ReconcileRange Match: %s", fullMatchKey)
 			return asset, true, nil
 		}
-		if k, _ := key(asset, providerIDMatchProps); k == providerIDMatchKey {
+		if k, err := key(asset, providerIDMatchProps); err != nil {
+			return nil, false, err
+		} else if k == providerIDMatchKey {
 			// Found a partial match. Save it until after all other options
 			// have been checked for full matches.
 			providerIDMatch = asset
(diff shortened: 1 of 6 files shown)


pkg/kubecost/asset.go
```diff
@@ -2341,7 +2348,20 @@ func (as *AssetSet) AggregateBy(props []AssetProperty, opts *AssetAggregationOpt
...
+	if aggStrings == nil {
+		props = nil
+	} else {
+		for _, s := range aggStrings {
+			if prop, err := ParseAssetProperty(s); err == nil {
+				props = append(props, prop)
+			}
+		}
+	}
+
```
SheepGotoHeaven, line 2362: This and others maybe better written like

```
var props []AssetProperty
if aggStrings != nil {
   props = []AssetProperty{}
    for _, s := range aggStrings {
        if prop, err := ParseAssetProperty(s); err == nil {
            props = append(props, prop)
        }
    }
}
```

?

nikovacevic (commented):
The syntax here looks natural enough that I like this combined props-and-labels approach. But we do lose some implicit instruction by going from `[]AssetProperty` to `[]string`, so we should add some examples to the documentation--particularly just the comments of the `AggregateBy` and `key` functions.

The only detail here that I can pick on is the name "aggStrings". The relevant Go idiom is to avoid "stuttering" names. This usually refers to naming packages and their exported functions; e.g. prefer `widget.Store` to `widget.WidgetStore`, etc. (see https://blog.golang.org/package-names) In this case, I'm getting a similar vibe from `aggStrings []string`. Any other ideas? Maybe keep "props" or "properties", or something more descriptive like "aggregateBy" or "aggBy" for short?

pkg/kubecost/asset.go
```diff
@@ -61,43 +61,49 @@ type Asset interface {
...
 			keys = append(keys, a.Properties().ProviderID)
-		case prop == AssetNameProp && a.Properties().Name != "":
+		case s == string(AssetNameProp) && a.Properties().Name != "":
 			keys = append(keys, a.Properties().Name)
+		case strings.HasPrefix(s, "label:"):
+			if label := a.Labels()[s[6:]]; label != "" {
+				keys = append(keys, label)
+			} else {
+				keys = append(keys, "__unallocated__")
+			}
```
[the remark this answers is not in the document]
SheepGotoHeaven: +1 to both. Logging at a minimum; I support going ahead and returning an error that gets bubbled back up to the HTTP response.
API handler can send back something like

```
400 Bad Request
Unsupported aggregation field X
```

One note is I don't think `all` is actually supported as an omni-aggregator, and grouping on-everything can be achieved by passing invalid props. So it'd be good to enable `all` or `__all__` or something if we're also enforcing errors on unrecognized keys.

pkg/kubecost/asset.go
```diff
@@ -2304,11 +2310,12 @@ func (sa *SharedAsset) String() string {
...
 type AssetSet struct {
 	sync.RWMutex
-	assets   map[string]Asset
-	props    []AssetProperty
-	Window   Window
-	Warnings []string
-	Errors   []string
+	aggStrings []string
+	assets     map[string]Asset
+	props      []AssetProperty
```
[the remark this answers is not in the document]
SheepGotoHeaven: Works for me! I didn't see any other purpose it served, but kept it around in case that was metadata we'd want on a given AssetSet. We can always put it back if it becomes relevant :)

pkg/kubecost/asset.go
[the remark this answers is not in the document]
SheepGotoHeaven: Here's another idea. We could have the function return `(string, []error)`, collect errors for any unparsable strings, return the same `"/"`-delimited key that is returned now, and also any errors. Let the caller decide what to do with the information. Something like:

```
var errors []error
switch true {
// ...
default:
    if errors == nil {
        errors = []error{}
    }
    errors = append(errors, fmt.Errorf("Invalid asset aggregation key: %s", s))
}
return strings.Join(keys, "/"), errors
```

I also like making the func signature `(string, error)` and returning `"", error` on first error found...not sure which is more in line with idiomatic Go.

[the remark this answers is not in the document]
nikovacevic: Definitely more idiomatic to simply return the first `error`. Once you've found one, no point in continuing in a known-to-be-error state. The user can re-request and find the next error. But I like the thought process.

pkg/kubecost/asset.go
```diff
@@ -57,51 +57,69 @@ type Asset interface {
...
+		case s == string(AssetServiceProp) && a.Properties().Service != "":
 			keys = append(keys, a.Properties().Service)
-		case prop == AssetProviderIDProp && a.Properties().ProviderID != "":
+		case s == string(AssetProviderIDProp) && a.Properties().ProviderID != "":
 			keys = append(keys, a.Properties().ProviderID)
-		case prop == AssetNameProp && a.Properties().Name != "":
+		case s == string(AssetNameProp) && a.Properties().Name != "":
 			keys = append(keys, a.Properties().Name)
+		case strings.HasPrefix(s, "label:"):
+			if label := a.Labels()[s[6:]]; label != "" {
```
nikovacevic, line 105: This works in this particular case, but be careful accessing strings by index in Go, as there are some gotchas relating to the fact that you're technically accessing bytes rather than runes. (Read this if you're curious: https://blog.golang.org/strings)

Instead, let's try to use the `strings` package:
https://golang.org/pkg/strings/#TrimPrefix

nikovacevic: Also, just a question about behavior. If you ask for "label:", then you will get `__unallocated__` for every asset, right? I don't know what the "right" answer is here, but would you say that is correct and would be WAI?

pkg/kubecost/asset.go
```diff
@@ -2392,26 +2416,26 @@ func (as *AssetSet) Clone() *AssetSet {
 	as.RLock()
 	defer as.RUnlock()
 
+	var aggregateBy []string
+	if as.aggregateBy != nil {
+		aggregateBy := []string{}
+		for _, s := range as.aggregateBy {
+			aggregateBy = append(aggregateBy, s)
+		}
+	}
```
nikovacevic, line 2425: Nice catch here

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2458,53 @@ func (as *AssetSet) End() time.Time {
 // FindMatch attempts to find a match in the AssetSet for the given Asset on
 // the provided properties and labels. If a match is not found, FindMatch
 // returns nil and a Not Found error.
-func (as *AssetSet) FindMatch(query Asset, props []AssetProperty) (Asset, error) {
+func (as *AssetSet) FindMatch(query Asset, aggregateBy []string) (Asset, error) {
 	as.RLock()
 	defer as.RUnlock()
 
-	matchKey := key(query, props)
+	matchKey, errors := key(query, aggregateBy)
```
nikovacevic, line 2465: `errors` plural?

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2458,53 @@ func (as *AssetSet) End() time.Time {
...
 	}
 
-	return nil, fmt.Errorf("Asset not found to match %s on %v", query, props)
+	return nil, fmt.Errorf("Asset not found to match %s on %v", query, aggregateBy)
 }
 
 // ReconciliationMatch attempts to find an exact match in the AssetSet on
 // (Category, ProviderID). If a match is found, it returns the Asset with the
-// intent to adjuts it. If no match exists, it attempts to find one on only
+// intent to adjut it. If no match exists, it attempts to find one on only
```
nikovacevic, line 2480: lol this should be "adjust", right?

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2458,53 @@ func (as *AssetSet) End() time.Time {
...
 // to insert the associated Cloud cost.
 func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
 	as.RLock()
 	defer as.RUnlock()
 
 	// Full match means matching on (Category, ProviderID)
-	fullMatchProps := []AssetProperty{AssetCategoryProp, AssetProviderIDProp}
-	fullMatchKey := key(query, fullMatchProps)
+	fullMatchProps := []string{string(AssetCategoryProp), string(AssetProviderIDProp)}
+	fullMatchKey, fullMatchErrors := key(query, fullMatchProps)
```
nikovacevic, line 2489: Would definitely recommend using the idiomatic `err` variable name, unless there's a naming collision.

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2458,53 @@ func (as *AssetSet) End() time.Time {
...
+
+	// This should never happen because we are using enumerated properties,
+	// but the check is here in case that changes
```
nikovacevic, line 2492: Good comment, and good habit to check always.

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2458,53 @@ func (as *AssetSet) End() time.Time {
...
+	if fullMatchErrors != nil {
+		return nil, false, fmt.Errorf("ReconciliationMatch: Invalid key(s) passed: %v", fullMatchErrors)
+	}
 
 	// Partial match means matching only on (ProviderID)
-	providerIDMatchProps := []AssetProperty{AssetProviderIDProp}
-	providerIDMatchKey := key(query, providerIDMatchProps)
+	providerIDMatchProps := []string{string(AssetProviderIDProp)}
+	providerIDMatchKey, _ := key(query, providerIDMatchProps)
```
nikovacevic, line 2499: I would say that this is acceptable, given that we're controlling the props, but I do slightly prefer to check like above. In any case, let's keep it consistent.

pkg/kubecost/asset.go
```diff
@@ -2512,7 +2545,10 @@ func (as *AssetSet) Insert(asset Asset) error {
 	defer as.Unlock()
 
 	// Determine key into which to Insert the Asset.
-	k := key(asset, as.props)
+	k, keyErrors := key(asset, as.aggregateBy)
```
nikovacevic, line 2548: `err`

pkg/kubecost/assetprops.go
```diff
@@ -55,7 +55,7 @@ func ParseAssetProperty(text string) (AssetProperty, error) {
 		return AssetProjectProp, nil
 	case "provider":
 		return AssetProviderProp, nil
-	case "providerID":
+	case "providerid":
```
nikovacevic, line 58: Excellent catch

pkg/kubecost/asset.go
```diff
@@ -57,51 +57,69 @@ type Asset interface {
...
-// no props should be used (e.g. to aggregate all assets). Passing one or more
-// props will key by only those props.
-func key(a Asset, props []AssetProperty) string {
+// to determine if two Assets should be combined. Passing `nil` `aggregateBy` indicates
+// that all available `AssetProperty` keys should be used. Passing empty `aggregateBy` indicates that
+// no key should be used (e.g. to aggregate all assets). Passing one or more `aggregateBy`
+// values will key by only those values.
+// Valid values of `aggregateBy` elements are strings which are an `AssetProperty`, and strings prefixed
+// with `"label:"`.
+func key(a Asset, aggregateBy []string) (string, []error) {
```
nikovacevic, line 66: Ah, sorry, some of my other comments reference naming and the plurality of `[]error`. Would strongly recommend early-return on first error encountered. (That said, always open to opposing arguments; but this is a pattern that generally complicates the error handling logic of the caller in messy ways.)

nikovacevic (changes requested):
Let me know if you have any questions about the feedback. On a good track here!

pkg/kubecost/asset.go
SheepGotoHeaven: lol, yes

pkg/kubecost/asset.go
SheepGotoHeaven: I think supplying `?aggregate=label:` returns `__unallocated__` for every asset, unless there's some way to give an asset a label where the key is empty string. I guess maybe it'd be better to just ignore the key in that case? Or send a `400 Empty labels not allowed`-type response.

I'll switch to using `TrimPrefix`. On the behavior, I think it depends on whether we want an API that says "you asked for assets to be grouped on label ' ', and you got it", or "that makes no sense, so it isn't allowed". I'd probably lean toward the latter now that you point it out.

pkg/kubecost/asset.go
SheepGotoHeaven: I'll make the change.

My thought was that making a bad call to an API, it can be nice to get feedback on everything you did wrong rather than

request -> 400 response
fix the bad key
request -> new 400 response
fix the other bad key

That could be done with one Error obj, by aggregating the keys and writing them in. But it sounds like the Go-y thing to do is break on the first encountered problem, so I'll do that.

Can always have the top-level check on the actual API handler look at all supplied keys to give that sort of response, if we want that.

pkg/kubecost/asset.go
```diff
@@ -57,51 +57,70 @@ type Asset interface {
...
+			key = a.Properties().ProviderID
+		case s == string(AssetNameProp):
+			key = a.Properties().Name
+		case strings.HasPrefix(s, "label:"):
+			if labelKey := strings.TrimPrefix(s, "label:"); labelKey != "" {
+				key = a.Labels()[labelKey]
+			} else {
+				// Don't allow aggregating on label ""
+				log.Errorf("Attempted to aggregate on empty label.")
+				return "", fmt.Errorf("Attempted to aggregate on invalid key: %s", s)
```
nikovacevic, line 110: Kind of a nit here, but best practice is to return the error without logging. If you log and return the error, you're likely to have duplicate logs (because the caller will get a non-nil `err` and presumably log and/or propagate the `error` up the stack, where it will presumably get logged &c.

Logging should generally happen when (1) there is a warning that you want to record, but you're not returning an error or (2) when an error has propagated all the way to the top of a given routine, such that you don't have an opportunity to `return nil, err` again.

pkg/kubecost/asset.go
```diff
@@ -57,51 +57,70 @@ type Asset interface {
...
+			log.Errorf("Attempted to aggregate on invalid key: %s", s)
+			return "", fmt.Errorf("Attempted to aggregate on invalid key: %s", s)
 		}
-	}
 
-	return strings.Join(keys, "/")
+		if key != "" {
+			keys = append(keys, key)
+		} else {
+			keys = append(keys, "__undefined__")
```
nikovacevic, line 120: Ooh I think I like "undefined" for Assets as opposed to "unallocated". In `allocation.go`, we define some consts like
```go
// UnallocatedSuffix indicates an unallocated allocation property
const UnallocatedSuffix = "__unallocated__"
```
Can we do that for `asset.go`, too? Just to formalize it.

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2459,59 @@ func (as *AssetSet) End() time.Time {
...
 	defer as.RUnlock()
 
-	matchKey := key(query, props)
+	matchKey, err := key(query, aggregateBy)
+	if err != nil {
+		return nil, err
+	}
 	for _, asset := range as.assets {
-		if key(asset, props) == matchKey {
+		if k, _ := key(asset, aggregateBy); k == matchKey {
```
nikovacevic, line 2471: This is tricky, actually--do you think we should check `err` and `return nil, err` here, now that `key` can return an error? I don't necessarily see a reason not to.

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2459,59 @@ func (as *AssetSet) End() time.Time {
...
 	}
 
-	return nil, fmt.Errorf("Asset not found to match %s on %v", query, props)
+	return nil, fmt.Errorf("Asset not found to match %s on %v", query, aggregateBy)
 }
 
 // ReconciliationMatch attempts to find an exact match in the AssetSet on
 // (Category, ProviderID). If a match is found, it returns the Asset with the
-// intent to adjuts it. If no match exists, it attempts to find one on only
+// intent to adjust it. If no match exists, it attempts to find one on only
```
nikovacevic, line 2481: yasss squash the typos!!

pkg/kubecost/asset.go
```diff
@@ -2434,44 +2459,59 @@ func (as *AssetSet) End() time.Time {
...
 
 	var providerIDMatch Asset
 	for _, asset := range as.assets {
-		if key(asset, fullMatchProps) == fullMatchKey {
+		if k, _ := key(asset, fullMatchProps); k == fullMatchKey {
 			log.DedupedInfof(10, "Asset ETL: Reconciliation[rcnw]: ReconcileRange Match: %s", fullMatchKey)
 			return asset, true, nil
 		}
-		if key(asset, providerIDMatchProps) == providerIDMatchKey {
+		if k, _ := key(asset, providerIDMatchProps); k == providerIDMatchKey {
```
nikovacevic, line 2514: Bolt would call checking the `err` here the "belt and suspenders" approach. I'm fine without, but just pointing out that we could check here if we really wanted to.

pkg/kubecost/asset.go
```diff
@@ -2614,12 +2662,10 @@ func (as *AssetSet) accumulate(that *AssetSet) (*AssetSet, error) {
 	}
 
 	// In the case of an AssetSetRange with empty entries, we may end up with
-	// an incoming as without props, even though we are trying to aggregate
-	// by props. This handles that case, assigning the correct props.
-	if !propsEqual(as.props, that.props) {
```
nikovacevic, line 2619: I think we need to keep this call to `!propsEqual`, and perhaps I should make the comment more clear. Basically, during accumulation, it's possible to accumulate over a range where the first `as` is empty (and thus has empty `as.props`), but at some point we hit "real" data that has non-empty `props`; e.g. accumulating some 5d range where the JSON would look like:
```json
[
  {},
  {},
  {
    "default": {...},
    "kubecost": {...}
  },
  {
    "default": {...},
    "kubecost": {...}
  },
  {
    "default": {...},
    "kubecost": {...}
  },
]
```
The first two sets don't have props, but the last three have `props == ["namespace"]`. So we want to persist those props, not the empty ones.

Does that make sense? I get some funky code smell from this, so we may just have to revisit with a less error-prone design at some point.

nikovacevic: I will make a note to go look in the unit tests to make sure we have coverage for this case. This is a nice little tricky spot to be able to rely on unit test coverage.

nikovacevic (changes requested):
Good progress. Let me know if you have any questions!

pkg/kubecost/asset.go
SheepGotoHeaven: Will do! Glad you like it. After one of our chats I was concerned that `__unallocated__` would have a kind of overloaded meaning (w/ one meaning in allocations context vs. another meaning here)

pkg/kubecost/asset.go
SheepGotoHeaven: Maybe I'm missing something, but here's what I was thinking:

We have

```
if !propsEqual(as.props, that.props) {
  if len(as.props) == 0 {
    as.props = that.props
  }
}
```

But `if len(as.props) == 0` then either

`!propsEqual(as.props, that.props) == true`  --> do the assignment
or
`propsEqual(as.props, that.props) == false`  --> both `as.props` and `that.props` are have `len() == 0` and the assignment is a no-op, so it's okay to do.

pkg/kubecost/asset.go
nikovacevic: I actually like the idea of `400 Bad Request: must provide a label`, just to explicitly avoid the issue altogether. Not a requirement (if users do this and get "unexpected" results, it's kind of on them...) but it's a good idea.

pkg/kubecost/asset.go
nikovacevic: Good teamwork :laughing:

pkg/kubecost/asset.go
nikovacevic: Agreed 100% that this would be a slick thing to add to an http handler

pkg/kubecost/asset.go
nikovacevic: This is a tricky one because some of the cases it handles are technically erroneous, and thus the expected behavior is ill-defined.

One distinction to note is that `len(as.props) == 0` actually handles two cases: `as.props == nil`  and `as.props == []string{}`. Different values, but both evaluate to length 0.

|  | as.props == that.props | as.props != that.props |
| --- | --- |---|
| len(as.props) > 0  | Consistent aggregation (keep accumulating) | Conflicting aggregation (error?) |
| len(as.props) == 0 | Consistent non-aggregation (keep accumulating) | Started without props, but ended up with some; most likely scenario is that we want to aggregate by `that.props` but the window reaches too far back, so we started with some `nil` `props`. Best to adopt the new, non-nil props. |

Again, the tricky part is that aggregating by `nil` causes no aggregation, whereas aggregating by `[]string{}` causes full aggregation. They are both generate `len(props) == 0`, but they cause opposite aggregation outcomes.

My takeaways are:
- Always question logic like this, because it does look suspect. But when in doubt, it's safer to maintain existing logic.
- We need to be able to reference a unit test in this comment that enforces expacted behavior. If that unit test fails, it means we broke it. If it passes, it means that the author of the change hasn't broken anything.
- It would be justified to revisit the design that makes this a sticking point. If we "define the problem out of existence" by picking a better architecture, then we won't have to worry about this in the future. I don't know the answer here, but it's worth getting it right.

pkg/kubecost/asset.go
```diff
@@ -57,51 +61,68 @@ type Asset interface {
...
-			AssetProviderProp,
-			AssetAccountProp,
-			AssetProjectProp,
-			AssetCategoryProp,
-			AssetClusterProp,
-			AssetTypeProp,
-			AssetServiceProp,
-			AssetProviderIDProp,
-			AssetNameProp,
+	if aggregateBy == nil {
```
nikovacevic, line 73: Note that we only do this when `aggregateBy == nil`, not when `len(aggregateBy) == 0`, or else we'd be revoking the ability to aggregate everything into one, nameless Asset. (Open question: maybe we _should_ merge the two and say that you can't aggregate all? We will need to be able to maintain non-aggregation, so that behavior has to stay.)

nikovacevic (commented):
Hope that commend about `propsEqual` makes sense--reach out if you have more questions. Maybe we just look into the unit tests to see if we have one that asserts proper aggregation and accumulation when spanning ranges that start and/or end with `nil` AssetSets? Then, we can simply prove to ourselves that the code works and call it good :smile:

nikovacevic:
> Hope that commend about `propsEqual` makes sense--reach out if you have more questions. Maybe we just look into the unit tests to see if we have one that asserts proper aggregation and accumulation when spanning ranges that start and/or end with `nil` AssetSets? Then, we can simply prove to ourselves that the code works and call it good

Ah, in fact we do have one! https://github.com/kubecost/cost-model/blob/develop/pkg/kubecost/asset_test.go#L852-L868

So, if that's passing, I think I'm happy with the change. We might also consider writing a few similar tests for edge-cases where `props` is at risk of being mis-matched, but that would be over-achievement.

pkg/kubecost/asset_test.go
```diff
@@ -675,24 +676,36 @@ func TestAssetSet_AggregateBy(t *testing.T) {
...
+	// 1e aggregateBy []string=["label:test"]
+	as = generateAssetSet(startYesterday)
+	err = as.AggregateBy([]string{"label:test"}, nil)
+	if err != nil {
+		t.Fatalf("AssetSet.AggregateBy: unexpected error: %s", err)
+	}
+	fmt.Println(as.assets)
+	assertAssetSet(t, as, "1e", window, map[string]float64{
+		"__undefined__": 53.00,
+		"test=test":     7.00,
```
nikovacevic, line 701: Boom, nice! :rocket: :boom:

nikovacevic (approved):
LGTM! Thank you!

PR #667, opened 2021-01-19
Ajay tripathy fix pvcalls

AjayTripathy:


AjayTripathy:
previously, this was cloning our cache on every call of the metrics emission loop. This is way faster-- cc @mbolt35

mbolt35 (approved):
LGTM

PR #669, opened 2021-01-19
Added env var for disabling OOC metric emissions

michaelmdresser:
See https://github.com/kubecost/kubecost-cost-model/pull/310

### DIFF
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -57,6 +57,7 @@ const (
 
 	UTCOffsetEnvVar = "UTC_OFFSET"
 
+	OutOfClusterEnabledEnvVar = "OUT_OF_CLUSTER_ENABLED"
 	CacheWarmingEnabledEnvVar = "CACHE_WARMING_ENABLED"
 	ETLEnabledEnvVar          = "ETL_ENABLED"
 )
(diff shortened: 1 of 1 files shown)


pkg/env/costmodelenv.go
```diff
@@ -57,6 +57,7 @@ const (
 
 	UTCOffsetEnvVar = "UTC_OFFSET"
 
+	OutOfClusterEnabledEnvVar = "OUT_OF_CLUSTER_ENABLED"
```
AjayTripathy, line 60: As in https://github.com/kubecost/cost-analyzer-helm-chart/pull/735 , we need to be a little more specific with naming to avoid confusion with our other ooc pipeline

pkg/env/costmodelenv.go
```diff
@@ -313,6 +314,11 @@ func GetParsedUTCOffset() time.Duration {
 	return offset
 }
 
+// IsOutOfClusterEnabled determines if OOC metrics (kubecost_cloud_credit_total and kubecost_cloud_expense_total) are emitted.
+func IsOutOfClusterEnabled() bool {
```
AjayTripathy, line 318: A little more specific of a name

AjayTripathy (approved):
Nits about naming but once that's fixed and made consistent with the helm chart, LGTM

michaelmdresser:
Closing this because the need for this variable is only in closed source. No reason to pass it through an extra layer of indirection for no reason.

PR #672, opened 2021-01-21
OOC Credits (GCP)

SheepGotoHeaven:
Initially tried to save the credit metadata in an array (name, type, amount). That blew my memory usage way up and pods kept getting evicted, so I trimmed it down to just store the cumulative credit amount for each Asset.

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -184,6 +184,71 @@ const (
 	SharedAssetType
 )
 
+// AssetCredit represents credit applied to the cost of an Asset
+type AssetCredit struct {
+	Amount float64
+	Name   string
+	Type   string
+}
+
+// Clone returns a copy of AssetCredit
+func (ac AssetCredit) Clone() AssetCredit {
+	return AssetCredit{
+		Amount: ac.Amount,
+		Name:   ac.Name,
+		Type:   ac.Type,
+	}
+}
+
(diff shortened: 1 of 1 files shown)


pkg/kubecost/asset.go
```diff
@@ -479,7 +480,7 @@ func (ca *Cloud) SetAdjustment(adj float64) {
 
 // TotalCost returns the Asset's total cost
 func (ca *Cloud) TotalCost() float64 {
-	return ca.Cost + ca.adjustment
+	return ca.Cost + ca.adjustment + ca.Credit
```
AjayTripathy, line 483: What unit is credits in? positive currency or negative?  Can you make a note on the Credit field?

pkg/kubecost/asset.go
```diff
@@ -550,7 +551,7 @@ func (ca *Cloud) Add(a Asset) Asset {
 	any.SetProperties(props)
 	any.SetLabels(labels)
 	any.adjustment = ca.Adjustment() + a.Adjustment()
-	any.Cost = (ca.TotalCost() - ca.Adjustment()) + (a.TotalCost() - a.Adjustment())
+	any.Cost = (ca.TotalCost() - ca.Adjustment() - ca.Credit) + (a.TotalCost() - a.Adjustment() - ca.Credit)
```
AjayTripathy, line 554: See above about positive/negative.

AjayTripathy (approved):

PR #673, opened 2021-01-22
Add provider ID to ClusterNode map key when possible

michaelmdresser:
## High-level
AWS sometimes re-uses node names, but the `provider_id` allows us to differentiate the two nodes. This PR updates `ClusterNodes` to return a map is, where possible, also keyed by `provider_id`, instead of just node name and cluster name.

## Implementation Details
I found the current implementation of `ClusterNodes` very difficult to work with for this change, so I refactored it to build the various maps of node info -> data separately and then merge them at the end. This approach is (in my opinion) cleaner, but the real benefit is it allows us to work with data that does not include `provider_id` alongside data that does. If you look at `buildNodeMap`, you'll see that the data which comes from metrics that do not have `provider_id` (e.g. the `kube_node_status_memory_bytes` metric and related extracted data in `ramBytesMap`) can be joined with data from metrics that do (e.g. the `node_gpu_hourly_cost` metric and related extracted data in `gpuCostMap`).

The idea of the join can be illustrated with this example:
```
"name1, cluster1" -> 10 cores  
"name1,cluster1,providerid1" -> $10  
"name1,cluster1,providerid1" -> $20  
....  
"name1,cluster1,providerid1" -> $10, 10 cores  
"name1,cluster1,providerid1" -> $20, 10 cores  
....
```

Also requires https://github.com/kubecost/kubecost-cost-model/pull/317 because of the change to keying.

## Background

Here is some data showing re-used `node` (the name) with differing `provider_id`.

`{endpoint="model",instance="ip-10-249-94-109.ec2.internal",instance_type="m5.4xlarge",job="kubecost-core-application",namespace="kubecost",node="ip-10-249-94-109.ec2.internal",pod="kubecost-core-application-0bedc1f-85b98bcd97-cn4ct",pod_name="kubecost-core-application-0bedc1f-85b98bcd97-cn4ct",provider_id="aws:///us-east-1b/i-0409c24ce282c8dab",region="us-east-1",service="kubecost-core-application"}	0.000274`

`{endpoint="model",instance="ip-10-249-94-109.ec2.internal",instance_type="m5.4xlarge",job="kubecost-core-application",namespace="kubecost",node="ip-10-249-94-109.ec2.internal",pod="kubecost-core-application-0bedc1f-85b98bcd97-s9t4l",pod_name="kubecost-core-application-0bedc1f-85b98bcd97-s9t4l",provider_id="aws:///us-east-1b/i-00873a38a553cb9f3",region="us-east-1",service="kubecost-core-application"}	0.0317`

Metrics used in `ClusterNode` and if they include `provider_id`:
| metric name                            | has provider_id |
| -------------------------------------- | --------------- |
| kube_node_status_capacity_cpu_cores    | no              |
| node_cpu_hourly_cost                   | yes             |
| kube_node_status_capacity_memory_bytes | no              |
| node_ram_hourly_cost                   | yes             |
| node_gpu_hourly_cost                   | yes             |
| kubecost_node_is_spot                  | yes             |
| node_cpu_seconds_total                 | no              |
| container_memory_working_set_bytes     | no              |
| node_total_hourly_cost                 | yes             |

Further information about re-use of node names: https://github.com/kubernetes/kubernetes/issues/39893


pkg/costmodel/cluster.go
```diff
@@ -469,343 +480,33 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset
...
-		}
-	}
-
-	// Mapping of cluster/node=cpu for computing resource efficiency
-	clusterNodeCPUTotal := map[string]float64{}
-	// Mapping of cluster/node:mode=cpu for computing resource efficiency
-	clusterNodeModeCPUTotal := map[string]map[string]float64{}
+	cpuCostMap, clusterAndNameToType1 := buildCPUCostMap(resNodeCPUCost, cp.ParseID)
+	ramCostMap, clusterAndNameToType2 := buildRAMCostMap(resNodeRAMCost, cp.ParseID)
+	gpuCostMap, clusterAndNameToType3 := buildGPUCostMap(resNodeGPUCost, cp.ParseID)
```
michaelmdresser, line 485: These three queries include node type information, so the map constructors also return a mapping from `(cluster name, node name)` to `node type`.

pkg/costmodel/cluster_helpers_test.go
```diff
@@ -0,0 +1,529 @@
...
+					ProviderID:   "prov_node1_B",
+					NodeType:     "type1",
+					CPUCost:      0.087,
+					CPUBreakdown: &ClusterCostsBreakdown{},
+					RAMBreakdown: &ClusterCostsBreakdown{},
+				},
+			},
+		},
+		{
+			name: "all fields + overlapping node names",
```
michaelmdresser, line 258: All of the test cases (this one in particular) are really long. I'm open to ideas about how to make them smaller or more readable.

pkg/costmodel/cluster_helpers.go
```diff
@@ -0,0 +1,631 @@
...
+// complexity.
+func buildNodeMap(
+	cpuCostMap, ramCostMap, gpuCostMap map[NodeIdentifier]float64,
+	cpuCoresMap, ramBytesMap, ramUserPctMap,
+	ramSystemPctMap map[nodeIdentifierNoProviderID]float64,
+	cpuBreakdownMap map[nodeIdentifierNoProviderID]*ClusterCostsBreakdown,
+	activeDataMap map[NodeIdentifier]activeData,
+	preemptibleMap map[NodeIdentifier]bool,
+	clusterAndNameToType map[nodeIdentifierNoProviderID]string,
+) map[NodeIdentifier]*Node {
```
mbolt35, line 567: Would love some other's input on this style - I'm definitely more of a fan of one parameter per line, but I think we're currently keeping ALL THE THINGS tm on a single line (which is miserable for a split screen setup -- would love to pay attention to horizontal margins). The only comment I can make about the current style we're using is "it's consistent." Just to be clear here, my style preferences have definitely shifted to more adaptive the older I get, so I yield to you guys for opinions. My only "must have" is that we choose something that can be auto-formatted via IDE and we keep it 100% consistent across all of our code. @nikovacevic @AjayTripathy @SheepGotoHeaven @Sean-Holcomb

michaelmdresser:
Fixing some issues introduced by a merge.

michaelmdresser:
Done. There were changes in `develop` that introduced a new map-building loop `resLabels` to fill in `Node.Labels`. Moved this into its own helper function and updated `buildNodeMap` and the test. The name for the preemptibility query also changed.

pkg/costmodel/cluster_helpers.go
AjayTripathy: Let's address style consistency in our Friday meeting and not block the review here.

AjayTripathy (approved):

PR #675, opened 2021-01-27
e2 fixes

AjayTripathy:
Manually tested against old e2 pricing and verified no change. @michaelmdresser can we get this in and later pick up a unit test for those special cases?


michaelmdresser:
What are those special cases?

michaelmdresser (approved):

PR #681, opened 2021-01-29
ComputeCostData gets CPU and RAM requests from the Kubernetes API

michaelmdresser:
The intended result of this change is reducing load on Prometheus for things it is not needed for. The one caveat of this change is a modification of the output of data for pods that stop existing in the cluster during the query lookback window (which is currently 2 minutes). Since we will be querying the k8s API at a time where the pod does not exist, it will not have any information about it. We choose to handle this by not outputting request information, only usage information.

I welcome feedback about the omission of request information in the case of the pod died within the query window. Prometheus's inherent "best guess" nature means this shouldn't introduce unexpected behavior. The biggest case I see this being a problem in is environments which have high pod churn and many low-duration pods. While this is a possible case, a user in this situation will have to begin tuning Prometheus to scrape faster to even have data available about such low-duration pods. If this becomes an issue, we can look at this again. I suspect the answer will just be to shorten the sleep for our computation of this metric.

The other argument for this not being a huge problem is that most users of Kubernetes don't over-request for pods, meaning usage data will almost always be the data that matters.

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -475,6 +475,12 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 				// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-types
 				// for the units of memory and CPU.
 				ramRequestBytes := container.Resources.Requests.Memory().Value()
+
+				// Because RAM (and CPU) information isn't coming from Prometheus, it won't
+				// have a timestamp associated with it. We need to provide a timestamp,
+				// otherwise the vector op that gets applied to take the max of usage
+				// and request won't work properly and will only take into account
+				// usage.
 				RAMReqV := []*util.Vector{
 					{
 						Value:     float64(ramRequestBytes),
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -381,8 +381,8 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 			continue // because ordering is important for the allocation model (all PV's applied to the first), just dedupe if it's already been added.
 		}
 		// The _else_ case for this statement is the case in which the container has been
-		// deleted so we have usage information but not request information. We insert partial
-		// data in that case.
+		// deleted so we have usage information but not request information. In that case,
+		// we return partial data for CPU and RAM: only usage and not requests.
 		if pod, ok := currentContainers[key]; ok {
 			podName := pod.GetObjectMeta().GetName()
 			ns := pod.GetObjectMeta().GetNamespace()
@@ -553,6 +553,9 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 			// CPU and RAM requests are obtained from the Kubernetes API.
 			// If this case has been reached, the Kubernetes API will not
 			// have information about the pod because it no longer exists.
+			//
+			// The case where this matters is minimal, mainly in environments
+			// with very short-lived pods that over-request resources.
 			RAMReqV := []*util.Vector{{}}
 			CPUReqV := []*util.Vector{{}}
 
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -469,8 +469,11 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 				// recreate the key and look up data for this container
 				newKey := NewContainerMetricFromValues(ns, podName, containerName, pod.Spec.NodeName, clusterID).Key()
 
-				// k8s.io/apimachinery/pkg/api/resource/amount.go for explanations of
-				// the types that memory/cpu requests are
+				// k8s.io/apimachinery/pkg/api/resource/amount.go and
+				// k8s.io/apimachinery/pkg/api/resource/quantity.go for
+				// details on the "amount" API. See
+				// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-types
+				// for the units of memory and CPU.
 				ramRequestBytes := container.Resources.Requests.Memory().Value()
 				RAMReqV := []*util.Vector{
 					{
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -477,7 +477,8 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 				ramRequestBytes := container.Resources.Requests.Memory().Value()
 				RAMReqV := []*util.Vector{
 					{
-						Value: float64(ramRequestBytes),
+						Value:     float64(ramRequestBytes),
+						Timestamp: float64(time.Now().UTC().Unix()),
 					},
 				}
 
@@ -485,7 +486,8 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 				cpuRequestMilliCores := container.Resources.Requests.Cpu().MilliValue()
 				CPUReqV := []*util.Vector{
 					{
-						Value: float64(cpuRequestMilliCores) / 1000,
+						Value:     float64(cpuRequestMilliCores) / 1000,
+						Timestamp: float64(time.Now().UTC().Unix()),
 					},
 				}
 
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -226,9 +226,7 @@ const (
 )
 
 func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyzerCloud.Provider, window string, offset string, filterNamespace string) (map[string]*CostData, error) {
-	queryRAMRequests := fmt.Sprintf(queryRAMRequestsStr, window, offset, window, offset)
 	queryRAMUsage := fmt.Sprintf(queryRAMUsageStr, window, offset, window, offset)
-	queryCPURequests := fmt.Sprintf(queryCPURequestsStr, window, offset, window, offset)
 	queryCPUUsage := fmt.Sprintf(queryCPUUsageStr, window, offset)
 	queryGPURequests := fmt.Sprintf(queryGPURequestsStr, window, offset, window, offset, 1.0, window, offset)
 	queryPVRequests := fmt.Sprintf(queryPVRequestsStr)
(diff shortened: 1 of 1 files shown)


pkg/costmodel/costmodel.go
```diff
@@ -487,21 +469,29 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
...
 				newKey := NewContainerMetricFromValues(ns, podName, containerName, pod.Spec.NodeName, clusterID).Key()
 
-				RAMReqV, ok := RAMReqMap[newKey]
-				if !ok {
-					klog.V(4).Info("no RAM requests for " + newKey)
-					RAMReqV = []*util.Vector{{}}
+				// k8s.io/apimachinery/pkg/api/resource/amount.go for explanations of
+				// the types that memory/cpu requests are
+				ramRequestBytes := container.Resources.Requests.Memory().Value()
+				RAMReqV := []*util.Vector{
```
michaelmdresser, line 475: I believe it is okay to only set `Value` because the main consumer of this data [only uses Value](https://github.com/kubecost/cost-model/blob/mmd-ccd-reduce-prom/pkg/costmodel/metrics.go#L936). The [only other consumer](https://github.com/kubecost/cost-model/blob/mmd-ccd-reduce-prom/pkg/costmodel/router.go#L421) is just an API handler.

dwbrown2:
Wow, this is so great! Two high-level questions on my end:

1. Is there anything we can say about api backwards compatibility? We commit to k8s 1.8 in open source and 1.10 I believe in Free open core. Would this change that?
2. Do this now mean that our emission loops "works" without a valid Prom endpoint?! Obviously it wouldn't have usage data included... but that would be big!

michaelmdresser:
> Wow, this is so great! Two high-level questions on my end:
> 
>     1. Is there anything we can say about api backwards compatibility? We commit to k8s 1.8 in open source and 1.10 I believe in Free open core. Would this change that?
> 
>     2. Do this now mean that our emission loops "works" without a valid Prom endpoint?! Obviously it wouldn't have usage data included... but that would be big!

1. I believe this won't be impacted by older versions of the Kubernetes API. I'm having trouble finding the exact API docs for v1.8, but the concept of CPU and memory requests is old.
2. Not yet. This just lightens the load we're putting on Prometheus by getting CPU and memory request information from the Kubernetes API instead of from Prometheus. We still rely on Prometheus for GPU requests, physical volume requests/usage/state, and network requests (zone, region, internet).

dwbrown2:
> > Wow, this is so great! Two high-level questions on my end:
> > ```
> > 1. Is there anything we can say about api backwards compatibility? We commit to k8s 1.8 in open source and 1.10 I believe in Free open core. Would this change that?
> > 
> > 2. Do this now mean that our emission loops "works" without a valid Prom endpoint?! Obviously it wouldn't have usage data included... but that would be big!
> > ```
> 
> 1. I believe this won't be impacted by older versions of the Kubernetes API. I'm having trouble finding the exact API docs for v1.8, but the concept of CPU and memory requests is old. The main question is whether the path to that information is the same in the API. I suspect it is.
> 2. Not yet. This just lightens the load we're putting on Prometheus by getting CPU and memory request information from the Kubernetes API instead of from Prometheus. We still rely on Prometheus for GPU requests, physical volume requests/usage/state, and network requests (zone, region, internet).

Cool! one follow up on 2. -- what happens currently if you don't have a valid PromQL endpoint today... do we just not emit GPU + PV metrics but still give you CPU/mem? Or do we panic / do something similar? I personally feel like moving towards the former is a big win all by itself, given how CPU/mem are the most core metrics here.

michaelmdresser:
> Cool! one follow up on 2. -- what happens currently if you don't have a valid PromQL endpoint today... do we just not emit GPU + PV metrics but still give you CPU/mem? Or do we panic / do something similar? I personally feel like moving towards the former is a big win all by itself, given how CPU/mem are the most core metrics here.

At the moment, [error handling for Prometheus queries](https://github.com/kubecost/cost-model/blob/mmd-ccd-reduce-prom/pkg/costmodel/costmodel.go#L288) means that any errors that occur during queries cause the entire `CostData` building function to fail. In this case, nothing is emitted. It would be possible to add logic return Kubernetes API-only information.

AjayTripathy:
> > Cool! one follow up on 2. -- what happens currently if you don't have a valid PromQL endpoint today... do we just not emit GPU + PV metrics but still give you CPU/mem? Or do we panic / do something similar? I personally feel like moving towards the former is a big win all by itself, given how CPU/mem are the most core metrics here.
> 
> At the moment, [error handling for Prometheus queries](https://github.com/kubecost/cost-model/blob/mmd-ccd-reduce-prom/pkg/costmodel/costmodel.go#L288) means that any errors that occur during queries cause the entire `CostData` building function to fail. In this case, nothing is emitted. It would be possible to add logic to return Kubernetes API-only information if a failure occurs.

I believe there's a subtlety here that's worth pointing out. I don't think that the emission fails if this errors, simply the Prometheus gauges don't update.

pkg/costmodel/costmodel.go
```diff
@@ -487,21 +469,29 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 				// recreate the key and look up data for this container
 				newKey := NewContainerMetricFromValues(ns, podName, containerName, pod.Spec.NodeName, clusterID).Key()
 
-				RAMReqV, ok := RAMReqMap[newKey]
-				if !ok {
-					klog.V(4).Info("no RAM requests for " + newKey)
-					RAMReqV = []*util.Vector{{}}
+				// k8s.io/apimachinery/pkg/api/resource/amount.go for explanations of
+				// the types that memory/cpu requests are
+				ramRequestBytes := container.Resources.Requests.Memory().Value()
```
AjayTripathy, line 474: I'm not sure what in amount.go guarantees this to be in bytes. is it actually resource/quantity.go that shims Value->ScaledValue(0) ? Does using ScaledValue(0) make more sense to ensure this is in bytes? (definitely not trying to make this a leading question, just wondering.)

AjayTripathy (commented):
Could you add some notes on how this was tested? Here's my suggested methodology:

Run kubecost on two GCP clusters
check the /model/metrics endpoint and validate that the requests and allocation stats 

(allocation is a computed stat that is max(usage and requests), so you also want to test a container without any requests set) are the same.

then post your findings in the review?

pkg/costmodel/costmodel.go
AjayTripathy: This seems fine, but let's make sure we shake this out in testing by deploying this and making a note in the review

michaelmdresser:
### Test Output (from /model/metrics)

Before is running CM@90986f41(develop) KCM@d36d7c9(develop)

After is running CM@46a44d8d(mmd-ccd-reduce-prom) KCM@d36d7c9(develop)

`hello-app` is an as-simple-as-possible server that exposes one endpoint. It shouldn't be under any load and doesn't request resources.

#### CPU

Before:  

Cost Model: `container_cpu_allocation{container="cost-model",instance="gke-michael-1-default-pool-d8a6f420-71cx",namespace="kubecost",node="gke-michael-1-default-pool-d8a6f420-71cx",pod="kubecost-cost-analyzer-7c96c5dbd6-s22kt"} 0.2`

Hello App: `container_cpu_allocation{container="hello-app",instance="gke-michael-1-default-pool-d8a6f420-jdn1",namespace="default",node="gke-michael-1-default-pool-d8a6f420-jdn1",pod="hello-app-deployment-6f44887c88-j4ms5"} 0.1`

After:  

Cost Model: `container_cpu_allocation{container="cost-model",instance="gke-michael-cm681-test-default-pool-3bdd5047-khrn",namespace="kubecost",node="gke-michael-cm681-test-default-pool-3bdd5047-khrn",pod="kubecost-cost-analyzer-68ff5859fb-nhzhd"} 0.0011457452263499044`

Hello App: `container_cpu_allocation{container="hello-app",instance="gke-michael-cm681-test-default-pool-3bdd5047-3830",namespace="default",node="gke-michael-cm681-test-default-pool-3bdd5047-3830",pod="hello-app-deployment-6f44887c88-cqw98"} 5.265937510648071e-07`

#### Memory

Before:  

Cost Model: `container_memory_allocation_bytes{container="cost-model",instance="gke-michael-1-default-pool-d8a6f420-71cx",namespace="kubecost",node="gke-michael-1-default-pool-d8a6f420-71cx",pod="kubecost-cost-analyzer-7c96c5dbd6-s22kt"} 5.767168e+07`

Hello App: `container_memory_allocation_bytes{container="hello-app",instance="gke-michael-1-default-pool-d8a6f420-jdn1",namespace="default",node="gke-michael-1-default-pool-d8a6f420-jdn1",pod="hello-app-deployment-6f44887c88-j4ms5"} 991232`

After:  

Cost Model: `container_memory_allocation_bytes{container="cost-model",instance="gke-michael-cm681-test-default-pool-3bdd5047-khrn",namespace="kubecost",node="gke-michael-cm681-test-default-pool-3bdd5047-khrn",pod="kubecost-cost-analyzer-68ff5859fb-nhzhd"} 1.3625344e+07`

Hello App: `container_memory_allocation_bytes{container="hello-app",instance="gke-michael-cm681-test-default-pool-3bdd5047-3830",namespace="default",node="gke-michael-cm681-test-default-pool-3bdd5047-3830",pod="hello-app-deployment-6f44887c88-cqw98"} 1.261568e+06`

pkg/costmodel/costmodel.go
michaelmdresser: Modified comment for a better explanation. I think using `Value()` is fine because the unit of memory in K8s [is bytes](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-types). It is just a shim to `ScaledValue(0)` but I think using value is clear given the expectation set by general K8s documentation.

michaelmdresser:
#### Test Results

Before is running develop/develop

After is running CM@dba6dd6 KCM@develop

Before: `container_memory_allocation_bytes{container="hello-app",instance="gke-michael-1-default-pool-d8a6f420-16jp",namespace="default",node="gke-michael-1-default-pool-d8a6f420-16jp",pod="hello-app-2g"} 2.147483648e+09`

After: `container_memory_allocation_bytes{container="hello-app",instance="gke-michael-cm681-test-default-pool-3bdd5047-3830",namespace="default",node="gke-michael-cm681-test-default-pool-3bdd5047-3830",pod="hello-app-2g"} 2.147483648e+09`

Before the latest commit, the allocation wasn't being calculated correctly because the new `CPUReqV` and `RAMReqV` were not setting `Timestamp`, so `getContainerAllocation` couldn't merge them correctly with `ApplyVectorOp`.

michaelmdresser:
Another test update after letting it sit for a while.

#### Before
cm681_allocation_before_week


#### After
cm681_allocation_after_week

AjayTripathy (approved):

PR #684, opened 2021-02-02
Include adjustment in idle computation

nikovacevic:
## Changes
- Compute `adjustmentRate` for scaling per-resource costs of idle allocation
- Remove profiling log

## Testing
- Unit test repro'd, and now passes


pkg/kubecost/allocation_test.go
```diff
@@ -901,34 +901,35 @@ func TestAllocationSet_ComputeIdleAllocations(t *testing.T) {
...
-	// +-----------------------------------------+------+------+------+------+
-	//   cluster1 subtotal                        100.00  50.00  40.00  10.00
-	// +-----------------------------------------+------+------+------+------+
-	//   cluster1 allocated                        48.00   6.00  16.00   6.00
-	// +-----------------------------------------+------+------+------+------+
-	//   cluster1 idle                             72.00  44.00  24.00   4.00
-	// +-----------------------------------------+------+------+------+------+
+	//     nodes                                  100.00  55.00  44.00  11.00      -10.00
+	// +-----------------------------------------+------+------+------+------+------------+
+	//   cluster1 subtotal                        100.00  50.00  40.00  10.00      -10.00
```
AjayTripathy, line 909: Is it quick to add a test for an adjustment in the positive direction also? I feel like these are conceptually the same, but I think it's worth calling it out in a test.

AjayTripathy (approved):

pkg/kubecost/allocation_test.go
nikovacevic: Yep, done!

PR #685, opened 2021-02-05
Read secret and save values to env variable

Sean-Holcomb:
Added methods for retrieving azure storage config values from secret
Tested by setting azure storage secret in values.yaml


AjayTripathy (approved):

mbolt35 (approved):
LGTM

PR #693, opened 2021-02-16
ComputeCostData: don't return on prometheus errors

michaelmdresser:
This will let cost-model run as only an exporter (without depending
on a Prometheus to query from). It will output allocation data
that is only comprised of _requests_ for CPU and memory. This is
possible because of the recent change to ComputeCostData that
removes the dependency on Prometheus for CPU and memory requests.

Also refactored install docs for more clarity and to highlight the new exporter-only install method.

Tested on my cluster and I start seeing `container_allocation` metrics (without a configured Prometheus) for CPU and memory after this change. 

The API of course does not behave properly and we could probably more gracefully signal that to the user. Thoughts? 

Should `ComputeCostDataRange` also have this change applied? I believe only `ComputeCostData` is used for metrics exporting.


CONTRIBUTING.md
```diff
@@ -13,10 +13,7 @@ Follow these steps to build from source and deploy:
 
 1. `docker build --rm -f "Dockerfile" -t <repo>/kubecost-cost-model:<tag> .`
 2. Edit the [pulled image](https://github.com/kubecost/cost-model/blob/master/kubernetes/deployment.yaml#L25) in the deployment.yaml to <repo>/kubecost-cost-model:<tag>
-3. Set [this environment variable](https://github.com/kubecost/cost-model/blob/master/kubernetes/deployment.yaml#L33) to the address of your prometheus server
-4. `kubectl create namespace cost-model`
```
dwbrown2, line 17: Let's touch base before we launch markdown changes!

michaelmdresser: Definitely. I don't want to break links from elsewhere especially.

I think the changes I've made make the flow clearer (in my head at least). I think https://github.com/kubecost/cost-model/blob/mmd-exporter-only/INSTALL.md looks better than the scattering of info between `deploying-as-a-pod.md`, `CONTRIBUTING.md`, etc.

michaelmdresser: Backed off the md changes in light of our discussion.

dwbrown2:
@nikovacevic @mbolt35 would either of you have time to give this a review this week?

nikovacevic (approved):
LGTM

(The only place we call this in `pkg/costmodel/metrics.go`, all we do is log the error on return anyways. So seems relatively harmless, but also like it wouldn't necessarily "do" anything.)

README.md
```diff
@@ -25,6 +25,8 @@ Here is a summary of features enabled by this cost model:
 You can deploy Kubecost on any Kubernetes 1.8+ cluster in a matter of minutes, if not seconds. 
 Visit the Kubecost docs for [recommended install options](https://docs.kubecost.com/install). Compared to building from source, installing from Helm is faster and includes all necessary dependencies. 
 
+If you prefer to install _only_ the open source code (or want to use `cost-model` as just a metric exporter), see [`INSTALL.md`](INSTALL.md) for instructions.
```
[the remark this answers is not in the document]
michaelmdresser: Ah I backed out the README changes after talking with Webb -- those will come later in a different PR.

PR #696, opened 2021-02-18
Make a single YAML for easy exporter-only deploy

michaelmdresser:
Hold for https://github.com/kubecost/cost-model/pull/693


dwbrown2 (approved):
LGTM, only question is do we want these in /exporter. Your call @michaelmdresser!

michaelmdresser:
I think keeping it under `kubernetes/` is good, and I can't think of a better name than `exporter/` for the subdirectory, so I'll stick with it.

PR #697, opened 2021-02-19
Instructions for exporter-only install

michaelmdresser:
Depends on https://github.com/kubecost/cost-model/pull/696

Also requires the changes from https://github.com/kubecost/cost-model/pull/693 to be published in the `cost-model` container.

### DIFF
--- a/kubecost-exporter.md
+++ b/kubecost-exporter.md
@@ -94,8 +94,6 @@ avg(count_over_time(container_memory_allocation_bytes{namespace="kube-system"}[1
 avg(avg_over_time(node_ram_hourly_cost[1d] )) by (node)
 ```
 
-#### Monthly cost of a label
-
 
 ## Setting Cost Alerts
 
--- a/kubecost-exporter.md
+++ b/kubecost-exporter.md
@@ -1,0 +1,117 @@
+# Running Kubecost as a Prometheus metric exporter
+
+Running Kubecost as a Prometheus metric exporter allows you to export various cost metrics to Prometheus without setting up any other Kubecost dependencies. Doing so lets you write PromQL queries to calculate the cost and efficiency of any Kubernetes concept, e.g. namespace, service, label, deployment, etc. You can also calculate the cost of different Kubernetes resources, e.g. nodes, PVs, LoadBalancers, and more. Finally, you can do other interesting things like create custom alerts via AlertManager and custom dashboards via Grafana. 
+
+## Installing
+
+> Note: all deployments of Kubecost function as a Prometheus metric exporter. We strongly recommend helm as an install path to take advantage of Kubecost’s full potential. [View recommended install](http://docs.kubecost.com/install).
+
+If you would prefer to not use the recommended install option and just deploy the Kubecost open source cost model as a metric exporter, you can follow these steps:
+
+
+1. Apply the combined YAML:
+
+    ```
+    kubectl apply -f https://raw.githubusercontent.com/kubecost/cost-model/develop/kubernetes/exporter/exporter.yaml --namespace cost-model
+    ```
+
+    > If you want to use a namespace other than `cost-model`, you will have to edit the `ClusterRoleBinding` after applying the YAML to change `subjects[0].namespace`. You can do this with `kubectl edit clusterrolebinding cost-model`.
+
+2. To verify that metrics are available:
+
+    ```
+    kubectl port-forward --namespace cost-model service/cost-model 9003
+    ```
+
+    Visit [http://localhost:9003/metrics](http://localhost:9003/metrics) to see exported metrics
+
+Add Kubecost scrape config to Prom ([more](https://prometheus.io/docs/introduction/first_steps/#configuring-prometheus))
+```
+- job_name: cost-model
+  scrape_interval: 1m
+  scrape_timeout: 10s
+  metrics_path: /metrics
+  scheme: http
+  static_configs:
+    - targets: [‘cost-model.cost-model.:9003/metrics’]
+```
+
+Done! Kubecost is now exporting cost metrics. See the following sections for different metrics available and query examples.
+
+## Available Prometheus Metrics 
+
+| Metric       | Description                                                                                            |
+| ------------ | ------------------------------------------------------------------------------------------------------ |
+| node_cpu_hourly_cost | Hourly cost per vCPU on this node  |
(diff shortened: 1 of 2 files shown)


kubecost-exporter.md
```diff
@@ -0,0 +1,117 @@
...
+If you would prefer to not use the recommended install option and just deploy the Kubecost open source cost model as a metric exporter, you can follow these steps:
+
+
+1. Apply the combined YAML:
+
+    ```
+    kubectl apply -f https://raw.githubusercontent.com/kubecost/cost-model/develop/kubernetes/exporter/exporter.yaml --namespace cost-model
+    ```
+
+    > If you want to use a namespace other than `cost-model`, you will have to edit the `ClusterRoleBinding` after applying the YAML to change `subjects[0].namespace`. You can do this with `kubectl edit clusterrolebinding cost-model`.
```
dwbrown2, line 18: I think most users would just modify the YAML before applying. I personally think it's ok to drop this in open source docs.

kubecost-exporter.md
```diff
@@ -0,0 +1,117 @@
...
+
+2. To verify that metrics are available:
```
dwbrown2, line 20: To me, this feels like just verifying that Step 1 worked. Where as adding the scrape config is the true Step 2. Are you thinking about this differently?

kubecost-exporter.md
```diff
@@ -0,0 +1,117 @@
...
+
+Done! Kubecost is now exporting cost metrics. See the following sections for different metrics available and query examples.
+
+## Available Prometheus Metrics 
+
+| Metric       | Description                                                                                            |
+| ------------ | ------------------------------------------------------------------------------------------------------ |
+| node_cpu_hourly_cost | Hourly cost per vCPU on this node  |
+| node_gpu_hourly_cost | Hourly cost per GPU on this node  |
+| node_ram_hourly_cost   | Hourly cost per Gb of memory on this node                       |
```
dwbrown2, line 47: We now have `kubecost_load_balancer_cost` to add to this list, yeah?

kubecost-exporter.md
```diff
@@ -0,0 +1,117 @@
...
+
+```
+avg_over_time(container_memory_allocation_bytes{namespace="kube-system",pod=~"fluentd.*"}[1d])
+  * on (pod,node) group_left
+avg(count_over_time(container_memory_allocation_bytes{namespace="kube-system"}[1d:1m])/60) by (pod,node)
+  * on (node) group_left
+avg(avg_over_time(node_ram_hourly_cost[1d] )) by (node)
+```
+
+#### Monthly cost of a label
```
dwbrown2, line 97: this is still a TODO as well

kubecost-exporter.md
michaelmdresser: My first instinct when following one of our other install instructions was to just modify `kubectl apply` with a different value for the `--namespace` flag.  I ran into this issue and it took me a while to fix it because everything was running but the `cost-model` pod was constantly logging failures (while staying up) because it wasn't getting the right permission scopes and I had to track down why.

Basically, I included this instruction because I would have benefited from it. Definitely interested in a better place for it, because right now I think it visually clutters up the "easy 2 step process" idea.

kubecost-exporter.md
michaelmdresser: I think adding the scrape config is definitely the true step 2, but I think I'm in favor of keeping it like this because this is a much more rapid-feedback way of ensuring that everything is running as expected and it gives a user the instinct for what to reach for if they need to debug.

kubecost-exporter.md
michaelmdresser: I'm not seeing it with the latest (public) `cost-model` container image. Is that a feature coming in this release?

kubecost-exporter.md
michaelmdresser: Right, I think I glossed over it. I'll try to craft this query tomorrow.

kubecost-exporter.md
dwbrown2: Should be available for the last 5+ releases: https://github.com/kubecost/cost-model/blob/f845eb02c200c4da993827ad2d62bf47b5922f84/pkg/costmodel/metrics.go#L636

michaelmdresser: Ah, it's probably not there for me because I don't have any load balancers so its not being emitted.

michaelmdresser: Are we documenting these metrics somewhere other than PROMETHEUS.md? It also doesn't have `kubecost_load_balancer_cost`, not to mention others like:
* `kubecost_node_is_spot`
* `kubecost_network_zone_egress_cost`
* `kubecost_network_region_egress_cost`
* `kubecost_network_internet_egress_cost`
* `kubecost_cluster_management_cost`
* `kubecost_cluster_info`

kubecost-exporter.md
dwbrown2: Happy to help if you need me! I'd start with the simplest version and then we can add complexity in future queries.

kubecost-exporter.md
dwbrown2: We have not documented anywhere yet! In this section, it seems like `kubecost_load_balancer_cost` and `kubecost_cluster_management_cost` are the most relevant? 

I haven't added network_egress metrics yet primarily because they are in a different repo.

dwbrown2 (approved):
LGTM, feel free to merge when ready or we can wait until just before launch event!

kubecost-exporter.md
```diff
@@ -0,0 +1,117 @@
...
+sum(kubecost_load_balancer_cost) by (namespace)
+```
+
+#### Monthly rate of each namespace’s CPU request
+
+```
+sum(container_cpu_allocation * on (node) group_left node_ram_hourly_cost) by (namespace) * 730
+```
+
+#### Historical spend for all `fluentd` pods in the `kube-system` namespace
```
michaelmdresser, line 87: This looks like its actually historical _memory_ spend. Is that right @dwbrown2?

dwbrown2: That's correct! Thanks for catching.

For these initial queries, I just wanted to keep it simple... as opposed to showing mem + cpu + loadbalancer, etc etc.

kubecost-exporter.md
michaelmdresser: I'll open a follow-up PR to add these metrics both here and to PROMETHEUS.md

PR #699, opened 2021-02-21
disable legacy api option for scale

AjayTripathy:
Testing done: disable this and data will not load on details.html


SheepGotoHeaven (approved):
LGTM

PR #712, opened 2021-02-25
Update Kubernetes libraries to v0.20.4

michaelmdresser:
Context needed to be added to a few API calls as a result. This upgrade
was performed by running:

`go get k8s.io/client-go@v0.20.4`

And then resolving compiler errors by adding `context.Background()` where needed.

Also updates `compat.go` to prefer the stable region label and not hardcode it. I opted to not update the instance type label at this time, but it should eventually prefer `v1.LabelInstanceTypeStable`.

Goes with https://github.com/kubecost/kubecost-cost-model/pull/345


michaelmdresser:
I had to update `golang.org/x/sync` and `github.com/prometheus/client_model` to get the build in CI to work (even though it was working locally with `docker build`).

michaelmdresser:
Tested with the open source-only and no Prometheus deploy path, which is working (`/metrics` looks good, implying the API watchers are working correctly) and the logs don't indicate anything out of order.

pkg/cloud/csvprovider.go
```diff
@@ -230,7 +230,11 @@ func NodeValueFromMapField(m string, n *v1.Node, useRegion bool) string {
 	mf := strings.Split(m, ".")
 	toReturn := ""
 	if useRegion {
-		toReturn = n.Labels[v1.LabelZoneRegion] + ","
+		if region, ok := util.GetRegion(n.Labels); ok {
+			toReturn = region + ","
+		} else {
+			klog.Infof("[ERROR] Getting region based on labels failed")
```
nikovacevic, line 236: Could use `log.Errorf(...)` here

pkg/util/compat.go
```diff
@@ -4,11 +4,13 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+// See https://kubernetes.io/docs/reference/labels-annotations-taints/
+
 func GetRegion(labels map[string]string) (string, bool) {
-	if _, ok := labels[v1.LabelZoneRegion]; ok {
+	if _, ok := labels[v1.LabelTopologyRegion]; ok { // Label as of 1.17
+		return labels[v1.LabelTopologyRegion], true
+	} else if _, ok := labels[v1.LabelZoneRegion]; ok { // deprecated label
```
nikovacevic, line 12: Was this, specifically, tested? If so, just done manually on one of our GCP clusters?

nikovacevic (approved):

pkg/util/compat.go
michaelmdresser: So, this is used here: https://github.com/kubecost/cost-model/blob/mmd-update-k8s-libraries/pkg/costmodel/costmodel.go#L977

This result is, eventually, consumed by the emitter of `kubecost_node_is_spot` (among other things) which includes region info. I am seeing that accurate on a GKE cluster (showing `us-east1` for a cluster running in `us-east1-b`).

michaelmdresser: This matches up with pre-change behavior, as far as I can tell with a fresh cluster and the regular helm chart.

michaelmdresser: It's also used in the providers, including GCP provider: https://github.com/kubecost/cost-model/blob/mmd-update-k8s-libraries/pkg/cloud/gcpprovider.go#L1180 and I'm not seeing related errors in the logs.

pkg/cloud/csvprovider.go
michaelmdresser: I agree, but I was copying the error log formatting elsewhere in the function. I'm in favor of switching all of them. Thoughts?

AjayTripathy: klog.Infof is legacy so please do migrate them where it's quick and you're in the code anyway.

michaelmdresser: Do we prefer stdlib `log`?

AjayTripathy: Sorry for the confusion. It's our own logging utility: 	"github.com/kubecost/cost-model/pkg/log"

michaelmdresser: Okay. Removed `klog` from the dependencies on `csvprovider.go` and updated the offending info logs to be error logs without the ERROR prefix.

pkg/util/compat.go
nikovacevic: Sounds good

AjayTripathy (approved):

PR #713, opened 2021-02-25
Ajay tripathy fix logspam

AjayTripathy:
* None of the namespace log errors were really informational. It's totally reasonable to not have namespace level labels or annotations.
* A storageclass can not exist for a variety of reasons, so this data was not that relevant


mbolt35 (approved):
lgtm

PR #714, opened 2021-02-26
Ajay tripathy fix flat filter

AjayTripathy:
Testing done: see unit test.

Filters were incorrectly assigning all shared costs to the namespace that was not filtered out.


nikovacevic (commented):
Going to run some more tests and follow up

AjayTripathy:
This feels fine, I will review this and merge. Thanks for picking it up!

PR #716, opened 2021-02-26
update error messages

kalsky:


michaelmdresser (approved):
Thanks for taking the time to improve the error messaging!

PR #717, opened 2021-03-01
fix panic on -1 index

hightoxicity:
This commit is fixing a panic when cost analyzer is fetching GCP pricing.


AjayTripathy:
Thanks so much for this. Something changed in the GCP Pricing API and we'll need to patch this back in.

AjayTripathy (approved):

AjayTripathy:
Hi @hightoxicity , thanks so much for this. We've updated 1.75.1 and 1.75.0 with this fix-- if you restart the pod it should get picked up.

hightoxicity:
Thx :-)

PR #722, opened 2021-03-03
Ajay tripathy fix index error

AjayTripathy:
Finish cleaning up the index error issue, as we did in our patch to master. This code is already running in master, for a hotfix. so this is just to sync our branches and catch develop up.


michaelmdresser (approved):
Based on https://github.com/kubecost/cost-model/pull/718/files for reference.

PR #725, opened 2021-03-06
NoStoreAPI Error Handling

mbolt35:
Add change that disallows query results to continue if there exists a NoStoreAPI warning. This error should be retriable in ETL pipeline, as no store api may just indicate a process restarting.

### DIFF
--- a/pkg/prom/error.go
+++ b/pkg/prom/error.go
@@ -12,6 +12,16 @@ import (
 // errorType used to check HasError
 var errorType = reflect.TypeOf((*error)(nil)).Elem()
 
+// NoStoreAPIWarning is a warning that we would consider an error. It returns partial data relating only to the
+// store apis which were reachable. In order to ensure integrity of data across all clusters, we'll need to identify
+// this warning and convert it to an error.
+const NoStoreAPIWarning string = "No StoreAPIs matched for this query"
+
+// IsNoStoreAPIWarning checks a warning to determine if it is equivalent to a no store API query.
+func IsNoStoreAPIWarning(warning string) bool {
+	return strings.EqualFold(warning, NoStoreAPIWarning)
+}
+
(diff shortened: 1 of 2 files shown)


AjayTripathy:
This LGTM, but a few things:

1. Were you able to test this? I think you can quickly test by turning off your thanos-store but leaving the sidecar running.
2. Does it make sense to also disable the sidecar-store connection altogether?
3. Any other warnings that should be spat out as errors? Or should at least propagate to diagnostics.html?

dwbrown2:
> This LGTM, but a few things:
> 
> 1. Were you able to test this? I think you can quickly test by turning off your thanos-store but leaving the sidecar running.
> 2. Does it make sense to also disable the sidecar-store connection altogether?
> 3. Any other warnings that should be spat out as errors? Or should at least propagate to diagnostics.html?

The second point feels like a separate issue/discussion to me

mbolt35:
> This LGTM, but a few things:
> 
> 1. Were you able to test this? I think you can quickly test by turning off your thanos-store but leaving the sidecar running and kicking off an ETL rebuild
> 2. Does it make sense to also disable the sidecar-as-a-store connection altogether?
> 3. Any other warnings that should be spat out as errors? Or should at least propagate to diagnostics.html?

1. Difficult to test this exact scenario without overlapping multiple stores. It may be possible to hard configure the store - I'll look into it a bit more. I was able to recreate using exact queries in the query interface.
2. Seems fine to me, but we've done a pretty good job including offset 3h in all of our queries (which will always exclude the sidecar). It doesn't seem _bad_ to have sidecar act as a querier, but it also doesn't hurt.
3. I only know of this one warning.

AjayTripathy (approved):

AjayTripathy:
This looks to be a major improvement so as long as it doesn't regress, I'm ok to ship.

PR #726, opened 2021-03-09
Attempt to emit gpu count

Sean-Holcomb:
Emit gpu count and use number to multiple gpu costs in cost model. Get accurate GPU count for Azure GPU nodes, previously hard coded to 1.

Testing:
The Emission was tested by the presence of the node_gpu_count value in prometheus.

Calculation of GPU cost should follow the formula node_gpu_count * node_gpu_hourly_cost * number_of_hours. To test this on AWS  nodes were given a fake type of p2.8xlarge, which has a node count of 8. Since the nodes with these fake types did not have GPUs before node_gpu_count and node_gpu_hourly_cost before they received the fake type the total gpu cost of one of these nodes should follow the above equations with the number_of_hours being equal to the amount of time they had the fake type. 8 * .793521 * ~57/60 =  ~6.03

Screen Shot 2021-03-11 at 11 50 19 AM
Screen Shot 2021-03-11 at 11 50 10 AM
Screen Shot 2021-03-11 at 12 04 17 PM

On Azure to validate the real GPU count being used I ran a NC12s. When I initialized it it showed 1 GPU in count and once the code was updated to GPU count for azure the node_gpu_count jumped to 2.

Screen Shot 2021-03-11 at 11 24 19 AM

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -452,6 +452,7 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset 
 	queryNodeRAMBytes := fmt.Sprintf(`avg_over_time(avg(kube_node_status_capacity_memory_bytes) by (cluster_id, node)[%s:%dm]%s)`, durationStr, minsPerResolution, offsetStr)
 	queryNodeGPUCost := fmt.Sprintf(`sum_over_time((avg(node_gpu_count) by (cluster_id, node) * on(node, cluster_id) group_right avg(node_gpu_hourly_cost) by (cluster_id, node, instance_type, provider_id))[%s:%dm]%s) * %f`, durationStr, minsPerResolution, offsetStr, hourlyToCumulative)
 	queryNodeGPUCount := fmt.Sprintf(`avg_over_time(avg(node_gpu_count) by (cluster_id, node)[%s:%dm]%s)`, durationStr, minsPerResolution, offsetStr)
+	queryNodeGPUHourlySum := fmt.Sprintf(`sum_over_time(avg(node_gpu_hourly_cost) by (cluster_id, node, instance_type, provider_id)[%s:%dm]%s) * %f`, durationStr, minsPerResolution, offsetStr, hourlyToCumulative)
 	queryNodeCPUModeTotal := fmt.Sprintf(`sum(rate(node_cpu_seconds_total[%s:%dm]%s)) by (kubernetes_node, cluster_id, mode)`, durationStr, minsPerResolution, offsetStr)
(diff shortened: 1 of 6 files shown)


pkg/costmodel/metrics.go
```diff
@@ -577,6 +578,11 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud
 			Help: "node_gpu_hourly_cost hourly cost for each gpu on this node",
 		}, []string{"instance", "node", "instance_type", "region", "provider_id"})
 
+		gpuCountGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "node_gpu_count",
+			Help: "node_gpu_count hourly cost for each gpu on this node",
```
mbolt35, line 583: GPU Count, not hourly cost.

mbolt35 (approved):
lgtm

pkg/costmodel/cluster.go
```diff
@@ -449,7 +450,8 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset
 	queryNodeCPUCores := fmt.Sprintf(`avg_over_time(avg(kube_node_status_capacity_cpu_cores) by (cluster_id, node)[%s:%dm]%s)`, durationStr, minsPerResolution, offsetStr)
 	queryNodeRAMCost := fmt.Sprintf(`sum_over_time((avg(kube_node_status_capacity_memory_bytes) by (cluster_id, node) * on(cluster_id, node) group_right avg(node_ram_hourly_cost) by (cluster_id, node, instance_type, provider_id))[%s:%dm]%s) / 1024 / 1024 / 1024 * %f`, durationStr, minsPerResolution, offsetStr, hourlyToCumulative)
 	queryNodeRAMBytes := fmt.Sprintf(`avg_over_time(avg(kube_node_status_capacity_memory_bytes) by (cluster_id, node)[%s:%dm]%s)`, durationStr, minsPerResolution, offsetStr)
-	queryNodeGPUCost := fmt.Sprintf(`sum_over_time((avg(node_gpu_hourly_cost * %d.0 / 60.0) by (cluster_id, node, provider_id))[%s:%dm]%s)`, minsPerResolution, durationStr, minsPerResolution, offsetStr)
+	queryNodeGPUCost := fmt.Sprintf(`sum_over_time((avg(node_gpu_count) by (cluster_id, node) * on(node, cluster_id) group_right avg(node_gpu_hourly_cost) by (cluster_id, node, instance_type, provider_id))[%s:%dm]%s) * %f`, durationStr, minsPerResolution, offsetStr, hourlyToCumulative)
```
AjayTripathy, line 453: Because this is a sort of data migration, I wouldn't do this in a query. I'd separately track cost and count, then modify cost * count in application code, not a prometheus query, because at startup this will be wrong since counts will be zero. Then in code if you can't find a "count" you can assume it is 1.

AjayTripathy (changes requested):

pkg/costmodel/cluster.go
Sean-Holcomb: Updated in most recent push

AjayTripathy (approved):

PR #727, opened 2021-03-09
Workflow to build on every PR to develop

michaelmdresser:
Since we've decided to use Actions for integration tests, I think we should standardize this repo's testing with Actions as well.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,0 +1,25 @@
+name: Develop PR - build, test
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


AjayTripathy:
This lgtm, just wondering how this works with the nightlies/what the image name will end up being?

michaelmdresser:
This is just a build for the sake of PR validation so we aren't depending on two different CI tools (dropping CircleCI). The image isn't pushed anywhere.

Sean-Holcomb (approved):

PR #732, opened 2021-03-10
fix allocation of shared overhead

AjayTripathy:
Tested via unit test and deploy to demo:

Screen Shot 2021-03-09 at 10 15 59 PM


nikovacevic (approved):
LGTM! Lovely solution. I just added a comment to the test case to explain how that final number is reached. It all adds up.

(Hopefully I can finally fix Properties soon, so that they're less painful to work with :sweat:)

PR #735, opened 2021-03-17
Add new screenshot

dwbrown2:


michaelmdresser:
Should this be displayed in the README?

dwbrown2:
Was thinking that would be part two of this! Just wanted to make sure this image looked good to you guys. The one drawback that I see is that it's rather large... I like the high res part though and felt like the tradeoff was worth it. Let me know what you think!

michaelmdresser (approved):
I find it clear, but I definitely think it would be improved for a viewer with a good description. I have the benefit of experience :smile:

dwbrown2:
Meaning in the README or want to see captions on the image? 

Another option is that we can replace this with a linked video in the future.

PR #736, opened 2021-03-17
Sean/aws lb recon

Sean-Holcomb:
Emit hostname of Load Balancer to Prometheus if IP address is missing and use that value to create a ProviderID in AWS for Load Balancers.

tested on EKS-Integration-Test which after changes began showing ingress-id in prometheus and a Provider ID for the Kubernetes service Load Balancer.

Screen Shot 2021-03-16 at 5 43 24 PM
Screen Shot 2021-03-16 at 5 19 11 PM

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -2286,6 +2286,5 @@ func (aws *AWS) ParseLBID(id string) string {
 		}
 		return id
 	}
-	log.Infof("awsprovider.ParseLBID: successful parse %s, %v", match[1], match)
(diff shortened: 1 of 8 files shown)


AjayTripathy:
Great start! To be clear, I'm not entirely sure this works for Application LBs https://docs.aws.amazon.com/eks/latest/userguide/alb-ingress.html

but let's start by reconciling what we can for NLBs
https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html

AjayTripathy (approved):

PR #740, opened 2021-03-17
Update Buffer to allocate a string to prevent []byte Pinning

mbolt35:
Comment Explaining Reasoning:

```go
        // The following commented code will map an existing byte slice's underlying array
	// to a string. The purpose is to avoid new allocation of a string from an existing
	// byte array. This is a solid optimization to save total allocations, but effectively
	// pins the string's existence to the underlying byte array. This can prevent GC of the
	// bytes in a few cases, and we should just return a newly allocated string to allow
	// future optimization.
	//return *(*string)(unsafe.Pointer(&b))
```

### DIFF
--- a/pkg/util/buffer.go
+++ b/pkg/util/buffer.go
@@ -384,9 +384,17 @@ func intDataSize(data interface{}) int {
 	return 0
 }
 
-// Direct byte to string conversion that doesn't allocate.
+// Conversion from byte slice to string
 func bytesToString(b []byte) string {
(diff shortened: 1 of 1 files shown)


nikovacevic (approved):
:fire: :fire: :fire:

AjayTripathy:
I did a quick back of the envelope test loading data on a scale cluster and it seems to have lower steady state memory at the cost of some additional CPU during load-- so lgtm

AjayTripathy (approved):

PR #742, opened 2021-03-17
Replace encoding/json with jsoniter

SheepGotoHeaven:
Pretty straightforward conversion of `json.Marshal` and `json.Unmarshal` to `jsoniter.ConfigCompatibleWithStandardLibrary.Marshal()` and `jsoniter.ConfigCompatibleWithStandardLibrary.Unmarshal()`.

## Implementation Details

Did not remove `encoding/json` completely because some modules use `json.Marshaler` interface or `json.NewDecoder`. These are not supplied by the `jsoniter` module.

Also, doing the exact drop-in mode described in the [jsoniter docs](https://github.com/json-iterator/go#usage):
```
import jsoniter "github.com/json-iterator/go"

var json = jsoniter.ConfigCompatibleWithStandardLibrary
```

results in issues when we have multiple files defining a package. Only one file in the package is allowed to declare the `json` top-level variable, and it seems strange to have the module globally available without an `import`, so instead I have changed instances of `json.(Unm|M)arshal` to `jsoniter.ConfigCompatibleWithStandardLibrary.(Unm|M)arshal`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/googleapis/gax-go v2.0.2+incompatible // indirect
 	github.com/gophercloud/gophercloud v0.2.0 // indirect
+	github.com/json-iterator/go v1.1.10
 	github.com/jszwec/csvutil v1.2.1
 	github.com/julienschmidt/httprouter v1.2.0
 	github.com/lib/pq v1.2.0
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/csv"
-	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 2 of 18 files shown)


mbolt35:
Just my 2 cents, but can we clean up the verbosity of the call? Either by what the suggest in their README:

```
import jsoniter "github.com/json-iterator/go"

var json = jsoniter.ConfigCompatibleWithStandardLibrary
// ...
json.Marshal(&data)
```

or by adding a `github.com/kubecost/cost-model/pkg/util/json`  and adding `Marshal` and `Unmarshal` methods that delegate the calls to `jsoniter.ConfigCompatibleWithStandardLibrary`?

Any other thoughts here?

SheepGotoHeaven:
Their suggested README example lead to some issues that might be part of my Go-newbishness or might be because their example is overly-simplistic.

A json util delegating to the library sounds like a good idea to me if that works for y'all

mbolt35:
> Their suggested README example lead to some issues that might be part of my Go-newbishness or might be because their example is overly-simplistic.
> 
> A json util delegating to the library sounds like a good idea to me if that works for y'all

I'm good with the delegation strategy - then I think the only updates to the source would be the import -- we could also control tests with other libraries (or swap back to the standard library) without any code changes going forward.

mbolt35 (approved):
Looks great!

PR #743, opened 2021-03-18
Load Balancer WIP

Sean-Holcomb:
Distribute Load Balancer total cost over Network costs of Allocation in Allocation API. Determines the number of hours that a given Allocation was up and had a Load Balancer applied to it, and distributes the total cost of the Load Balancer proportionate to that length of time for each Allocation that used it.

Methodology:
During a given period of time (Query Window) the start and end of the Load Balancer service is determined along with the start and end times of any allocation that has the LB service's selector on it. For each of these allocations the overlap between the allocation's start and end time and the LB service's start and end time is determined in hours. Once the total of these hours is calculated, for each allocation, the LoadBalancerCost is set with Total_LB_Cost * Overlap_Hours / Total_Hours

Limitations: 
This is only available from the Allocation Api so it can only be seen on the Reports page or multi-day queries with etl enabled on details.html. Additionally aggregation by service on Reports does not accurately aggregate the cost of the load balancer into the correct service. This is best view via drill down to the pod level.

Testing: 
Tested on the EKS-Integration-Test cluster, which has a single load balancer applied to the 3 containers of cost-analyzer. With a single day query from the allocations compute endpoint, your can see the total cost calculated by the hourly rate.
Screen Shot 2021-03-19 at 4 46 54 PM
 And the cost of the LB being distributed to the Allocations it is applied to.
 
Screen Shot 2021-03-19 at 4 46 54 PM

Here are the results on the Reports page:
Screen Shot 2021-03-19 at 4 50 48 PM


AjayTripathy:
Could you explain where the costs are being allocated a bit more? The loadbalancer cost gets placed into the network category? Does it make sense to put it on a new, separate category called "LoadBalancer"? Do we have the ability to disentangle pod network egress costs from loadbalancers in this setup?

pkg/costmodel/allocation.go
```diff
@@ -1586,6 +1603,77 @@ func applyUnmountedPVCs(window kubecost.Window, podMap map[podKey]*Pod, pvcMap m
...
+				e = lb.End
+			}
+			hours := e.Sub(s).Hours()
+			totalHours += hours
+			allocHours[alloc] = hours
+		}
+
+		// Distribute cost of service once total hours in calculated
+		for alloc, hours := range allocHours {
+			alloc.NetworkCost += lb.Cost * hours / totalHours
```
nikovacevic, line 1671: Did we end up just deciding to lump this into `NetworkCost`, or should we add a `LBCost` field? I would probably advocate for the latter. @AjayTripathy @mbolt35

pkg/costmodel/allocation.go
```diff
@@ -1586,6 +1603,77 @@ func applyUnmountedPVCs(window kubecost.Window, podMap map[podKey]*Pod, pvcMap m
...
+			// given LB and the associated Allocation so that a precise
+			// number of hours can be used to compute cumulative cost.
+			s, e := alloc.Start, alloc.End
+			if lb.Start.After(alloc.Start) {
+				s = lb.Start
+			}
+			if lb.End.Before(alloc.End) {
+				e = lb.End
+			}
+			hours := e.Sub(s).Hours()
```
nikovacevic, line 1664: Might want to consider a negative-value check here

nikovacevic (changes requested):
Great start, looking solid. I see Ajay and I had the same question about schema haha.

I would also say that more rigorous testing should be required. Can we confirm that this works for at least two load balancers? And perhaps we can test the period of time where a load balancer is created, and then removed some amount of time later, to make sure that all the numbers add up for the full lifecycle?

pkg/costmodel/allocation.go
```diff
@@ -1586,6 +1603,77 @@ func applyUnmountedPVCs(window kubecost.Window, podMap map[podKey]*Pod, pvcMap m
 	}
 }
 
+type LB struct {
+	Cost float64
```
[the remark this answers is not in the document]
nikovacevic: Also, a public struct should be commented. Does your IDE give you warnings about that?

Sean-Holcomb:
I separated Load Balancer costs into their own category, and did the updates Niko suggested. Still have not gotten to the more robust testing that Niko outlined

nikovacevic:
Thanks for the update @Sean-Holcomb. Let me know if I can help with anything!

Also, planting a flag to talk about `bingen` soon. I see you have changes committed to the codec, but they're so marginal that I can't tell if you used the code generator or did that manually. (If you are already on top of the code generator, then kudos!) At any rate, we'll need to bump the version of the codec, too, and any of use (@AjayTripathy, @mbolt35, and myself) can help with that.

Sean-Holcomb:
Multi-load balancer test: 
From March 23rd-25th there were two load balancers running on the cluster with the 24th being the only day where both load balancers were running the entire day. As expected, over the course of this time period the load balancer cost is elevated. 
Screen Shot 2021-03-30 at 3 48 27 PM
For reference the cost on the 20th-22nd were constant at .60 per day while only a single LB was running on the cluster. 
Screen Shot 2021-03-30 at 4 05 10 PM
And on the 24th when 2 load balancers were running all day the LB costs was 1.20
Screen Shot 2021-03-30 at 3 48 56 PM
In prometheus we can see two distinct signals for the kubecost_load_balancer_cost metric during the lifespan of the second load balancer. Both signals end on the 25th when the cluster went down

Screen Shot 2021-03-30 at 4 13 15 PM

Sean-Holcomb:
Life-cycle test:
For this test the load balancer was running on the cluster for 24 hours 45 minutes 39 seconds or 24.761 hours
Screen Shot 2021-03-30 at 4 33 30 PM
Screen Shot 2021-03-30 at 4 33 40 PM
with an hourly cost of $0.025 the expected LB cost for this time period is $0.619

Screen Shot 2021-03-30 at 4 33 00 PM

nikovacevic (approved):
Awesome work on this, @Sean-Holcomb. Thanks for all the testing, too; it makes a huge difference.

LGTM!

PR #744, opened 2021-03-18
Fix Reservation Diagnostics

Sean-Holcomb:
Prevent failing Athena Queries from running and provide more useful logs to replace failures. Also prevent reservation pricing Diagnostic if there is no reservation data to be queried

Tested on EKS-Integration-Test cluster where there is no Reservation data in the CUR
Screen Shot 2021-03-18 at 11 01 44 AM

Screen Shot 2021-03-18 at 11 01 57 AM

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -1789,14 +1789,20 @@ func (a *AWS) GetReservationDataFromAthena() error {
 	if cfg.AthenaBucketName == "" {
 		return fmt.Errorf("No Athena Bucket configured")
 	}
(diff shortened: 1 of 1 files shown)


AjayTripathy:
LGTM, just double chcking this didn't regress-- did you test this with an unauthorized key?

AjayTripathy (approved):

Sean-Holcomb:
> LGTM, just double chcking this didn't regress-- did you test this with an unauthorized key?
The unauthorized key that I was using caused all Athena queries to fail, the changes made here do not have an effect on that

AjayTripathy (approved):

PR #741, opened 2021-03-17
Open source /allocation/compute

nikovacevic:
Requires https://github.com/kubecost/kubecost-cost-model/pull/360

## Changes
- Move `/model/allocation/compute` to open source

## Testing
- Manual on gcp-niko

PR #746, opened 2021-03-19
Open source Reports UI

SheepGotoHeaven:
Initial open-source implementation of the Allocation reporting UI.

## Screenshots

Chome:
Screenshot from 2021-03-19 14-34-17

Firefox (with menu open):

Screenshot from 2021-03-19 14-35-01

Detail Modal Open:
Screenshot from 2021-03-19 14-36-20


ui/src/components/DetailsDialog.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import { get } from 'lodash'
+import Button from '@material-ui/core/Button'
+import Dialog from '@material-ui/core/Dialog'
+import DialogActions from '@material-ui/core/DialogActions'
+import DialogContent from '@material-ui/core/DialogContent'
+import DialogTitle from '@material-ui/core/DialogTitle'
+import Details from './Details'
+
+const DetailsDialog = ({
```
nikovacevic, line 10: No need for this without drill-down

ui/README.md
```diff
@@ -0,0 +1,23 @@
+# Kubecost Open Source UI
```
dwbrown2, line 1: This documentation LGTM, and we can plan to link from the repo's primary readme in a future release too. 

I may just add the note at the top of this section about our recommended install path! 
https://github.com/kubecost/cost-model/blob/develop/kubecost-exporter.md#installing

michaelmdresser: Definitely agree about highlighting the regular install path.

ui/README.md
```diff
@@ -0,0 +1,23 @@
...
+npx parcel src/index.html
+```
+
+This will launch a development server, serving the UI at `http://localhost:1234` and targeting the data for an instance of
+Kubecost running at `http://localhost:9090`. To access an arbitrary Kubecost install, you can use
+
+```
+kubectl port-forward deployment/kubecost-cost-analyzer 9090
+```
+
```
michaelmdresser, line 22: I assume the eventual goal is to package this up more nicely for an in-cluster deploy? Obviously no pressure to do that now, just wanted to know.

michaelmdresser:
Does this depend on https://github.com/kubecost/cost-model/pull/741?

ui/src/spa_index.jsx
```diff
@@ -0,0 +1,509 @@
...
+    display: 'flex',
+    flexFlow: 'row',
+    padding: 24,
+  },
+  titles: {
+    flexGrow: 1,
+  },
+})
+
+const ReportsPage = () => {
```
nikovacevic, line 74: Trying to remember, because I think we talked about it.. why are we using `.jsx` now? Is this for syntax highlighting, or some functionality? If so, should we convert all files to `.jsx` in the interest of uniformity?

ui/src/spa_index.jsx
```diff
@@ -0,0 +1,509 @@
...
+
+    setLoading(false)
+    setFetch(false)
+  }
+  return (
+    <Page active="reports.html">
+      <Header breadcrumbs={[
+        { 'name': 'Reports', 'href': 'new_index.html#/reports' },
+        { 'name': title, 'href': `new_index.html#/reports` },
+      ]}>
```
nikovacevic, line 406: I think we can drop the breadcrumb. What do you think?

nikovacevic (commented):
Amazing start on such a short time-scale! If you want to talk through some of these questions, or tap me in for some of the clean-up work, let me know. Otherwise, I think we will be all set at any point to call it, merge these changes, and start filing tickets for improvements.

ui/README.md
SheepGotoHeaven: Yep, that's what I'm thinking. In the future it can launch as a part of the cluster.

SheepGotoHeaven:
@michaelmdresser 

> Does this depend on #741?

It does, yes.

ui/src/components/Controls/index.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react'
+import { makeStyles } from '@material-ui/styles'
+import EditControl from './Edit'
+import DownloadControl from './Download'
+
+const useStyles = makeStyles({
+  controls: {
+    flexGrow: 0,
+    minWidth: 200,
```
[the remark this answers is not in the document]
SheepGotoHeaven: Maybe. I haven't tried it without this definition. `classes.controls` is used in the div wrapping the root of the `Controls` component.

ui/src/components/Controls/index.js
```diff
@@ -0,0 +1,71 @@
...
+        accumulate={accumulate}
+        setAccumulate={setAccumulate}
+        idleOptions={idleOptions}
+        idle={idle}
+        setIdle={setIdle}
+        filters={filters}
+        setFilters={setFilters}
+        currency={currency}
+        clearContext={clearContext}
+      />
```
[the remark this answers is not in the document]
SheepGotoHeaven: I think that'd be pretty straightforward and agree that is nicer from a UI/UX perspective.

I had thought maube it makes sense to leave the UX a little more painful here for the sake of consistency. Like, if someone goes from OSUI to trying out the closed-source product, that menu is in the same place but with more stuff.

You could also make a case though that the best way to *show* them that there are new options in the closed source is to make that experience different (OS, all options are laid out flat, CS "where are my options? oh they're in this buTTON WITH NEW OPTIONS?!").

So now that you say it, I'm on board. Happy to make the change

ui/src/components/AllocationReport.js
```diff
@@ -0,0 +1,200 @@
...
+              return (
+                <TableRow key={key} hover style={{ cursor: 'pointer' }} onClick={() => drillDown(row)}>
+                  <TableCell align="left">{row.name}</TableCell>
+                  <TableCell align="right">{toCurrency(row.cpuCost, currency)}</TableCell>
+                  <TableCell align="right">{toCurrency(row.ramCost, currency)}</TableCell>
+                  <TableCell align="right">{toCurrency(row.pvCost, currency)}</TableCell>
+                  <TableCell align="right">{efficiency}%</TableCell>
+                  <TableCell align="right">{toCurrency(row.totalCost, currency)}</TableCell>
+                </TableRow>
+              )
```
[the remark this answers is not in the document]
SheepGotoHeaven: Riiiiiight. Darn. I thought it was so cool that this feature just "came along for the ride".

But yeah, without filters we're just changing the aggBy and showing the wrong data. Welp, taking this out is at least easy.

Good catch!

ui/src/spa_index.jsx
SheepGotoHeaven: I thought it was correcting some minor syntax highlighting at the time, but looking at it again now I don't think it does.

Googled around some and it seems like there's no really great reason to prefer one extension or the other. The most in-depth conversation of it I could find was here: https://github.com/airbnb/javascript/pull/985

It's basically just a bunch of people saying "we should standardize on .js" and one mad Airbnb person being like "Returning inline html elements is not \~javascript\~"

I'm game to just keep `.js` so we don't have to rename a ton of files and because it seems like there's really not a significant difference.

ui/src/spa_index.jsx
SheepGotoHeaven: If we aren't drilldown-ing then absolutely. So yes.

SheepGotoHeaven:
All original comments should now be addressed. New Screenshots:

Chrome:
Screen Shot 2021-03-22 at 5 04 01 PM


Safari:
Screen Shot 2021-03-22 at 5 04 20 PM


Firefox:
Screen Shot 2021-03-22 at 5 05 44 PM

nikovacevic (changes requested):
@SheepGotoHeaven this is looking great so far! Love the flattening of the form controls. Feels very simple, clean and natural.

One bug I just ran into, hoping you will be able to repro and investigate: try using the custom date range picker to select a window. I'm getting:
```
Uncaught TypeError: utils.getDayText is not a function
```
Screenshot from 2021-03-23 13-14-25

As always, let me know if I can help!

ui/src/spa_index.jsx
nikovacevic: Okay, interesting, that's a good find. Just know, I'm open to opinions either way, but I do like keeping things standard, so `.js` is fine with me, but just holler if you want to go `.jsx`.

ui/src/components/AllocationReport.js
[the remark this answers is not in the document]
nikovacevic: We'll bring it back soon enough, I'm sure! Just a few more features to add first :smile:

ui/src/components/Controls/index.js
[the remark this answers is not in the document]
nikovacevic: Haha I like thinking this way about what users' experiences might be. Consistency is generally a good argument, too, but in this case I think we are okay with a break because the closed-source UI will always have more features, and in some cases the OS might even have features the CS doesn't have (difficult to think of one, but possible). If we were talking "free" to "enterprise", where fluidity between tiers was expected, then consistency feels way more important to me.

At any rate, thank you for the changes--I think it's looking absolutely fantastic.

ui/src/components/Controls/index.js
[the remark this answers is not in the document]
nikovacevic: Right on. Looks like both went, and things still look good, so thank you! I should have been more clear: the `minWidth` was causing a weird right-hand margin, but it's even nicer to remove it all :+1:

SheepGotoHeaven:
> @SheepGotoHeaven this is looking great so far! Love the flattening of the form controls. Feels very simple, clean and natural.
> 
> One bug I just ran into, hoping you will be able to repro and investigate: try using the custom date range picker to select a window. I'm getting:
> 
> ```
> Uncaught TypeError: utils.getDayText is not a function
> ```
> 
> Screenshot from 2021-03-23 13-14-25
> 
> As always, let me know if I can help!

Thanks and nice catch. It looks like this is because the closed-source UI has `@date-io/date-fns` at version `1.3.13` and the open source has version `2.10.8`.

I was hoping I could just iterate

-> Run build
-> Get errors
-> install deps

But I guess if you don't have a required dependency, Parcel just installs `latest` for you >_>

Will check the rest of these and resubmit.

SheepGotoHeaven:
We also have React v17 vs v16 and Reacharts 2.0 vs 1.8 but those don't seem to be causing issues. Do you think we should put this PR on the older library versions for consistency? Or is it alright to float the newer ones if there are no discernable problems?

@nikovacevic

dwbrown2:
> > @SheepGotoHeaven this is looking great so far! Love the flattening of the form controls. Feels very simple, clean and natural.
> > One bug I just ran into, hoping you will be able to repro and investigate: try using the custom date range picker to select a window. I'm getting:
> > ```
> > Uncaught TypeError: utils.getDayText is not a function
> > ```
> > 
> > 
> > Screenshot from 2021-03-23 13-14-25
> > As always, let me know if I can help!
> 
> Thanks and nice catch. It looks like this is because the closed-source UI has `@date-io/date-fns` at version `1.3.13` and the open source has version `2.10.8`.
> 
> I was hoping I could just iterate
> 
> -> Run build
> -> Get errors
> -> install deps
> 
> But I guess if you don't have a required dependency, Parcel just installs `latest` for you >_>
> 
> Will check the rest of these and resubmit.

Proposal is that we can ship with custom data later too if this is complex or time-consuming...

SheepGotoHeaven:
> > > @SheepGotoHeaven this is looking great so far! Love the flattening of the form controls. Feels very simple, clean and natural.
> > > One bug I just ran into, hoping you will be able to repro and investigate: try using the custom date range picker to select a window. I'm getting:
> > > ```
> > > Uncaught TypeError: utils.getDayText is not a function
> > > ```
> > > 
> > > 
> > > Screenshot from 2021-03-23 13-14-25
> > > As always, let me know if I can help!
> > 
> > 
> > Thanks and nice catch. It looks like this is because the closed-source UI has `@date-io/date-fns` at version `1.3.13` and the open source has version `2.10.8`.
> > I was hoping I could just iterate
> > -> Run build
> > -> Get errors
> > -> install deps
> > But I guess if you don't have a required dependency, Parcel just installs `latest` for you >_>
> > Will check the rest of these and resubmit.
> 
> Proposal is that we can ship with custom data later too if this is complex or time-consuming...

Just re-pushed and it should be fixed now -- there was just a version mismatch between the closed and open source for a Date library we use

nikovacevic (approved):
LGTM! Thanks, Neal, excellent work.

PR #747, opened 2021-03-20
Drop now optional requirements...

dwbrown2:


README.md
```diff
@@ -16,9 +16,7 @@ Here is a summary of features enabled by this cost model:
 ## Requirements
 
 - Kubernetes version 1.8 or higher
-- kube-state-metrics
-- Node exporter
```
michaelmdresser, line 20: This is true in the context of exporter-only, right? But not regular installs?

dwbrown2: So node-exporter is 100% not required in any mode.

I believe that ksm is "optional" but it does give you extra metrics, e.g. PVs. So maybe we drop node-exporter and caveat ksm?

dwbrown2: cc @AjayTripathy just in case he has a different perspective

michaelmdresser: > So maybe we drop node-exporter and caveat ksm?

I agree with this idea.

michaelmdresser (approved):

PR #749, opened 2021-03-24
Ajay tripathy fix multilabels

AjayTripathy:
Needs multi-product-label tests, but should work now.


pkg/kubecost/asset.go
```diff
@@ -143,9 +143,15 @@ func AssetToExternalAllocation(asset Asset, aggregateBy []string) (*Allocation,
 		// labelName should be derived from the mapping of properties to
 		// label names, unless the aggBy is explicitly a label, in which
 		// case we should pull the label name from the aggBy string.
+		// Unless this matches a special aggregation, as we have that mapping already transformed...
 		labelName := aggBy
+		agglName := aggBy
```
nikovacevic, line 148: How is the use of `agglName` different from `labelName` here? Seems like this could have been achieved by just setting `labelName = v` instead of `agglName = v` on line 153?

nikovacevic (approved):
LGTM with caveat of rewriting soon

nikovacevic:
@AjayTripathy It would be nice to have a test case with multiple labels pointing at "product" so that we can assert something about what we expect to have happen. But then again, it may not be possible because it seems like the behavior is random?

PR #750, opened 2021-03-24
Upgrading to Parcel 2

SheepGotoHeaven:
Dependabot found a security issue with Parcel 1.12.5, but that is the latest version of Parcel 1 and no new versions will be released.


AjayTripathy (approved):

nikovacevic (approved):
LGTM! Excited to do the same for the other repo soon, too :smile:

PR #753, opened 2021-03-29
Add Usage section to README!

dwbrown2:


README.md
```diff
@@ -27,7 +27,12 @@ Here is a summary of features enabled by this cost model:
 You can deploy Kubecost on any Kubernetes 1.8+ cluster in a matter of minutes, if not seconds. 
 Visit the Kubecost docs for [recommended install options](https://docs.kubecost.com/install). Compared to building from source, installing from Helm is faster and includes all necessary dependencies. 
 
-> If you want to deploy cost model in Prometheus exporter only mode, check out [kubecost-exporter.md](kubecost-exporter.md).
+## Usage
+
+* User interface
```
michaelmdresser, line 32:
```suggestion
* User interface (`kubectl port-forward --namespace kubecost deployment/kubecost-cost-analyzer 9090`)
```

michaelmdresser (approved):

README.md
michaelmdresser: I think linking is definitely a better idea. Do we have docs for "accessing kubecost" in general somewhere?

michaelmdresser: Hmm. I think the majority of users will want instructions for the normal UI, not the OSS one, but we would want to include instructions for both.

PR #754, opened 2021-03-29
Soften generateKey in AggregateBy to use empty strings for missing properties instead of erroring

nikovacevic:
## Changes
- Soften generateKey in AggregateBy to use empty strings for missing properties instead of erroring

## Testing
- Unit testing

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -1113,6 +1113,10 @@ func computeIdleCoeffs(properties Properties, options *AllocationAggregationOpti
 }
 
 func (a *Allocation) generateKey(properties Properties) string {
+	if a == nil {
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Looks like there is regression testing, but is it quick to write a unit test for the case where nodes are missing?

nikovacevic:
Yep, they were just in progress :smile: Done now!

AjayTripathy (approved):
It LGTM but would we still want a warning when this case happens?

nikovacevic:
Responded in Slack
> I don't think we need one, if it's now normal working-as-intended behavior.
> We'll still get logs if the node doesn't get set, but we probably don't need one on aggregation.

PR #762, opened 2021-04-05
remove groupright for scaling, use minutes in only one query

AjayTripathy:
Screen Shot 2021-04-05 at 11 36 50 AM
At a steady state, this does not regress. More testing is needed, but want feedback on approach before going too far down this road.

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -442,6 +442,15 @@ func costTimesMinuteAndCount(activeDataMap map[NodeIdentifier]activeData, costMa
 	}
 }
 
+func costTimesMinute(activeDataMap map[NodeIdentifier]activeData, costMap map[NodeIdentifier]float64) {
+	for k, v := range activeDataMap {
+		if cost, ok := costMap[k]; ok {
+			minutes := v.minutes
+			costMap[k] = cost * (minutes / 60)
+		}
+	}
+}
+
 func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset time.Duration) (map[NodeIdentifier]*Node, error) {
 	durationStr := fmt.Sprintf("%dm", int64(duration.Minutes()))
 	offsetStr := fmt.Sprintf(" offset %dm", int64(offset.Minutes()))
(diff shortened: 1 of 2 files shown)


pkg/costmodel/cluster.go
```diff
@@ -438,20 +455,15 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset
...
-	queryNodeGPUHourlySum := fmt.Sprintf(`sum_over_time(avg(node_gpu_hourly_cost) by (cluster_id, node, instance_type, provider_id)[%s:%dm]%s) * %f`, durationStr, minsPerResolution, offsetStr, hourlyToCumulative)
+	queryNodeCPUCost := fmt.Sprintf(`avg(avg_over_time(node_cpu_hourly_cost[%s]%s)) by (cluster_id, node, instance_type, provider_id)`, durationStr, offsetStr)
+	queryNodeCPUCores := fmt.Sprintf(`avg(avg_over_time(kube_node_status_capacity_cpu_cores[%s]%s)) by (cluster_id, node)`, durationStr, offsetStr)
+	queryNodeRAMCost := fmt.Sprintf(`avg(avg_over_time(node_ram_hourly_cost[%s]%s)) by (cluster_id, node, instance_type, provider_id) / 1024 / 1024 / 1024`, durationStr, offsetStr)
+	queryNodeRAMBytes := fmt.Sprintf(`avg(avg_over_time(kube_node_status_capacity_memory_bytes[%s]%s)) by (cluster_id, node)`, durationStr, offsetStr)
+	queryNodeGPUCount := fmt.Sprintf(`avg(avg_over_time(node_gpu_count[%s]%s)) by (cluster_id, node, provider_id)`, durationStr, offsetStr)
+	queryNodeGPUHourlySum := fmt.Sprintf(`avg(avg_over_time(node_gpu_hourly_cost[%s]%s)) by (cluster_id, node, instance_type, provider_id)`, durationStr, offsetStr)
```
nikovacevic, line 466: Query changes look good. Being picky, I'd propose that we rename some of the variables, too, to reflect the change to average hourly cost of the (thing) over the window, whereas some of them used to query for cumulative amounts.

pkg/costmodel/cluster.go
```diff
@@ -425,6 +425,23 @@ type nodeIdentifierNoProviderID struct {
...
+		if cost, ok := costMap[k]; ok {
+			minutes := v.minutes
+			count := 1.0
+			if c, ok := resourceCountMap[keyNon]; ok {
+				count = c
+			}
+			costMap[k] = cost * (minutes / 60) * count
+		}
+	}
+}
```
nikovacevic, line 443: Just to be clear, the units here are: `hourly cost per one instance of resource`, `hours` (derived from `minutes`), and `number of instances of the resource`. That sound right?

nikovacevic (approved):
Methodology looks right to me. Agree with the PR description w/r/t the idea that more testing might be nice.

pkg/costmodel/cluster.go
AjayTripathy: yep!

pkg/costmodel/cluster.go
```diff
@@ -518,10 +541,13 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, duration, offset
 	ramSystemPctMap := buildRAMSystemPctMap(resNodeRAMSystemPct)
 
 	cpuBreakdownMap := buildCPUBreakdownMap(resNodeCPUModeTotal)
-	activeDataMap := buildActiveDataMap(resActiveMins, resolution, cp.ParseID)
 	preemptibleMap := buildPreemptibleMap(resIsSpot, cp.ParseID)
 	labelsMap := buildLabelsMap(resLabels)
 
+	costTimesMinuteAndCount(activeDataMap, cpuCostMap, cpuCoresMap)
+	costTimesMinuteAndCount(activeDataMap, ramCostMap, ramBytesMap)
+	costTimesMinute(activeDataMap, gpuCostMap) // there's no need to do a weird "nodeIdentifierNoProviderID" type match since gpuCounts have a providerID
```
nikovacevic, line 549: So, each individual GPU gets counted separately? Interesting, I don't think I knew that.

Sean-Holcomb: The GPU counts get multiplied in at line 531

Sean-Holcomb:
GPU pricing tests:
What is being tested here is that the conversion from summing gpu hourly cost data points to the average hourly rate multiplied by hours running produces total GPU costs that make sense. 

AWS: 
Using a p2.8xlarge node running these changes GPU price is correctly counting with minutes running matching the the length of time that GPU metric were being emitted in Prometheus. Additionally the total price adds up with Node_Hourly_Rate * Node_Count * Hours_Running. In this case. $0.642627 * 8 * 158/60  = $13.5380088. 

Screen Shot 2021-04-06 at 1 03 47 PM

Screen Shot 2021-04-06 at 1 08 13 PM

GCP: 
We see similar results on GCP with a NC12s_v2 node. In this case the node is given the default gpu price of $0.95. In this case $0.95 * 2 * 102/60 = $3.23. While these number do add up it is unclear if the default gpu price is correct in this case.

Screen Shot 2021-04-06 at 12 47 15 PM

Screen Shot 2021-04-06 at 12 43 28 PM

Sean-Holcomb (approved):
Updated Variable names to match the rest of the naming scheme

pkg/costmodel/cluster.go
nikovacevic: Ah, ok, the PR has changed since this comment. Will re-review.

AjayTripathy:
Thank you so much for picking this up @Sean-Holcomb and running with the testing!

PR #763, opened 2021-04-07
Add CPU and RAM usage maximums to Allocation

michaelmdresser:
This PR adds CPU and RAM usage maximums to the `Allocation` type for eventual usage by a new version of Savings's request right-sizing.

## Getting Usage

RAM usage max was an easy edit to our current average usage query, changing `avg(avg_over_time())` to `max(max_over_time())`.

CPU usage max was a little more complicated. Because we use `rate` to calculate average CPU usage, we have to use a subquery or a recording rule to take a maximum of rate over time. Fortunately, we already have a recording rule called `kubecost_savings_container_cpu_usage_seconds` that has the necessary information. See code comments for more commentary on rationale.

## Adding Usage to `Allocation`

- I changed a few query, function, and parameter names to make clear what is average usage and what is max usage.
- I duplicated the relevant `apply` functions to convert Prom responses into struct fields
- I modified the behavior of `Allocation.add()` to consider two cases: accumulation and aggregation

## Accumulation and Aggregation

We have two sets of behavior that could previously be combined into the same logic.

Accumulation is the combination of `Allocation`s that refer to the same thing, without a loss of granularity. Imagine needing to combine two `Allocation`s that refer to the same container. In this case, usage maxima need to be maxed together. For example, if container C used a maximum of X CPU in allocation A1 and container C used a maximum of Y CPU in allocation A2, then the resulting allocation A3 should have container C's max CPU as `max(X, Y)`.

Aggregation is the combination of `Allocation`s that refer to different things under the same aggregation umbrella (i.e. with a loss of granularity). Imagine combining containers under the same namespace. In this case, usage maxima need to be added together. For example, if container C1 used a maximum of X CPU in A1 and container C2 used a maximum of Y CPU in A2 (and they are the only two containers in the namespace), then the allocation aggregated for C1 and C2's namespace should have a maximum CPU of `X + Y`.

Because most of the rest of the `Allocation.add()` logic did not need to change, I added a flag that tells `add()` whether it is performing an accumulating add or an aggregating add. Because `AllocationSet.insert()` uses `Allocation.add()`, it also has this flag. To make this clear in the public API, I removed `Allocation.Add()` and `AllocationSet.Insert()` and replaced them with `Allocation.AddAccumulate()`, `Allocation.AddAggregate()`, `AllocationSet.InsertAccumulate()`, and `AllocationSet.InsertAggregate()`. I think this is a pretty good way of handling this, but I'm absolutely open to better approaches.

# Tests

## No ETL, fields set
To check that the field was being set properly, I ran a query against `/allocation/compute` which hits `ComputeAllocation` directly without dealing with the ETL.

```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation/compute | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "cpuCoreUsageMax": .cpuCoreUsageMax,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "ramByteUsageMax": .ramByteUsageMax,
           }
            )
        )'
```

[Results](https://gist.github.com/michaelmdresser/666ad543b25ade2fc280adebe546b897) (large)

## ETL, fields set

To also verify that the fields were being set in the ETL, I also had to force a rebuild:
```
curl -G \
    --connect-timeout 3 \
    --max-time 120 \
    -d 'window=7d' \
    -d 'commit=true' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/etl/allocation/rebuild
```

The test query:
```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "cpuCoreUsageMax": .cpuCoreUsageMax,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "ramByteUsageMax": .ramByteUsageMax,
           }
            )
        )'
```

[Results](https://gist.github.com/michaelmdresser/dd9e52e35cdd52c52215185c7aa17acd) (large)

## Non-ETL, aggregation logic

```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    -d 'aggregate=namespace' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation/compute | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "cpuCoreUsageMax": .cpuCoreUsageMax,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "ramByteUsageMax": .ramByteUsageMax,
           }
            )
        )'
```

```
[
  {
    "cost-model": {
      "cpuCoreUsageAverage": 0.003388,
      "cpuCoreUsageMax": 0.005153,
      "ramByteUsageAverage": 278031103.110494,
      "ramByteUsageMax": 284053504
    },
    "kube-system": {
      "cpuCoreUsageAverage": 0.051783,
      "cpuCoreUsageMax": 0.062117,
      "ramByteUsageAverage": 439955257.97804,
      "ramByteUsageMax": 451493888
    },
    "kubecost": {
      "cpuCoreUsageAverage": 0.029408,
      "cpuCoreUsageMax": 0.292005,
      "ramByteUsageAverage": 517551154.179154,
      "ramByteUsageMax": 710434816
    },
    "kubecost-staging": {
      "cpuCoreUsageAverage": 0.023558,
      "cpuCoreUsageMax": 0.044541,
      "ramByteUsageAverage": 399534795.95991,
      "ramByteUsageMax": 425254912
    }
  }
]
```

## ETL, aggregation logic

```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    -d 'aggregate=namespace' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "cpuCoreUsageMax": .cpuCoreUsageMax,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "ramByteUsageMax": .ramByteUsageMax,
           }
            )
        )'
```

```
[
  {
    "__idle__": {
      "cpuCoreUsageAverage": 0,
      "cpuCoreUsageMax": 0,
      "ramByteUsageAverage": 0,
      "ramByteUsageMax": 0
    },
    "cost-model": {
      "cpuCoreUsageAverage": 0.003474,
      "cpuCoreUsageMax": 0.004702,
      "ramByteUsageAverage": 271591911.483692,
      "ramByteUsageMax": 276983808
    },
    "kube-system": {
      "cpuCoreUsageAverage": 0.051683,
      "cpuCoreUsageMax": 0.064119,
      "ramByteUsageAverage": 439029513.20743,
      "ramByteUsageMax": 453189632
    },
    "kubecost": {
      "cpuCoreUsageAverage": 0.025604,
      "cpuCoreUsageMax": 0.164336,
      "ramByteUsageAverage": 525091844.350198,
      "ramByteUsageMax": 923942912
    },
    "kubecost-staging": {
      "cpuCoreUsageAverage": 0.023704,
      "cpuCoreUsageMax": 0.044678,
      "ramByteUsageAverage": 396685146.036298,
      "ramByteUsageMax": 420511744
    }
  },
  {
    "__idle__": {
      "cpuCoreUsageAverage": 0,
      "cpuCoreUsageMax": 0,
      "ramByteUsageAverage": 0,
      "ramByteUsageMax": 0
    },
    "cost-model": {
      "cpuCoreUsageAverage": 0.003393,
      "cpuCoreUsageMax": 0.005153,
      "ramByteUsageAverage": 278466523.984012,
      "ramByteUsageMax": 284553216
    },
    "kube-system": {
      "cpuCoreUsageAverage": 0.051784,
      "cpuCoreUsageMax": 0.061107,
      "ramByteUsageAverage": 439987553.92533,
      "ramByteUsageMax": 451862528
    },
    "kubecost": {
      "cpuCoreUsageAverage": 0.029126,
      "cpuCoreUsageMax": 0.285256,
      "ramByteUsageAverage": 514780208.328162,
      "ramByteUsageMax": 694689792
    },
    "kubecost-staging": {
      "cpuCoreUsageAverage": 0.023542,
      "cpuCoreUsageMax": 0.043447,
      "ramByteUsageAverage": 399383724.328601,
      "ramByteUsageMax": 423763968
    }
  }
]
```

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -714,7 +714,13 @@ func applyCPUCoresUsedMax(podMap map[podKey]*Pod, resCPUCoresUsedMax []*prom.Que
 			pod.AppendContainer(container)
 		}
 
-		pod.Allocations[container].CPUCoreUsageMax = res.Values[0].Value
+		if pod.Allocations[container].RawAllocationOnly == nil {
+			pod.Allocations[container].RawAllocationOnly = &kubecost.RawAllocationOnlyData{
+				CPUCoreUsageMax: res.Values[0].Value,
+			}
+		} else {
+			pod.Allocations[container].RawAllocationOnly.CPUCoreUsageMax = res.Values[0].Value
+		}
 	}
 }
 
@@ -844,7 +850,13 @@ func applyRAMBytesUsedMax(podMap map[podKey]*Pod, resRAMBytesUsedMax []*prom.Que
 			pod.AppendContainer(container)
 		}
 
-		pod.Allocations[container].RAMBytesUsageMax = res.Values[0].Value
+		if pod.Allocations[container].RawAllocationOnly == nil {
+			pod.Allocations[container].RawAllocationOnly = &kubecost.RawAllocationOnlyData{
+				RAMBytesUsageMax: res.Values[0].Value,
+			}
+		} else {
+			pod.Allocations[container].RawAllocationOnly.RAMBytesUsageMax = res.Values[0].Value
+		}
 	}
 }
 
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -17,13 +17,30 @@ import (
 )
 
 const (
-	queryFmtPods                  = `avg(kube_pod_container_status_running{}) by (pod, namespace, cluster_id)[%s:%s]%s`
-	queryFmtRAMBytesAllocated     = `avg(avg_over_time(container_memory_allocation_bytes{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
-	queryFmtRAMRequests           = `avg(avg_over_time(kube_pod_container_resource_requests_memory_bytes{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
-	queryFmtRAMUsage              = `avg(avg_over_time(container_memory_working_set_bytes{container_name!="", container_name!="POD", instance!=""}[%s]%s)) by (container_name, pod_name, namespace, instance, cluster_id)`
-	queryFmtCPUCoresAllocated     = `avg(avg_over_time(container_cpu_allocation{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
-	queryFmtCPURequests           = `avg(avg_over_time(kube_pod_container_resource_requests_cpu_cores{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
-	queryFmtCPUUsage              = `avg(rate(container_cpu_usage_seconds_total{container_name!="", container_name!="POD", instance!=""}[%s]%s)) by (container_name, pod_name, namespace, instance, cluster_id)`
+	queryFmtPods              = `avg(kube_pod_container_status_running{}) by (pod, namespace, cluster_id)[%s:%s]%s`
+	queryFmtRAMBytesAllocated = `avg(avg_over_time(container_memory_allocation_bytes{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
+	queryFmtRAMRequests       = `avg(avg_over_time(kube_pod_container_resource_requests_memory_bytes{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
+	queryFmtRAMUsageAvg       = `avg(avg_over_time(container_memory_working_set_bytes{container_name!="", container_name!="POD", instance!=""}[%s]%s)) by (container_name, pod_name, namespace, instance, cluster_id)`
+	queryFmtRAMUsageMax       = `max(max_over_time(container_memory_working_set_bytes{container_name!="", container_name!="POD", instance!=""}[%s]%s)) by (container_name, pod_name, namespace, instance, cluster_id)`
+	queryFmtCPUCoresAllocated = `avg(avg_over_time(container_cpu_allocation{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
+	queryFmtCPURequests       = `avg(avg_over_time(kube_pod_container_resource_requests_cpu_cores{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
+	queryFmtCPUUsageAvg       = `avg(rate(container_cpu_usage_seconds_total{container_name!="", container_name!="POD", instance!=""}[%s]%s)) by (container_name, pod_name, namespace, instance, cluster_id)`
+
+	// This query could be written without the recording rule
+	// "kubecost_savings_container_cpu_usage_seconds", but we should
+	// only do that when we're ready to incur the performance tradeoffs
+	// with subqueries which would probably be in the world of hourly
+	// ETL.
+	//
+	// See PromQL subquery documentation for a rate example:
+	// https://prometheus.io/blog/2019/01/28/subquery-support/#examples
+	queryFmtCPUUsageMax = `
+max(
+  max_over_time(
+    kubecost_savings_container_cpu_usage_seconds[%s]%s
+  )
+) by (container_name, pod_name, namespace, instance, cluster_id)`
+
 	queryFmtGPUsRequested         = `avg(avg_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, cluster_id)`
 	queryFmtNodeCostPerCPUHr      = `avg(avg_over_time(node_cpu_hourly_cost[%s]%s)) by (node, cluster_id, instance_type)`
 	queryFmtNodeCostPerRAMGiBHr   = `avg(avg_over_time(node_ram_hourly_cost[%s]%s)) by (node, cluster_id, instance_type)`
@@ -107,8 +124,11 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	queryRAMRequests := fmt.Sprintf(queryFmtRAMRequests, durStr, offStr)
 	resChRAMRequests := ctx.Query(queryRAMRequests)
 
-	queryRAMUsage := fmt.Sprintf(queryFmtRAMUsage, durStr, offStr)
-	resChRAMUsage := ctx.Query(queryRAMUsage)
+	queryRAMUsageAvg := fmt.Sprintf(queryFmtRAMUsageAvg, durStr, offStr)
+	resChRAMUsageAvg := ctx.Query(queryRAMUsageAvg)
+
+	queryRAMUsageMax := fmt.Sprintf(queryFmtRAMUsageMax, durStr, offStr)
+	resChRAMUsageMax := ctx.Query(queryRAMUsageMax)
 
 	queryCPUCoresAllocated := fmt.Sprintf(queryFmtCPUCoresAllocated, durStr, offStr)
 	resChCPUCoresAllocated := ctx.Query(queryCPUCoresAllocated)
@@ -116,8 +136,11 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	queryCPURequests := fmt.Sprintf(queryFmtCPURequests, durStr, offStr)
 	resChCPURequests := ctx.Query(queryCPURequests)
 
-	queryCPUUsage := fmt.Sprintf(queryFmtCPUUsage, durStr, offStr)
-	resChCPUUsage := ctx.Query(queryCPUUsage)
+	queryCPUUsageAvg := fmt.Sprintf(queryFmtCPUUsageAvg, durStr, offStr)
+	resChCPUUsageAvg := ctx.Query(queryCPUUsageAvg)
+
+	queryCPUUsageMax := fmt.Sprintf(queryFmtCPUUsageMax, durStr, offStr)
+	resChCPUUsageMax := ctx.Query(queryCPUUsageMax)
 
 	queryGPUsRequested := fmt.Sprintf(queryFmtGPUsRequested, durStr, offStr)
 	resChGPUsRequested := ctx.Query(queryGPUsRequested)
@@ -202,10 +225,12 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 
 	resCPUCoresAllocated, _ := resChCPUCoresAllocated.Await()
 	resCPURequests, _ := resChCPURequests.Await()
-	resCPUUsage, _ := resChCPUUsage.Await()
+	resCPUUsageAvg, _ := resChCPUUsageAvg.Await()
+	resCPUUsageMax, _ := resChCPUUsageMax.Await()
 	resRAMBytesAllocated, _ := resChRAMBytesAllocated.Await()
 	resRAMRequests, _ := resChRAMRequests.Await()
-	resRAMUsage, _ := resChRAMUsage.Await()
+	resRAMUsageAvg, _ := resChRAMUsageAvg.Await()
+	resRAMUsageMax, _ := resChRAMUsageMax.Await()
 	resGPUsRequested, _ := resChGPUsRequested.Await()
 
 	resNodeCostPerCPUHr, _ := resChNodeCostPerCPUHr.Await()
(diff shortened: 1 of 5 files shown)


pkg/costmodel/allocation.go
```diff
@@ -17,13 +17,30 @@ import (
...
+	// ETL.
+	//
+	// See PromQL subquery documentation for a rate example:
+	// https://prometheus.io/blog/2019/01/28/subquery-support/#examples
+	queryFmtCPUUsageMax = `
+max(
+  max_over_time(
+    kubecost_savings_container_cpu_usage_seconds[%s]%s
+  )
+) by (container_name, pod_name, namespace, instance, cluster_id)`
```
nikovacevic, line 42: Dig this query, glad you could use the `kubecost_savings` recording rule.

As a matter of code style, do you prefer this multi-line approach? I go back and forth, but it might be nice to keep the style consistent, but open to contrarian arguments.

pkg/kubecost/allocation.go
```diff
@@ -435,6 +489,19 @@ func (a *Allocation) add(that *Allocation) {
...
+		// maximum is the correct resulting value.
+		a.CPUCoreUsageMax = math.Max(a.CPUCoreUsageMax, that.CPUCoreUsageMax)
+		a.RAMBytesUsageMax = math.Max(a.RAMBytesUsageMax, that.RAMBytesUsageMax)
+	} else if addType == aggregation {
+		// Aggregation-type additions imply different resources, same
+		// aggregation group, so the correct maximum is actually the sum
+		// of maxima.
+		a.CPUCoreUsageMax += that.CPUCoreUsageMax
+		a.RAMBytesUsageMax += that.RAMBytesUsageMax
+	}
```
nikovacevic, line 504: Totally understand the reason this was introduced, but I have a few concerns. (Not least that we used to do this, and it made working with the `Allocation` type more complicated having to know _how_ you wanted to add two of them together. Things are very nice when `Add` is smart and the type is resilient.)

1. We (probably) have internal information that allows us to make this decision. We have `window` to apply exactly this set of logic; i.e. if the windows do not overlap, then we are accumulating; otherwise, we are aggregating.
2. This proposed logic (or even following the logic above) may not work. Consider aggregating by controller, and we come across the `kubecost-cost-analyzer` pod. We find 10 pods, and we're aggregating, so we add their maxes. But they ran serially! This is also by-definition, not coincidence, because they share the same controller. So we'll be giving 10x recommendations for this controller. There are further issues we may run into on this front, which we should discuss, but if you start by simply trying to add the maxes for everything in a window, you might end up with funky results.

nikovacevic (commented):
Exciting start. Let's talk tomorrow!

pkg/costmodel/allocation.go
michaelmdresser: I had it in multi-line format mostly for my own debugging sanity. As long as we're confident it is the right query (I am) I think moving it back to single line is totally fine.

michaelmdresser:
After a discussion with Niko, aggregation/accumulation logic does not make sense for maximum usage values. We currently combine `Allocation`s using `Allocation.add()` iteratively. Combining usage maximums cannot be done iteratively. I have included a small example situation that should illustrate this in a comment that is now part of the PR.

We will instead maintain a new struct of data (`RawAllocationOnlyData`) within `Allocation` that is only valid (non-`nil`) if the `Allocation` has never been combined with another, i.e. `.add()` has not been called on it.

We no longer need to change `Add()` or `Insert()` so the PR becomes much simpler.

# Tests

## No ETL, fields set

```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation/compute | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "rawAllocationOnly": .rawAllocationOnly,
           }
            )
        )'
```

[Results](https://gist.github.com/michaelmdresser/ae725101d5ea0f2a1b7d3a3a8b4aa19d)

## ETL, fields set
Same rebuild logic.

```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "rawAllocationOnly": .rawAllocationOnly,
           }
            )
        )'
```

[Results](https://gist.github.com/michaelmdresser/7daa28385a848dc8978ba90e23ffc64c)

## Non-ETL: aggregate causes nulls

```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    -d 'aggregate=namespace' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation/compute | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "rawAllocationOnly": .rawAllocationOnly,
           }
            )
        )'
```

:warning: :warning: 
**I'm not sure why only the `cost-model` namespace has a non-null `rawAllocationOnly`**.
:warning: :warning: 

```
[
  {
    "cost-model": {
      "cpuCoreUsageAverage": 0.003384,
      "ramByteUsageAverage": 285717822.577778,
      "rawAllocationOnly": {
        "cpuCoreUsageMax": 0.0045490115895717815,
        "ramByteUsageMax": 291217408
      }
    },
    "kube-system": {
      "cpuCoreUsageAverage": 0.052128,
      "ramByteUsageAverage": 442533775.204633,
      "rawAllocationOnly": null
    },
    "kubecost": {
      "cpuCoreUsageAverage": 0.036283,
      "ramByteUsageAverage": 506717201.956025,
      "rawAllocationOnly": null
    },
    "kubecost-staging": {
      "cpuCoreUsageAverage": 0.023531,
      "ramByteUsageAverage": 404499937.355509,
      "rawAllocationOnly": null
    }
  }
]
```

## ETL: aggregate causes nulls

```
curl -G \
    --connect-timeout 3 \
    --max-time 5 \
    -d 'window=1d' \
    -d 'aggregate=namespace' \
    http://localhost:8080/api/v1/namespaces/kubecost/services/kubecost-cost-analyzer:model/proxy/allocation | \
    jq '.data |
       map(
         map_values(
           {
             "cpuCoreUsageAverage": .cpuCoreUsageAverage,
             "cpuCoreUsageMax": .cpuCoreUsageMax,
             "ramByteUsageAverage": .ramByteUsageAverage,
             "ramByteUsageMax": .ramByteUsageMax,
           }
            )
        )'
```

```
[
  {
    "__idle__": {
      "cpuCoreUsageAverage": 0,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 0,
      "ramByteUsageMax": null
    },
    "cost-model": {
      "cpuCoreUsageAverage": 0.00339,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 279263818.97016,
      "ramByteUsageMax": null
    },
    "kube-system": {
      "cpuCoreUsageAverage": 0.051777,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 440331963.974926,
      "ramByteUsageMax": null
    },
    "kubecost": {
      "cpuCoreUsageAverage": 0.027913,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 514374945.759034,
      "ramByteUsageMax": null
    },
    "kubecost-staging": {
      "cpuCoreUsageAverage": 0.023545,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 400120018.512576,
      "ramByteUsageMax": null
    }
  },
  {
    "__idle__": {
      "cpuCoreUsageAverage": 0,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 0,
      "ramByteUsageMax": null
    },
    "cost-model": {
      "cpuCoreUsageAverage": 0.003382,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 285931037.328969,
      "ramByteUsageMax": null
    },
    "kube-system": {
      "cpuCoreUsageAverage": 0.052201,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 442545641.439202,
      "ramByteUsageMax": null
    },
    "kubecost": {
      "cpuCoreUsageAverage": 0.03744,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 505892400.885643,
      "ramByteUsageMax": null
    },
    "kubecost-staging": {
      "cpuCoreUsageAverage": 0.023537,
      "cpuCoreUsageMax": null,
      "ramByteUsageAverage": 404448589.76716,
      "ramByteUsageMax": null
    }
  }
]
```

pkg/kubecost/allocation.go
```diff
@@ -188,6 +233,22 @@ func (a *Allocation) Equal(that *Allocation) bool {
...
+
+	if a.RawAllocationOnly != nil && that.RawAllocationOnly != nil {
+		if !util.IsApproximately(a.RawAllocationOnly.CPUCoreUsageMax, that.RawAllocationOnly.CPUCoreUsageMax) {
+			return false
+		}
+		if !util.IsApproximately(a.RawAllocationOnly.RAMBytesUsageMax, that.RawAllocationOnly.RAMBytesUsageMax) {
+			return false
+		}
+	}
```
nikovacevic, line 250: This isn't ideal. Is it worth keeping those number directly on the type and checking mere equality?

pkg/kubecost/allocation.go
```diff
@@ -71,6 +71,38 @@ type Allocation struct {
...
+//
+// The logical maximum CPU usage is 5, but this cannot be calculated iteratively,
+// which is how we calculate aggregations and accumulations of Allocations currently.
+// This becomes a problem I could call "maximum sum of overlapping intervals" and is
+// essentially a variant of an interval scheduling algorithm.
+//
+// If we had types to differentiate between regular Allocations and AggregatedAllocations
+// then this type would be unnecessary and its fields would go into the regular Allocation
+// and not in the AggregatedAllocation.
+type RawAllocationOnlyData struct {
```
nikovacevic, line 103: My only challenge here is: why not put these on the main `Allocation` type with a comment about zeroing them out on `AggregateBy` and a `json:"cpuCoreUsageMax,omitempty"` so that they aren't encoded they're zero?

That has the advantage of keeping the type flatter and simpler to interact with. It has the disadvantage of perhaps misleading one into thinking max usage is actually zero, when in fact it's just not set.

nikovacevic (approved):
LGTM. Just let me know what you think about the `Allocation` type feedback.

pkg/kubecost/allocation.go
michaelmdresser: I like the idea of keeping `Allocation` flatter like you've suggested, but it does overload 0: I'm not sure if it would ever happen, but I can imagine a legitimate `Allocation` that has actual 0 usage. When this is expanded to include variance, 0 is an even more legitimate and possible value.

I dislike the "pointer to a struct because of how Go does JSON marshaling" thing that I'm doing right now. I think the right alternative if we want to keep the type flatter is to stick a `bool` flag in it and check that instead of relying on Go's default values.

What do you think?

pkg/kubecost/allocation.go
michaelmdresser: See below.

pkg/kubecost/allocation.go
nikovacevic: The point about variance using 0 as a distinct and legitimate value is compelling. Argh. The bool flag is kinda nasty, too, from an internal-API perspective. (Imagine accessing one of those numbers, but not realizing that you were supposed to check a separate bool value first, just to be sure.)

AWS will often use types like `*string` or `*int` to give that extra `nil` option. We could do that, but... I also don't love that.

The only other idea that comes to mind is to make the fields flat and unexported, then provide accessor functions like `CPUCoreMaxUsage() (float64, error)` so that an error can be returned when accessing on an aggregated `Allocation`. Maybe better? Maybe over-engineered?

michaelmdresser: Yeah I'm not a huge fan of just raw `*string`. Private fields and accessor functions is a decent idea, but it makes deserialization even more of a pain which currently just works https://github.com/kubecost/cost-model/blob/develop/pkg/kubecost/allocation_test.go#L434 I think without us writing a manual `UnmarshalJSON`.

I really don't like any of our options here. Thoughts?

nikovacevic: Well, it'd be a `*float64` in this case, but I take it that you still don't like that. `nil`s are a bummer, but I guess they're also kind of what we need (something to differentiate from `0`).

While I've never let free serialization stand between me and the design I want, it's also probably not the perfect solution. If it were me, I'd probably punt and just go with something simple (probably `*float64`) or perhaps I'd go with the accessor idea and punt on unmarshaling, but at this point I would just trust you to make the decision whatever that may be.

michaelmdresser: I'm going to stick with what I've got right now because it tries to force (in a less-than-ideal way) the user to check that the allocation is unaggregated and the user only has to do so once and then can have confidence accessing the fields.

PR #768, opened 2021-04-13
Sean/refactor properties

Sean-Holcomb:
This is an attempt to refactor the generic properties class which was a map[string]interface{} into a struct called AllocationProperties along the same lines as AssetProperties. 

Additionally calls to Aggregate via the `allocation` api no use a []string based argument rather than a properties object.

Testing:
Relevant unit and integration tests passing locally. Manual testing of Details and Reports pages, querying all aggregation options

Closed source PR:
https://github.com/kubecost/kubecost-cost-model/pull/379


pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,308 @@
...
+		return false
+	}
+
+	if p.ProviderID != that.ProviderID {
+		return false
+	}
+
+	pLabels := p.Labels
+	thatLabels := that.Labels
+	if len(pLabels) != len(thatLabels) {
```
nikovacevic, line 142: If the lengths are different, don't we already know that the two sets are not equal?

pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	strs = append(strs, fmt.Sprintf("Labels:{%s}", strings.Join(strs, ",")))
+
+	var AnnotationStrs []string
+	for k, prop := range p.Annotations {
+		AnnotationStrs = append(AnnotationStrs, fmt.Sprintf("%s:%s", k, prop))
+	}
+	strs = append(strs, fmt.Sprintf("Annotations:{%s}", strings.Join(strs, ",")))
+
+	return strings.Join(strs, ",")
```
nikovacevic, line 236: This is probably inconsequential, but I imagine that this could return some strings with strange combinations of commas and semi-colons. For complicated composite types like this, I'd tend to stick with semi-colons between fields and commas between list items within fields, e.g.
```
Cluster:cluster1;Node:node1;...;Services:service1,service2;Labels:{app:app1,env:env1}
```

(I might be wrong, but I think the current function would generate something like this:)
```
Cluster:cluster1,Node:node1,...,Services:service1;service2,Labels:{app:app1,env:env1}
```

pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// AggregationStrings converts a AllocationProperties object into a slice of strings
+// representing a request to aggregate by certain properties.
+// NOTE: today, the ordering of the properties *has to match the ordering
+// of the allocaiton function generateKey*
```
nikovacevic, line 242: *allocation

pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	if p.ProviderID != "" {
+		aggStrs = append(aggStrs, AllocationProviderIDProp.String())
+	}
+	if len(p.Services) > 0 {
+		aggStrs = append(aggStrs, AllocationServiceProp.String())
+	}
+	if len(p.Services) > 0 {
+		aggStrs = append(aggStrs, AllocationServiceProp.String())
+	}
```
nikovacevic, line 278: Dupe?

pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,308 @@
...
+	if p.Namespace != "" {
+		aggStrs = append(aggStrs, AllocationNamespaceProp.String())
+	}
+	if p.Pod != "" {
+		aggStrs = append(aggStrs, AllocationPodProp.String())
+	}
+	if p.ProviderID != "" {
+		aggStrs = append(aggStrs, AllocationProviderIDProp.String())
+	}
```
nikovacevic, line 272: Perhaps I'm confusing things, but we're probably going to want to talk about ordering here.

pkg/kubecost/allocation.go
```diff
@@ -358,28 +360,32 @@ func (a *Allocation) add(that *Allocation) {
...
+		if aNode == thatNode {
+			if a.Properties == nil {
+				a.Properties = &AllocationProperties{Node: aNode}
+			} else {
+				a.Properties.Node = aNode
+			}
 		}
+	} else {
+		a.Properties = nil
 	}
```
nikovacevic, line 386: This, we'll want to discuss. We only do this because of the old, terrible, map implementation.

nikovacevic (commented):
Awesome, Sean, this makes me so happy! The only thing we might want to discuss is whether or not the new `AllocationProperties` struct makes a good fit for the `AggregateBy` properties. It might, or it might not, but we can discuss the pros and cons, as well as alternatives.

pkg/kubecost/allocation.go
```diff
@@ -1194,140 +1177,120 @@ func (a *Allocation) generateKey(properties Properties) string {
...
-			controllerKind, err := a.Properties.GetControllerKind()
-			if err == nil {
-				names = append(names, controllerKind)
+	// Search for special case label for ETL conversion
+	aggControllerKind := ""
+	for _, agg := range aggregateBy {
+		if strings.HasPrefix(agg, "controllerKind:") {
+			aggControllerKind = strings.Split(agg, ":")[1]
+		}
+	}
```
nikovacevic, line 1186: I have a sense, but can you document how this code block is designed to work, precisely? (e.g. what are the legal values and what happens when legal and illegal values are passed?) And is there a reason for doing it this way, as opposed to simply letting users pass `daemonset`, `deployment`, or `statefulset` as opposed to `controllerKind:...`?

pkg/kubecost/allocation.go
```diff
@@ -1194,140 +1177,120 @@ func (a *Allocation) generateKey(properties Properties) string {
...
+
+	for _, agg := range aggregateBy {
+		switch true {
+		case agg == string(AllocationClusterProp):
```
nikovacevic, line 1190: We can potentially make these just type `string` so that type casts aren't required. Have a sense of whether or not that's a good idea?

pkg/kubecost/allocation.go
```diff
@@ -1194,140 +1177,120 @@ func (a *Allocation) generateKey(properties Properties) string {
...
-			annotationNames := []string{}
-
-			aggAnnotations, _ := properties.GetAnnotations()
-			for annotationName := range aggAnnotations {
-				if val, ok := annotations[annotationName]; ok {
-					annotationNames = append(annotationNames, fmt.Sprintf("%s=%s", annotationName, val))
-				} else if indexOf(UnallocatedSuffix, annotationNames) == -1 { // if UnallocatedSuffix not already in names
-					annotationNames = append(annotationNames, UnallocatedSuffix)
+		case strings.HasPrefix(agg, "controllerKind:"):
+			continue
```
nikovacevic, line 1235: This is a little confusing to read because it relies on a separate block above, but it's not exactly obvious how/why unless you understand the design of that property.

pkg/kubecost/allocation_test.go
```diff
@@ -856,7 +864,7 @@ func TestAllocationSet_AggregateBy(t *testing.T) {
 
 	// 1a AggregationProperties=(Cluster)
 	as = generateAllocationSet(start)
-	err = as.AggregateBy(Properties{ClusterProp: ""}, nil)
+	err = as.AggregateBy([]string{string(AllocationClusterProp)}, nil)
```
nikovacevic, line 867: Understand why this is the way it is, but I think it's also okay to just use `[]string{"cluster"}` here for readability. Or we can change the type of `AllocationClusterProp` to be `string`. What do you think?

pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,354 @@
...
+	pLabels := p.Labels
+	thatLabels := that.Labels
+	if len(pLabels) == len(thatLabels) {
+		for k, pv := range pLabels {
+			tv, ok := thatLabels[k]
+			if !ok || tv != pv {
+				return false
+			}
+		}
+	}
```
nikovacevic, line 163: I think we need an else here?

pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,354 @@
...
+		for i, pv := range pServices {
+			tv := thatServices[i]
+			if tv != pv {
+				return false
+			}
+		}
+	} else {
+		return false
+	}
+	return true
```
nikovacevic, line 192: Very nit-picky here, but it's nice to use spacing to communicate "paragraphs" of code that relate to each other. Did a double-take here, thinking this final return was a part of the services idea, then realized it was just the final return of the whole function. Something to keep in mind.

pkg/kubecost/allocationprops.go
```diff
@@ -0,0 +1,354 @@
...
+	strs = append(strs, fmt.Sprintf("Annotations:{%s}", strings.Join(strs, ",")))
+
+	return fmt.Sprintf("{%s}", strings.Join(strs, "; "))
+}
+
+// AggregationStrings converts a AllocationProperties object into a slice of strings
+// representing a request to aggregate by certain properties.
+// NOTE: today, the ordering of the properties *has to match the ordering
+// of the allocation function generateKey*
+func (p *AllocationProperties) AggregationStrings() []string {
```
nikovacevic, line 291: Can we delete this function now? We don't want people doing this, right?

nikovacevic (changes requested):
Just a few small things to address, some nits, but otherwise this is looking fantastic. Very stoked with how this turned out!

Let me know if you want to discuss anything (particularly the `controllerKind` syntax) or if you disagree and want to keep everything as-is.

pkg/kubecost/allocation_test.go
Sean-Holcomb: I do not see a problem with this especially since, all of these are going through parseProperty anyway

pkg/costmodel/allocation.go
```diff
@@ -1001,11 +1001,10 @@ func resToPodAnnotations(resPodAnnotations []*prom.QueryResult) map[podKey]map[s
 func applyLabels(podMap map[podKey]*Pod, namespaceLabels map[namespaceKey]map[string]string, podLabels map[podKey]map[string]string) {
 	for podKey, pod := range podMap {
 		for _, alloc := range pod.Allocations {
-			allocLabels, err := alloc.Properties.GetLabels()
-			if err != nil {
-				allocLabels = map[string]string{}
+			allocLabels := alloc.Properties.Labels
+			if allocLabels == nil {
+				allocLabels = make(map[string]string)
```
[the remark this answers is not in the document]
Sean-Holcomb: I was getting nil panics when I launched my pod

nikovacevic (approved):
LGTM!

If you don't mind, can you `go fmt` this as well before merging? Could just be GitHub, but **pkg/kubecost/allocation.go** looks like it might have some funny formatting.

PR #770, opened 2021-04-13
Added Chinese pricing url for Chinese region

Sean-Holcomb:
Update AWS pricing so nodes in Chinese regions use Chinese pricing URL. It is worth mentioning that the multi-region pricing endpoint for china only includes the two Chinese regions cn-north-1 and cn-northwest-1, while the other multi-region endpoint includes all other regions but not the Chinese ones. One edge cast of note here is that if no region is detected on any nodes then it will default to the non-Chinese multi-region endpoint, which could cause issues for a multi-region Chinese cluster. It is not clear at this point if this is an actual concern though.

Testing:
To test this I hard coded the Chinese regions and checked that the pricing api was being hit correctly.
No hard coding:
Screen Shot 2021-04-13 at 11 06 43 AM

Hard coded "cn-north-1"
Screen Shot 2021-04-13 at 11 07 28 AM

While this did work here it does cause issues down the line when the nodes are actually looking for their pricing data in the returned json:
Screen Shot 2021-04-13 at 11 13 41 AM

Hard coded "cn-northwest-1"
Screen Shot 2021-04-13 at 11 15 06 AM

Hard coded "cn-northwest-"
This is an invalid region and causes Kubecost to crash. In this case the above queries hung without completing and this was show when visiting Kubecost
Screen Shot 2021-04-13 at 11 21 51 AM

Edit:
Returned prices for Chinese regions are in CNY

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -191,15 +191,28 @@ type AWSOfferTerm struct {
 	PriceDimensions map[string]*AWSRateCode `json:"priceDimensions"`
 }
 
+func (ot *AWSOfferTerm) String() string {
+	var strs []string
+	for k, rc := range ot.PriceDimensions {
+		strs = append(strs, fmt.Sprintf("%s:%s", k, rc.String()))
+	}
+	return fmt.Sprintf("%s:%s", ot.Sku, strings.Join(strs, ","))
+}
+
 // AWSRateCode encodes data about the price of a product
 type AWSRateCode struct {
 	Unit         string          `json:"unit"`
 	PricePerUnit AWSCurrencyCode `json:"pricePerUnit"`
 }
 
+func (rc *AWSRateCode) String() string {
+	return fmt.Sprintf("{unit: %s, pricePerUnit: %v", rc.Unit, rc.PricePerUnit)
+}
+
 // AWSCurrencyCode is the localized currency. (TODO: support non-USD)
 type AWSCurrencyCode struct {
-	USD string `json:"USD"`
+	USD string `json:"USD,omitempty"`
+	CNY string `json:"CNY,omitempty"`
 }
 
 // AWSProductTerms represents the full terms of the product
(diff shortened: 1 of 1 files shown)


dwbrown2:
> While this did work here it does cause issues down the line when the nodes are actually looking for their pricing data in the returned json:

Were you able to verify an actual pricing match in any form? That seems valuable to just confirm that the structure of the JSON data is as expected..

> This is an invalid region and causes Kubecost to crash. In this case the above queries hung without completing and this was show when visiting Kubecost

Maybe out of scope for this PR, but is this the behavior we want?

nikovacevic (commented):
Looks like a good start. Do we know if there are use-cases today that would require us to parse both China and non-China responses for some reason? And how to we know that the condition `strings.Contains(currentNodeRegion, "cn-")` works for identifying that?

Sean-Holcomb:
> Looks like a good start. Do we know if there are use-cases today that would require us to parse both China and non-China responses for some reason? And how to we know that the condition `strings.Contains(currentNodeRegion, "cn-")` works for identifying that?

The list of Nodes that comes into this method is on a per Cluster basis, so there should not be an issue with cross region, I was just making a note of it. I also update `strings.Contains()` to `strings.HasPrefix()` so that it is more specific

Sean-Holcomb:
> > While this did work here it does cause issues down the line when the nodes are actually looking for their pricing data in the returned json:
> 
> Were you able to verify an actual pricing match in any form? That seems valuable to just confirm that the structure of the JSON data is as expected..
> 
> > This is an invalid region and causes Kubecost to crash. In this case the above queries hung without completing and this was show when visiting Kubecost
> 
> Maybe out of scope for this PR, but is this the behavior we want?

There was in fact an issue parsing the JSON related to special aws codes and the currency

In terms of that behavior, I would say no, definitely is not what we want. I think it is reasonable to say that the api could fail in a different way and this could happen and that we should be more defensive about it.

AjayTripathy:
> Do we know if there are use-cases today that would require us to parse both China and non-China responses for some reason?

IMO this should really not be possible/encouraged/supported for firewall reasons.

Sean-Holcomb:
@dwbrown2 
Looks like the bad URL was causing an infinite loop, hence the crash that is fixed now.

When node is hard coded to say it is from Chinese region, `model/allNodePricing` returned this showing that the Chinese nodes and volumes are parsing from the JSON
Screen Shot 2021-04-14 at 10 05 45 AM

AjayTripathy (approved):

PR #773, opened 2021-04-19
Add check for Chinese prices on Node creation

Sean-Holcomb:
Added in check for Chinese pricing on node before error is thrown. Region is not readily available in this method, so checking the pricing dimension map for the SKU with both the Chinese and US suffixes seems to be the best method here.
Testing:
Run AWS cluster without issues, unable to test effectiveness on Chinese cluster, will wait for feed back from user

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -854,11 +854,11 @@ func (aws *AWS) DownloadPricingData() error {
 							cost = offerTerm.PriceDimensions[sku.(string)+OnDemandRateCodeCn+HourlyRateCodeCn].PricePerUnit.CNY
 						}
 						if strings.Contains(key, "EBS:VolumeP-IOPS.piops") {
-						// If the specific UsageType is the per IO cost used on io1 volumes
-						// we need to add the per IO cost to the io1 PV cost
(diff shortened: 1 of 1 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -1071,11 +1071,20 @@ func (aws *AWS) createNode(terms *AWSProductTerms, usageType string, k Key) (*No
...
-	if !ok {
-		return nil, fmt.Errorf("Could not fetch data for \"%s\"", k.ID())
+	if ok {
+		cost = c.PricePerUnit.USD
+	} else {
+		// Check for Chinese pricing before throwing error
+		c, ok = terms.OnDemand.PriceDimensions[terms.Sku+OnDemandRateCodeCn+HourlyRateCodeCn]
+		if ok {
+			cost = c.PricePerUnit.CNY
+		} else{
```
nikovacevic, line 1083: Super-nit, but can we `go fmt` this?

nikovacevic (approved):
LGTM (besides the formatting, which isn't blocking, but would be a nice touch-up)

nikovacevic:
Formatting looks great, thanks!

PR #775, opened 2021-04-21
Fixed a bad URL in the allocations UI

SheepGotoHeaven:
URL was pointing at `/model/allocation/compute` but needs `/allocation/compute`


nikovacevic (approved):
LGTM!

PR #779, opened 2021-04-22
Alternate source for PV provider ID on aws

Sean-Holcomb:
Check PV spec for alternate location of Provider ID. `pv.Spec.CSI.VolumeHandle` is specified in the AWS API so there should be no problem accessing the value described in https://github.com/kubecost/cost-model/issues/771


nikovacevic (approved):
LGTM. Were we able to test this on an actual volume?

Sean-Holcomb:
> LGTM. Were we able to test this on an actual volume?

No actual volume, just against the user spec and matching aws api. I am pretty confident in this one just because of its simplicity and the fact that it is the second case in the control flow, so it should have no effect on PVs that already have provider ids.

PR #790, opened 2021-05-05
Sean/pv alloc recon

Sean-Holcomb:
This PR Updates the Allocation Reconciliation process to read in Disk assets and adjust PV costs of allocation accordingly.

Methodology:
PV usage is distributed evenly to the containers within the pods that they are attached to via PVC. This is reflected in an Allocation's PVByteHour value. However given that a Pods can have multiple PVCs more information is needed on which PV is responsible for what usage and cost of an Allocation. To address this I have added the PVBreakdown field to the Allocation Properties which is a map of PV name to byteHours and cost that PV is responsible for. Using this addition information, a proper adjustment can be made to the cost of the Allocation for each PV it has byteHours for. The equation is ALLOC_PVADJUSTMENT = (PV_COST + PV_ADJUSTMENT) * ALLOC_PVBYTEHOURS / PV_BYTEHOURS - ALLOC_PVCOST

Testing:
Added Unit tests for multiple PV attachments
Screen Shot 2021-05-04 at 6 13 21 PM

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -390,10 +390,10 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 					alloc.PVCost += cost / count
 
 					// record the amount of total PVBytes Hours attributable to a given PV
-					if alloc.Properties.PVBreakDown == nil {
-						alloc.Properties.PVBreakDown = map[string]kubecost.PVUsage{}
+					if alloc.Properties.PVBreakdown == nil {
+						alloc.Properties.PVBreakdown = map[string]kubecost.PVUsage{}
 					}
-					alloc.Properties.PVBreakDown[pvc.Volume.Name] = kubecost.PVUsage{
+					alloc.Properties.PVBreakdown[pvc.Volume.Name] = kubecost.PVUsage{
 						ByteHours: pvc.Bytes * hrs / count,
 						Cost:      cost / count,
 					}
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -361,7 +361,6 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 			alloc.CPUCost = alloc.CPUCoreHours * node.CostPerCPUHr
 			alloc.RAMCost = (alloc.RAMByteHours / 1024 / 1024 / 1024) * node.CostPerRAMGiBHr
 			alloc.GPUCost = alloc.GPUHours * node.CostPerGPUHr
-
 			if pvcs, ok := podPVCMap[podKey]; ok {
 				for _, pvc := range pvcs {
 					// Determine the (start, end) of the relationship between the
@@ -389,6 +388,15 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 					// weighted by count (i.e. the number of containers in the pod)
 					alloc.PVByteHours += pvc.Bytes * hrs / count
 					alloc.PVCost += cost / count
+
+					// record the amount of total PVBytes Hours attributable to a given PV
+					if alloc.Properties.PVBreakDown == nil {
+						alloc.Properties.PVBreakDown = map[string]kubecost.PVUsage{}
+					}
+					alloc.Properties.PVBreakDown[pvc.Volume.Name] = kubecost.PVUsage{
+						ByteHours: pvc.Bytes * hrs / count,
+						Cost:      cost / count,
+					}
 				}
 			}
 
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -1615,14 +1615,14 @@ func (a *Allocation) reconcileNodes(nodeByProviderID map[string]*Node) {
 }
 
 func (a *Allocation) reconcileDisks(diskByName map[string]*Disk) {
-	pvBreakDown := a.Properties.PVBreakDown
-	if pvBreakDown == nil {
+	pvBreakdown := a.Properties.PVBreakdown
+	if pvBreakdown == nil {
 		// No PV usage to reconcile
 		return
 	}
 	// Set PV Adjustment for allocation to 0 for idempotency
 	a.PVCostAdjustment = 0.0
-	for pvName, pvUsage := range pvBreakDown {
+	for pvName, pvUsage := range pvBreakdown {
 		disk, ok := diskByName[pvName]
 		if !ok {
 			// Failed to find disk in assets
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -67,6 +67,7 @@ type Allocation struct {
 	LoadBalancerCost       float64               `json:"loadBalancerCost"`
 	PVByteHours            float64               `json:"pvByteHours"`
 	PVCost                 float64               `json:"pvCost"`
+	PVAdjustment           float64               `json:"pvAdjustment"`
 	RAMByteHours           float64               `json:"ramByteHours"`
 	RAMBytesRequestAverage float64               `json:"ramByteRequestAverage"`
 	RAMBytesUsageAverage   float64               `json:"ramByteUsageAverage"`
@@ -74,7 +75,6 @@ type Allocation struct {
 	RAMAdjustment          float64               `json:"ramAdjustment"`
 	SharedCost             float64               `json:"sharedCost"`
 	ExternalCost           float64               `json:"externalCost"`
-
 	// RawAllocationOnly is a pointer so if it is not present it will be
 	// marshalled as null rather than as an object with Go default values.
 	RawAllocationOnly *RawAllocationOnlyData `json:"rawAllocationOnly"`
@@ -155,6 +155,7 @@ func (a *Allocation) Clone() *Allocation {
 		LoadBalancerCost:       a.LoadBalancerCost,
 		PVByteHours:            a.PVByteHours,
 		PVCost:                 a.PVCost,
+		PVAdjustment:           a.PVAdjustment,
 		RAMByteHours:           a.RAMByteHours,
 		RAMBytesRequestAverage: a.RAMBytesRequestAverage,
 		RAMBytesUsageAverage:   a.RAMBytesUsageAverage,
@@ -232,6 +233,9 @@ func (a *Allocation) Equal(that *Allocation) bool {
 	if !util.IsApproximately(a.PVCost, that.PVCost) {
 		return false
 	}
+	if !util.IsApproximately(a.PVAdjustment, that.PVAdjustment) {
+		return false
+	}
 	if !util.IsApproximately(a.RAMByteHours, that.RAMByteHours) {
 		return false
 	}
@@ -269,7 +273,7 @@ func (a *Allocation) Equal(that *Allocation) bool {
 
 // TotalCost is the total cost of the Allocation
 func (a *Allocation) TotalCost() float64 {
-	return a.CPUTotalCost() + a.GPUTotalCost() + a.RAMTotalCost() + a.PVCost + a.NetworkCost + a.SharedCost + a.ExternalCost + a.LoadBalancerCost
+	return a.CPUTotalCost() + a.GPUTotalCost() + a.RAMTotalCost() + a.PVTotalCost() + a.NetworkCost + a.SharedCost + a.ExternalCost + a.LoadBalancerCost
 }
 
 func (a *Allocation) CPUTotalCost() float64 {
@@ -282,6 +286,10 @@ func (a *Allocation) GPUTotalCost() float64 {
 
 func (a *Allocation) RAMTotalCost() float64 {
 	return a.RAMCost + a.RAMAdjustment
+}
+
+func (a *Allocation) PVTotalCost() float64 {
+	return a.PVCost + a.PVAdjustment
 }
 
 // CPUEfficiency is the ratio of usage to request. If there is no request and
@@ -374,6 +382,7 @@ func (a *Allocation) MarshalJSON() ([]byte, error) {
 	jsonEncodeFloat64(buffer, "pvBytes", a.PVBytes(), ",")
 	jsonEncodeFloat64(buffer, "pvByteHours", a.PVByteHours, ",")
 	jsonEncodeFloat64(buffer, "pvCost", a.PVCost, ",")
+	jsonEncodeFloat64(buffer, "pvAdjustment", a.PVAdjustment, ",")
 	jsonEncodeFloat64(buffer, "ramBytes", a.RAMBytes(), ",")
 	jsonEncodeFloat64(buffer, "ramByteRequestAverage", a.RAMBytesRequestAverage, ",")
 	jsonEncodeFloat64(buffer, "ramByteUsageAverage", a.RAMBytesUsageAverage, ",")
@@ -513,6 +522,7 @@ func (a *Allocation) add(that *Allocation) {
 	a.CPUAdjustment += that.CPUAdjustment
 	a.RAMAdjustment += that.RAMAdjustment
 	a.GPUAdjustment += that.GPUAdjustment
+	a.PVAdjustment += that.PVAdjustment
 
 	// Any data that is in a "raw allocation only" is not valid in any
 	// sort of cumulative Allocation (like one that is added).
@@ -1179,7 +1189,7 @@ func computeIdleCoeffs(options *AllocationAggregationOptions, as *AllocationSet,
 		} else {
 			coeffs[clusterID][name]["cpu"] += alloc.CPUTotalCost()
 			coeffs[clusterID][name]["gpu"] += alloc.GPUTotalCost()
-			coeffs[clusterID][name]["ram"] += alloc.RAMCost
+			coeffs[clusterID][name]["ram"] += alloc.RAMTotalCost()
 
 			totals[clusterID]["cpu"] += alloc.CPUTotalCost()
 			totals[clusterID]["gpu"] += alloc.GPUTotalCost()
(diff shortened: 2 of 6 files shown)


pkg/kubecost/kubecost_codecs.go
```diff
@@ -545,6 +572,34 @@ func (target *AllocationProperties) UnmarshalBinary(data []byte) (err error) {
...
+
+			// --- [begin][read][struct](PVUsage) ---
+			dd := &PVUsage{}
+			ee := buff.ReadInt()     // byte array length
+			ff := buff.ReadBytes(ee) // byte array
+			errA := dd.UnmarshalBinary(ff)
+			if errA != nil {
+				return errA
+			}
+			zzz := *dd
```
Sean-Holcomb, line 594: Had to manually change = to := here

pkg/kubecost/bingen.go
```diff
@@ -25,5 +25,6 @@ package kubecost
 // @bingen:generate:AllocationLabels
 // @bingen:generate:AllocationAnnotations
 // @bingen:generate:RawAllocationOnlyData
+// @bingen:generate:PVUsage
 
-//go:generate bingen -package=kubecost -version=12 -buffer=github.com/kubecost/cost-model/pkg/util
+//go:generate bingen -package=kubecost -version=13 -buffer=github.com/kubecost/cost-model/pkg/util
```
Sean-Holcomb, line 30: @AjayTripathy I did end up needing to rebuild bingen here

pkg/kubecost/kubecost_codecs.go
nikovacevic: That's not good. We'll have to see what's up, but manually adjusting the codec is a red flag.

nikovacevic: cc @mbolt35

pkg/kubecost/allocationprops.go
```diff
@@ -74,6 +74,7 @@ type AllocationProperties struct {
 	ProviderID     string                `json:"providerID,omitempty"`
 	Labels         AllocationLabels      `json:"allocationLabels,omitempty"`
 	Annotations    AllocationAnnotations `json:"allocationAnnotations,omitempty"`
+	PVBreakdown    map[string]PVUsage    `json:"pvBreakDown,omitempty"`
```
nikovacevic, line 77: Is the idea here that PVCost can go away, even? Is in, could we do something like:
```go
func (a *Allocation) PVCost() float64 {
    cost := 0.0
    for _, pv := range a.Properties.PVBreakdown {
        cost += pv.Cost
    }
    return cost
}
```
Or am I misunderstanding what this does?

nikovacevic: Also, I would argue this belongs on `Allocation`, not on `AllocationProperties`, if it is storing allocation data as opposed to metadata. How would you feel about that change?

nikovacevic: In general (performance not withstanding) I do like the idea of breaking down PV data on each Allocation for the sake of detail.

nikovacevic (commented):
Few questions to address, then I'll re-review!

Sean-Holcomb:
New closed source PR https://github.com/kubecost/kubecost-cost-model/pull/397

pkg/kubecost/allocationprops.go
Sean-Holcomb: Yeah definitely, I did the same for PVByteHours(). I think this is an improvement but there are some api repercussions that I am unsure about.

pkg/kubecost/kubecost_codecs.go
```diff
@@ -302,45 +324,70 @@ func (target *Allocation) UnmarshalBinary(data []byte) (err error) {
...
+
+			// --- [begin][read][struct](PVUsage) ---
+			ee := &PVUsage{}
+			ff := buff.ReadInt()     // byte array length
+			gg := buff.ReadBytes(ff) // byte array
+			errE := ee.UnmarshalBinary(gg)
+			if errE != nil {
+				return errE
+			}
+			z := *ee
```
Sean-Holcomb, line 346: @mbolt35 this is now where I had to change it from = to :=

pkg/costmodel/allocation.go
```diff
@@ -1665,8 +1670,13 @@ func applyUnmountedPVs(window kubecost.Window, podMap map[podKey]*Pod, pvMap map
...
 		podMap[key].Allocations[container].Properties.Container = container
-		podMap[key].Allocations[container].PVByteHours = unmountedPVBytes[cluster] * window.Minutes() / 60.0
-		podMap[key].Allocations[container].PVCost = amount
+		unmountedBreakDown := map[string]kubecost.PVUsage{
+			kubecost.UnmountedSuffix: {
+				ByteHours: unmountedPVBytes[cluster] * window.Minutes() / 60.0,
+				Cost:      amount,
+			},
+		}
+		podMap[key].Allocations[container].AddPVBreakDown(unmountedBreakDown)
```
Sean-Holcomb, line 1679: @nikovacevic I am unsure about this update that was necessary because I removed PVCost and PVByteHours

pkg/kubecost/allocationprops.go
Sean-Holcomb: Looks like the refactor did cause some memory issues, cost-analyser pod is now failing to schedule. and it seems to be caused by the pod exceeding memory limits, I am also seeing a lot of failed health checks to prometheus

nikovacevic: On your cluster or on the scale cluster? That seems pretty strange tbh for such a small cluster. Have you confirmed that these changes are the cause of that? Can you resolve the failing-to-schedule by using a different image?

Sean-Holcomb: Yes I have an image from Tuesday afternoon before the refactor that schedules and works correctly. I am honestly still not 100% sure what is going on with it.

pkg/kubecost/allocation.go
```diff
@@ -65,16 +65,15 @@ type Allocation struct {
 	GPUCostAdjustment      float64               `json:"gpuCostAdjustment"`
 	NetworkCost            float64               `json:"networkCost"`
 	LoadBalancerCost       float64               `json:"loadBalancerCost"`
-	PVByteHours            float64               `json:"pvByteHours"`
-	PVCost                 float64               `json:"pvCost"`
+	PVBreakdown            map[string]PVUsage    `json:"pvBreakDown"`
```
nikovacevic, line 68: What do you think about just calling this `PVs`? Not sure what "Breakdown" adds, and I could see us adding information to this, turning it into a more robust record of the related PVs.

pkg/kubecost/allocation.go
```diff
@@ -108,6 +107,13 @@ type RawAllocationOnlyData struct {
 	RAMBytesUsageMax float64 `json:"ramByteUsageMax"`
 }
 
+// PVUsage contains the byte hour usage
+// and cost of an Allocation for a single PV
+type PVUsage struct {
```
nikovacevic, line 112: This name I'd also like to think about. "Usage" has a loaded meaning in Kubecost; namely, a resource "allocation" is the max of "usage" and "request." I wonder if we could get away with just calling this `PV` or `PVC` or perhaps `PVAllocation`? I'm foreseeing us potentially looking to expand what we store in this, so a more general name might age better.

Also, should we keep the name on the object as `Name string`? For the purpose of this PR, you have name on the `PVBreakdown` map, but that makes individual instances of this type less useful stand-alone.

pkg/kubecost/allocation.go
```diff
@@ -284,6 +306,26 @@ func (a *Allocation) RAMTotalCost() float64 {
...
+	return cost
+}
+
+func (a *Allocation) PVByteHours() float64 {
+	byteHours := 0.0
+	for _, pv := range a.PVBreakdown {
+		byteHours += pv.ByteHours
+	}
+	return byteHours
+}
```
nikovacevic, line 327: Cool cool. I like having these as function. Do you mind commenting them, as they are public?

If your IDE doesn't warn you about un-commented exported functions, we can try to set that up, too. Everyone who uses VSCode with the Go plugin will get warned about this, so it'd be good to get parity across everyone's IDEs.

pkg/kubecost/allocation.go
```diff
@@ -381,8 +423,10 @@ func (a *Allocation) MarshalJSON() ([]byte, error) {
 	jsonEncodeFloat64(buffer, "networkCost", a.NetworkCost, ",")
 	jsonEncodeFloat64(buffer, "loadBalancerCost", a.LoadBalancerCost, ",")
 	jsonEncodeFloat64(buffer, "pvBytes", a.PVBytes(), ",")
-	jsonEncodeFloat64(buffer, "pvByteHours", a.PVByteHours, ",")
-	jsonEncodeFloat64(buffer, "pvCost", a.PVCost, ",")
+	jsonEncodeFloat64(buffer, "pvByteHours", a.PVByteHours(), ",")
+	jsonEncodeFloat64(buffer, "pvCost", a.PVCost(), ",")
+	jsonEncode(buffer, "pvBreakdown", a.PVBreakdown, ",")
```
nikovacevic, line 428: I didn't see this field in the payload you listed in the PR description. It'll be cool to see!

pkg/kubecost/allocation.go
```diff
@@ -506,28 +550,49 @@ func (a *Allocation) add(that *Allocation) {
...
+			apv, ok := a.PVBreakdown[pvName]
+			if !ok {
+				apv = PVUsage{}
+			}
+			apv.Cost += pv.Cost
+			apv.ByteHours += pv.ByteHours
+			a.PVBreakdown[pvName] = apv
+		}
+	}
+}
```
nikovacevic, line 594: I think this function would make more sense as a function called `Add` on the type `PVBreakdown`. It would be a shame to want to add two `PVBreakdown` instances and have to put them in `Allocation` instances to do so.

pkg/kubecost/allocation.go
```diff
@@ -506,28 +550,49 @@ func (a *Allocation) add(that *Allocation) {
...
 	a.GPUCost += that.GPUCost
 	a.RAMCost += that.RAMCost
-	a.PVCost += that.PVCost
 	a.NetworkCost += that.NetworkCost
 	a.LoadBalancerCost += that.LoadBalancerCost
 	a.SharedCost += that.SharedCost
 	a.ExternalCost += that.ExternalCost
 
+	// Sum PV Breakdown
+	a.AddPVBreakDown(that.PVBreakdown)
```
nikovacevic, line 564: See comment about this.

pkg/kubecost/allocation_test.go
```diff
@@ -938,7 +965,7 @@ func TestAllocationSet_AggregateBy(t *testing.T) {
 	//         container8: an[team=team2]          6.00   1.00   1.00   1.00   1.00   1.00   1.00
 	//         container9: an[team=team1]          6.00   1.00   1.00   1.00   1.00   1.00   1.00
 	// +----------------------------------------+------+------+------+------+------+------+------+
-	//   cluster2 subtotal                        46.00  11.00  11.00   6.00   6.00   6.00   6.00
+	//   cluster2 subtotal                        46.00  11.00  11.00   6.00  6.00   6.00   6.00
```
nikovacevic, line 968: Why this? Haha

pkg/kubecost/allocation_test.go
```diff
@@ -1751,6 +1778,12 @@ func TestAllocationSet_ReconcileAllocations(t *testing.T) {
 	for key := range as.idleKeys {
 		as.Delete(key)
 	}
+	// add reconcilable pvs to pod-mno
+	for _, a := range as.allocations {
+		if a.Properties.Pod == "pod-mno" {
+			a.AddPVBreakDown(map[string]PVUsage{"disk1": {Cost: 2.5, ByteHours: 2.5 * gb}, "disk2": {Cost: 5, ByteHours: 5 * gb}})
```
nikovacevic, line 1784: This is a pretty brutal line. Maybe we can multi-line and indent it for readability?

pkg/costmodel/allocation.go
nikovacevic: See comment about this function

pkg/kubecost/allocation.go
nikovacevic: Also, do you have strong feelings that this should be a value type `PVUsage` as opposed to a pointer-type `*PVUsage`? Might make things _slightly_ more performant to use `map[string]*PVUsage`.

nikovacevic: Also also, I think we might need to have this key off `Cluster` and `Name`, so probably going to need something like
```go
map[pvKey]PVUsage

type pvKey struct {
    cluster string
    name    string
}
```

nikovacevic (changes requested):
Looking good so far! Let me know if you have any questions, or want to discuss anything face-to-face.

pkg/kubecost/allocation.go
Sean-Holcomb: Will replace, once I get a pod up and running

pkg/kubecost/allocation.go
Sean-Holcomb: I turned it on and cranked the warning level up to error
Screen Shot 2021-05-06 at 5 54 09 PM

pkg/kubecost/allocation.go
```diff
@@ -65,7 +65,7 @@ type Allocation struct {
 	GPUCostAdjustment      float64               `json:"gpuCostAdjustment"`
 	NetworkCost            float64               `json:"networkCost"`
 	LoadBalancerCost       float64               `json:"loadBalancerCost"`
-	PVBreakdown            map[string]PVUsage    `json:"pvBreakDown"`
+	PVs                    PV
```
nikovacevic, line 68: We'll probably want this to have `json:"-"` to explicitly indicate not to JSON encode/decode, and for consistency.

pkg/kubecost/allocation.go
```diff
@@ -107,9 +107,52 @@ type RawAllocationOnlyData struct {
 	RAMBytesUsageMax float64 `json:"ramByteUsageMax"`
 }
 
-// PVUsage contains the byte hour usage
+// PV is a map of Disk Asset Identifiers to the
+// usage of them by an Allocation as recorded in a PVAllocation
+type PV map[PVKey]*PVAllocation
```
nikovacevic, line 112: Sorry to be nitpicky, but this name being singular looks misleading now. "PVs" or "PVAllocations" might be better imo. I like the way that it looks with that name on the `Allocation` type. What do you think?

pkg/kubecost/allocation.go
```diff
@@ -107,9 +107,52 @@ type RawAllocationOnlyData struct {
...
+// PVKey for identifying Disk type assets
+type PVKey struct {
+	Cluster string `json:"cluster"`
+	Name    string `json:"name"`
+}
+
+// PVAllocation contains the byte hour usage
 // and cost of an Allocation for a single PV
-type PVUsage struct {
+type PVAllocation struct {
```
nikovacevic, line 158: Should we consider adding "Name" here? I know it's in the map, but do you see us wanting to use this struct standalone at some point?

nikovacevic (commented):
Looking good! Just these last few small comments to address, and discussing performance and bingen issues out-of-band.

pkg/kubecost/allocation.go
Sean-Holcomb: I don't really see it as ever becoming a stand alone, at this point I think it would be better to keep it more light weight, if we do need to add it in the future its not like it will be particularly difficult.

Sean-Holcomb:
Memory Test:
To make sure that the new schema does not use excessive memory I ran a comparison of Asset and Allocation ETL rebuilds on the EKS-Integration-Test cluster. 
In the below images the green line is the new image while the brown line is the develop line. Overall there does appear to be an increase in memory usage with the new schema which is to be expected, The first peaks on both lines are the Asset ETL run and the second peaks are the Allocation ETL rebuild. Besides the slightly higher general memory usage it is also worth noting that the green Allocation ETL rebuild is much more pronounced of a peak than that on the brown line and does not appear to have the same memory usage tapper off. In general the memory increase doesn't not seem to be highly significant and was not the cause of the scheduling issues I experienced in some of my earlier testing
Screen Shot 2021-05-07 at 3 37 25 PM
Screen Shot 2021-05-07 at 3 37 09 PM

pkg/kubecost/kubecost_codecs.go
```diff
@@ -302,45 +342,87 @@ func (target *Allocation) UnmarshalBinary(data []byte) (err error) {
...
+		for i := 0; i < dd; i++ {
+			// --- [begin][read][struct](PVKey) ---
+			ee := &PVKey{}
+			ff := buff.ReadInt()     // byte array length
+			gg := buff.ReadBytes(ff) // byte array
+			errE := ee.UnmarshalBinary(gg)
+			if errE != nil {
+				return errE
+			}
+			v := *ee
```
Sean-Holcomb, line 362: added semi-colon here

nikovacevic (approved):
Thanks for your hard work @Sean-Holcomb LGTM

pkg/kubecost/kubecost_codecs.go
Sean-Holcomb: @mbolt35

Sean-Holcomb:
@mbolt35 released bingen v0.0.4 which fixes that problem

I have also added additional adjustment fields to the allocation object to reduce future, bingen version updates

nikovacevic (approved):
I'm happy to approve this, but I'll just say that the `Allocation` type is starting to look pretty bloated to me. Not sure what we can do about it, given that we've just decided to run wild with the idea of "adjustments" for everything. But I will say that doubling the number of fields is not helping our case w/r/t payload size. If it becomes an issue, we can always revisit later (or perhaps define more "views" with less data, although that has downsides, too) but want to flag sooner than later.

cc @AjayTripathy

Sean-Holcomb:
@nikovacevic In terms of optic maybe we could move all the adjustments into an AllocationAdjustments struct, The other option, if we think that Adjustments for every cost category is overkill is to just dump them all into one Adjustment field, which I think would still work.

PR #796, opened 2021-05-11
Cluster Overhead Allocation

Sean-Holcomb:
Used by: https://github.com/kubecost/kubecost-cost-model/pull/414

This update pulls cost values from the Asset ETL for Cluster Management Fees and Attached Disks and applies them to allocations as a shared cost adjustments.  

Methodology:
These shared cost values are evenly split among the tenants of the node or cluster, because of this we first need to count the number of tenants per node and cluster before adding on the shared cost adjustment to know how much of the split that each allocation should be receiving. Unfortunately this adds an additional allocations loop to this query time process, and requires additional consideration whether it will be performant enough for our largest clients. It may be possible to put this value behind an additional flag additionally there has been some discussion creating a shared cluster cost allocation row, which might be better to add in at ETL build time.

Limitation:
Since this implementation gives equal portions of the shared resource to each of its tenants within a given time frame allocations that run many times such as jobs can receive an excessive share of the cost. 

Testing:
Added additional attached disks and cluster management types to the test assets in the unit tests to ensure that cost were being distributed correctly.
Results returning on AWS cluster:
Screen Shot 2021-05-10 at 5 43 25 PM

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -388,7 +388,7 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 					// weighted by count (i.e. the number of containers in the pod)
 					// record the amount of total PVBytes Hours attributable to a given PV
 					if alloc.PVs == nil {
-						alloc.PVs = kubecost.PV{}
+						alloc.PVs = kubecost.PVAllocations{}
 					}
 					pvKey := kubecost.PVKey{
 						Cluster: pvc.Cluster,
@@ -1678,7 +1678,7 @@ func applyUnmountedPVs(window kubecost.Window, podMap map[podKey]*Pod, pvMap map
 			Cluster: cluster,
 			Name:    kubecost.UnmountedSuffix,
 		}
-		unmountedBreakDown := kubecost.PV{
+		unmountedBreakDown := kubecost.PVAllocations{
 			pvKey: {
 				ByteHours: unmountedPVBytes[cluster] * window.Minutes() / 60.0,
 				Cost:      amount,
@@ -1730,7 +1730,7 @@ func applyUnmountedPVCs(window kubecost.Window, podMap map[podKey]*Pod, pvcMap m
 			Cluster: cluster,
 			Name:    kubecost.UnmountedSuffix,
 		}
-		unmountedBreakDown := kubecost.PV{
+		unmountedBreakDown := kubecost.PVAllocations{
 			pvKey: {
 				ByteHours: unmountedPVCBytes[key] * window.Minutes() / 60.0,
 				Cost:      amount,
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -387,10 +387,14 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 					// Apply the size and cost of the PV to the allocation, each
 					// weighted by count (i.e. the number of containers in the pod)
 					// record the amount of total PVBytes Hours attributable to a given PV
-					if alloc.PVBreakdown == nil {
-						alloc.PVBreakdown = map[string]kubecost.PVUsage{}
+					if alloc.PVs == nil {
+						alloc.PVs = kubecost.PV{}
 					}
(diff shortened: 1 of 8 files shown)


nikovacevic:
Why are these costs considered "shared cost adjustments" and not just shared costs or something else? I don't understand exactly how these relate to the process of reconciliation, which is what adjustments are generally for.

Sean-Holcomb:
I would say that the process of Asset Reconciliation is cost information moving from a CUR to the Asset ETL. Allocation Reconciliation is cost information moving from the Asset ETL to Allocation ETL. Any addition or subtraction of cost that happens through this process is an adjustment as it is a change to the initial estimate even if that initial estimate is zero. If you are willing to grant that definition then I think it is fair to have these values set by the reconciliation process as adjustments. Maybe there are other places where we are extracting information from the Asset ETL and moving it into Allocations ETL and if this information was put there as its own special "overhead" allocation like _idle_ it would no longer be an adjustment.

In terms of whether or not they should be shared adjustments, I think it makes intuitive sense for these to be shared cost, but the current usage of shared cost seems to have some conflict with these values, so I can see how you have hesitations about calling them that.

nikovacevic:
Not sure I entirely agree with this characterization of Asset-reconciliation and Allocation-reconciliation, but I see where you're coming from. I think where data is coming from should not be relevant to how we categorize it. For example, we could choose to reconcile a CUR with Allocation data without causing trouble; we could let someone manually tweak Asset data without involving a CUR; and if we use Asset data as a source for Allocation data, as we do for idle cost, it doesn't have to be categorized as an adjustment.

Perhaps the team will disagree, but I would argue that the distinction here is between a "measured" or "predicted" cost and a "corrected" or "actual" cost. It's not exactly black-and-white with cluster management, but we're measuring/predicting cluster management fees using a Prometheus metric that we export, so it doesn't feel like an adjustment to me. It could easily be a separate row (like idle) that can optionally be shared among the other allocations, or perhaps it could be something more novel for which there's no precedent.

Another way to think about this is to ask the question: what if we need to reconcile cluster management fees? We measure the fee at $30, but find out via the CUR that the actual cost is $25. Traditionally, we'd subtract $5 from the row using an adjustment column, but if the original cost is recorded under `sharedCostAdjustment`, where would that adjustment to that go?

Sean-Holcomb:
Your points are well taken, we could very well make another field OverheadCost with an OverheadCostAdjustment and populate the OverheadCost with the this asset costs and the adjustment field with the adjustment for those assets. I am not sure that anyone would care for that particular Adjustment field, but I think we all agree this probably isn't the approach that we want given it will just contribute to the growing number of fields on the Allocation Struct.

I think the idea of an Overhead Allocation is an alright idea, but then we add query options for sharing and splitting it, much like those for Idle, and when we end up sharing it it goes to sharedCost anyways. This might be a better approach and also forgoes sharedCostAdjustment, which could then be used in the way that you described it. This could also probably use the same loops as the idle calculations, but I think it is also time to think about moving all of this stuff away from query time as it seems that everyone is uneasy about it.

dwbrown2:
I'm supportive of whatever you guys think is best, but just know that if we
lump this into shared costs we are going to the get users asking for a
breakdown/more detail. I know this because this is already a feature
request for shared costs today :)

Webb
Kubecost, Co-founder

ᐧ

Sean-Holcomb:
@nikovacevic and I just had a good discussion about this and we came to the conclusion that this overhead value will go in shared cost. We considered the options of additional fields in allocation and/or an overhead allocation which could be kept separate or shared across other allocations. However the conclusion that we came to was that the drawbacks of these other approaches was greater than the added complexity to the shared filed. For example adding a new overhead field to allocations would require an additional column in the UI and increase the memory footprint which has already been quickly growing. Adding an Overhead Allocation would continue distort the meaning of what an allocation is, while also probably requiring the overhead field in addition, since the cost would have to go somewhere and it would not fall into any of the other categories (cpu, gpu,  etc).

Sean-Holcomb:
added closed source PR: https://github.com/kubecost/kubecost-cost-model/pull/414

Sean-Holcomb:
Closing this pullrequest because my new PR which branches from it has some critical updates to cost distribution
https://github.com/kubecost/cost-model/pull/807

PR #799, opened 2021-05-13
Add quotes around dataset

AjayTripathy:
If your GCP dataset starts with a number, I believe the query parser does not interpret the dataset as a string. Should fix

```I0512 23:57:24.127986       1 gcpprovider.go:407] Error querying gcp:
googleapi: Error 400: Syntax error: Missing whitespace between literal
and alias at [6:32], invalidQuery
```

repro:
Screen Shot 2021-05-12 at 8 24 02 PM

and after, expected since I made up a dataset:

Screen Shot 2021-05-12 at 10 01 16 PM

### DIFF
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -361,16 +361,17 @@ func (gcp *GCP) ExternalAllocations(start string, end string, aggregators []stri
 		s = append(s, gcpOOC...)
 		qerr = err
 		*/
-		queryString := fmt.Sprintf(`(
+		queryString := `(
 			SELECT
 				service.description as service,
 				TO_JSON_STRING(labels) as keys,
 				SUM(cost) as cost
-			FROM  "%s"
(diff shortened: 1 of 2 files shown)


pkg/cloud/gcpprovider.go
```diff
@@ -366,7 +366,7 @@ func (gcp *GCP) ExternalAllocations(start string, end string, aggregators []stri
 				service.description as service,
 				TO_JSON_STRING(labels) as keys,
 				SUM(cost) as cost
-			FROM  %s
+			FROM  "%s"
```
joshughes, line 369: Based on my tests this needs to be ` back ticks not "

AjayTripathy: That is true. https://stackoverflow.com/questions/51143932/syntax-error-unexpected-string-literal-93868086-ga-sessions-at-1244-big

Thank you for the review!

michaelmdresser (approved):

PR #800, opened 2021-05-13
Ajay tripathy alerts

AjayTripathy:


pkg/costmodel/clusters/clustermap.go
```diff
@@ -183,7 +183,7 @@ func (pcm *PrometheusClusterMap) refreshClusters() {
 	updated, err := pcm.loadClusters()
 	if err != nil {
 		log.Errorf("Failed to load cluster info via query after %d retries", LoadRetries)
-		return
+		updated = make(map[string]*ClusterInfo) // If we fail to query prometheus, register no clusters, as we use this to gauge availability of clusters.
```
mbolt35, line 186: We can't use this in that way. For prometheus, it might make sense, but prometheus here will usually only have 1 entry (the current cluster), but for thanos, I don't think it's wise to use the state of the clusters 3h ago as the current cluster availability.

mbolt35: Additionally, imagine a query fails to thanos. That doesn't really mean that all the other clusters are unavailable.

AjayTripathy: Re: unavailability-- that's the intention. All alerts firing if a query fails to thanos is ok.

The state of clusters 3h ago does make sense to use since we should have separate "prometheus up" alerts on each client. This is saying essentially-- what clusters does thanos have data for as of 3h ago. We can make a separate map and functions for this, but it seems to make sense here.

dwbrown2: Re: unavailability -- should our initial diagnostic alert not catch this? I agree with Bolt that is seems strange to report that all clusters stopped reporting data, when in this case we would no be sure of that...

dwbrown2:
@AjayTripathy can you share the output of this alert in Slack/email?

pkg/costmodel/clusters/clustermap.go
AjayTripathy: There's not really a great way to distinguish those states, unless you assume the existence of a persistent volume and use it to write down the clusters, which you can't. If thanos returns empty or error we're never really sure if its down, returning partial data from unavailable stores, or if all clusters really have gone away.

dwbrown2:
I can live with a duplicate alert if @mbolt35 can... unless there's something else here I'm missing

AjayTripathy:
@dwbrown2  , in slack: 

```
kc  9:07 PM
Clusters added: cluster-one
```

@mbolt35 based on our discussions, I think we've come to the decision that something is better thannothing and these should ship.

pkg/costmodel/clusters/clustermap.go
nikovacevic: This looks kind of suspect to me, too. Is it true that the other PR (https://github.com/kubecost/kubecost-cost-model/pull/410/files) will catch all instances of clusters coming up and going down, even without this? And if this goes in, doesn't that mean that all the other things that ClusterMap is used for will break, until Thanos comes back up?

nikovacevic: If we have a clear-eyed vision of how this is an improvement over the basic version of the PR, and have made peace with the other things that will break as a consequence, then I don't want to obstruct this. But this has the look of a design decision that may yield unexpected results elsewhere, in addition to generating a confusing message in the first place for the cluster alert. (i.e. why send an alert that all the clusters are simultaneously down, when it's paradoxical and easily verifiable that only Thanos is down--and perhaps 3h after the actual event?)

AjayTripathy: > If we have a clear-eyed vision of how this is an improvement over the basic version of the PR

This PR is 
* an error case handler to make sure the alert won't panic
* will refresh the clustermap to be empty in case prometheus isn't contacted/down

Of these two additions, I think the only controversy might be the second point. If there are any usecases where clustermap is expected to have an uptime dependent only on the initialized state, I think we can make a justification to narrow this to just make sure to update the clustermap if its nil?

> why send an alert that all the clusters are simultaneously down ... when only Thanos is down

Thanos can be "up" (pods running with no errors) and still return no data, so you want at least some alert that verifies from a blackbox perspective that clusters are available. the 3h delay is unfortunate but consistent, at least.

nikovacevic: Why would the alert panic without this? That sounds to me like something the alert code needs to handle, not a different function in a different repo, right?

As far as what this does to the behavior of this specific alert, I can be persuaded. Sounds like we're just looking for an imperfect compromise, anyways. What still concerns me is (1) that this behavior is unexpected, and will become something we stumble over, and (2) that this causes other bugs where Thanos is down for a while, then 3h later a feature that depends on the ClusterMap stops working because we blew away a perfectly valid map. Do you share that concern, or do you think that this won't cause problems like that?

AjayTripathy: Ok, let's just take those unintended consequences off the table by creating a second field and using it in the alert.

AjayTripathy: Or, if that feels too messy (I'd need to create a whole parallel system of accesses), I guess investigate all usage of ClusterMap for unintended consequences?

nikovacevic: The unintended consequences of emptying the ClusterMap, off the top of my head, would be breaking a few view things on the FE, as well as some API filtering mechanisms that rely on ClusterMap to let you filter by the name (not just the ID). Not sure what else. I'm not sure what you mean about creating a second field, but certainly open other solutions.

mbolt35: @AjayTripathy This is precisely why I disagree with the use case for cluster map. However, I've conceded on the `3h` delay on cluster up/down, but this change is not really one that I can support:

> * an error case handler to make sure the alert won't panic

It is not the responsibility of the `ClusterMap` to ensure that the alert doesn't panic. 

> * will refresh the clustermap to be empty in case prometheus isn't contacted/down

We use the `ClusterMap` in multiple places on the backend and the frontend (which accesses cluster ID/Name mappings via endpoint). This proposed "clearing" of the map on an error completely disrupts all other use-cases and hijacks the design intent. It's not the role of the `ClusterMap` to meet the requirements of the alert, it's the meet the design requirements of providing a cluster-id/name label based on the most recent successful state. 

If you want the alert to issue "all clusters are down" when thanos returns no data, or when thanos is actually down, then the alert should implement that functionality. This approach is actually _better_ in that in the event thanos is unreachable, you could still use the cluster map to get the last known list of cluster ids to dispatch in the alert.

AjayTripathy: Ok, I'll close this. It's a little annoying to have to replicate this functionality and duplicate a query to thanos in the alert but I think I can live with that.

PR #801, opened 2021-05-13
Enable a custom cluster ID label for by clauses

michaelmdresser:
Based on: https://github.com/kubecost/cost-model/pull/782, forked from it to handle merge conflicts.

Allows a user to specify an alternative cluster ID label with the environment variable `PROM_CLUSTER_ID_LABEL` and updates all PromQL `by` clauses to use that label.

### Testing

`go vet -printf ./...` doesn't have any errors.

Tested by deploying without setting the `PROM_CLUSTER_ID_LABEL` var. Spot check of Cost Allocation and Savings suggests everything is running normally.

Updated the deployment to set `PROM_CLUSTER_ID_LABEL=cluster_id`. Spot check of Cost Allocation and Savings also looked good.


nikovacevic (approved):
LGTM

PR #802, opened 2021-05-14
Sean/idle by node

Sean-Holcomb:
Closed Source PR: https://github.com/kubecost/kubecost-cost-model/pull/408

This PR adds a query time parameter to the Allocations which allows for Idle allocations to be calculated at the node level rather than at the cluster level, with the intent that when Idle values are aggregated, filtered or shared, the result will be equivalent and when idle is split it is shown on a per node basis rather that a per cluster basis. 

Testing:
Unit tests have been added for creation of idle allocation as nodes along with tests for the various transformation that can take place on allocation. Additionally this PR includes a refactor of the aggregateBy test so that it now uses the subtests rather than being a monolithic test. The purpose of this is for greater visibility into failing test cases, and improved organization.

Azure Cluster:
SplitIdle=true, IdleByNode=true
Screen Shot 2021-05-14 at 12 10 47 PM
SplitIdle=false, IdleByNode=false
Screen Shot 2021-05-14 at 12 16 04 PM
One thing of concern here is the presence of a negative idle allocation CPU cost in the node breakdown. Whether or not there is such a thing as a negative idle cost is up for question, but the sum of the node idle is equal to the total of the unsplit idle which suggests a problem with how idle is being calculated in general.
GCP Cluster:
SplitIdle=true, IdleByNode=true
Screen Shot 2021-05-14 at 11 44 44 AM
AWS Cluster:
SplitIdle=true, IdleByNode=true
Screen Shot 2021-05-14 at 2 50 51 PM
SplitIdle=false, IdleByNode=false
Screen Shot 2021-05-14 at 3 00 44 PM

This PR also includes a slackening on the requirement that all allocations must be assignable to a cluster or node and if they are not gives a warning and skips that allocation when generating the node. This is in part to the anomalous allocation with no name in the last photo which has a cluster but no provider Id and was causing allocation calls to fail entirely. It is my belief that a warning is sufficient in cases such as these because It is unlikely that an issue like this will be a large part of the cluster costs. It is also unclear if this is an issue with another area of how allocations are calculated. Here is a more detailed view of the allocation in question.
Screen Shot 2021-05-14 at 3 00 32 PM

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -930,8 +930,8 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	for _, alloc := range as.allocations {
 		idleKey, err := alloc.getIdleKey(options)
 		if err != nil {
-			log.Warningf("AllocationSet.AggregateBy: missing idleKey for allocation: %s", alloc.Name)
-			return err
+			log.DedupedInfof(5,"AllocationSet.AggregateBy: missing idleKey for allocation: %s", alloc.Name)
+			continue
 		}
 
 		skip := false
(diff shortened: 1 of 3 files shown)


AjayTripathy (approved):
Small change but other than that LGTM

PR #810, opened 2021-05-24
Allow unmounted resources past idle filters

Sean-Holcomb:
Previously resources without idleKeys (Cluster or Provider ID) would derail the idle allocations process, this was not a problem when Cluster was the only possible key however with with Provider ID, allocations may be missing this value. Rather than ending the process these instances now trigger logs  and let the process continue. The only case of this happening so far is Unmounted PVs which do not contribute to any computational resources (CPU, GPU or RAM). However if it were to take place on a resource with computational resources, this would be indicative of an issue with Provider ID being distributed correctly.

I have added logging in the places where it makes sense to draw attention to these resources without being too repetitive and given the unmounted idleKey to items missing their own idleKey

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -965,7 +965,11 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 			for _, idleAlloc := range idleSet.allocations {
 				// Only share idle if the idleKey matches; i.e. the allocation
 				// is from the same idleKey as the idle costs
-				iaIdleKey, _ := idleAlloc.getIdleKey(options)
+				iaIdleKey, err := idleAlloc.getIdleKey(options)
+				if err != nil {
+					log.Errorf("AllocationSet.AggregateBy: Idle allocation is missing idlekey %s", idleAlloc.Name)
+					return err
+				}
 
 				if iaIdleKey != idleKey {
 					continue
@@ -1082,7 +1086,11 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	if len(aggSet.idleKeys) > 0 && groupingIdleFiltrationCoeffs != nil {
 		for idleKey := range aggSet.idleKeys {
 			idleAlloc := aggSet.Get(idleKey)
-			iaIdleKey, _ := idleAlloc.getIdleKey(options)
+			iaIdleKey, err := idleAlloc.getIdleKey(options)
+			if err != nil {
+				log.Errorf("AllocationSet.AggregateBy: Idle allocation is missing idlekey %s", idleAlloc.Name)
+				return err
+			}
 
 			if resourceCoeffs, ok := groupingIdleFiltrationCoeffs[iaIdleKey]; ok {
 				idleAlloc.CPUCost *= resourceCoeffs["cpu"]
@@ -1228,7 +1236,6 @@ func computeIdleCoeffs(options *AllocationAggregationOptions, as *AllocationSet,
 
 		idleKey, err := alloc.getIdleKey(options)
 		if err != nil {
-			// Use Unallocated for allocations without idle key
 			log.DedupedWarningf(3, "Missing Idle Key for %s", alloc.Name)
 		}
 
@@ -1276,7 +1283,6 @@ func computeIdleCoeffs(options *AllocationAggregationOptions, as *AllocationSet,
 		// idleKey will be providerId or cluster
 		idleKey, err := alloc.getIdleKey(options)
 		if err != nil {
-			// Use Unallocated for allocations without idle key
 			log.DedupedWarningf(3, "Missing Idle Key in share set for %s", alloc.Name)
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/kubecost/allocation.go
```diff
@@ -1018,18 +1015,14 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		for _, alloc := range shareSet.allocations {
 			idleKey, err := alloc.getIdleKey(options)
 			if err != nil {
-				log.DedupedWarningf(5, "AllocationSet.AggregateBy: missing idleKey for allocation: %s", alloc.Name)
-				continue
+				log.DedupedWarningf(3, "AllocationSet.AggregateBy: missing idleKey for allocation: %s", alloc.Name)
```
nikovacevic, line 1018: So, if we aren't continuing any longer, then what happens? How does this affect idle by cluster?

pkg/kubecost/allocation.go
```diff
@@ -966,10 +965,8 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 			for _, idleAlloc := range idleSet.allocations {
 				// Only share idle if the idleKey matches; i.e. the allocation
 				// is from the same idleKey as the idle costs
-				iaIdleKey, err := idleAlloc.getIdleKey(options)
-				if err != nil {
-					return err
-				}
+				iaIdleKey, _ := idleAlloc.getIdleKey(options)
```
nikovacevic, line 968: Should we only swallow this error if we're doing idle by node? Are we sure that we want to alter the old behavior to support the new behavior?

pkg/kubecost/allocation.go
```diff
@@ -1347,13 +1336,13 @@ func (a *Allocation) getIdleKey(options *AllocationAggregationOptions) (string,
 		// Key allocations to ProviderId to match against node
 		idleKey = a.Properties.ProviderID
 		if idleKey == "" {
-			return idleKey, fmt.Errorf("ProviderId is not set")
+			return UnmountedSuffix, fmt.Errorf("ProviderId is not set")
```
nikovacevic, line 1339: Do we know that this will always be the case? Isn't it possible (e.g. in an error state) to be missing a provider ID and not because the allocation is unmounted?

pkg/kubecost/allocation.go
```diff
@@ -1347,13 +1336,13 @@ func (a *Allocation) getIdleKey(options *AllocationAggregationOptions) (string,
...
 		}
 	} else {
 		// key the allocations by cluster id
 		idleKey = a.Properties.Cluster
 		if idleKey == "" {
-			return idleKey, fmt.Errorf("ClusterProp is not set")
+			return UnmountedSuffix, fmt.Errorf("ClusterProp is not set")
```
nikovacevic, line 1345: Why is it the case that we should return "unmounted" and an error in all cases where an allocation is without a cluster ID?

nikovacevic (commented):
@AjayTripathy may feel differently, but I have some questions about this before we merge, even if it fixes some short-term trouble.

pkg/kubecost/allocation.go
nikovacevic: Another way of phrasing this is: how can we be certain that everything missing a ProviderID (or a Cluster in the other case) is, in fact, an Unmounted PV?

pkg/kubecost/allocation.go
Sean-Holcomb: I am ignoring these because they imply a node without a cluster or a node without a provider ID which from what I can tell does not exist

pkg/kubecost/allocation.go
Sean-Holcomb: So all of these continues actually came from the last PR on this [802](https://github.com/kubecost/cost-model/pull/802/files) and before they actually returned. The behavior from cluster id that you are talking about preserving is a blank allocations screen with an error in the network tab. I don't think that was the right behavior to begin with but it also stood in the way of this feature so I changed it.

pkg/kubecost/allocation.go
Sean-Holcomb: I am not married to using the unmountedSuffix here, I could just as well switch it back to returning "", it just seemed more accurate given the the only instance of missing idleKey I have seen

AjayTripathy:
> @AjayTripathy may feel differently, but I have some questions about this before we merge, even if it fixes some short-term trouble.

No need to rush this. I'll live with unmounted disappearing from this view if we want to take the time to get this right.

pkg/kubecost/allocation.go
nikovacevic: Hm. I'm not sure I agree that returning an error and failing the query is wrong here. (If this implies a node without a cluster, and that shouldn't be possible, then I think the only correct behavior upon encountering an error here is to return it.) But I don't want to obstruct. Can we at least log the error before moving on?

pkg/kubecost/allocation.go
nikovacevic: Okay, thanks for the context.

pkg/kubecost/allocation.go
nikovacevic: If you don't mind, then let's switch back to `""`. Just because at this moment that is the case, doesn't mean it will always be the case, and I don't see another reason to believe that "no provider ID" should always imply "unmounted PV".

pkg/kubecost/allocation.go
nikovacevic: Same as above.

nikovacevic (approved):
I don't want to obstruct getting this fixed, so approving with the hope that we can at least remove the "unmounted" bit.

pkg/kubecost/allocation.go
Sean-Holcomb: Switched back now

pkg/kubecost/allocation.go
Sean-Holcomb: added back in since it will not interfere with the feature

PR #819, opened 2021-06-01
better logging in query. mirrors queryrange logging but adds req.URL

AjayTripathy:
As we support more promql providers, we should harden some of this error code. This seems like an oversight as this was in the queryrange logic but not in query logic and we migrated a bunch of things away from queryrange.

### DIFF
--- a/pkg/prom/query.go
+++ b/pkg/prom/query.go
@@ -186,6 +186,12 @@ func (ctx *Context) query(query string) (interface{}, prometheus.Warnings, error
 
 		return nil, warnings, fmt.Errorf("query error %d: '%s' fetching query '%s'", resp.StatusCode, err.Error(), query)
 	}
+	// Unsuccessful Status Code, log body and status
+	statusCode := resp.StatusCode
+	statusText := http.StatusText(statusCode)
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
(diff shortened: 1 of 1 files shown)


michaelmdresser:
While we're looking at this kind of thing, I've seen some very long error messages (e.g. a failure from Chronosphere) get elided so we miss important information about the error. https://github.com/kubecost/cost-model/blob/develop/pkg/costmodel/allocation.go#L261-L267 it may be a problem with the error collection idea.

nikovacevic (approved):

michaelmdresser:
> @michaelmdresser curious, what do you think is happening to cause those error messages to get elided?

I'm not totally sure. It could be that `context` limits the stored errors or (more likely) that whatever library we're using to call Prometheus is doing it. The exact error message I saw get elided one time looked like an HTTP error page (this was while debugging Chronosphere IIRC) and I believe @AjayTripathy was there too. It looked something like `Failed to query <somehtmltag>...rest_of_html`; the ellipses were part of the error message that was logged and there was obviously missing content that I assumed was within them.

Also, here's a better link to the code (permalink) so we don't lose what I'm talking about: https://github.com/kubecost/cost-model/blob/b53060b0bd513e7b57fdab787e40239da014b1e8/pkg/costmodel/allocation.go#L275-L281

AjayTripathy:
lets merge and revisit other elided error, think @mbolt35 is working on something there

PR #820, opened 2021-06-02
Replaced old KSM request metrics with the new metric

michaelmdresser:
Replaced `kube_pod_container_resource_requests_cpu_cores`
and `kube_pod_container_resource_requests_memory_bytes`
with `kube_pod_container_resource_requests{resource="cpu", unit="core"}`
and `kube_pod_container_resource_requests{resource="memory", unit="byte"}` respectively.

Necessary because in KSM v2.0+, the old metrics have been removed. This particular migration is easy because we whitelisted `kube_pod_container_resource_requests` many months ago, at the same time we whitelisted the old metrics being replaced in this commit.

Achieved with:

```
fd --type f | xargs sed -i 's/kube_pod_container_resource_requests_memory_bytes{/kube_pod_container_resource_requests{resource="memory", unit="byte", /g'

fd --type f | xargs sed -i 's/kube_pod_container_resource_requests_cpu_cores{/kube_pod_container_resource_requests{resource="cpu", unit="core", /g'
```

And a small amount of minor manual modification.

## Testing
See closed-source PR.


mbolt35 (approved):
These look good to me. Pretty reasonable change. Approving here, but I am assuming we have some in the helm chart as well?

PR #821, opened 2021-06-03
Move mock generation to mock.go and clean up old code

Sean-Holcomb:
Used by: https://github.com/kubecost/kubecost-cost-model/pull/430
This PR moves mock Asset and Allocation generation code used in unit tests to a separate file so that it can be exported and used in closed source. Additionally it removes functionality that is being moved to closed source.
Testing: 
Unit tests pass after refactor


pkg/kubecost/mock.go
```diff
@@ -0,0 +1,667 @@
...
+import (
+	"fmt"
+	"time"
+)
+const gb = 1024 * 1024 * 1024
+const day = 24 * time.Hour
+var disk = PVKey{}
+
+// NewUnitAllocation creates an *Allocation with all of its float64 values set to 1 and generic properties if not provided in arg
+func NewUnitAllocation(name string, start time.Time, resolution time.Duration, props *AllocationProperties) *Allocation {
```
nikovacevic, line 12: @Sean-Holcomb how do you feel about this name, now that it's exported from the package? Do you think it's clear enough that it's a test function, or do you think we should try to add the word "Mock" or "Test" in there to make it clear?

I'm imagining someone importing and using the `kubecost` package finding `kubecost.NewUnitAllocation(...)` and not understanding that it's a test-only function.

nikovacevic (approved):
Generally, I think these are great changes. I only worry about naming of these exported functions being confused for non-test functions.

pkg/kubecost/mock.go
Sean-Holcomb: I think that is a good point, I will go over them

PR #823, opened 2021-06-07
emit pv id even if there is not yet cost data

AjayTripathy:


michaelmdresser (approved):
Looks good. What was the bug that necessitated this?

Sean-Holcomb (approved):
Seems good

AjayTripathy:
> Looks good. What was the bug that necessitated this?

There are certain PV classes (CSI providers) for which we don't pull pricing data from AWS but instead rely on reconciliation to update pricing.

PR #830, opened 2021-06-10
Emit kube_pod_labels with all label names

michaelmdresser:
In KSM v2, kube_pod_labels was restricted to only emit labels
if they are explicitly whitelisted. See
https://github.com/kubernetes/kube-state-metrics/issues/1270#issuecomment-712986441

We need labels so we can match pods with services, among other things.

This commit enables cost-model to emit kube_pod_labels as if it were
pre-KSM v2. This should not cause problems if KSM v1 is deployed because
we are just emitting the same data (which will have a different job label)
and it isn't used for things like sum/rate/etc. queries.

[Design doc](https://docs.google.com/document/d/1Mr0JpZ4PTjrOMQtOMgTo2PeveB1s2Ukd646uP3tJ3Cw/edit#bookmark=id.gh1ogpwg2kwt) (internal)

## Testing

I diffed the `kube_pod_labels` output of our `/metrics` page with the `kube_pod_labels` from a KSMv1 installed on the same cluster and got identical results, except our metric also has the (new in KSMv2) `uid` label.

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -1,7 +1,6 @@
 package costmodel
 
 import (
-	"fmt"
 	"math"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


AjayTripathy (commented):
Pretty sure you need to sanitize label names, if that is not being done implicitly somewhere I missed.

https://github.com/kubecost/cost-model/blob/d14596125f0a5200dbcbfbe2cfc970ee1c86ad93/pkg/prom/metrics.go#L104

AjayTripathy (approved):

PR #833, opened 2021-06-14
Switch KSM v1 metric emission to be one env var

michaelmdresser:
Simplifies configuration. A valid reduction of config, too, because
there should not be a situation where we want to emit only some of
the removed KSM v1 metrics.

Ref: https://github.com/kubecost/cost-analyzer-helm-chart/pull/930#discussion_r649569856

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -1068,25 +1068,16 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud
 			})
 		}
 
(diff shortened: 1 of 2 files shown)


AjayTripathy:
LGTM but can you quickly share how this was tested?

AjayTripathy (approved):

nikovacevic (approved):
Dig it

michaelmdresser:
Tested by deploying to one of my test clusters and checking that our `/metrics` contained the `kube_pod_labels`, `kube_node_labels`, `kube_node_status_capacity_cpu_cores`, and `kube_node_status_memory_bytes` metrics.

PR #840, opened 2021-06-25
Remove Prometheus duration prefix, remove duplicate method

Sean-Holcomb:
- Refactored time.go so that I could delete duplicates of methods that it contained and moved from package util to timeutil
- Added tests for ParseDuration
- Create method to remove Prometheus "offset " prefix from stings and added to some methods
- Replace string parameters with duration parameter

CS PR: https://github.com/kubecost/kubecost-cost-model/pull/458

Testing: 
Tested on GCP cluster, removing error messages

```
Error returned to client: error parsing offset (offset 3h): error parsing duration: offset 3h did not match expected format [0-9+](s|m|d|h)
```

and

```
Error] Asset ETL: ComputeAssets: clusterManagementQuery: Prometheus communication error: 400 (Bad Request) URL: 'http://kubecost-thanos-query-frontend-http.kubecost:10902/api/v1/query?max_source_resolution=5m&query=sum_over_time%28%28avg%28kubecost_cluster_management_cost%29+by+%28cluster_id%29%29%5B0m%3A1m%5D+offset+180m%29+%2A+0.016667' Headers: '{ Access-Control-Allow-Origin: [ * ], Content-Length: [ 103 ], Content-Type: [ application/json ], Date: [ Fri, 25 Jun 2021 17:00:29 GMT ], Vary: [ Accept-Encoding ], Access-Control-Allow-Headers: [ Accept, Accept-Encoding, Authorization, Content-Type, Origin ], Access-Control-Allow-Methods: [ GET, OPTIONS ], Access-Control-Expose-Headers: [ Date ], Cache-Control: [ no-store ] }', Body: '{"status":"error","errorType":"bad_data","error":"1:71: parse error: duration must be greater than 0"}
```
 and confirmed that affected endpoint clusterCostsOverTime and clusterCoster handle the offset and window parameters in the same way as they did before

### DIFF
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -135,7 +135,7 @@ func (gcp *GCP) GetLocalStorageQuery(window, offset time.Duration, rate bool, us
 		baseMetric = "container_fs_usage_bytes"
 	}
 
-	fmtOffset := timeutil.DurationToPromString(offset)
+	fmtOffset := timeutil.DurationToPromOffsetString(offset)
 
 	fmtCumulativeQuery := `sum(
 		sum_over_time(%s{device!="tmpfs", id="/"}[%s:1m]%s)
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -3,6 +3,7 @@ package cloud
 import (
 	"context"
 	"fmt"
+	"github.com/kubecost/cost-model/pkg/util/timeutil"
 	"io"
 	"io/ioutil"
 	"math"
@@ -124,7 +125,7 @@ func gcpAllocationToOutOfClusterAllocation(gcpAlloc gcpAllocation) *OutOfCluster
 
 // GetLocalStorageQuery returns the cost of local storage for the given window. Setting rate=true
 // returns hourly spend. Setting used=true only tracks used storage, not total.
-func (gcp *GCP) GetLocalStorageQuery(window, offset string, rate bool, used bool) string {
+func (gcp *GCP) GetLocalStorageQuery(window, offset time.Duration, rate bool, used bool) string {
 	// TODO Set to the price for the appropriate storage class. It's not trivial to determine the local storage disk type
 	// See https://cloud.google.com/compute/disks-image-pricing#persistentdisk
 	localStorageCost := 0.04
@@ -134,10 +135,7 @@ func (gcp *GCP) GetLocalStorageQuery(window, offset string, rate bool, used bool
 		baseMetric = "container_fs_usage_bytes"
 	}
 
-	fmtOffset := ""
-	if offset != "" {
-		fmtOffset = fmt.Sprintf("offset %s", offset)
-	}
+	fmtOffset := timeutil.DurationToPromString(offset)
 
 	fmtCumulativeQuery := `sum(
 		sum_over_time(%s{device!="tmpfs", id="/"}[%s:1m]%s)
@@ -151,8 +149,9 @@ func (gcp *GCP) GetLocalStorageQuery(window, offset string, rate bool, used bool
 	if rate {
 		fmtQuery = fmtMonthlyQuery
 	}
+	fmtWindow := timeutil.DurationString(window)
 
-	return fmt.Sprintf(fmtQuery, baseMetric, window, fmtOffset, env.GetPromClusterLabel(), localStorageCost)
+	return fmt.Sprintf(fmtQuery, baseMetric, fmtWindow, fmtOffset, env.GetPromClusterLabel(), localStorageCost)
 }
 
 func (gcp *GCP) GetConfig() (*CustomPricing, error) {
--- a/pkg/costmodel/aggregation.go
+++ b/pkg/costmodel/aggregation.go
@@ -196,7 +196,7 @@ func GetTotalContainerCost(costData map[string]*CostData, rate string, cp cloud.
 	return totalContainerCost
 }
 
-func (a *Accesses) ComputeIdleCoefficient(costData map[string]*CostData, cli prometheusClient.Client, cp cloud.Provider, discount float64, customDiscount float64, windowString, offset string) (map[string]float64, error) {
+func (a *Accesses) ComputeIdleCoefficient(costData map[string]*CostData, cli prometheusClient.Client, cp cloud.Provider, discount float64, customDiscount float64, window, offset time.Duration) (map[string]float64, error) {
 	coefficients := make(map[string]float64)
 
 	profileName := "ComputeIdleCoefficient: ComputeClusterCosts"
@@ -204,12 +204,12 @@ func (a *Accesses) ComputeIdleCoefficient(costData map[string]*CostData, cli pro
 
 	var clusterCosts map[string]*ClusterCosts
 	var err error
-
-	key := fmt.Sprintf("%s:%s", windowString, offset)
+	fmtWindow, fmtOffset := timeutil.DurationOffsetStrings(window, offset)
+	key := fmt.Sprintf("%s:%s", fmtWindow, fmtOffset)
 	if data, valid := a.ClusterCostsCache.Get(key); valid {
 		clusterCosts = data.(map[string]*ClusterCosts)
 	} else {
-		clusterCosts, err = a.ComputeClusterCosts(cli, cp, windowString, offset, false)
+		clusterCosts, err = a.ComputeClusterCosts(cli, cp, window, offset, false)
 		if err != nil {
 			return nil, err
 		}
@@ -225,7 +225,7 @@ func (a *Accesses) ComputeIdleCoefficient(costData map[string]*CostData, cli pro
 		}
 
 		if costs.TotalCumulative == 0 {
-			return nil, fmt.Errorf("TotalCumulative cluster cost for cluster '%s' returned 0 over window '%s' offset '%s'", cid, windowString, offset)
+			return nil, fmt.Errorf("TotalCumulative cluster cost for cluster '%s' returned 0 over window '%s' offset '%s'", cid, fmtWindow, fmtOffset)
 		}
 
 		totalContainerCost := 0.0
@@ -1492,11 +1492,9 @@ func (a *Accesses) ComputeAggregateCostModel(promClient prometheusClient.Client,
 			}
 		}
 
-		// Convert to Prometheus-compatible strings
-		durStr, offStr := timeutil.DurationOffsetStrings(dur, off)
-
-		idleCoefficients, err = a.ComputeIdleCoefficient(costData, promClient, a.CloudProvider, discount, customDiscount, durStr, offStr)
+		idleCoefficients, err = a.ComputeIdleCoefficient(costData, promClient, a.CloudProvider, discount, customDiscount, dur, off)
 		if err != nil {
+			durStr, offStr := timeutil.DurationOffsetStrings(dur, off)
 			log.Errorf("ComputeAggregateCostModel: error computing idle coefficient: duration=%s, offset=%s, err=%s", durStr, offStr, err)
 			return nil, "", err
 		}
@@ -1744,10 +1742,16 @@ func (a *Accesses) warmAggregateCostModelCache() {
 	// for the given duration. Cache is intentionally set to expire (i.e. noExpireCache=false) so that
 	// if the default parameters change, the old cached defaults with eventually expire. Thus, the
 	// timing of the cache expiry/refresh is the only mechanism ensuring 100% cache warmth.
-	warmFunc := func(duration, durationHrs, offset string, cacheEfficiencyData bool) (error, error) {
+	warmFunc := func(duration, offset time.Duration, cacheEfficiencyData bool) (error, error) {
+		if a.ThanosClient != nil {
+			duration = thanos.OffsetDuration()
+			log.Infof("Setting Offset to %s", duration)
+		}
+		fmtDuration, fmtOffset := timeutil.DurationOffsetStrings(duration, offset)
+		durationHrs, err := timeutil.DayDurationToHourDuration(fmtDuration)
 		promClient := a.GetPrometheusClient(true)
 
-		windowStr := fmt.Sprintf("%s offset %s", duration, offset)
+		windowStr := fmt.Sprintf("%s fmtOffset %s", fmtDuration, fmtOffset)
 		window, err := kubecost.ParseWindowUTC(windowStr)
 		if err != nil {
 			return nil, fmt.Errorf("invalid window from window string: %s", windowStr)
@@ -1778,17 +1782,15 @@ func (a *Accesses) warmAggregateCostModelCache() {
 
 		aggKey := GenerateAggKey(window, field, subfields, aggOpts)
 		log.Infof("aggregation: cache warming defaults: %s", aggKey)
-		key := fmt.Sprintf("%s:%s", durationHrs, offset)
+		key := fmt.Sprintf("%s:%s", durationHrs, fmtOffset)
 
 		_, _, aggErr := a.ComputeAggregateCostModel(promClient, window, field, subfields, aggOpts)
 		if aggErr != nil {
 			log.Infof("Error building cache %s: %s", window, aggErr)
 		}
-		if a.ThanosClient != nil {
-			offset = thanos.Offset()
-			log.Infof("Setting offset to %s", offset)
-		}
-		totals, err := a.ComputeClusterCosts(promClient, a.CloudProvider, durationHrs, offset, cacheEfficiencyData)
+
+
+		totals, err := a.ComputeClusterCosts(promClient, a.CloudProvider, duration, offset, cacheEfficiencyData)
 		if err != nil {
 			log.Infof("Error building cluster costs cache %s", key)
 		}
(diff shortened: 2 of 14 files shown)


pkg/util/mapper/mapper.go
```diff
@@ -530,6 +530,8 @@ func durationString(duration time.Duration) string {
 }
 
 func parseDuration(duration string) (time.Duration, error) {
+	// Trim prefix of Prometheus format duration
+	duration = strings.TrimPrefix(duration, "offset ")
```
michaelmdresser, line 534: Should leading whitespace be trimmed before trimming the prefix? I have no idea, not sure how this is used.

pkg/util/time.go
```diff
@@ -66,6 +67,8 @@ func DurationOffsetStrings(duration, offset time.Duration) (string, string) {
 
 // ParseDuration converts a Prometheus-style duration string into a Duration
 func ParseDuration(duration string) (*time.Duration, error) {
+	// Trim prefix of Prometheus format duration
+	duration = strings.TrimPrefix(duration, "offset ")
```
michaelmdresser, line 71: Likewise on trimming whitespace before.

michaelmdresser (approved):
A couple of small questions. Did you test to make sure cluster-sizing is working?

nikovacevic:
Echoing @michaelmdresser, it'd be good to have an idea of how this was tested. (IMO this is a great candidate for unit testing! Shame on me for not writing them in the first place for this function.) Otherwise, it all looks good to me.

One question: did you consider adding these checks to `ParseTimeRange(duration, offset string)` instead? If so, what are the tradeoffs you considered? The reason I ask is that in that function, we know that one of the two represents an offset, so we can be forgiving if someone passes "offset 2d" instead of "2d". In `ParseDuration(duration string)`, it's a bit more strange IMO to check for the word "offset", as it's a much more general function. Curious to hear your thoughts, as the author of the fix!

AjayTripathy:
Let's make sure this is tested, merge this to master, and patch please! Thanks all.

pkg/costmodel/cluster.go
```diff
@@ -71,7 +71,8 @@ type ClusterCostsBreakdown struct {
 // NewClusterCostsFromCumulative takes cumulative cost data over a given time range, computes
 // the associated monthly rate data, and returns the Costs.
 func NewClusterCostsFromCumulative(cpu, gpu, ram, storage float64, window, offset string, dataHours float64) (*ClusterCosts, error) {
-	start, end, err := util.ParseTimeRange(window, offset)
+	offset = timeutil.CleanDurationString(offset)
+	start, end, err := timeutil.ParseTimeRange(window, offset)
```
Sean-Holcomb, line 75: Cleaning offset here because it is used later in this method

pkg/costmodel/cluster.go
```diff
@@ -702,7 +703,8 @@ func ClusterLoadBalancers(cp cloud.Provider, client prometheus.Client, duration,
 // ComputeClusterCosts gives the cumulative and monthly-rate cluster costs over a window of time for all clusters.
 func (a *Accesses) ComputeClusterCosts(client prometheus.Client, provider cloud.Provider, window, offset string, withBreakdown bool) (map[string]*ClusterCosts, error) {
 	// Compute number of minutes in the full interval, for use interpolating missed scrapes or scaling missing data
-	start, end, err := util.ParseTimeRange(window, offset)
+	offset = timeutil.CleanDurationString(offset)
+	start, end, err := timeutil.ParseTimeRange(window, offset)
```
Sean-Holcomb, line 707: Cleaning offset here because it is used later in this method

pkg/util/timeutil/timeutil.go
```diff
@@ -223,3 +150,92 @@ func (jt *JobTicker) TickIn(d time.Duration) {
...
+		return nil, fmt.Errorf("error parsing duration: %s did not match expected format [0-9+](s|m|d|h)", duration)
+	}
+
+	dur := time.Duration(amount) * unit
+	return &dur, nil
+}
+
+// CleanDurationString removes prometheus formatted prefix "offset " allong with leading a trailing whitespace
+// from duration string, leaving behind a string with format [0-9+](s|m|d|h)
+func CleanDurationString(duration string) string {
```
Sean-Holcomb, line 190: This is the main new method

pkg/costmodel/cluster.go
```diff
@@ -702,7 +703,8 @@ func ClusterLoadBalancers(cp cloud.Provider, client prometheus.Client, duration,
 // ComputeClusterCosts gives the cumulative and monthly-rate cluster costs over a window of time for all clusters.
 func (a *Accesses) ComputeClusterCosts(client prometheus.Client, provider cloud.Provider, window, offset string, withBreakdown bool) (map[string]*ClusterCosts, error) {
```
mbolt35, line 704: Looping in @nikovacevic here because I know he normalize a lot of the window/offset input, but should we aim to accept a `time.Duration` in the non-http handler functions? With the motivation to remove the awkward inclusion of a `timeutil.CleanDurationString` call? The http handlers can use `queryParams.GetDuration(...)` (which should include CleanDurationString internally), and the rest of the APIs can just reliably accept a `time.Duration`. Appreciate thoughts from @michaelmdresser and @Sean-Holcomb as well.

Sean-Holcomb: I'm onboard with that idea. I already felt like I was getting out of scope for this bug fix with the package change, but that sounds like the right approach to something I felt was a little awkward too. Just sounds like some more hack and slash clean up.

nikovacevic: Agree that `timeutil.CleanDurationString` is awkward, and requires every engineer both to know and to remember to call that function all the time. Ideally this kind of thing would be patched in one place and "just work." (For instance, in a previous review of this PR, I asked why this logic couldn't simply be added to the beginning of `ParseTimeRange` for only the `offset string` parameter, and never got a response, but the question stands.)

I'm 100% in favor of @mbolt35's recommendation of only allowing `window` and `offset` to be passed as strings where required by the circumstance (e.g. HTTP parameters) and to be of type `time.Duration` everywhere else. That said, I also don't want to blow up scope.

@Sean-Holcomb in your judgment, is this a pervasive and serious enough issue to warrant a deep clean up effort right now? Or can we fix the originally-scoped bug with a cleaner, simpler implementation and leave deep-cleaning refactors of duration and offset strings for a future clean-up task?

mbolt35: I think it's worth the time if you're already in there - This already makes a big difference IMO - the duplication of this method in a few places was starting to become unwieldy.

Sean-Holcomb: @nikovacevic sorry for not responding to that comment, but you should know that it did lead me to delete some more duplicate code. Moving it ParseTimeRange method doesn't really change the need for the awkward placements at higher up in the call chain, because the string itself is causing problems when it is used for Prometheus queries it becomes "offset offset 3h". Both of this methods return a time that is used in a different way

mbolt35:
Thing change is long overdue, thank you @Sean-Holcomb for taking it on. Great work! 

Generally, do we feel like we've covered the main use-cases in our tests? I agree with @nikovacevic that this is a perfect example of functionality we should have have good unit test coverage for.

pkg/costmodel/cluster.go
michaelmdresser: Also strongly in favor of parsing strings into useful types in advance so we don't have to later. I'll defer to @Sean-Holcomb to determine the scoping of immediately changing this across the codebase vs. delaying.

pkg/util/timeutil/timeutil.go
```diff
@@ -1,8 +1,9 @@
-package util
+package timeutil
```
michaelmdresser, line 1: @nikovacevic @mbolt35 I don't really want to block on this, but this is once again an instance of package naming being really annoying. I think it would be awesome to name this package `time` and then use `util.time.Foo` but darn Go would make us `import timeutil "github.com/kubecost/cost-model/pkg/util/time"` instead pretty much everywhere. What do you think of, in general, this kind of package naming to avoid overloading stdlib stuff vs. aliasing imports?

mbolt35: I am fine either way -- in general, I suggested the `timeutil` package because go conventions use `http` as an example. They say "don't name your package `util` and don't name it `http` (stdlib collision), but name it `httputil`" -- I, of course, agree with you, but I am just defaulting to their suggestions in the event that it somehow makes life easier. I don't know though :(

mbolt35: https://blog.golang.org/package-names

"Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them."

and ...

"Avoid unnecessary package name collisions. While packages in different directories may have the same name, packages that are frequently used together should have distinct names. This reduces confusion and the need for local renaming in client code. For the same reason, avoid using the same name as popular standard packages like io or http."

nikovacevic: I believe @mbolt35 found a Go-related resource that espoused the idea of naming packages, e.g. `httputil` instead of `util`. Which is an idea I like, for the sake of specificity. So I actually think the right move is to name the package `timeutil`, and to refactor the rest of `pkg/util` into `pkg/util/xyzutil`. What do you think about that idea?

pkg/cloud/gcpprovider.go
```diff
@@ -134,10 +135,7 @@ func (gcp *GCP) GetLocalStorageQuery(window, offset string, rate bool, used bool
 		baseMetric = "container_fs_usage_bytes"
 	}
 
-	fmtOffset := ""
-	if offset != "" {
-		fmtOffset = fmt.Sprintf("offset %s", offset)
-	}
+	fmtOffset := timeutil.DurationToPromString(offset)
```
nikovacevic, line 138: Why does this use `timeutil.DurationToPromString`, but for `window` below we use `timeutil.DurationString`?

pkg/costmodel/aggregation.go
```diff
@@ -1743,10 +1742,16 @@ func (a *Accesses) warmAggregateCostModelCache() {
...
+		if a.ThanosClient != nil {
+			duration = thanos.OffsetDuration()
+			log.Infof("Setting Offset to %s", duration)
+		}
+		fmtDuration, fmtOffset := timeutil.DurationOffsetStrings(duration, offset)
+		durationHrs, err := timeutil.DayDurationToHourDuration(fmtDuration)
 		promClient := a.GetPrometheusClient(true)
 
-		windowStr := fmt.Sprintf("%s offset %s", duration, offset)
+		windowStr := fmt.Sprintf("%s fmtOffset %s", fmtDuration, fmtOffset)
```
nikovacevic, line 1754: This looks suspicious.

nikovacevic: Was this particular path tested? Seems difficult, perhaps, but also we should be extremely careful with sweeping changes like the ones in this PR because if little errors like this one.

pkg/costmodel/aggregation.go
```diff
@@ -1743,10 +1742,16 @@ func (a *Accesses) warmAggregateCostModelCache() {
...
 	// if the default parameters change, the old cached defaults with eventually expire. Thus, the
 	// timing of the cache expiry/refresh is the only mechanism ensuring 100% cache warmth.
-	warmFunc := func(duration, durationHrs, offset string, cacheEfficiencyData bool) (error, error) {
+	warmFunc := func(duration, offset time.Duration, cacheEfficiencyData bool) (error, error) {
+		if a.ThanosClient != nil {
+			duration = thanos.OffsetDuration()
+			log.Infof("Setting Offset to %s", duration)
+		}
+		fmtDuration, fmtOffset := timeutil.DurationOffsetStrings(duration, offset)
+		durationHrs, err := timeutil.DayDurationToHourDuration(fmtDuration)
```
nikovacevic, line 1751: At a glance, I'm not sure I understand what the function `DayDurationToHourDuration` would do. Convert from a `string` to a `time.Duration`? `string` to `string`?

pkg/util/timeutil/timeutil.go
```diff
@@ -50,22 +52,47 @@ func DurationString(duration time.Duration) string {
...
 }
 
+// DurationToPromString returns a Prometheus formatted string with leading offset or empty string if given a negative duration
+func DurationToPromString(duration time.Duration) string {
+	dirStr := DurationString(duration)
+	if dirStr != "" {
+		dirStr = fmt.Sprintf("offset %s", dirStr)
+	}
+	return dirStr
+}
```
nikovacevic, line 70: If I give this function `duration = 30 * time.Minute`, then I get back the string `"offset 30m"`? FWIW the name `DurationToPromString` is misleading here, as not all Prometheus durations need to be prepended with "offset", which also makes it confusing to read the code where this function is used.

nikovacevic: While I'm open to being convinced otherwise, I tend to think we don't need an entire separate function for prepending "offset" on a duration string.

pkg/util/timeutil/timeutil.go
```diff
@@ -50,22 +52,47 @@ func DurationString(duration time.Duration) string {
 			// convert to mins
 			durStr = fmt.Sprintf("%dm", durSecs/SecsPerMin)
 		} else if durSecs > 0 {
-			// default to mins, as long as duration is positive
+			// default to secs, as long as duration is positive
```
nikovacevic, line 55: Good catch!

pkg/costmodel/cluster.go
```diff
@@ -775,10 +770,7 @@ func (a *Accesses) ComputeClusterCosts(client prometheus.Client, provider cloud.
 		queryTotalLocalStorage = fmt.Sprintf(" + %s", queryTotalLocalStorage)
 	}
 
-	fmtOffset := ""
-	if offset != "" {
-		fmtOffset = fmt.Sprintf("offset %s", offset)
-	}
+	fmtOffset := timeutil.DurationToPromString(offset)
```
nikovacevic, line 773: This looks suspicious, unless `timeutil.DurationToPromString` happens to add the string "offset " to the given duration, which would be unexpected.

nikovacevic (commented):
Several comments to address. In particular, there is at least one spot where I believe a bug may be introduced, so we'll want to double-check that.

In general, I'm torn about these kinds of changes. On the one hand, I agree in principle with everyone commenting here that it'd have been better to design all these functions to take `time.Duration` from the beginning. No argument here. And if we could guarantee no bugs, then I'd be excited to welcome these changes wholeheartedly. Unfortunately, I am worried about unintended consequences, especially given that many of these are old functions that are difficult to test. My conservative preference would be to leave most of them alone, only fixing the specific places we know that something is broken, or to thoroughly test every function, which I imagine would take quite some time. It might even be worth discussing whether or not to deprecate as many of these functions as possible, rather than trying to retrofit them. (e.g. do we use aggAPI cache warming any more?)

I'd love opinions from @mbolt35 and @AjayTripathy on this. Perhaps it's just that I'm sensitive because of all the bugs we've encountered this week, but this reading this PR gives me the feeling like it has the potential to do more harm than it prevents.

pkg/util/timeutil/timeutil.go
michaelmdresser: SGTM! I'm glad there's at least some canonical Go advice here. I wouldn't run out and refactor all of `pkg/util` but good to keep in mind for future changes like this.

pkg/util/timeutil/timeutil.go
Sean-Holcomb: I changed the name to DurationToPromOffsetString which I hope is a bit more clear how it is different from the other method. The function does a little more than just prepend, it also checks for empty strings. I agree that the fact that it through you off is a problem, and if you think even with the new name it isn't clear enough what it does, I will remove it. I do still like it better as a utility function.

pkg/costmodel/aggregation.go
Sean-Holcomb: I changed the name to FormatDurationStringDaysToHours

pkg/costmodel/aggregation.go
Sean-Holcomb: It was not, thank you for catching, I fixes and tested the aggregate, but I am not sure that is exactly what needs to be tested with this method

Sean-Holcomb:
@nikovacevic There is definitely the possibility that the refactoring here may cause some bugs, but the thing I think you are overlooking is that previous state of these functions did cause a bug and that is what is being fixed here. If these methods had been dealing in durations to begin with the bug never would have happened. The question is are we okay with leaving bug prone code in the codebase with some messy bandaids on top of it to avoid bugs in the short term, or is it better to risk the short term bugs to try and prevent similar situations down the line? My point here is that there is no guarantee of no bugs. 

I am also all for depreciation since it seems like we have better solutions to these problems now

nikovacevic:
@Sean-Holcomb that makes sense, I definitely appreciate that argument. I do still think that sweeping changes tend to invite new bugs, whereas targeted changes to fix specific, reported bugs tend to have a lower chance of inviting new bugs. At any rate, I support your convictions here and it looks like we support this PR as a team. I'm happy to join Michael in approving this PR now, with the note that we should do an extra-thorough job of regression testing this release. And I hope we can revisit some of these old features with the intent to deprecate soon!

nikovacevic (approved):

PR #842, opened 2021-06-25
Add json unmarshaling for all Asset types and AssetSet

kaelanspatel:
Implemented `UnmarshalJSON` for all asset types and for `AssetSet`, and wrote tests for each in `asset_unmarshal_test.go`. To test, run ` go test -run Test[ASSET TYPE HERE]_Unmarshal` e.g. `go test -run TestNetwork_Unmarshal`.

Exceptions are:
- A marshaled `Disk` cannot be recreated exactly from the values in the json. The `Local` field is left unset when unmarshaling. This is accounted for in the test, but it will mean that any `Disk` you get back from unmarshaling is not going to be the disk you put in.
- `AssetSet` only json-encodes the `asset` list; it should unmarshal to `AssetSetResponse`, which only contains that list.

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -2021,15 +2021,13 @@ func (n *Node) MarshalJSON() ([]byte, error) {
 }
 
 func (n *Node) UnmarshalJSON(b []byte) error {
-	//f := reflect.New(reflect.TypeOf((*Node)(nil)).Elem()).Interface().(interface{ UnmarshalBinary([]byte) error })
-	//var f map[string]*json.RawMessage
+
 	var f interface{}
-	//fmt.Println(fmt.Sprintf("%T", f))
+
 	err := json.Unmarshal(b, &f)
 	if err != nil {
 		return err
 	}
-	//fmt.Println(f)
 
 	fmap := f.(map[string]interface{})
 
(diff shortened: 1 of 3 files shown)


mbolt35:
My only request here is if we could potentially separate this into a separate source file? The high volume of encoding/decoding code intuitively feels like it should be organized in different source just to follow the binary marshal/unmarshal implementation.

pkg/kubecost/asset.go
```diff
@@ -2515,28 +2386,18 @@ func (sa *SharedAsset) Equal(a Asset) bool {
...
 	return toString(sa)
 }
 
+// This type exists because only the assets map of AssetSet is marshaled to
+// json, which makes it impossible to recreate an AssetSet struct. Thus,
+// the type when unmarshaling a marshaled AssetSet,is AssetSetResponse
+type AssetSetResponse struct {
+	Assets map[string]Asset
+}
+
```
michaelmdresser, line 2400: If this is used for json unmarshaling, it should have a `json:"fieldnamehere"` tag on the `Assets` field.

pkg/kubecost/asset.go
```diff
@@ -3045,6 +2899,14 @@ func (asr *AssetSetRange) MarshalJSON() ([]byte, error) {
...
 }
 
+// As with AssetSet, AssetSetRange does not serialize all its fields,
+// making it impossible to reconstruct the AssetSetRange from its json.
+// Therefore, the type a marshaled AssetSetRange unmarshals to is
+// AssetSetRangeResponse
+type AssetSetRangeResponse struct {
+	Assets []*AssetSetResponse
+}
+
```
michaelmdresser, line 2909: Same here.

michaelmdresser (approved):

pkg/kubecost/asset.go
michaelmdresser: Actually, never mind, because you define unmarshal manually. Ignore me!

PR #845, opened 2021-06-28
Added NetworkTransferBytes and NetworkReceiveBytes to Allocation

mbolt35:
CAdvisor Queries for tracking Pod Total Transferred/Received Bytes

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -63,6 +63,8 @@ type Allocation struct {
 	GPUHours                   float64               `json:"gpuHours"`
 	GPUCost                    float64               `json:"gpuCost"`
 	GPUCostAdjustment          float64               `json:"gpuCostAdjustment"`
+	NetworkTransferBytes       float64               `json:"networkTransferBytes"`
+	NetworkReceiveBytes        float64               `json:"networkReceiveBytes"`
 	NetworkCost                float64               `json:"networkCost"`
 	NetworkCostAdjustment      float64               `json:"networkCostAdjustment"`
 	LoadBalancerCost           float64               `json:"loadBalancerCost"`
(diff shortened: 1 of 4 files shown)


nikovacevic:
Any notes on how this was tested?

pkg/kubecost/allocation.go
```diff
@@ -63,6 +63,8 @@ type Allocation struct {
 	GPUHours                   float64               `json:"gpuHours"`
 	GPUCost                    float64               `json:"gpuCost"`
 	GPUCostAdjustment          float64               `json:"gpuCostAdjustment"`
+	NetworkTransferBytes       float64               `json:"networkTransferBytes"`
+	NetworkReceiveBytes        float64               `json:"networkReceiveBytes"`
```
nikovacevic, line 67: Should we also add these to the `MarshalJSON`, or do you want to keep them out of the JSON payload for now?

nikovacevic (commented):
Code looks great! Just the two questions about testing and JSON.

pkg/kubecost/allocation.go
mbolt35: Great Googly Moogly

mbolt35: > Any notes on how this was tested?

I had logging after the query that seemed to indicate the data coming through was correct, but after adding JSON encoding, I'm seeing 0s for both on the allocation request for 7d, so it seems I need to find a better way to end-to-end test.

Suggestions other than observing on the JSON result?

mbolt35: Added that, the fields to the equality check, and also to `add()`

mbolt35:
image

nikovacevic (approved):

PR #848, opened 2021-06-29
Check GPU label value in azure provider

Sean-Holcomb:
Check value of GPU Label and GPU count before emitting GPU cost for Azure nodes

Testing:
To test this feature I did four stages, first was an old image with no GPU label which is the first section of 0 GPU cost on the graph below, second was this up date with the new branch and the GPU label set to a label that was present on both labels with correct value, this is the first half of the plateau in the graph below. Third was develop branch with the label being present with the incorrect value, this is a reproduction of the bug and the second half of the plateau. Forth I kept the labels the same as the last step and switch back to the image with the these changes and the cost dropped back to 0. 

Screen Shot 2021-06-30 at 9 11 52 AM

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -210,11 +210,16 @@ func (k *azureKey) Features() string {
 	return fmt.Sprintf("%s,%s,%s", region, instance, usageType)
 }
 
+// GPUType returns value of GPULabel if present
 func (k *azureKey) GPUType() string {
 	if t, ok := k.Labels[k.GPULabel]; ok {
 		return t
 	}
 	return ""
+}
+
+func (k *azureKey) isValidGPUNode() bool {
+	return k.GPUType() == k.GPULabelValue && k.GetGPUCount() != "0"
 }
 
 func (k *azureKey) ID() string {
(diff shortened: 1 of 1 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -734,24 +739,30 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
 	c, err := az.GetConfig()
 	if err != nil {
 		return nil, fmt.Errorf("No default pricing data available")
 	}
-	if key.GPUType() != "" {
+	if azKey.isValidGPUNode()  {
 		return &Node{
 			VCPUCost: c.CPU,
 			RAMCost:  c.RAM,
+			UsesBaseCPUPrice: true,
```
AjayTripathy, line 763: I don't think we want this? You want to use the price from the API, not force the base price. I believe the CPU/RAM/GPU are just there to enforce ratios when splitting the whole cost among the node.

AjayTripathy: Or are you saying the code above returns the price as computed from the API?

Sean-Holcomb: Both this and the else clause below return default prices, it doesn't seem like the presence of a GPU would make this false so I added it in to match the case below.

AjayTripathy (approved):

PR #850, opened 2021-06-30
Check key existence in ClusterLoadBalancers

nikovacevic:
Fixes https://github.com/kubecost/cost-model/issues/849

## Changes
- Only try to set fields on a member of a map after checking it exists

## Testing
- Unable to reproduce, but manually regression tested

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -687,14 +687,16 @@ func ClusterLoadBalancers(cp cloud.Provider, client prometheus.Client, duration,
 			continue
 		}
 
-		s := time.Unix(int64(result.Values[0].Timestamp), 0)
-		e := time.Unix(int64(result.Values[len(result.Values)-1].Timestamp), 0)
-		mins := e.Sub(s).Minutes()
+		if lb, ok := loadBalancerMap[key]; ok {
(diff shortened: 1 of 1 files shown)


pkg/costmodel/cluster.go
```diff
@@ -687,14 +687,16 @@ func ClusterLoadBalancers(cp cloud.Provider, client prometheus.Client, duration,
...
+			s := time.Unix(int64(result.Values[0].Timestamp), 0)
+			e := time.Unix(int64(result.Values[len(result.Values)-1].Timestamp), 0)
+			mins := e.Sub(s).Minutes()
 
-		loadBalancerMap[key].Start = s
-		loadBalancerMap[key].Minutes = mins
+			lb.Start = s
+			lb.Minutes = mins
+		} else {
+			log.Warningf("ClusterLoadBalancers: found minutes for key that does not exist: %s", key)
```
michaelmdresser, line 698: Dedupe?

michaelmdresser (approved):
LGTM

pkg/costmodel/cluster.go
nikovacevic: Yeahhh probably smart

AjayTripathy (approved):

PR #855, opened 2021-07-01
Exclude unmounted PVs from shared costs

kaelanspatel:
Small change excluding allocations representing unmounted PVs from the calculation of shared costs; this should mean that the shared costs column for the Unmounted PVs row should always be zero. Total costs should still remain consistent.


kaelanspatel:
This change in tandem with https://github.com/kubecost/kubecost-cost-model/pull/460, which prevents Cluster Management costs from being distributed to unmounted PVs.

Sean-Holcomb (approved):
LGTM

PR #856, opened 2021-07-01
Multi-Cloud updates

Sean-Holcomb:
Required By: https://github.com/kubecost/kubecost-cost-model/pull/445
Related to: https://github.com/kubecost/cost-analyzer-helm-chart/pull/961

The main purpose of this PR is to expand the definition of the `AzureStorageConfig` to include the account name which is used as the providerKey for Azure.

It also cleans up some logs and adds an Azure provider to a method for generating `cloud.Provider`

Update:
To allow Reconciliation to work properly for multi-cloud the parsing of provider id's had to be moved out of the individual provider implementations and into helper functions. This is because when the federated data is read from Thanos, provider id parsing happened with the method of Provider type of the cluster where Kubecost is running. Because of this, Kubernetes assets from differing cloud providers would not have their provider ids parsed correctly and would not reconcile.

Screen Shot 2021-07-01 at 3 35 18 PM

While this solution is less flexible in that the provider id has to go through the parse functions of every provider type, in practice this is not a problem because all of the parse functions so far contain substrings that identify the provider type (ie "gcp:", "aws//") and all other parse functions return the string as is. The result is proper reconciliation on a mutli-cloud federated cluster.

Screen Shot 2021-07-07 at 5 56 02 PM
Screen Shot 2021-07-07 at 5 57 13 PM
Screen Shot 2021-07-07 at 5 57 50 PM


michaelmdresser (approved):
LGTM

nikovacevic (approved):
LGTM!

pkg/cloud/provider.go
```diff
@@ -501,3 +504,58 @@ func GetOrCreateClusterMeta(cluster_id, cluster_name string) (string, string, er
...
+	match = rx.FindStringSubmatch(id)
+	if len(match) >= 2 {
+		return match[1]
+	}
+
+	// Return id for Azure Provider, CSV Provider and Custom Provider
+
+	return id
+
+
```
nikovacevic, line 529: Let's try to be a little tighter with code formatting here. This is a lot of empty space. (I thought `go fmt` deleted excessive whitespace, but maybe that's just a setting I'm using.)

pkg/costmodel/allocation.go
```diff
@@ -1383,16 +1382,15 @@ func applyNodeCostPerCPUHr(nodeMap map[nodeKey]*NodePricing, resNodeCostPerCPUHr
 			nodeMap[key] = &NodePricing{
 				Name:       node,
 				NodeType:   instanceType,
-				ProviderID: providerIDParser(providerID),
+				ProviderID: cloud.ParseID(providerID),
```
nikovacevic, line 1385: Is there something we could do so that we don't have to parse the `providerID` at this point? Feels like a step that we could (maybe even should) be doing upstream of this.

nikovacevic (commented):
It's definitely nicer to let each cloud provider implement its own version of those functions, but I trust your judgment here. Just a few questions.

pkg/cloud/provider.go
michaelmdresser: I'm pretty sure `go fmt` does, though I think it only would delete line 529 here, maybe 528 too because it's the end of the function?

pkg/costmodel/allocation.go
Sean-Holcomb: The step before this is metric emission into Prometheus, I really like the idea of all the providerIDs going into Prometheus being pre parsed. Two things about that though. There would be a coverage gap, where we would have to parse it on both sides until we were sure that all the unparsed providers ids were out of everyones prometheus. I know we also don't emit all of the metrics that we use, if there were some that has provider_id and we were not in control we would still have to parse them after which would end up being a mixed solution with some pre-parsing and some post-parsing, which I would rather not do. Off hand do you know if that is the case? If we were able to do pre parsing I could move the parse functions back into the provider implementations which would be nice and allow for meaningful logging on parsing too.

michaelmdresser: `resNodeCostPerCPUHr` could theoretically contain results for multiple cloud providers though, right?

michaelmdresser: Ah, looks like I was a little slow. That makes perfect sense Sean.

PR #860, opened 2021-07-05
fix ServiceAccountCheck struct missing json tag quote

kitianFresh:
Hi, team

Found a little bug for json struct.


AjayTripathy:
Thanks for this fix @kitianFresh !

AjayTripathy (approved):

PR #861, opened 2021-07-05
deal with prometheus client init error and fatal exit when there has error

kitianFresh:
deal with prometheus client init error and fatal exit when there has error; when no prometheus, the server should exit and show reasons or it will crash when requests some api when server started


nikovacevic (approved):
Looks great, @kitianFresh. Thank you!

PR #864, opened 2021-07-13
Add ShareTenancyCosts to configuration, defaulting to true

nikovacevic:
Requires https://github.com/kubecost/kubecost-cost-model/pull/471
Requires https://github.com/kubecost/cost-analyzer-frontend/pull/328
Requires https://github.com/kubecost/cost-analyzer-helm-chart/pull/973

## Changes
- Add `ShareTenancyCosts` to configuration, with default value of `"true"` until it is set manually.

## Testing
In conjunction with the other PRs, the experience should look like this:

Default to sharing tenancy costs, with a chip in the UI denoting that. A tooltip explains what it means and indicates that it can be disabled in the Settings page.
Screenshot from 2021-07-13 13-47-33

Visiting the Settings page surfaces this setting and allows it to be turned off.
Screenshot from 2021-07-13 13-42-07
Screenshot from 2021-07-13 13-44-09

With the setting turned off, sharing tenancy costs stops, allowing us to get $0 in shared cost with none of the sharing features enabled, which is not possible today.
Screenshot from 2021-07-13 13-44-26

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -366,14 +366,17 @@ func (aws *AWS) GetManagementPlatform() (string, error) {
 
 func (aws *AWS) GetConfig() (*CustomPricing, error) {
 	c, err := aws.Config.GetCustomPricingData()
+	if err != nil {
+		return nil, err
+	}
 	if c.Discount == "" {
 		c.Discount = "0%"
 	}
 	if c.NegotiatedDiscount == "" {
 		c.NegotiatedDiscount = "0%"
 	}
-	if err != nil {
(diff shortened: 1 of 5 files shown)


michaelmdresser (approved):
LGTM

Sean-Holcomb (approved):
This seems fine, I do wonder if it is necessary to propagate, the shareTenencyCost through the OS given that it is a closed source functionality. I am assuming it is to the end of handling the configurations in a uniform fashion, even if it doesn't mean anything at this level.

nikovacevic:
@Sean-Holcomb yep, that's exactly it.

PR #870, opened 2021-07-20
Add helm config parameter for shared overhead costs

kaelanspatel:
Ref https://github.com/kubecost/cost-analyzer-helm-chart/issues/972
Requires X

Logic for dealing with the `sharedOverhead` parameter, adding it to the `SharedCosts` map in `CustomPricing`.


pkg/cloud/providerconfig.go
```diff
@@ -187,6 +187,23 @@ func DefaultPricing() *CustomPricing {
...
+
+		if reflect.ValueOf(obj.SharedCosts["total"]).Type() != val.Type() {
+			return fmt.Errorf("cannot insert value into custom pricing shared costs")
+		}
+
+		obj.SharedCosts["total"] = value
+
+		return nil
+
+	}
```
kaelanspatel, line 205: I don't like this at all, but I don't see any other place to put it.

michaelmdresser: I just talked with Niko about this. A possible alternative is to actually add a `SharedOverhead` field to the `CustomPricing` struct and then check all uses of the `SharedCosts` map and replace with `SharedOverhead`. According to Niko, even though `SharedCosts` is a map it only ever has one thing in it (the shared overhead cost). You'll have to confirm that. If that is indeed the case then `SharedCosts` can be removed entirely and can be replaced with `SharedOverhead` which doesn't have to be a map.

Doing this would totally eliminate this nastiness but it does involve a more substantial code change (in other repos, too). I think it is a good idea for long term code health.

Does that make sense? What do you think?

kaelanspatel: Yes, that makes complete sense. I'll confirm that there's no other elements being put in `SharedCosts` and, if that's the case, see about replacing it with `SharedOverhead`.

kaelanspatel:
Recent commit requires https://github.com/kubecost/cost-analyzer-frontend/pull/352, https://github.com/kubecost/kubecost-cost-model/pull/487

Remove references to single-value map `SharedCosts` and replaces them with `SharedOverhead`, removing conditional logic block in pricing config.

pkg/cloud/provider.go
```diff
@@ -130,56 +130,56 @@ type OutOfClusterAllocation struct {
...
+	DefaultIdle                  string `json:"defaultIdle"`
+	AzureSubscriptionID          string `json:"azureSubscriptionID"`
+	AzureClientID                string `json:"azureClientID"`
+	AzureClientSecret            string `json:"azureClientSecret"`
+	AzureTenantID                string `json:"azureTenantID"`
+	AzureBillingRegion           string `json:"azureBillingRegion"`
+	CurrencyCode                 string `json:"currencyCode"`
+	Discount                     string `json:"discount"`
+	NegotiatedDiscount           string `json:"negotiatedDiscount"`
+	SharedOverhead               string `json:"sharedOverhead"`
```
nikovacevic, line 175: Do you know if there will be any ramifications to renaming this? Since there was not a helm value prior, it seems uncontroversial, but good to double-check.

pkg/costmodel/aggregation.go
```diff
@@ -1451,17 +1452,19 @@ func (a *Accesses) ComputeAggregateCostModel(promClient prometheusClient.Client,
...
-			durationCoefficient := window.Hours() / timeutil.HoursPerMonth
-			if err != nil {
-				return nil, "", fmt.Errorf("unable to parse shared cost %s: %s", val, err)
-			}
-			sc[key] = &SharedCostInfo{
-				Name: key,
-				Cost: cost * durationCoefficient,
-			}
+
+		overheadVal := c.SharedOverhead
```
nikovacevic, line 1456: Again, could probably get away without assigning a new variable here, but that's more of a nit.

pkg/cloud/providerconfig.go
```diff
@@ -187,6 +187,7 @@ func DefaultPricing() *CustomPricing {
 }
 
 func SetCustomPricingField(obj *CustomPricing, name string, value string) error {
+
```
nikovacevic, line 190: :eyes:

nikovacevic (commented):
With this one, in particular, can you share some details about how this was tested? Was it tested manually? If so, on which providers? Just want to make sure we have our bases covered, and it's good practice to describe that.

pkg/cloud/provider.go
kaelanspatel: I _should_ have removed all references to `SharedCosts` across all repos and replaced them + their logic with the new `SharedOverhead`.

pkg/costmodel/aggregation.go
kaelanspatel: You're right about that, there are actually a few cases where the refactor could go further, but I think you could also go so far you end up changing fundamental structs and I honestly didn't want to overreach.

kaelanspatel:
> With this one, in particular, can you share some details about how this was tested? Was it tested manually? If so, on which providers? Just want to make sure we have our bases covered, and it's good practice to describe that.

Tbh I tested it manually with only GCP. The logic was and is identical for all providers in adding shared overhead, so I didn't think it was necessary to go further.

If you think that it would benefit from further testing I can definitely get on that, it's just I can't think of a way that it would be different when the pre-existing code already works. One of the key things about this refactor is (imo) removing those extra logic blocks in `provider.go` etc and making use of the logic that already exists (i.e. use of `SetCustomPricingField`).

Definitely robustness-wise I understand testing further, though.

pkg/costmodel/aggregation.go
nikovacevic: What do you mean by that?

nikovacevic:
It should be very quick to test on AWS and Azure, as long as you have a kube-config for each of them. (Do you? If not, let me know and I can share one with you.) When the test is so simple to carry out, it's nice to have the peace of mind IMO. It's not uncommon to end up finding a type or some other surprise as a consequence of what feels like mundane testing.

pkg/costmodel/aggregation.go
kaelanspatel: Ah, I read that as "you could get away _with_ assigning a new variable here", my bad.

PR #871, opened 2021-07-20
Sean/azure spot clean up

Sean-Holcomb:
Apply code suggestions from Kubecost team to @subreptivus 's PR [832](https://github.com/kubecost/cost-model/pull/832)
to improve spot pricing estimates by using the Azure retail price.

Testing: 
Created new cluster with a spot vmss with an A2 instance and progressively applied configurations and this PR to see the effect on the price of the node being emitted. The result were mixed as adding these changes actually increased the price of the spot node being reported. In the below graph the first step is when the pricing API for the Azure cluster is configured and you can see the A2 instance remains much lower than the DS2 instance. The second increase in price happens when the changes from this PR are applied as the A2 price increases almost to the same as the Ds2
Screen Shot 2021-07-19 at 5 12 54 PM
Screen Shot 2021-07-19 at 5 13 05 PM

Looking at the logs I am able to confirm that the retail pricing API is successfully being hit.
Screen Shot 2021-07-19 at 5 23 45 PM

To test further I created a non-spot A2 VMSS which remained at the initial price after api configuration. Before moving on with this PR I would like to see how this new A2 node's cost is adjusted. And also see how the pricing for the spot node looks in the Azure cost analyzer to confirm if the price jump was accurate or not.

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
@@ -149,6 +151,64 @@ func getRegions(service string, subscriptionsClient subscriptions.Client, provid
 	}
 }
 
+func getRetailPrice(region string, skuName string, currencyCode string) (string, error) {
+	pricingURL := "https://prices.azure.com/api/retail/prices?$skip=0&currencyCode='" + currencyCode + "'"
+	filterBase := "&$filter="
+	var filterParams []string
+
+	if region != "" {
+		regionParam := "armRegionName eq '" + region + "'"
+		filterParams = append(filterParams, regionParam)
+	}
+
+	if skuName != "" {
+		skuNameParam := "armSkuName eq '" + skuName + "'"
+		filterParams = append(filterParams, skuNameParam)
+	}
+
+	if len(filterParams) > 0 {
+		filterParamsEscaped := url.QueryEscape(strings.Join(filterParams[:], " and "))
+		pricingURL += filterBase + filterParamsEscaped
+	}
+
+	klog.V(4).Infof("starting download retail price payload from \"%s\"", pricingURL)
+	resp, err := http.Get(pricingURL)
+
+	if err != nil {
+		klog.V(4).Infof("Bogus fetch of \"%s\": %v", pricingURL, err)
+		return "", err
+	}
+
+	pricingPayload := AzureRetailPricing{}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		klog.V(4).Infof("Error getting response: %v", err)
+	}
+
+	jsonErr := json.Unmarshal(body, &pricingPayload)
+	if jsonErr != nil {
+		klog.V(4).Infof("Error unmarshalling data: %v", jsonErr)
+	}
+
+	retailPrice := ""
+
+	for i := range pricingPayload.Items {
+		if !(pricingPayload.Items[i].Type != "Consumption" || strings.Contains(pricingPayload.Items[i].ProductName, "Windows")) {
+			retailPrice = fmt.Sprintf("%f", pricingPayload.Items[i].RetailPrice)
+		}
+	}
+
+	klog.V(4).Infof("done parsing retail price payload from \"%s\"\n", pricingURL)
+
+	if retailPrice == "" {
+		return retailPrice, fmt.Errorf("Couldn't find price for product \"%s\" in \"%s\" region", skuName, region)
+	} else {
+		return retailPrice, nil
+	}
+
+}
+
 func toRegionID(meterRegion string, regions map[string]string) (string, error) {
 	var rp regionParts = strings.Split(strings.ToLower(meterRegion), " ")
 	regionCode := regionCodeMappings[rp[0]]
@@ -180,6 +240,39 @@ func checkRegionID(regionID string, regions map[string]string) bool {
 		}
 	}
 	return false
+}
+
+type AzureRetailPricing struct {
+	BillingCurrency    string                         `json:"BillingCurrency"`
+	CustomerEntityId   string                         `json:"CustomerEntityId"`
+	CustomerEntityType string                         `json:"CustomerEntityType"`
+	Items              []AzureRetailPricingAttributes `json:"Items"`
+	NextPageLink       string                         `json:"NextPageLink"`
+	Count              int                            `json:"Count"`
+}
+
+type AzureRetailPricingAttributes struct {
+	CurrencyCode         string     `json:"currencyCode"`
+	TierMinimumUnits     float32    `json:"tierMinimumUnits"`
+	RetailPrice          float32    `json:"retailPrice"`
+	UnitPrice            float32    `json:"unitPrice"`
+	ArmRegionName        string     `json:"armRegionName"`
+	Location             string     `json:"location"`
+	EffectiveStartDate   *time.Time `json:"effectiveStartDate"`
+	EffectiveEndDate     *time.Time `json:"effectiveEndDate"`
+	MeterId              string     `json:"meterId"`
+	MeterName            string     `json:"meterName"`
+	ProductId            string     `json:"productId"`
+	SkuId                string     `json:"skuId"`
+	ProductName          string     `json:"productName"`
+	SkuName              string     `json:"skuName"`
+	ServiceName          string     `json:"serviceName"`
+	ServiceId            string     `json:"serviceId"`
+	ServiceFamily        string     `json:"serviceFamily"`
+	UnitOfMeasure        string     `json:"unitOfMeasure"`
+	Type                 string     `json:"type"`
+	IsPrimaryMeterRegion bool       `json:"isPrimaryMeterRegion"`
+	ArmSkuName           string     `json:"armSkuName"`
 }
 
 // AzurePricing either contains a Node or PV
@@ -703,6 +796,7 @@ func (az *Azure) DownloadPricingData() error {
 						Node: &Node{
 							Cost:         priceStr,
 							BaseCPUPrice: baseCPUPrice,
+							UsageType:    usageType,
 						},
 					}
 				}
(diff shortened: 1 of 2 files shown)


Sean-Holcomb:
Update on testing:
The issue I was seeing was that the results of the API were not being filtered strictly enough to grab the correct spot price. I added a bool argument to the getRetailPrice function `spot` which if true filters for the word " Spot" in the item skuName and if it is not true skips any spot items. The result was a drop in price below the none spot A2 node.
Screen Shot 2021-07-23 at 11 21 43 AM
Additionally the price that is being grabbed from the Api result now matches with the daily cost of the spot node seen on the azure console 0.028938 * 24 = 0.694512
Screen Shot 2021-07-22 at 8 04 31 PM
Screen Shot 2021-07-23 at 11 48 37 AM

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +839,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
 	if !ok {
 		return nil, fmt.Errorf("azure: NodePricing: key is of type %T", key)
 	}
+	config, _ := az.GetConfig()
+	if config.SpotLabel != "" && config.SpotLabelValue != "" {
+		labelsMapStr := fmt.Sprintf("%s", azKey)
+		spotLabelMapStr := fmt.Sprintf(" %s:%s ", config.SpotLabel, config.SpotLabelValue)
+		if strings.Contains(labelsMapStr, spotLabelMapStr) {
+			features := strings.Split(azKey.Features(), ",")
+			spotFeatures := strings.ReplaceAll(azKey.Features(), ",ondemand", ",spot")
```
michaelmdresser, line 853: Is this `ReplaceAll` robust to the features being just `ondemand` and not `foo,bar,ondemand`?

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +839,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
+
 	azKey, ok := key.(*azureKey)
 	if !ok {
 		return nil, fmt.Errorf("azure: NodePricing: key is of type %T", key)
 	}
+	config, _ := az.GetConfig()
+	if config.SpotLabel != "" && config.SpotLabelValue != "" {
+		labelsMapStr := fmt.Sprintf("%s", azKey)
+		spotLabelMapStr := fmt.Sprintf(" %s:%s ", config.SpotLabel, config.SpotLabelValue)
+		if strings.Contains(labelsMapStr, spotLabelMapStr) {
```
michaelmdresser, line 851: I'm a little lost here. Does `azKey.String()` (which I think is what gets called when you `%s` format it?) just output space-separated labels of the node?

If that assumption is right, does it always have a leading/trailing space? This is the same kind of question with the placement of `ondemand` in the `ReplaceAll`, I just want to make sure we're robust to the placement of the specific label within the list.

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +839,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
+			features := strings.Split(azKey.Features(), ",")
+			spotFeatures := strings.ReplaceAll(azKey.Features(), ",ondemand", ",spot")
+			if n, ok := az.Pricing[spotFeatures]; ok {
+				klog.V(4).Infof("Returning pricing for node %s: %+v from key %s", azKey, n, spotFeatures)
```
michaelmdresser, line 855: Could this cause log spam?

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +839,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
+				if azKey.isValidGPUNode() {
+					n.Node.GPU = "1" // TODO: support multiple GPUs
+				}
+				return n.Node, nil
+			}
```
michaelmdresser, line 860: I assume `az.Pricing` is some sort of cache of node data? Why does the `Node.GPU` need to be updated if its already been cached?

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +839,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
+			klog.V(1).Infof("[Info] found spot instance, trying to get retail price for %s: %s, ", spotFeatures, azKey)
+
+			spotCost, err := getRetailPrice(features[0], features[1], config.CurrencyCode, true)
```
michaelmdresser, line 863: Is the order (mapping of index -> feature) of `features` always the same? Is there any danger of `len(features) < 2`?

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +839,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
+			if err != nil {
+				log.DedupedWarningf(5, "failed to retrieve spot retail pricing")
```
michaelmdresser, line 865: Is the intent of not returning here to fall back on regular pricing if the spot lookup fails?

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +839,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
+						gpu = "1"
+					}
+					allPrices.(map[string]*AzurePricing)[spotFeatures] = &AzurePricing{
+						Node: &Node{
+							Cost:      spotCost,
+							UsageType: "spot",
+							GPU:       gpu,
+						},
+					}
+					az.Pricing = allPrices.(map[string]*AzurePricing)
```
michaelmdresser, line 880: Is there any mutex or concurrency control on `az.Pricing` that should be checked here? I see a mutex being used in `AllNodePricing` for something like this.

--

Also, what if there are 2 different unknown node types? Won't this set of logic do the following?
1. create a new allprices with node type 1
2. update `az.Pricing` with the new allprices with node type 1
3. lookup node type 2 and get back in this logic
4. create a new allprices with node type 2 (and not node type 1)
5. update `az.Pricing` with the new allprices that has node type 1 but not node type 2

Would it be better to just merge the current `az.Pricing` map with the new, updated map?

michaelmdresser (commented):
Thanks for all of the testing commentary, have some questions. Let me know if it'd help to hop on a call to discuss.

pkg/cloud/azureprovider.go
```diff
@@ -149,6 +154,64 @@ func getRegions(service string, subscriptionsClient subscriptions.Client, provid
...
+	}
+
+	jsonErr := json.Unmarshal(body, &pricingPayload)
+	if jsonErr != nil {
+		return "", fmt.Errorf("Error unmarshalling data: %v", jsonErr)
+	}
+
+	retailPrice := ""
+	for _, item := range pricingPayload.Items {
+		if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
```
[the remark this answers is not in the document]
Sean-Holcomb: The "Consumption" on could, but the the filter language does not include "not" so it is pretty limited. This also stopped me from including spot in the filters

pkg/cloud/azureprovider.go
Sean-Holcomb: Here is @Subreptivus on the topic
> Padding is there, because labelsMapStr will have all the labels separated by space, and substr can find needed string even in the middle of another string (this is kind of just for safety, because you can set spot labels while deploying cluster and it could be something really simple, which could be already there within another label, like: kubernetes.azure.com/scalesetpriority: mylabel and kubernetes.azure.com/scalesetpriority: mylabelspot).

however maybe doing a split on " " and searching for the unpadded string in the resulting slice would be more robust

pkg/cloud/azureprovider.go
Sean-Holcomb: That is correct

pkg/cloud/azureprovider.go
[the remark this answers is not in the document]
michaelmdresser: Ah, unfortunate. Well, no biggy.

pkg/cloud/azureprovider.go
michaelmdresser: Hmm yeah this kind of string manipulation just doesn't feel right to me. I'm particularly not a fan of the `fmt.Sprintf("%s", azKey)`, which seems to be relying on behavior of `azKey.String()` (I think?) that I wouldn't expect.

pkg/cloud/azureprovider.go
michaelmdresser: Would you drop a comment for that?

pkg/cloud/azureprovider.go
Sean-Holcomb: The same mutex is being locked for this function
```
func (az *Azure) NodePricing(key Key) (*Node, error) {
	az.DownloadPricingDataLock.RLock()
	defer az.DownloadPricingDataLock.RUnlock()
```

What I am unsure of is since `az.AllNodePricing()` is called here and contains a `defer az.DownloadPricingDataLock.RUnlock()` it seems like it is unlocking the mutex prematurely

michaelmdresser: That's only a read lock. I think you need to be using `.Lock()` when editing `az.Pricing` and `.RLock()` when reading.

pkg/cloud/azureprovider.go
```diff
@@ -149,6 +154,72 @@ func getRegions(service string, subscriptionsClient subscriptions.Client, provid
...
+	klog.V(4).Infof("starting download retail price payload from \"%s\"", pricingURL)
+	resp, err := http.Get(pricingURL)
+
+	if err != nil {
+		return "", fmt.Errorf("bogus fetch of \"%s\": %v", pricingURL, err)
+	}
+
+	if resp.StatusCode != 200 {
+		log.DedupedInfof(5, "retail price responded with status code %d", resp.StatusCode)
+	}
```
michaelmdresser, line 190: I'm pretty sure you should return an error here instead of just logging, unless a non-200 response code is expected.

pkg/cloud/azureprovider.go
```diff
@@ -740,10 +847,55 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
 	if !ok {
 		return nil, fmt.Errorf("azure: NodePricing: key is of type %T", key)
 	}
+	config, _ := az.GetConfig()
+	if slv, ok := azKey.Labels[config.SpotLabel];
+	ok && slv == config.SpotLabelValue && config.SpotLabel != "" && config.SpotLabelValue != "" {
+		features := strings.Split(azKey.Features(), ",")
+		region := features[0]
+		instance := features[1]
+		spotFeatures := fmt.Sprintf("%s,%s,%s", region, instance, "spot")
```
michaelmdresser, line 861: This is way better! Thanks!

michaelmdresser (approved):

PR #875, opened 2021-07-23
Add cloud status types

Sean-Holcomb:
required for:  https://github.com/kubecost/kubecost-cost-model/pull/481

### DIFF
--- a/pkg/kubecost/status.go
+++ b/pkg/kubecost/status.go
@@ -33,3 +33,31 @@ type FileStatus struct {
 	Errors       []string          `json:"errors,omitempty"`
 	Warnings     []string          `json:"warnings,omitempty"`
 }
+
+// CloudStatus describes CloudStore metadata
+type CloudStatus struct {
+	CloudAssets    *CloudAssetStatus     `json:"cloudAssets,omitempty`
+	Reconciliation *ReconciliationStatus `json:"reconciliation,omitempty"`
+}
+
+// CloudAssetStatus describes CloudAsset metadata of a CloudStore
+type CloudAssetStatus struct {
+	Coverage    Window            `json:"coverage"`
+	LastRun     time.Time         `json:"lastRun"`
+	NextRun     time.Time         `json:"NextRun"`
+	Progress    float64           `json:"progress"`
+	RefreshRate string            `json:"refreshRate"`
+	Resolution  string            `json:"resolution"`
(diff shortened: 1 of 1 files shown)


pkg/kubecost/status.go
```diff
@@ -33,3 +33,31 @@ type FileStatus struct {
...
+}
+
+// ReconciliationStatus describes Reconcilation metadata of a CloudStore
+type ReconciliationStatus struct {
+	Coverage    Window            `json:"coverage"`
+	LastRun     time.Time         `json:"lastRun"`
+	NextRun     time.Time         `json:"NextRun"`
+	Progress    float64           `json:"progress"`
+	RefreshRate string            `json:"refreshRate"`
+	Resolution  string            `json:"resolution"`
```
michaelmdresser, line 61: Why is `Resolution` a string and not a `time.Duration`? Looks like this is a pattern, so probably a question for @nikovacevic

michaelmdresser (approved):
Why do the new statuses define `NextRun` but `ETLStatus` doesn't?

Why don't they define `MaxBatch`, `UTCOffset`, and `Backup` when `ETLStatus` does?

 Just want to make sure I understand!

pkg/kubecost/status.go
nikovacevic: It's a string in the status struct so that it's human-readable. Otherwise, it just looks like an integer, which is hard to interpret in status responses. It's a `time.Duration` in the ETLs, themselves.

michaelmdresser: Ah, gotcha! Thanks.

pkg/kubecost/status.go
```diff
@@ -33,3 +33,31 @@ type FileStatus struct {
...
+type CloudStatus struct {
+	CloudAssets    *CloudAssetStatus     `json:"cloudAssets,omitempty`
+	Reconciliation *ReconciliationStatus `json:"reconciliation,omitempty"`
+}
+
+// CloudAssetStatus describes CloudAsset metadata of a CloudStore
+type CloudAssetStatus struct {
+	Coverage    Window            `json:"coverage"`
+	LastRun     time.Time         `json:"lastRun"`
+	NextRun     time.Time         `json:"NextRun"`
```
nikovacevic, line 47: Probably better to lower-case the "N" here and in the other struct, too. Would you agree?

PR #877, opened 2021-07-26
Add Start, End, and Minutes to AllocationSetRange

michaelmdresser:
These methods are useful because they enable more accurate rate
calculations. Instead of relying on the ASR's Window (which generally
maps to the _query_ window) these are computed from the Allocations
in the ASR. So if Kubecost has been running for only a day and 30d
of data is queried, the AllocationSetRange's Window will be ~30d but
the new Start(), End(), and Minutes() will reflect the data within
the ASR (i.e. they will reflect 1d of duration).

At some point we may want to modify the API response to include metadata
like this for AllocationSetRanges and AllocationSets.

Tested with the unit tests!

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2456,16 +2456,16 @@ func (asr *AllocationSetRange) End() (time.Time, error) {
 
 // Minutes returns the duration, in minutes, between the earliest start
 // and the latest end of all allocations in the AllocationSetRange.
-func (asr *AllocationSetRange) Minutes() (float64, error) {
+func (asr *AllocationSetRange) Minutes() float64 {
 	start, err := asr.Start()
 	if err != nil {
-		return 0, fmt.Errorf("failed to calculate start: %s", err)
+		return 0
 	}
 	end, err := asr.End()
 	if err != nil {
-		return 0, fmt.Errorf("failed to calculate end: %s", err)
+		return 0
 	}
 	duration := end.Sub(start)
 
-	return duration.Minutes(), nil
+	return duration.Minutes()
 }
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2405,3 +2405,67 @@ func (asr *AllocationSetRange) Window() Window {
 
 	return NewWindow(&start, &end)
 }
+
+// Start returns the earliest start of all Allocations in the AllocationSetRange.
+// It returns an error if there are no allocations.
+func (asr *AllocationSetRange) Start() (time.Time, error) {
+	start := time.Time{}
+	firstStartNotSet := true
+	asr.Each(func(i int, as *AllocationSet) {
+		as.Each(func(s string, a *Allocation) {
+			if firstStartNotSet {
+				start = a.Start
+				firstStartNotSet = false
+			}
+			if a.Start.Before(start) {
+				start = a.Start
+			}
+		})
+	})
+
+	if firstStartNotSet {
+		return start, fmt.Errorf("had no data to compute a start from")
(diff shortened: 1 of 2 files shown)


pkg/kubecost/allocation.go
```diff
@@ -2405,3 +2405,67 @@ func (asr *AllocationSetRange) Window() Window {
 
 	return NewWindow(&start, &end)
 }
+
+// Start returns the earliest start of all Allocations in the AllocationSetRange.
+// It returns an error if there are no allocations.
+func (asr *AllocationSetRange) Start() (time.Time, error) {
+	start := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
```
nikovacevic, line 2412: Why this time and not the zero value for `time.Time`?

pkg/kubecost/allocation.go
```diff
@@ -2405,3 +2405,67 @@ func (asr *AllocationSetRange) Window() Window {
...
+	})
+
+	if firstStartNotSet {
+		return start, fmt.Errorf("had no data to compute a start from")
+	}
+
+	return start, nil
+}
+
+// End returns the lateste end of all Allocations in the AllocationSetRange.
```
nikovacevic, line 2433: `:%s/lateste/latest`

pkg/kubecost/allocation.go
```diff
@@ -2405,3 +2405,67 @@ func (asr *AllocationSetRange) Window() Window {
...
+// and the latest end of all allocations in the AllocationSetRange.
+func (asr *AllocationSetRange) Minutes() (float64, error) {
+	start, err := asr.Start()
+	if err != nil {
+		return 0, fmt.Errorf("failed to calculate start: %s", err)
+	}
+	end, err := asr.End()
+	if err != nil {
+		return 0, fmt.Errorf("failed to calculate end: %s", err)
+	}
```
nikovacevic, line 2467: What do you think about just returning `0` if there are no allocations?

pkg/kubecost/allocation.go
```diff
@@ -2405,3 +2405,67 @@ func (asr *AllocationSetRange) Window() Window {
...
+		return start, fmt.Errorf("had no data to compute a start from")
+	}
+
+	return start, nil
+}
+
+// End returns the lateste end of all Allocations in the AllocationSetRange.
+// It returns an error if there are no allocations.
+func (asr *AllocationSetRange) End() (time.Time, error) {
+	end := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
```
nikovacevic, line 2436: Just curious why you picked this time, and not the zero value for `time.Time`?

nikovacevic (approved):
Couple of Qs, but the design decisions are up to you, so feel free to merge if you're happy.

pkg/kubecost/allocation.go
nikovacevic: (Oops, dupe question!)

pkg/kubecost/allocation.go
michaelmdresser: No reason in particular. Switched to zero value.

pkg/kubecost/allocation.go
michaelmdresser: Because the methods this calls (`Start` and `End`) can error, I think I have to propagate that error because otherwise I make an assumption that the only error states `Start` and `End` will be in are "no allocations."

OTOH I do think it is entirely reasonable for 0 to be returned here in the event of an error on start and end, as the value itself isn't actually invalid and callers currently have no need of knowing if the 0 is the result of an error or not. I'll change it.

PR #878, opened 2021-07-26
Add node type to allocation properties

michaelmdresser:
NodeType gives us type information in the Allocation which avoids
a full query and iteration over Asset data. This is useful for a
variety of recommendation features to improve performance and
reduce code complexity.


pkg/kubecost/allocationprops.go
```diff
@@ -93,6 +96,7 @@ func ParseProperty(text string) (string, error) {
 type AllocationProperties struct {
 	Cluster        string                `json:"cluster,omitempty"`
 	Node           string                `json:"node,omitempty"`
+	NodeType       string                `json:"nodeType,omitempty"`
```
nikovacevic, line 99: Reading this kind of gets me excited, actually--I think there could be some interesting Allocation-related analysis done with this field.

pkg/kubecost/bingen.go
```diff
@@ -29,4 +29,4 @@ package kubecost
 // @bingen:generate:PVKey
 // @bingen:generate:PVAllocation
 
-//go:generate bingen -package=kubecost -version=15 -buffer=github.com/kubecost/cost-model/pkg/util
+//go:generate bingen -package=kubecost -version=16 -buffer=github.com/kubecost/cost-model/pkg/util
```
nikovacevic, line 32: Flagging for @AjayTripathy that this will require a rebuild

pkg/kubecost/mock.go
```diff
@@ -20,6 +20,7 @@ func NewMockUnitAllocation(name string, start time.Time, resolution time.Duratio
 	if props == nil {
 		properties.Cluster = "cluster1"
 		properties.Node = "node1"
+		properties.NodeType = "kc1-super-medium"
```
nikovacevic, line 23: lol

nikovacevic (approved):
LGTM!

michaelmdresser:
Just in case someone decides to merge this — I want to check Allocation.add first, I think it needs to be updated

nikovacevic:
Thanks Michael. Going to wait on this until after the current release, anyways.

AjayTripathy:
I want to be really disciplined about schema changes. How much does this improve performance at read time? As it is, it isn't really worth it to trade a full allocation rebuild for code simplicity. That may change as we make allocation rebuilds better, but can we just pursue the feature you had in mind without this PR?

michaelmdresser:
Without this change, a full scan of all assets is required for each controller considered by the spot checklist. So this should scale poorly (`O(|assets| * |controllers|)` is my rough guess at runtime complexity). This change is still not _strictly_ necessary and we can go on without it; the checklist estimates work. I'd be curious to hear how it is performing for our very large customers.

michaelmdresser:
Closing due to age and a lack of focus on this issue right now. We can reopen this or a similar PR once we reevaluate the speed of Spot Checklist and identify this as a bottleneck.

PR #883, opened 2021-08-02
Changes for edge case where idle would not be shared between allocations.

kaelanspatel:
Requires https://github.com/kubecost/cost-analyzer-frontend/pull/354

The way in which idle is being shared between allocations means that if a resource (cpu/gpu/ram) does not have any cost other than idle, the idle costs for that resource would disappear. E.g. below there are no GPU costs but the idle one, so currently that $27 would be distributed by weight to each allocation. However, since there are no other GPU costs, the weights would all be zero, effectively deleting the cost.
image

With this + the frontend pull, the costs instead go into a manufactured idle row which only keeps the costs for a resource which would not be distributed:
image

Caveats:
- Because the idle is not being distributed, there's no way to normalize it, so the total idle either appears or doesn't.
- There is also no way AFAIK to associate the costs with a specific asset besides the share, so it may be that, depending on the idle sharing type, the total cost of an allocation view drill-down might be inconsistent with the asset view for that asset. E.g. if you share by cluster then breakdown by node, even if the undistributed idle is associated with that node, it will still be filtered out when drilling down.

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -752,8 +752,10 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	//     the output (i.e. they can be used to generate a valid key for
 	//     the given properties) then aggregate; otherwise... ignore them?
 	// 10. If the merge idle option is enabled, merge any remaining idle
-	//     allocations into a single idle allocation
-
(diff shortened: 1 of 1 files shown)


Sean-Holcomb (approved):
LGTM, Great Comments

PR #886, opened 2021-08-03
Omit PV Hourly Metrics for !Available and !Bound

mbolt35:
Fairly easy filter to avoid including PV Hourly Cost metrics when the status is anything other than Available or Bound

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -1381,8 +1381,8 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 
 			pvs := cmme.KubeClusterCache.GetAllPersistentVolumes()
 			for _, pv := range pvs {
-				// Omit pv_hourly_cost if the volume status is not available or unbound
-				if pv.Status.Phase != v1.VolumeAvailable && pv.Status.Phase != v1.VolumeBound {
+				// Omit pv_hourly_cost if the volume status is failed
+				if pv.Status.Phase == v1.VolumeFailed {
 					continue
 				}
 
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -1381,6 +1381,11 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 
 			pvs := cmme.KubeClusterCache.GetAllPersistentVolumes()
 			for _, pv := range pvs {
+				// Omit pv_hourly_cost if the volume status is not available or unbound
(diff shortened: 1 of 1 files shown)


nikovacevic:
Looks like a simple and reasonable solution. Have we been able to repro and test that it does, in fact, filter out the PVs we're targeting? Or do we need a more broken env to test on?

AjayTripathy:
The solution seems right but I do think we want to charge for Unbound PVs? I think we literally just don't want to charge for the failed state, as unbound by definition I believe implies that you _are_ paying. Failure is poorly defined, so we don't know if you're paying or not.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#phase

mbolt35:
> The implementation seems right but I do think we want to charge for Unbound PVs? I think we literally just don't want to charge for the failed state, as unbound by definition I believe implies that you _are_ paying. Failure is poorly defined, so we don't know if you're paying or not.
> 
> https://kubernetes.io/docs/concepts/storage/persistent-volumes/#phase

You are correct, and this logic only filters un

> The implementation seems right but I do think we want to charge for Unbound PVs? I think we literally just don't want to charge for the failed state, as unbound by definition I believe implies that you _are_ paying. Failure is poorly defined, so we don't know if you're paying or not.
> 
> https://kubernetes.io/docs/concepts/storage/persistent-volumes/#phase

So if it's not bound, it's available? It's unclear to me what each phase maps to in terms of provider disks. I, maybe naively, assumed Available meant without a claim, bound meant claimed, released meant deleted, and failed meant non existent. These definitions don't add much clarity, but I believe we should simply add !Released (this has been unclaimed, but isn't yet Available). It's difficult to assume what Failed means, but this state only really seems reasonable if the disk failed to create. Is that a safe bet?

mbolt35:
> Looks like a simple and reasonable solution. Have we been able to repro and test that it does, in fact, filter out the PVs we're targeting? Or do we need a more broken env to test on?

Good question concerning PVs we're targetting - I don't really understand the context in which we are adding this change unless there is a report specifically outlining the evidence it exists. However, logically, this seems reasonable -- if there are solid reasons to believe this could create unexpected results, I am definitely tabling it.

AjayTripathy (approved):

AjayTripathy:
> These definitions don't add much clarity, but I believe we should simply add !Released (this has been unclaimed, but isn't yet Available).

I think that's reasonable too, but isn't !Failed logically equivalent? I also think we want to _blacklist_ certain states, not _whitelist_ in this case, because this is based purely on observational data (I noted that many failed PVs were not being charged for on flavors of AKS)

PR #891, opened 2021-08-09
rewrite group-right query for simplicity and performance

AjayTripathy:
This removes the group-right query without adding any additional queries. It vastly simplifies our PV build to just be 
cost per GB/hour * GB * hours.

Testing done: run both builds and note a very small rounding difference between the two:

prod:
Screen Shot 2021-08-08 at 9 41 05 PM

this PR:
Screen Shot 2021-08-08 at 9 44 16 PM

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -228,36 +228,6 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, duration, o
 		diskMap[key].Bytes = bytes
 	}
 
-	for _, result := range resPVCost {
-		cluster, err := result.GetString(env.GetPromClusterLabel())
-		if err != nil {
-			cluster = env.GetClusterID()
-		}
-
-		name, err := result.GetString("persistentvolume")
-		if err != nil {
-			log.Warningf("ClusterDisks: PV cost data missing persistentvolume")
-			continue
-		}
-
-		// TODO niko/assets storage class
-
-		cost := result.Values[0].Value
-		key := fmt.Sprintf("%s/%s", cluster, name)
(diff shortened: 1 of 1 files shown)


dwbrown2:
This feels like something we should test very carefully...

pkg/costmodel/cluster.go
```diff
@@ -1151,3 +1065,96 @@ func ClusterCostsOverTime(cli prometheus.Client, provider cloud.Provider, startS
...
+
+		name, err := result.GetString("persistentvolume")
+		if err != nil {
+			log.Warningf("ClusterDisks: PV cost data missing persistentvolume")
+			continue
+		}
+
+		// TODO niko/assets storage class
+
+		cost := result.Values[0].Value
```
nikovacevic, line 1145: Is it worth checking the length here?

nikovacevic: My only other thought is that units might be nice. Like `costPerGBHr` so that the arithmetic below is more clear.

nikovacevic (approved):
Looks right to me. What is our most unusual set-up in terms of PVs? Or do we have a user we know to have an unorthodox PV set-up? I feel relatively confident that this will work for 99% of users, but agree with Webb that it would be good to tread carefully here.

AjayTripathy:
> Looks right to me. What is our most unusual set-up in terms of PVs? Or do we have a user we know to have an unorthodox PV set-up? I feel relatively confident that this will work for 99% of users, but agree with Webb that it would be good to tread carefully here.

I guess "most unusual" would be PVs of type __unallocated__ ... but because this is an assets query I don't think much would change.

PR #892, opened 2021-08-09
Fix overreporting gpu requests for vgpus

kaelanspatel:
Ref: https://github.com/kubecost/cost-model/issues/884

Changes to vgpu logic to prevent over-allocation. Now, if there are vgpu requests, we divide gpu count by a coefficient, which is the total number of allocable vgpus in the `aws-virtual-gpu-device-plugin-daemonset` to get the proportion of physical gpu used.

If the vgpu count can't be parsed, the coefficient defaults to `10.0`, the default [here](https://github.com/awslabs/aws-virtual-gpu-device-plugin/blob/master/manifests/device-plugin.yml).

Tested locally on an aws cluster with multiple replicas of pods requesting vgpus and observing `\model\metrics` endpoint as well as changes on frontend.

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -311,6 +311,17 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 		log.Warningf("ComputeCostData: continuing despite error parsing normalization values from %s: %s", queryNormalization, err.Error())
 	}
 
+	// Determine if there are vgpus configured and if so get the total allocatable number
+	// If there are no vgpus, the coefficient is set to 1.0
+	vgpuCount, err := getAllocatableVGPUs(cm.Cache)
+	vgpuCoeff := 1.0
+	if err != nil {
+		log.Warningf("ComputeCostData: unable to set allocable vgpus from daemonset: " + err.Error())
+	}
+	if vgpuCount > 0.0 {
+		vgpuCoeff = vgpuCount
+	}
+
 	nodes, err := cm.GetNodeCost(cp)
 	if err != nil {
 		log.Warningf("GetNodeCost: no node cost model available: " + err.Error())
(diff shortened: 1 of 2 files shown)


pkg/costmodel/costmodel.go
```diff
@@ -928,6 +940,15 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 	nodeList := cm.Cache.GetAllNodes()
 	nodes := make(map[string]*costAnalyzerCloud.Node)
 
+	vgpuCount, err := getAllocatableVGPUs(cm.Cache)
+	vgpuCoeff := 1.0
```
AjayTripathy, line 944: Shouldn't this default to 10 since that's the AWS default?

pkg/costmodel/costmodel.go
```diff
@@ -2193,6 +2214,30 @@ func getStatefulSetsOfPod(pod v1.Pod) []string {
...
+	vgpuCount := 0.0
+	for _, ds := range daemonsets {
+		dsContainerList := &ds.Spec.Template.Spec.Containers
+		for _, ctnr := range *dsContainerList {
+			if ctnr.Args != nil {
+				for _, arg := range ctnr.Args {
+					if strings.Contains(arg, "--vgpu=") {
+						vgpus, err := strconv.ParseFloat(arg[strings.IndexByte(arg, '=')+1:], 64)
+						if err != nil {
+							return 10.0, fmt.Errorf("cannot parse vgpu allocation string %s: %v", arg, err)
```
AjayTripathy, line 2228: I think you'd rather continue here. Just because one node has an improperly formatted vgpu setting doesn't mean they all do.

pkg/costmodel/costmodel.go
```diff
@@ -928,6 +940,15 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 	nodeList := cm.Cache.GetAllNodes()
 	nodes := make(map[string]*costAnalyzerCloud.Node)
 
+	vgpuCount, err := getAllocatableVGPUs(cm.Cache)
```
AjayTripathy, line 943: This can be an expensive operation, I suspect you only want to compute this for nodes claiming to have a vGPU.

kaelanspatel: AFAIK `GetNodeCost` doesn't run on every node, so this would run once and merely apply it to every node requesting a vGPU. Is that not the case? I was under the impression that this function is called only two single times (once in `GetNodeCost` and once in `ComputeCostData`) in `CostModelMetricsEmitter.Start()`.

AjayTripathy: ComputeCostData is itself called in a loop for metrics emission.

pkg/costmodel/costmodel.go
kaelanspatel: Yes, fixed.

pkg/costmodel/costmodel.go
AjayTripathy: after sync-ing offline, this operation is probably fine so let's not prematurely optimize.

pkg/costmodel/costmodel.go
kaelanspatel: Fixed.

AjayTripathy (approved):

PR #893, opened 2021-08-16
Fix label alias aggregation on sanitized labels

nikovacevic:
## Changes
- In `AggregateBy`, sanitize the label name behind the alias before finding a label match in `Allocation` properties.

## Testing
- Unit tests
- Manual testing (see below)

### Manual testing
Changed Product to `app.kubernetes.io/name` and tested that aggregating by Product works:
Screenshot from 2021-08-16 10-56-51
Screenshot from 2021-08-16 10-56-39

### DIFF
--- a/pkg/kubecost/allocation_test.go
+++ b/pkg/kubecost/allocation_test.go
@@ -503,6 +503,34 @@ func TestAllocationSet_generateKey(t *testing.T) {
 	if key != "dept1/envt1/ownr1/prod1/team1" {
 		t.Fatalf("generateKey: expected \"dept1/envt1/ownr1/prod1/team1\"; actual \"%s\"", key)
 	}
+
+	// Ensure that labels with illegal Prometheus characters in LabelConfig
+	// still match their sanitized values.
+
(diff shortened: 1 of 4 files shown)


pkg/kubecost/allocation_test.go
```diff
@@ -503,6 +503,34 @@ func TestAllocationSet_generateKey(t *testing.T) {
...
+
+	alloc.Properties = &AllocationProperties{
+		Cluster:   "cluster1",
+		Namespace: "namespace1",
+		Labels: map[string]string{
+			"prom_illegal_department": "dept1",
+			"env":                     "envt1",
+			"_owner_":                 "ownr1",
+		},
+	}
```
nikovacevic, line 522: @AjayTripathy does this look like a reasonable set of test scenarios?

AjayTripathy: I think you want one test case with a "." in it, like app.kubernetes/io

AjayTripathy (approved):

PR #894, opened 2021-08-16
Alloc.add handles changes in only controller name

michaelmdresser:
The new `AddDifferentController` test failed before the change to
Allocation.add and outlines the situation. We want to be able
to handle a situation that the Kubernetes spec does not expect
but occurs in the wild: a pod whose controller's name changes
but nothing else does. We'd like the two _technically_ separate
allocations that this generates to be added as if they were the
same without empty-stringing the controller name field. We opted
to pick the alphabetically first string but another strategy
could be used instead.

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -3,6 +3,7 @@ package kubecost
 import (
 	"bytes"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -601,8 +602,37 @@ func (a *Allocation) add(that *Allocation) {
 		log.Warningf("Allocation.AggregateBy: trying to add a nil receiver")
 		return
 	}
+
+	// Save the original properties so we can set a controller name in the event of
+	// equality in all fields except the controller name
+	leftProperties := a.Properties
+	rightProperties := that.Properties
+
 	// Preserve string properties that are matching between the two allocations
 	a.Properties = a.Properties.Intersection(that.Properties)
+
+	// Overwrite regular intersection logic for the controller name property in the
+	// case that all properties except controller name are the same.
+	if leftProperties != nil &&
+		rightProperties != nil &&
+		leftProperties.Controller != rightProperties.Controller &&
+		leftProperties.Controller != "" &&
+		rightProperties.Controller != "" {
+
+		lpNoController := leftProperties.Clone()
+		lpNoController.Controller = ""
+		rpNoController := rightProperties.Clone()
+		rpNoController.Controller = ""
+
+		if lpNoController.Equal(rpNoController) {
+			controllers := []string{
+				leftProperties.Controller,
+				rightProperties.Controller,
+			}
+			sort.Strings(controllers)
+			a.Properties.Controller = controllers[0]
+		}
+	}
 
 	// Expand the window to encompass both Allocations
 	a.Window = a.Window.Expand(that.Window)
--- a/pkg/kubecost/allocation_test.go
+++ b/pkg/kubecost/allocation_test.go
@@ -355,6 +355,30 @@ func TestAllocation_Share(t *testing.T) {
 	}
 }
 
+func TestAllocation_AddDifferentController(t *testing.T) {
+	a1 := &Allocation{
+		Properties: &AllocationProperties{
+			Container:  "container",
+			Pod:        "pod",
+			Namespace:  "ns",
+			Cluster:    "cluster",
+			Controller: "controller 1",
+		},
+	}
+	a2 := a1.Clone()
+	a2.Properties.Controller = "controller 2"
+
+	result, err := a1.Add(a2)
+	if err != nil {
+		t.Fatalf("Allocation.Add: unexpected error: %s", err)
+	}
+
+	if result.Properties.Controller == "" {
+		t.Errorf("Adding allocations whose properties only differ in controller name should not result in an empty string controller name.")
+	}
+
+}
+
 func TestAllocation_MarshalJSON(t *testing.T) {
 	start := time.Date(2021, time.January, 1, 0, 0, 0, 0, time.UTC)
 	end := time.Date(2021, time.January, 2, 0, 0, 0, 0, time.UTC)


pkg/kubecost/allocation.go
```diff
@@ -601,9 +602,38 @@ func (a *Allocation) add(that *Allocation) {
...
 	// Preserve string properties that are matching between the two allocations
 	a.Properties = a.Properties.Intersection(that.Properties)
 
+	// Overwrite regular intersection logic for the controller name property in the
+	// case that all properties except controller name are the same.
+	if leftProperties != nil &&
+		rightProperties != nil &&
+		leftProperties.Controller != rightProperties.Controller &&
+		leftProperties.Controller != "" &&
+		rightProperties.Controller != "" {
```
nikovacevic, line 620: I don't know the answer here, but are we certain that we want to do this only when _both_ Controllers are `""`?

nikovacevic: i.e. when one is empty and the other is not, we're happy with the current logic?

pkg/kubecost/allocation.go
```diff
@@ -601,9 +602,38 @@ func (a *Allocation) add(that *Allocation) {
...
+		rpNoController.Controller = ""
+
+		if lpNoController.Equal(rpNoController) {
+			controllers := []string{
+				leftProperties.Controller,
+				rightProperties.Controller,
+			}
+			sort.Strings(controllers)
+			a.Properties.Controller = controllers[0]
+		}
```
nikovacevic, line 634: I'm slightly worried that this will happen a _lot_ (i.e. the general case of adding Allocations where Controllers don't match is very common) and that Cloning both sets of properties, then doing a deep comparison will slow down `add` considerably. Do we have any notion of how much this will cost?

nikovacevic: In this analysis, I'm not counting the bit inside the `if` statement, assuming that that case will not happen very often.

pkg/kubecost/allocation.go
michaelmdresser: That's a good call out. I think it's reasonable to also do this when only one is empty string but we should think about edge cases. I don't have time to do so today, but I can check it out in the morning.

michaelmdresser: (if we do include the empty string case, note to self to update the sort logic so that we don't pick the empty string)

pkg/kubecost/allocation.go
michaelmdresser: That did occur to me. I'm really not sure what this will cost and that worries me, too.

We could rewrite this to compare just the properties we care about (`cluster`, `namespace`, `pod`, `container`) which would could avoid the cloning and `.Equal` too if we just keep a reference to the original `a.Properties` (aka `leftProperties`).

AjayTripathy: Could the top level if check the keys or podName? The case where the key of two containers matches but the controllers do not should be relatively rare so we may just want a special case for that

michaelmdresser: Some rough benchmarking is below. It only covers the overlapping case (where we enter the if statement).

## The benchmarks (same for benching each version of `.add`)

```go
func BenchmarkAllocation_addDifferentController(b *testing.B) {
	a1 := &Allocation{
		Properties: &AllocationProperties{
			Container:  "container",
			Pod:        "pod",
			Namespace:  "ns",
			Cluster:    "cluster",
			Controller: "controller 1",
			Services:   []string{"svc1", "svc2"},
			Labels: map[string]string{
				"label1":            "value1",
				"label2":            "value2",
				"label3":            "value3",
				"label4":            "value4",
				"label5":            "value5",
				"label6":            "value6",
				"label7":            "super long value 7",
				"label very long 8": "wow incredibly yep long wow value 8",
			},
		},
	}
	a2 := a1.Clone()
	a2.Properties.Controller = "controller 2"

	for i := 0; i < b.N; i++ {
		a1Again := a1.Clone()
		_, err := a1Again.Add(a2)
		if err != nil {
			b.Fatalf("unexpected error adding: %s", err)
		}
	}
}
```

```go
func BenchmarkAllocation_addDifferentControllerFewLabels(b *testing.B) {
	a1 := &Allocation{
		Properties: &AllocationProperties{
			Container:  "container",
			Pod:        "pod",
			Namespace:  "ns",
			Cluster:    "cluster",
			Controller: "controller 1",
			Services:   []string{"svc1", "svc2"},
			Labels: map[string]string{
				"label1": "value1",
				"label2": "value2",
				"label3": "value3",
			},
		},
	}
	a2 := a1.Clone()
	a2.Properties.Controller = "controller 2"

	for i := 0; i < b.N; i++ {
		a1Again := a1.Clone()
		_, err := a1Again.Add(a2)
		if err != nil {
			b.Fatalf("unexpected error adding: %s", err)
		}
	}
}
```

```go
func BenchmarkAllocation_addDifferentControllerManyLabels(b *testing.B) {
	a1 := &Allocation{
		Properties: &AllocationProperties{
			Container:  "container",
			Pod:        "pod",
			Namespace:  "ns",
			Cluster:    "cluster",
			Controller: "controller 1",
			Services:   []string{"svc1", "svc2"},
			Labels:     map[string]string{},
		},
	}

	for i := 0; i < 500; i++ {
		a1.Properties.Labels[fmt.Sprintf("label%d", i)] = fmt.Sprintf("value%d", i)
	}
	a2 := a1.Clone()
	a2.Properties.Controller = "controller 2"

	for i := 0; i < b.N; i++ {
		a1Again := a1.Clone()
		_, err := a1Again.Add(a2)
		if err != nil {
			b.Fatalf("unexpected error adding: %s", err)
		}
	}
}
```

## Versions of `add` benchmarked

### Key generation (`add`)

```go
	// Generate keys for each allocation to allow for special logic to set the controller
	// in the case of keys matching but controllers not matching.
	aggByForKey := []string{"cluster", "node", "namespace", "pod", "container"}
	leftKey := a.generateKey(aggByForKey, nil)
	rightKey := a.generateKey(aggByForKey, nil)
	leftProperties := a.Properties
	rightProperties := that.Properties

	// Preserve string properties that are matching between the two allocations
	a.Properties = a.Properties.Intersection(that.Properties)

	// Overwrite regular intersection logic for the controller name property in the
	// case that the Allocation keys are the same but the controllers are not.
	if leftKey == rightKey &&
		leftProperties != nil &&
		rightProperties != nil &&
		leftProperties.Controller != rightProperties.Controller {
		if leftProperties.Controller == "" {
			a.Properties.Controller = rightProperties.Controller
		} else if rightProperties.Controller == "" {
			a.Properties.Controller = leftProperties.Controller
		} else {
			controllers := []string{
				leftProperties.Controller,
				rightProperties.Controller,
			}
			sort.Strings(controllers)
			a.Properties.Controller = controllers[0]
		}
	}
```

### Initial commit, clone and equality (`add2`)

```go
	// Save the original properties so we can set a controller name in the event of
	// equality in all fields except the controller name
	leftProperties := a.Properties
	rightProperties := that.Properties

	// Preserve string properties that are matching between the two allocations
	a.Properties = a.Properties.Intersection(that.Properties)

	// Overwrite regular intersection logic for the controller name property in the
	// case that all properties except controller name are the same.
	if leftProperties != nil &&
		rightProperties != nil &&
		leftProperties.Controller != rightProperties.Controller &&
		leftProperties.Controller != "" &&
		rightProperties.Controller != "" {

		lpNoController := leftProperties.Clone()
		lpNoController.Controller = ""
		rpNoController := rightProperties.Clone()
		rpNoController.Controller = ""

		if lpNoController.Equal(rpNoController) {
			controllers := []string{
				leftProperties.Controller,
				rightProperties.Controller,
			}
			sort.Strings(controllers)
			a.Properties.Controller = controllers[0]
		}
	}

```

### Custom if (`add3`)

```go
	leftProperties := a.Properties
	rightProperties := that.Properties

	// Preserve string properties that are matching between the two allocations
	a.Properties = a.Properties.Intersection(that.Properties)

	// Overwrite regular intersection logic for the controller name property in the
	// case that the Allocation keys are the same but the controllers are not.
	if leftProperties != nil && rightProperties != nil &&
		leftProperties.Pod == rightProperties.Pod &&
		leftProperties.Node == rightProperties.Node &&
		leftProperties.Cluster == rightProperties.Cluster &&
		leftProperties.Namespace == rightProperties.Namespace &&
		leftProperties.Container == rightProperties.Container &&
		leftProperties.Controller != rightProperties.Controller {
		if leftProperties.Controller == "" {
			a.Properties.Controller = rightProperties.Controller
		} else if rightProperties.Controller == "" {
			a.Properties.Controller = leftProperties.Controller
		} else {
			controllers := []string{
				leftProperties.Controller,
				rightProperties.Controller,
			}
			sort.Strings(controllers)
			a.Properties.Controller = controllers[0]
		}
	}
```

## Results
```
→ go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/kubecost/cost-model/pkg/kubecost
cpu: AMD Ryzen 9 5950X 16-Core Processor            
BenchmarkAllocation_addDifferentController-32               	 582687	     2760 ns/op
BenchmarkAllocation_add2DifferentController-32              	 312595	     3261 ns/op
BenchmarkAllocation_add3DifferentController-32              	 696004	     2011 ns/op
BenchmarkAllocation_addDifferentControllerFewLabels-32      	 467998	     2531 ns/op
BenchmarkAllocation_add2DifferentControllerFewLabels-32     	 409444	     2917 ns/op
BenchmarkAllocation_add3DifferentControllerFewLabels-32     	 821022	     1617 ns/op
BenchmarkAllocation_addDifferentControllerManyLabels-32     	   7105	   145035 ns/op
BenchmarkAllocation_add2DifferentControllerManyLabels-32    	   3691	   312633 ns/op
BenchmarkAllocation_add3DifferentControllerManyLabels-32    	   8056	   138765 ns/op
```

## Conclusion

The custom `if` (`add3`) is naturally more performant and predictable, but it has a brittleness that the key generation method (`add`) does not. I would prefer either over the cloning/equality method (`add2`) mainly because I would be worried about unpredictable performance of the `.Clone` in high-label-cardinality situations.

After the discussion above (covering `add2`), this branch is currently using the key generation method (`add`) as of commit 2b993e3.

AjayTripathy (approved):

nikovacevic (approved):
Thanks for adding the benchmarking, Michael. I like this solution. Avoiding the extra Clone calls has the advantage of saving us from memory allocation, too, so I think this looks very reasonable and the benchmarking bears that out.

PR #912, opened 2021-08-31
configurable configmapnames

AjayTripathy:
## What does this PR change?
Makes the configmapname configurable

No

Allows users to set configmap name instead of hardcoding it.

## How was this PR tested?
See
https://github.com/kubecost/cost-analyzer-helm-chart/pull/1052


nikovacevic (approved):
LGTM

PR #913, opened 2021-09-01
Add prometheus metrics endpoint

Sean-Holcomb:
## What does this PR change?
This PR exposes the `/diagnostics/prometheusMetrics` endpoint which runs a series of prometheus queries which were previously being run on the front end which verify that prometheus metrics required by Kubecost are being emitted.

- https://github.com/kubecost/cost-analyzer-frontend/pull/374

Exposes endpoint for verifying that prometheus metrics are properly emitting 

## How was this PR tested?
This PR was manually tested on a federated cluster


pkg/prom/diagnostics.go
```diff
@@ -66,3 +66,109 @@ func LogPrometheusClientState(client prometheus.Client) {
...
+			Label:       "cAdvsior metrics available",
+			Description: "Determine if cAdvisor metrics are available during last 5 minutes.",
+			DocLink:     fmt.Sprintf("%s#cadvisor-metrics-available", docs),
+		},
+		{
+			ID:          "ksmMetric",
+			Query:       fmt.Sprintf(`absent_over_time(kube_pod_container_resource_requests{resource="memory", unit="byte"}[5m]  %s)`, offset),
+			Label:       "Kube-state-metrics available",
+			Description: "Determine if metrics from kube-state-metrics are available during last 5 minutes.",
+			DocLink:     fmt.Sprintf("%s#kube-state-metrics-metrics-available", docs),
```
michaelmdresser, line 88: @mbolt35 with the changes to metric emission/necessity of KSM, is this still the right metric to be checking? Do you think its the right description?

pkg/prom/diagnostics.go
```diff
@@ -66,3 +66,109 @@ func LogPrometheusClientState(client prometheus.Client) {
...
+		{
+			ID:          "neMetric",
+			Query:       fmt.Sprintf(`absent_over_time(node_cpu_seconds_total[5m]  %s)`, offset),
+			Label:       "Node-exporter metrics available",
+			Description: "Determine if metrics from node-exporter are available during last 5 minutes.",
+			DocLink:     fmt.Sprintf("%s#node-exporter-metrics-available", docs),
+		},
+		{
+			ID:          "recordingMetric",
+			Query:       fmt.Sprintf(`absent_over_time(node_cpu_seconds_total[5m]  %s)`, offset),
```
michaelmdresser, line 105: I don't think this is the right query for our recording rules. I think (?) these are our recording rules: https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L412

pkg/prom/diagnostics.go
```diff
@@ -66,3 +66,109 @@ func LogPrometheusClientState(client prometheus.Client) {
...
+			Label:       "Recording rules are available",
+			Description: "Determine if metrics defined by kubecost recording rules are available during last 5 minutes.",
+			DocLink:     fmt.Sprintf("%s#recording-rules-are-available", docs),
+		},
+		{
+			ID:          "cadvisorLabel",
+			Query:       fmt.Sprintf(`absent_over_time(container_cpu_usage_seconds_total{container_name!="",pod_name!="",instance!~"10.*|192.*|172.*"}[5m]  %s)`, offset),
```
michaelmdresser, line 112: @AjayTripathy did you make changes to the cAdvisor label setup? I don't remember, but I do recall it was mentioned at some point. Want to make sure this is still an appropriate check.

pkg/prom/diagnostics.go
```diff
@@ -66,3 +66,109 @@ func LogPrometheusClientState(client prometheus.Client) {
...
+			Label:       "Expected cAdvsior labels available",
+			Description: "Determine if expected cAdvisor labels are present during last 5 minutes.",
+			DocLink:     fmt.Sprintf("%s#cadvisor-metrics-available", docs),
+		},
+		{
+			ID:          "ksmVersion",
+			Query:       fmt.Sprintf(`absent_over_time(kube_persistentvolume_capacity_bytes[5m]  %s)`, offset),
+			Label:       "Expected kube-state-metrics version found",
+			Description: "Determine if metric in required kube-state-metrics version are present during last 5 minutes.",
+			DocLink:     fmt.Sprintf("%s#expected-kube-state-metrics-version-found", docs),
```
michaelmdresser, line 122: @mbolt35 same question w.r.t. this being the right check after our metric emission change.

pkg/prom/diagnostics.go
```diff
@@ -66,3 +66,109 @@ func LogPrometheusClientState(client prometheus.Client) {
...
+}
+
+// executePrometheusDiagnosticQuery executes a PrometheusDiagnostic query using the given context
+func (pd *PrometheusDiagnostic) executePrometheusDiagnosticQuery(ctx *Context) error {
+	resultCh := ctx.Query(pd.Query)
+	result, err := resultCh.Await()
+	if err != nil {
+		return fmt.Errorf("prometheus diagnostic %s failed with error: %s", pd.ID, err)
+	}
+    if result == nil {
```
michaelmdresser, line 168: This could use a `go fmt`

pkg/prom/diagnostics.go
```diff
@@ -66,3 +66,109 @@ func LogPrometheusClientState(client prometheus.Client) {
...
+    	result = []*QueryResult{}
+	}
+	pd.Result = result
+	pd.Passed = len(result) == 0
```
michaelmdresser, line 172: Why does a result length of 0 mean the check passed?

michaelmdresser (approved):
I love this change!! There's one thing that has to be changed (a wrong query, I think) and then I just had a few questions for others to confirm we're making the right queries. Overall, though, this looks excellent. Approving because the change should be minor.

pkg/prom/diagnostics.go
dwbrown2: Yes, this should be able to be removed. @AjayTripathy can confirm though.

pkg/prom/diagnostics.go
Sean-Holcomb: These queries are meant to come back empty most of which with the `absent_over_time` function.

pkg/prom/diagnostics.go
Sean-Holcomb: In general response to these comment. This particular PR is more of a port of what existed. The discussion about which metrics are incorrect is useful and I will use it later

pkg/prom/diagnostics.go
michaelmdresser: Ahh, gotcha that makes sense.

PR #915, opened 2021-09-02
clean log spam

AjayTripathy:
## What does this PR change?
Cleans log spam 

No

Easier debugging through log cleanups

## How was this PR tested?
Manually inspecting logs.


AjayTripathy:
Ah, this pretty much only spams if you forget to drop in a pricing key. I'd suggest we make this change as some people don't mind running in that mode to test.

michaelmdresser (approved):

michaelmdresser:
@AjayTripathy it looks like we've got conflicts here and haven't looked at this PR in a while. Would you like to update it, or should we just close it?

michaelmdresser:
Closing because stale

PR #916, opened 2021-09-02
Kubecost Resources StringTable + Staggered VersionSet Support

mbolt35:
# Update Generated Kubecost Resources

## What does this PR change?
Update bingen generation to v0.0.6, which allows defining version sets for allocation and assets.

Encoding of ETL resources will now use a base string table which is 20-30% less memory and faster on decode.

## How was this PR tested?
Staggering the versions on a test server, observe rebuilding, and benchmarked string table encode/decode via resource file test app.

### DIFF
--- a/pkg/kubecost/bingen.go
+++ b/pkg/kubecost/bingen.go
@@ -1,5 +1,10 @@
 package kubecost
 
+// Default Version Set (uses -version flag passed) includes shared resources
+// @bingen:generate:Window
+
+// Asset Version Set: Includes Asset pipeline specific resources
+// @bingen:set[name=Assets,version=15]
 // @bingen:generate:Any
 // @bingen:generate:Asset
 // @bingen:generate:AssetLabels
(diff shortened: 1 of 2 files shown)


pkg/kubecost/bingen.go
```diff
@@ -1,5 +1,10 @@
 package kubecost
 
+// Default Version Set (uses -version flag passed) includes shared resources
+// @bingen:generate:Window
```
michaelmdresser, line 4: Is there any danger of a version mismatch here in the future because asset and allocation both require `Window`?

michaelmdresser (approved):
I've only reviewed the gen calls and the first bit of the generated code, but it looks good.

pkg/kubecost/bingen.go
mbolt35: If you update window's version, assets and allocation will fail to load, and you'll basically get an error back from decoding causing a rebuild. Basically, each resource has it's own version.

michaelmdresser: Killer :+1:

PR #917, opened 2021-09-03
ClusterMapInfo Injects Local Cluster Info on Empty

mbolt35:
## What does this PR change?
On initial startup, or if cost-model had not been running for the last 5 minutes, the initial prometheus query to get cluster info would always return 0 data (since the cost-model emits this metric). To account for variance here, we always include the local cluster info in the event that the map is empty. 

* Bug Fix for improperly displaying cluster id/name labels

## How was this PR tested?
Turning off cost-model deployment for 10 minutes, starting up. Observing the bug no longer occurs.


michaelmdresser (approved):
Looks solid to me! Niko certainly knows this stuff better than I do, though.

PR #920, opened 2021-09-07
Fix AggregateCostModel's parsing of configured SharedOverhead

nikovacevic:
## What does this PR change?
- Implement `GetSharedOverheadCostPerMonth` as a clear and safe way to access the configured shared overhead field.

- When shared overhead is an empty string, the value will default to `0.0` instead of creating errors.

## How was this PR tested?
- Manually on gcp-niko

### Before
Screenshot from 2021-09-07 13-52-17

### After
Screenshot from 2021-09-07 13-48-37

Also confirmed that setting non-zero shared costs still works, then reset to $0.00 and confirmed that that still worked, too.
Screenshot from 2021-09-07 14-33-02

### DIFF
--- a/pkg/cloud/provider.go
+++ b/pkg/cloud/provider.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


kaelanspatel (approved):
LGTM

AjayTripathy (approved):

AjayTripathy:
Thanks, let's get this merged for the upcoming release!

PR #922, opened 2021-09-07
List AWS Reserved Instances

nealormsbee:
Minor adjustments to allow listing of active AWS Reserved Instances.

## What does this PR change?
- Adds a function receiver to AWSAccessKey type which returns the key values.
- Makes awsRegions variable public by renaming it to AWSRegions
- Adds v2 of the golang aws sdk

No

N/A

N/A

## How was this PR tested?
Manually, by deploying to an AWS cluster and seeing that API responses still return as normal

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -3,6 +3,7 @@ package cloud
 import (
 	"bytes"
 	"compress/gzip"
+	"context"
 	"encoding/csv"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -87,7 +91,7 @@ const SpotRefreshDuration = 15 * time.Minute
 
 const defaultConfigPath = "/var/configs/"
 
-var awsRegions = []string{
+var AWSRegions = []string{
```
nealormsbee, line 94: Made this variable public so that it is accessible from client packages.

michaelmdresser (approved):

nikovacevic (approved):

PR #924, opened 2021-09-08
Config Path Overlap Fixes

mbolt35:
## What does this PR change?
Removes all hard-coded paths which involve the default CONFIG_PATH environment variable, and replaces with the value from the env package. Since we do not allow the config path to be changed currently from the helm chart, there is currently no effect on the dependent code. However, as we push to build a more reliable configuration, these smaller fixes will become more valuable.

## How was this PR tested?
Helm deployment on AWS and observation.

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -43,14 +44,22 @@ import (
...
 const SpotInfoUpdateType = "spotinfo"
 const AthenaInfoUpdateType = "athenainfo"
 const PreemptibleType = "preemptible"
 
 const APIPricingSource = "Public API"
 const SpotPricingSource = "Spot Data Feed"
+const SpotRefreshDuration = 15 * time.Minute // How often spot data is refreshed
 const ReservedInstancePricingSource = "Savings Plan, Reserved Instance, and Out-Of-Cluster"
 
+func getAtenaConfigFilePath() string {
```
nikovacevic, line 59: Atena -> Athena :smile:

nikovacevic (changes requested):
Lil typo, then LGTM!

michaelmdresser:
Seems like a good change, @mbolt35 are you okay with resolving conflicts (or re-opening this PR) so it can be properly reviewed?

mbolt35:
This is no longer required - will probably be rolled into a bigger change at some point. Thanks for the quick review :)

PR #931, opened 2021-09-10
Fix issue with query

AjayTripathy:
## What does this PR change?
Fix regression where we relied on legacy cadvisor metrics in diagnostics.
Screen Shot 2021-09-09 at 5 25 23 PM


mbolt35 (approved):
LGTM

PR #935, opened 2021-09-13
Stop emission of spurious price values

nealormsbee:
## What does this PR change?
Modifies metric emission loop not to emit costs that are > 30x a running average of that cost for CPU, RAM, GPU.

No.

Prevents cost data from being corrupted by incorrect resource capacity emissions from the k8s API. See issue described here: https://github.com/kubecost/cost-model/issues/927

- Resolves https://github.com/kubecost/cost-model/issues/927

## How was this PR tested?
Deployed to a cluster, and mis-reports were simulated by giving metrics a high probability of emitting massive cost spikes.


pkg/costmodel/metrics.go
```diff
@@ -450,17 +464,67 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
+
+				avgCosts, ok := nodeCostAverages[labelKey]
+
+				// initialize average cost tracking for this node if there is none
+				if !ok {
+					avgCosts = NodeCostAverages{
+						CpuCostAverage:         cpuCost,
+						RamCostAverage:         ramCost,
+						GpuCostAverage:         gpuCost,
+						TotalCostAverage:       totalCost,
```
AjayTripathy, line 477: I don't think we have any evidence that totalcost is exploding, just ram and cpu unit costs when the k8s api reports odd data. Let's limit the radius of this change to known issues.

AjayTripathy (changes requested):

nealormsbee:
Updated! Still declining to record a `totalCost` datapoint if CPU or RAM datapoint is an outlier, since `totalCost` is defined in part by those values and would itself be an outlier. No longer keeping a running average of totalCost or GPU costs, though

pkg/costmodel/metrics.go
```diff
@@ -450,17 +460,52 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
-				cmme.NodeTotalPriceRecorder.WithLabelValues(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID).Set(totalCost)
+
+				const outlierFactor float64 = 30
+				// don't record cpuCost, ramCost, or gpuCost in the case of wild outliers
+				// k8s api sometimes causes cost spikes as described here:
+				// https://github.com/kubecost/cost-model/issues/927
+				if cpuCost < outlierFactor*avgCosts.CpuCostAverage {
+					cmme.CPUPriceRecorder.WithLabelValues(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID).Set(cpuCost)
+					avgCosts.CpuCostAverage = (avgCosts.CpuCostAverage*avgCosts.NumCpuDataPoints + cpuCost) / (avgCosts.NumCpuDataPoints + 1)
+					avgCosts.NumCpuDataPoints += 1
```
michaelmdresser, line 487: You are using `+=` but because `avgCosts` is not a pointer, this isn't actually updating the `nodeCostAverages` map. I think you should add a `nodeCostAverages[labelKey] = avgCosts` after performing these updates (either within each `if` block, or after all avgs have been handled.

Example: https://play.golang.org/p/tt8WGrExVLz

michaelmdresser (changes requested):

michaelmdresser:
Also, looks like you've got an unnecessary commit in this branch (not a big deal): 73b8da2

pkg/costmodel/metrics.go
nealormsbee: Oof. These scripting languages have ruined me

michaelmdresser (approved):

pkg/costmodel/metrics.go
michaelmdresser: :smile: perhaps not ruined. There's a world of language semantics discussion on mutable vs immutable values. It's just what you've gotten used to, especially given that we use pointers to structs _everywhere_ in the code.

nealormsbee:
I'm unable to merge here because Github still telling me Ajay requested changes. @AjayTripathy I think you'll have to come back and give this the ✔️ before it'll let me merge.

AjayTripathy (approved):

PR #937, opened 2021-09-14
Added StringTable Encoding/Decoding to AssetSet and AllocationSet

mbolt35:
## What does this PR change?
Enables StringTable encoding/decoding on AssetSet and AllocationSet to reduce file size and speed up IO.

* From this point forward, all ETL data will be encoding using a string table format which reduces file size 20-30% and speeds up encoding and decoding 20%

## How was this PR tested?
* Against scale cluster data, and with sandboxed bingen tests


nikovacevic (approved):
LGTM!

PR #938, opened 2021-09-16
Add provider regions

nealormsbee:
Adds a list of supported regions to each provider, and adds a `Regions()` getter to the provider interface. Makes awsRegions var private again.

## What does this PR change?
Providers now implement a `Regions()` method, which returns a list of valid regions.

No

N/A

N/A

## How was this PR tested?
Manually

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -73,6 +73,77 @@ var (
 	mtStandardN, _ = regexp.Compile(`^Standard_N[C|D|V]\d+r?[_v\d]*[_Promo]*$`)
 )
 
+var azureRegions = []string{
+	"eastus",
+	"eastus2",
+	"southcentralus",
+	"westus2",
+	"westus3",
+	"australiaeast",
+	"southeastasia",
+	"northeurope",
+	"swedencentral",
+	"uksouth",
+	"westeurope",
+	"centralus",
+	"northcentralus",
+	"westus",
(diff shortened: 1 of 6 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -73,6 +73,77 @@ var (
 	mtStandardN, _ = regexp.Compile(`^Standard_N[C|D|V]\d+r?[_v\d]*[_Promo]*$`)
 )
 
+var azureRegions = []string{
```
michaelmdresser, line 76: Would you include a comment explaining how to generate this list again in the event of Azure's regions changing?

pkg/cloud/gcpprovider.go
```diff
@@ -34,6 +34,37 @@ import (
 const GKE_GPU_TAG = "cloud.google.com/gke-accelerator"
 const BigqueryUpdateType = "bigqueryupdate"
 
+var gcpRegions = []string{
```
michaelmdresser, line 37: Same idea here with a comment.

michaelmdresser (approved):
Two small things, functionality looks good.

pkg/cloud/azureprovider.go
nealormsbee: Done! And glad you asked -- I had used a less-authoritative source initially and it was missing regions. New list is taken from the Azure CLI

PR #941, opened 2021-09-16
Ajay tripathy etl only mode [DRAFT]

AjayTripathy:
## What does this PR change?
This splits the deployment of the data gathering agent layer from the read layer from a permissions perspective.


mbolt35:
@AjayTripathy I have a really good idea of how to handle this case. It separates the configmap watching from the rest of the cluster cache (which I think it necessary anyways)

AjayTripathy:
That'd be awesome. We want to separate this to build some sort of HA-metrics emitter as well...

michaelmdresser:
I believe this functionality was addressed in https://github.com/kubecost/cost-model/pull/1188, so I'm going to close this PR.

PR #942, opened 2021-09-17
prevent cloud assets from being considered for reconciliation

Sean-Holcomb:
## What does this PR change
Update reconciliation match check to ignore cloud assets. Cloud Assets do not need to be reconciled with the CUR because that is their source, additionally this lack of filtering appears to be having a negative interaction with the reserved instances in the AWS provider. The enforced compute category of this query is causing partial matches to happen where full matches should be taking place, and inserting duplicate cloud assets into the Asset ETL 
Screen Shot 2021-09-16 at 5 47 38 PM

Faster Reconciliation, remove duplication

## How was this PR tested?
Tested on federated cluster

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -2727,7 +2727,7 @@ func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
 	for _, asset := range as.assets {
 		// Ignore cloud assets when looking for reconciliation matches
 		if asset.Type() == CloudAssetType {
-			continue;
+			continue
 		}
 		if k, err := key(asset, fullMatchProps); err != nil {
 			return nil, false, err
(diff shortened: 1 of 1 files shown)


pkg/kubecost/asset.go
```diff
@@ -2725,6 +2725,10 @@ func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
 
 	var providerIDMatch Asset
 	for _, asset := range as.assets {
+		// Ignore cloud assets when looking for reconciliation matches
+		if asset.Type() == CloudAssetType {
+			continue;
```
michaelmdresser, line 2730: Tiniest nit imaginable: I don't think the `;` is necessary.

michaelmdresser (approved):

pkg/kubecost/asset.go
Sean-Holcomb: I've been doing to much FE work the last couple weeks

nikovacevic (approved):
Great catch here, Sean.

PR #943, opened 2021-09-17
Add .gitignore for large AWS reserved instance offering file

nikovacevic:
This keeps the file `awsreservationofferings.json` out of git, but you still need to add it yourself to the `./configs` directory.

Required for https://github.com/kubecost/kubecost-cost-model/pull/539

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 ui/.cache
 ui/dist
 ui/node_modules/
+configs/awsreservationofferings.json


AjayTripathy:
Do you need this to build and run the application though? I think we should leave this in for build automation and simplicity even if it increases the size of the git repo?

nikovacevic:
The file is too big, so GH won't let me push it.

nikovacevic:
We either need to do this (https://docs.github.com/en/repositories/working-with-files/managing-large-files/configuring-git-large-file-storage) or, preferably, thin out all the stuff we don't need from that file and compress it so that we can just add to git.

nealormsbee:
@nikovacevic I think for now we can roll a compressed version. I'm retaining a master file with data for all of the offering permutations, but since the current implementation of RI recommendations disregards tenancy type, offering class, and a bunch of other things, I built a version that's just a bunch of key:value pairs of:

`{ "<region>/<instanceType>": lowestHourlyCostReservationForThatRegionAndInstanceTypePair }`

Still something to thing about because ofc as we start leveraging more information in the recommendation, the data needed will grow.

nealormsbee:
Closing this because the file size was reduced and file is added in https://github.com/kubecost/cost-model/pull/949

PR #944, opened 2021-09-17
Support Allocation GPU utilization/efficiency through integration with Nvidia GPU Operator/DCGM.

kaelanspatel:
## What does this PR change?
Changes the `Allocation` struct to features new fields `GPURequestAverage` and `GPUUsageAverage`, the later of which is supported through a Prometheus query for the Nvidia dcgm-exporter metric `DCGM_FI_DEV_GPU_UTIL`. Also adds `Allocation.GPUEfficiency()`, which calculates the ratio of GPU usage to requests, as with CPU and RAM efficiencies.

Notes
- These features will only work if the Nvidia `dcgm-exporter` pod is running on the node. Otherwise, the value `GPUUsageAverage` will be zero. This means that even if you are utilizing a GPU by allocating it, you may have zero usage. Setup is not entirely trivial and is detailed [here](https://docs.google.com/document/d/1fsbV55wTbpfWy4m9_FVETqHgtTMahfUa2w5MzV9gngc/edit?usp=sharing).

- https://github.com/kubecost/cost-analyzer-helm-chart/pull/1071

Adds support for integrating Nvidia DCGM GPU metrics into Kubecost and calculating Allocation GPU efficiency.

- 
- 

## How was this PR tested?
Updated `allocation_test.go` to feature checks for new `Allocation` fields, including efficiency. Additionally, created a gpu test load generator pod on cluster and observed metrics in Prometheus + through the api.

Prometheus shows `DCGM_FI_DEV_GPU_UTIL` being autodetected and whitelisted:
gpu_prom

`Allocation` for load-generating pod shows consistent  `GPURequestAverage` and `GPUUsageAverage`:
image

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -30,7 +30,7 @@ const (
 	queryFmtCPUUsageMax              = `max(rate(container_cpu_usage_seconds_total{container!="", container_name!="POD", container!="POD"}[%s]%s)) by (container_name, container, pod_name, pod, namespace, instance, %s)`
 	queryFmtGPUsRequested            = `avg(avg_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, %s)`
 	queryFmtGPUsAllocated            = `avg(avg_over_time(container_gpu_allocation{container!="", container!="POD", node!=""}[%s]%s)) by (container, pod, namespace, node, %s)`
-	queryFmtGPUUsageAvg              = `avg(rate(DCGM_FI_DEV_GPU_UTIL{container!=""}[%s]%s)) by (container, pod, namespace, %s)`
+	queryFmtGPUUsageAvg              = `avg(avg_over_time(DCGM_FI_DEV_GPU_UTIL{container!=""}[%s]%s)) by (container, pod, namespace, %s)`
 	queryFmtNodeCostPerCPUHr         = `avg(avg_over_time(node_cpu_hourly_cost[%s]%s)) by (node, %s, instance_type, provider_id)`
 	queryFmtNodeCostPerRAMGiBHr      = `avg(avg_over_time(node_ram_hourly_cost[%s]%s)) by (node, %s, instance_type, provider_id)`
 	queryFmtNodeCostPerGPUHr         = `avg(avg_over_time(node_gpu_hourly_cost[%s]%s)) by (node, %s, instance_type, provider_id)`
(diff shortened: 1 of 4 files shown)


pkg/costmodel/allocation.go
```diff
@@ -954,6 +960,33 @@ func applyGPUsAllocated(podMap map[podKey]*Pod, resGPUsRequested []*prom.QueryRe
 
 		hrs := pod.Allocations[container].Minutes() / 60.0
 		pod.Allocations[container].GPUHours = res.Values[0].Value * hrs
+		pod.Allocations[container].GPURequestAverage = res.Values[0].Value
```
michaelmdresser, line 963: Is this an actual average? Should we be querying this more like CPU/RAM request averages?

pkg/costmodel/allocation.go
```diff
@@ -954,6 +960,33 @@ func applyGPUsAllocated(podMap map[podKey]*Pod, resGPUsRequested []*prom.QueryRe
...
+		if err != nil {
+			log.DedupedWarningf(10, "CostModel.ComputeAllocation: GPU usage avg query result missing 'container': %s", key)
+			continue
+		}
+
+		if _, ok := pod.Allocations[container]; !ok {
+			pod.AppendContainer(container)
+		}
+
+		pod.Allocations[container].GPUUsageAverage = res.Values[0].Value * 0.01
```
michaelmdresser, line 989: Why is this multiplied by 0.01? A comment would be helpful.

michaelmdresser (commented):
Couple of comments. I'd also like to see a documentation update (allocation.md, probably) that highlights the new fields and why usage will be 0 (and efficiency 0) when the NVIDIA metrics are not available.

Also, @mbolt35 is this the correct way of version-bumping the schema now? I haven't had to do it since we split the Asset and Allocation versions.

pkg/costmodel/allocation.go
kaelanspatel: It is. Can you expand on this? To my knowledge it is being queried and handled like RAM requests.

pkg/costmodel/allocation.go
kaelanspatel: I'll add one. For posterity, it's because (for whatever reason) the "percentage" expressed by this metric is in whole numbers, rather than the decimals we're using for CPU/RAM i.e. 100% is `100`, not `1`.

pkg/costmodel/allocation.go
michaelmdresser: This function does some funky stuff with `queryFmtGPUsAllocated` -- look at the first `if` statement in this function, I think you _might_ run into trouble because requests is a separate array and can get overwritten by allocation.

However, I'm not super familiar with GPU. This might not be a problem at all, in which case :+1: 

Other resources (RAM, CPU, etc) have separate function for applying requests, usage, and average. We don't need to adhere to that style if this is correct, its just what tipped me off that this looked a little odd.

pkg/costmodel/allocation.go
michaelmdresser: Aha, of course. Excellent.

pkg/costmodel/allocation.go
kaelanspatel: Ah, yes I see. I believe this is consistent with what we do throughout, which is override requests with limits (which should also be what is happening in any pod using a GPU regardless; you're just supposed to have limits).

There may be problems with this particular function that I am not aware of, but the value of "GPUs allocated per allocation" right now is back-calculated from `GPUHours` anyway, so really all this change does is make that a separate field in Allocation.

PR #946, opened 2021-09-20
Metrics Exporter

mbolt35:
## What does this PR change?
* Refactors to reduce boilerplate required for startup 
    * Added `HTTPService` and `HTTPServices` interfaces as a prototype for future service based architecture
        * Moved `ClusterManager` (used to provided cluster definitions to frontend cluster select page) to an `HTTPService` implementation as it was the only API with 0 dependencies
    * Added a general API for the ConfigMap watcher API
    * Moved ScrapeInterval lookup for prometheus to a separate prom utility function.
* Kubecost-Metrics Exporter now includes all Kubecost metrics

* Separation of read and write components to allow for more robust deployments. 

## How was this PR tested?
* Multicluster setup involving separate read and write installs

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -11,8 +11,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/kubecost/cost-model/pkg/services"
 	"github.com/kubecost/cost-model/pkg/util/httputil"
 	"github.com/kubecost/cost-model/pkg/util/timeutil"
+	"github.com/kubecost/cost-model/pkg/util/watcher"
 
 	"k8s.io/klog"
 
@@ -22,7 +24,6 @@ import (
 
 	"github.com/kubecost/cost-model/pkg/cloud"
 	"github.com/kubecost/cost-model/pkg/clustercache"
-	cm "github.com/kubecost/cost-model/pkg/clustermanager"
 	"github.com/kubecost/cost-model/pkg/costmodel/clusters"
 	"github.com/kubecost/cost-model/pkg/env"
 	"github.com/kubecost/cost-model/pkg/errors"
@@ -32,9 +33,7 @@ import (
 	"github.com/kubecost/cost-model/pkg/thanos"
 	"github.com/kubecost/cost-model/pkg/util/json"
 	prometheus "github.com/prometheus/client_golang/api"
-	prometheusClient "github.com/prometheus/client_golang/api"
 	prometheusAPI "github.com/prometheus/client_golang/api/prometheus/v1"
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/patrickmn/go-cache"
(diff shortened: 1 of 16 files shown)


mbolt35:
Have a few unit tests that are not yet committed because of a few issues. Working on getting these fixed :)

pkg/services/services.go
```diff
@@ -0,0 +1,65 @@
...
+// HTTPService defines an implementation prototype for an object capable of registering
+// endpoints on an http router which provide an service relevant to the cost-model.
+type HTTPService interface {
+	// Register assigns the endpoints and returns an error on failure.
+	Register(*httprouter.Router) error
+}
+
+// HTTPServices defines an implementation prototype for an object capable of managing and registering
+// predefined HTTPService routes
+type HTTPServices interface {
```
michaelmdresser, line 19: Why should this be an interface and not just a struct?

pkg/services/services.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	dhs.Lock()
+	defer dhs.Unlock()
+
+	dhs.services = append(dhs.services, service)
+}
+
+// RegisterAll registers all the services added with the provided router
+func (dhs *defaultHTTPServices) RegisterAll(router *httprouter.Router) error {
```
michaelmdresser, line 46: If this can't return an error, why require it?

pkg/services/services.go
```diff
@@ -0,0 +1,65 @@
...
+
+// NewCostModelServices creates an HTTPServices implementation containing any predefined
+// http services used with the cost-model
+func NewCostModelServices() HTTPServices {
+	return &defaultHTTPServices{
+		services: []HTTPService{
+			NewClusterManagerService(),
+		},
+	}
+}
```
michaelmdresser, line 65: Is there a good way to document this new pattern? Something like "all new services should be `HTTPService`s and they should be added to the default set of services created by `NewCostModelServices`". I know we don't really have good dev documentation right now, but this is something that could go in a `CONTRIBUTING.md` or `DEVELOPMENT.md`.

pkg/costmodel/router.go
```diff
@@ -1004,59 +971,40 @@ func Initialize(additionalConfigWatchers ...ConfigWatchers) *Accesses {
 
 	timeout := 120 * time.Second
 	keepAlive := 120 * time.Second
-	scrapeInterval, _ := time.ParseDuration("1m")
```
michaelmdresser, line 1007: This is... interesting. Any idea why it was like this before? Legacy from parsing a string from a config or Prom?

michaelmdresser (approved):
LGTM from a high-level code standpoint. There are parts of this that I'm not intimately familiar with, but I trust that the reshuffling of components was tested thoroughly.

A few tiny questions, but nothing blocking.

pkg/services/services.go
mbolt35: Always good practice to create a contract for a general piece of architecture sitting at the top level. Also, should any other API choose to incorporate an implementation, it makes writing tests a whole lot easier.

pkg/services/services.go
mbolt35: Because I designed a contract that expects this behavior for long term. It doesn't yet, but it will.

pkg/services/services.go
mbolt35: I'm going to defer immediate documentation for this specific design for a few reasons:

1. I don't have any expectations (currently) of enforcing this pattern on any existing http handlers delegating an internal API. The reason why is that the one service I added was an API that stuck out as odd in the router and in the pkg. It appears to be a modifiable resource (which it was designed that way initially), but it was written down to just configmap proxy (sadly). It's a unique sort of self-contained API that will likely be used in our stack later.
2. This particular API is something I am imagining being something on our hosted solution (real REST implementation with actual stateful resources). 

So the reasoning behind the shifting around here is to remove the black sheep API from front and center, and to prototype a general RESTful http service springboard for the future. Hope that all makes sense :)

pkg/costmodel/router.go
mbolt35: I think it was written in a hurry or at the end of the day. Either way, I was happy to improve it :)

PR #949, opened 2021-09-22
Added minimized aws reservation data file

nealormsbee:
## What does this PR change?
Adds a json file with AWS reserved instance offering data

No

Allows AWS RI offerings to be compared against On-Demand counterparts

N/A

## How was this PR tested?
Manually


nealormsbee:
@nikovacevic this would obviate https://github.com/kubecost/cost-model/pull/943 because the file has now been made small enough to actually store in git, whereas previously it was too large.

nikovacevic (approved):

PR #951, opened 2021-09-22
Fix CVE for set-value

mbolt35:
## What does this PR change?
Updates `set-value` transitive dependencies to use 4.0.1

## How was this PR tested?
Need assistance here. @nealormsbee Do we have anything automated here? How to best test this update?


nealormsbee (approved):
LGTM. The project still builds fine and this gets the vulnerable version out of our project.

nealormsbee:
@mbolt35 I guess to answer questions about how to best test:

I ran this:

```bash
> npm list set-value
cost-analyzer-frontend@1.87.2 /Users/neal/code/kubecost/cost-analyzer-frontend
└─┬ parcel-bundler@1.12.4
  └─┬ micromatch@3.1.10
    └─┬ snapdragon@0.8.2
      └─┬ base@0.11.2
        └─┬ cache-base@1.0.1
          ├── set-value@2.0.1 
          └─┬ union-value@1.0.1
            └── set-value@2.0.1  deduped
```

so it's a sub-dependency of our build tool, parcel. From there all I did was ensure that the project still builds with the updated version. I don't know of a good way to do deeper testing that isn't totally insane...like getting `cache-base@1.0.1`, learning about what it does and how it uses `set-value`, and then testing for compatibility with the new `set-value` version.

Ideally we'd just see all of these updated regularly as Parcel releases updates, but Parcel v1 is end of life and not maintained. I think we're stuck with this relatively high-level testing approach until we migrate to a different build tool like Parcel v2 (alpha) or Webpack.

PR #952, opened 2021-09-22
Fill out Pricing Source Api for Azure provider for rate card api

Sean-Holcomb:
## What does this PR change?
This PR fills out the PricingSourceStatus api for the Azure Provider so that information about the state of the connection of the Rate Card API can be accessed

- https://github.com/kubecost/cost-analyzer-frontend/pull/402

Azure Users will now be able to view the status of their Rate Card API connection from the Diagnostic page

## How was this PR tested?
This PR was tested on the an Azure cluster, but modifying the configuration values for the Rate Card API on the settings page, to generate failing states of the Rate Card API

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -382,7 +382,7 @@ type Azure struct {
 	Clientset               clustercache.ClusterCache
 	Config                  *ProviderConfig
 	ServiceAccountChecks    map[string]*ServiceAccountCheck
-	RateCardPricingStatus   string
+	RateCardPricingError    error
 }
 
 type azureKey struct {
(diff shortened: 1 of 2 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -1379,8 +1382,25 @@ func (az *Azure) ServiceAccountStatus() *ServiceAccountStatus {
...
 
+const rateCardPricingSource = "Rate Card API"
+
+// PricingSourceStatus returns the status of the rate card api
 func (az *Azure) PricingSourceStatus() map[string]*PricingSource {
-	return make(map[string]*PricingSource)
+	sources := make(map[string]*PricingSource)
+	rcps := &PricingSource {
+		Name: rateCardPricingSource,
+		Error: az.RateCardPricingStatus,
```
michaelmdresser, line 1392: Does this imply that `RateCardPricingStatus` should really be of type `error`? The only place it seems to be used is here to convey error information.

michaelmdresser (approved):

pkg/cloud/azureprovider.go
Sean-Holcomb: It is only used for that, I was following a president set in AWS implementation, but they should probably change

PR #957, opened 2021-10-01
Avoid nil panic on non-GKE GCP clusters (K3s)

michaelmdresser:
## What does this PR change?

A user running K3s on GCP is encountering a nil panic
caused by our parsing logic for the node.kubernetes.io/instance-type
label. On GCP, we expect GCP node type formats (xx-xx-xx) with
one or two dashes. The K3s cluster sets this value as just "k3s" in the
user's cluster.

By adding a guard on the length of the parsed result we avoid the panic.

N/A

Fix a bug where a nil panic would occur when running on K3s clusters on GCP.

- https://kubecost.zendesk.com/agent/tickets/864

## How was this PR tested?

Deployed to a GKE cluster to confirm nothing broke (no errors observed) and added unit test for the user's specific case.

N/A


mbolt35 (approved):
LGTM

PR #971, opened 2021-10-20
Summary view

nikovacevic:
## Changes
* `SummaryAllocation` types, for use with the Summary Allocation API.
* `ResourceTotals` types, for pre-computing and caching structures that allow quick access to computing idle, sharing coefficients, etc.

## Testing
* https://github.com/kubecost/kubecost-integration-tests/pull/26
* https://github.com/kubecost/kubecost-cost-model/pull/562 has some unit tests, written off of the 
* Unit testing of `SummaryAllocation` and `ResourceTotals` types would be worthwhile, but to expedite the process we're just integration testing that the entire pipeline generates matching results to the Allocation API.
* Benchmarking available here: https://github.com/kubecost/kubecost-cost-model/pull/562 (H/T @michaelmdresser)

## Related PRs
* Branches off https://github.com/kubecost/cost-model/pull/960
* Required with https://github.com/kubecost/kubecost-cost-model/pull/562

## Documentation
* We won't be documenting this feature at first, but documentation will come eventually.

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -350,46 +350,82 @@ func (a *Allocation) Equal(that *Allocation) bool {
 
 // TotalCost is the total cost of the Allocation including adjustments
 func (a *Allocation) TotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.CPUTotalCost() + a.GPUTotalCost() + a.RAMTotalCost() + a.PVTotalCost() + a.NetworkTotalCost() + a.LBTotalCost() + a.SharedTotalCost() + a.ExternalCost
 }
 
 // CPUTotalCost calculates total CPU cost of Allocation including adjustment
 func (a *Allocation) CPUTotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.CPUCost + a.CPUCostAdjustment
 }
 
 // GPUTotalCost calculates total GPU cost of Allocation including adjustment
 func (a *Allocation) GPUTotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.GPUCost + a.GPUCostAdjustment
 }
 
 // RAMTotalCost calculates total RAM cost of Allocation including adjustment
 func (a *Allocation) RAMTotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.RAMCost + a.RAMCostAdjustment
 }
 
 // PVTotalCost calculates total PV cost of Allocation including adjustment
 func (a *Allocation) PVTotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.PVCost() + a.PVCostAdjustment
 }
 
 // NetworkTotalCost calculates total Network cost of Allocation including adjustment
 func (a *Allocation) NetworkTotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.NetworkCost + a.NetworkCostAdjustment
 }
 
 // LBTotalCost calculates total LB cost of Allocation including adjustment
 func (a *Allocation) LBTotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.LoadBalancerCost + a.LoadBalancerCostAdjustment
 }
 
 // SharedTotalCost calculates total shared cost of Allocation including adjustment
 func (a *Allocation) SharedTotalCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.SharedCost
 }
 
 // PVCost calculate cumulative cost of all PVs that Allocation is attached to
 func (a *Allocation) PVCost() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	cost := 0.0
 	for _, pv := range a.PVs {
 		cost += pv.Cost
@@ -399,6 +435,10 @@ func (a *Allocation) PVCost() float64 {
 
 // PVByteHours calculate cumulative ByteHours of all PVs that Allocation is attached to
 func (a *Allocation) PVByteHours() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	byteHours := 0.0
 	for _, pv := range a.PVs {
 		byteHours += pv.ByteHours
@@ -410,6 +450,10 @@ func (a *Allocation) PVByteHours() float64 {
 // no usage or cost, then efficiency is zero. If there is no request, but there
 // is usage or cost, then efficiency is 100%.
 func (a *Allocation) CPUEfficiency() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	if a.CPUCoreRequestAverage > 0 {
 		return a.CPUCoreUsageAverage / a.CPUCoreRequestAverage
 	}
@@ -425,6 +469,10 @@ func (a *Allocation) CPUEfficiency() float64 {
 // no usage or cost, then efficiency is zero. If there is no request, but there
 // is usage or cost, then efficiency is 100%.
 func (a *Allocation) RAMEfficiency() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	if a.RAMBytesRequestAverage > 0 {
 		return a.RAMBytesUsageAverage / a.RAMBytesRequestAverage
 	}
@@ -439,6 +487,10 @@ func (a *Allocation) RAMEfficiency() float64 {
 // TotalEfficiency is the cost-weighted average of CPU and RAM efficiency. If
 // there is no cost at all, then efficiency is zero.
 func (a *Allocation) TotalEfficiency() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	if a.RAMTotalCost()+a.CPUTotalCost() > 0 {
 		ramCostEff := a.RAMEfficiency() * a.RAMTotalCost()
 		cpuCostEff := a.CPUEfficiency() * a.CPUTotalCost()
@@ -482,6 +534,10 @@ func (a *Allocation) PVBytes() float64 {
 
 // ResetAdjustments sets all cost adjustment fields to zero
 func (a *Allocation) ResetAdjustments() {
+	if a == nil {
+		return
+	}
+
 	a.CPUCostAdjustment = 0.0
 	a.GPUCostAdjustment = 0.0
 	a.RAMCostAdjustment = 0.0
@@ -550,27 +606,47 @@ func (a *Allocation) IsAggregated() bool {
 
 // IsExternal is true if the given Allocation represents external costs.
 func (a *Allocation) IsExternal() bool {
+	if a == nil {
+		return false
+	}
+
 	return strings.Contains(a.Name, ExternalSuffix)
 }
 
 // IsIdle is true if the given Allocation represents idle costs.
 func (a *Allocation) IsIdle() bool {
+	if a == nil {
+		return false
+	}
+
 	return strings.Contains(a.Name, IdleSuffix)
 }
 
 // IsUnallocated is true if the given Allocation represents unallocated costs.
 func (a *Allocation) IsUnallocated() bool {
+	if a == nil {
+		return false
+	}
+
 	return strings.Contains(a.Name, UnallocatedSuffix)
 }
 
 // IsUnmounted is true if the given Allocation represents unmounted volume costs.
 func (a *Allocation) IsUnmounted() bool {
+	if a == nil {
+		return false
+	}
+
 	return strings.Contains(a.Name, UnmountedSuffix)
 }
 
 // Minutes returns the number of minutes the Allocation represents, as defined
 // by the difference between the end and start times.
 func (a *Allocation) Minutes() float64 {
+	if a == nil {
+		return 0.0
+	}
+
 	return a.End.Sub(a.Start).Minutes()
 }
 
@@ -594,6 +670,10 @@ func (a *Allocation) Share(that *Allocation) (*Allocation, error) {
 
 // String represents the given Allocation as a string
 func (a *Allocation) String() string {
+	if a == nil {
+		return "<nil>"
+	}
+
 	return fmt.Sprintf("%s%s=%.2f", a.Name, NewWindow(&a.Start, &a.End), a.TotalCost())
 }
 
@@ -743,16 +823,16 @@ func NewAllocationSet(start, end time.Time, allocs ...*Allocation) *AllocationSe
 // succeeds, the allocation is marked as a shared resource. ShareIdle is a
 // simple flag for sharing idle resources.
 type AllocationAggregationOptions struct {
-	AllocationResourceTotalsStore ResourceTotalsStore
-	FilterFuncs                   []AllocationMatchFunc
-	IdleByNode                    bool
-	LabelConfig                   *LabelConfig
-	MergeUnallocated              bool
-	ShareFuncs                    []AllocationMatchFunc
-	ShareIdle                     string
-	ShareSplit                    string
-	SharedHourlyCosts             map[string]float64
-	SplitIdle                     bool
+	AllocationTotalsStore AllocationTotalsStore
+	FilterFuncs           []AllocationMatchFunc
+	IdleByNode            bool
+	LabelConfig           *LabelConfig
+	MergeUnallocated      bool
+	ShareFuncs            []AllocationMatchFunc
+	ShareIdle             string
+	ShareSplit            string
+	SharedHourlyCosts     map[string]float64
+	SplitIdle             bool
 }
 
 // AggregateBy aggregates the Allocations in the given AllocationSet by the given
@@ -893,8 +973,6 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 			}
 		}
 	}
-
-	// TODO shouldn't we just return all shared cost in this case below?
 
 	// It's possible that no more un-shared, non-idle, non-external allocations
 	// remain at this point. This always results in an emptySet, so return early.
@@ -1235,8 +1313,6 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		}
 	}
 
-	// TODO move the following to a self-contained function?
-
 	// (11) In the edge case that some idle has not been distributed because
 	// there is no usage of that resource type, add idle back to
 	// aggregations with only that cost applied.
@@ -1257,40 +1333,36 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	// __idle__ $0      $12     $0
 	// kubecost $12     $0      $7
 
-	if idleSet.Length() > 0 && !options.SplitIdle {
-		if undistributedIdleMap["cpu"] || undistributedIdleMap["gpu"] || undistributedIdleMap["ram"] {
-			for _, idleAlloc := range idleSet.allocations {
-				skip := false
+	hasUndistributedIdle := undistributedIdleMap["cpu"] || undistributedIdleMap["gpu"] || undistributedIdleMap["ram"]
+	if idleSet.Length() > 0 && hasUndistributedIdle {
+		for _, idleAlloc := range idleSet.allocations {
+			// if the idle does not apply to the non-filtered values, skip it
+			skip := false
+			for _, ff := range options.FilterFuncs {
+				if !ff(idleAlloc) {
+					skip = true
+					break
+				}
+			}
+			if skip {
+				continue
+			}
 
-				// if the idle does not apply to the non-filtered values, skip it
-				for _, ff := range options.FilterFuncs {
-					if !ff(idleAlloc) {
-						skip = true
-						break
-					}
+			// if the idle doesn't have a cost to be shared, also skip it
+			if idleAlloc.CPUCost != 0 && idleAlloc.GPUCost != 0 && idleAlloc.RAMCost != 0 {
+				// artificially set the already shared costs to zero
+				if !undistributedIdleMap["cpu"] {
+					idleAlloc.CPUCost = 0
+				}
+				if !undistributedIdleMap["gpu"] {
+					idleAlloc.GPUCost = 0
+				}
+				if !undistributedIdleMap["ram"] {
+					idleAlloc.RAMCost = 0
 				}
 
-				if skip {
-					continue
-				}
-
-				// if the idle doesn't have a cost to be shared, also skip it
-				if idleAlloc.CPUCost != 0 && idleAlloc.GPUCost != 0 && idleAlloc.RAMCost != 0 {
-
-					// artificially set the already shared costs to zero
-					if !undistributedIdleMap["cpu"] {
-						idleAlloc.CPUCost = 0
-					}
-					if !undistributedIdleMap["gpu"] {
-						idleAlloc.GPUCost = 0
-					}
-					if !undistributedIdleMap["ram"] {
-						idleAlloc.RAMCost = 0
-					}
-
-					idleAlloc.Name = IdleSuffix
-					aggSet.Insert(idleAlloc)
-				}
+				idleAlloc.Name = IdleSuffix
+				aggSet.Insert(idleAlloc)
 			}
 		}
 	}
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -743,15 +743,16 @@ func NewAllocationSet(start, end time.Time, allocs ...*Allocation) *AllocationSe
 // succeeds, the allocation is marked as a shared resource. ShareIdle is a
 // simple flag for sharing idle resources.
 type AllocationAggregationOptions struct {
-	FilterFuncs       []AllocationMatchFunc
-	IdleByNode        bool
-	LabelConfig       *LabelConfig
-	MergeUnallocated  bool
-	SharedHourlyCosts map[string]float64
-	ShareFuncs        []AllocationMatchFunc
-	ShareIdle         string
-	ShareSplit        string
-	SplitIdle         bool
+	AllocationResourceTotalsStore ResourceTotalsStore
+	FilterFuncs                   []AllocationMatchFunc
+	IdleByNode                    bool
+	LabelConfig                   *LabelConfig
+	MergeUnallocated              bool
+	ShareFuncs                    []AllocationMatchFunc
+	ShareIdle                     string
+	ShareSplit                    string
+	SharedHourlyCosts             map[string]float64
+	SplitIdle                     bool
 }
 
 // AggregateBy aggregates the Allocations in the given AllocationSet by the given
@@ -759,33 +760,47 @@ type AllocationAggregationOptions struct {
 // given AllocationProperty; e.g. Containers can be divided by Namespace, but not vice-a-versa.
 func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAggregationOptions) error {
 	// The order of operations for aggregating allocations is as follows:
+	//
 	//  1. Partition external, idle, and shared allocations into separate sets.
 	//     Also, create the aggSet into which the results will be aggregated.
+	//
 	//  2. Compute sharing coefficients for idle and shared resources
 	//     a) if idle allocation is to be shared, compute idle coefficients
 	//     b) if idle allocation is NOT shared, but filters are present, compute
 	//        idle filtration coefficients for the purpose of only returning the
 	//        portion of idle allocation that would have been shared with the
 	//        unfiltered results. (See unit tests 5.a,b,c)
-	//     c) generate shared allocation for then given shared overhead, which
+	//     c) generate shared allocation for them given shared overhead, which
 	//        must happen after (2a) and (2b)
 	//     d) if there are shared resources, compute share coefficients
+	//
 	//  3. Drop any allocation that fails any of the filters
+	//
 	//  4. Distribute idle allocations according to the idle coefficients
+	//
 	//  5. Generate aggregation key and insert allocation into the output set
+	//
 	//  6. If idle is shared and resources are shared, some idle might be shared
 	//     with a shared resource. Distribute that to the shared resources
 	//     prior to sharing them with the aggregated results.
+	//
 	//  7. Apply idle filtration coefficients from step (2b)
+	//
 	//  8. Distribute shared allocations according to the share coefficients.
+	//
 	//  9. If there are external allocations that can be aggregated into
 	//     the output (i.e. they can be used to generate a valid key for
 	//     the given properties) then aggregate; otherwise... ignore them?
+	//
 	// 10. If the merge idle option is enabled, merge any remaining idle
 	//     allocations into a single idle allocation. If there was any idle
 	//	   whose costs were not distributed because there was no usage of a
 	//     specific resource type, re-add the idle to the aggregation with
 	//     only that type.
+	//
+	// 11. Distribute any undistributed idle, in the case that idle
+	//     coefficients end up being zero and some idle is not shared.
+
 	if as.IsEmpty() {
 		return nil
 	}
@@ -879,6 +894,8 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		}
 	}
 
+	// TODO shouldn't we just return all shared cost in this case below?
+
 	// It's possible that no more un-shared, non-idle, non-external allocations
 	// remain at this point. This always results in an emptySet, so return early.
 	if len(as.allocations) == 0 {
@@ -963,7 +980,7 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	}
 
 	// (2c) Convert SharedHourlyCosts to Allocations in the shareSet. This must
-	// come after idle coefficients are computes so that allocations generated
+	// come after idle coefficients are computed so that allocations generated
 	// by shared overhead do not skew the idle coefficient computation.
 	for name, cost := range options.SharedHourlyCosts {
 		if cost > 0.0 {
@@ -1218,7 +1235,9 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		}
 	}
 
-	// In the edge case that some idle has not been distributed because
+	// TODO move the following to a self-contained function?
+
+	// (11) In the edge case that some idle has not been distributed because
 	// there is no usage of that resource type, add idle back to
 	// aggregations with only that cost applied.
 
@@ -1240,9 +1259,7 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 
 	if idleSet.Length() > 0 && !options.SplitIdle {
 		if undistributedIdleMap["cpu"] || undistributedIdleMap["gpu"] || undistributedIdleMap["ram"] {
-
 			for _, idleAlloc := range idleSet.allocations {
-
 				skip := false
 
 				// if the idle does not apply to the non-filtered values, skip it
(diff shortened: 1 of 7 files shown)


pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+				shouldShare = true
+				break
+			}
+		}
+		if shouldShare {
+			sa := NewSummaryAllocation(alloc, reconcile, reconcileNetwork)
+			sa.Share = true
+			sas.Insert(sa)
+			continue
+		}
```
michaelmdresser, line 268: What does inserting a summary allocation with `Share = true` mean?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	}
+
+	for key := range as.externalKeys {
+		sas.externalKeys[key] = true
+	}
+
+	for key := range as.idleKeys {
+		sas.idleKeys[key] = true
+	}
```
michaelmdresser, line 296: I think I understand why we have these maps but this feels like another abstraction leaking (perhaps not the right phrase for what's going on here?) because of the mismatch we have between "reporting all costs in allocation" and "allocations are kubernetes resource allocations." I think I may be beating a dead horse at this point!

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+		return that, nil
+	}
+
+	if that == nil || len(that.SummaryAllocations) == 0 {
+		return sas, nil
+	}
+
+	// Set start, end to min(start), max(end)
+	start := *sas.Window.Start()
+	end := *sas.Window.End()
```
michaelmdresser, line 312: Any danger of a nil panic here?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+}
+
+// SummaryAllocationSet stores a set of SummaryAllocations, each with a unique
+// name, that share a window. An AllocationSet is mutable, so treat it like a
+// threadsafe map.
+type SummaryAllocationSet struct {
+	sync.RWMutex
+	externalKeys       map[string]bool
+	idleKeys           map[string]bool
+	SummaryAllocations map[string]*SummaryAllocation `json:"allocations"`
```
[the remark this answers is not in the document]
michaelmdresser: My guess is that this unfortunate circumstance is because of the auto-serialization approach of `encoding/json` that requires exported fields (yay, Golang).

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	//     shared allocations before sharing them with the aggregated
+	//     allocations.
+	//
+	//  #. Apply idle filtration
+	//
+	//  #. Distribute shared allocations
+	//
+	//  #. Distribute shared tenancy costs
+	//
+	//  TODO
```
michaelmdresser, line 377: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	//     the output (i.e. they can be used to generate a valid key for
+	//     the given properties) then aggregate; otherwise... ignore them?
+	//
+	//  #. If the merge idle option is enabled, merge any remaining idle
+	//     allocations into a single idle allocation. If there was any idle
+	//	   whose costs were not distributed because there was no usage of a
+	//     specific resource type, re-add the idle to the aggregation with
+	//     only that type.
+	//
+	//  TODO
```
michaelmdresser, line 388: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+
+	// Check if we have any work to do; if not, then early return. If
+	// aggregateBy is nil, we don't aggregate anything. On the other hand,
+	// an empty slice implies that we should aggregate everything. (See
+	// generateKey for why that makes sense.)
+	shouldAggregate := aggregateBy != nil
+	shouldShare := len(options.SharedHourlyCosts) > 0 || len(options.ShareFuncs) > 0
+	if !shouldAggregate && !shouldShare {
+		return nil
+	}
```
michaelmdresser, line 416: Does this have identical behavior to `AllocationSet` in which an empty aggregation is not the same as the "default" (from the HTTP API's perspective) aggregation of `cluster/node/namespace/...`?

pkg/kubecost/summaryallocation.go
michaelmdresser: On further reading, it seems like this has implications on `AggregateBy`, and that, like the allocation API, interacting with the summary API in an understandable way requires `AggregateBy` to be called. Do you think we should explicitly encourage this behavior in comments (or some other way) or is it enough to rely on reference usages?

I don't think this is worth a change in this PR. Too much work. But it feels like another candidate for revisiting this interface when we get some more space.

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	// shareSet will be shared among aggSet after initial aggregation
+	// is complete
+	shareSet := &SummaryAllocationSet{
+		Window: sas.Window.Clone(),
+	}
+
+	sas.Lock()
+	defer sas.Unlock()
+
+	// TODO comment
```
michaelmdresser, line 443: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	sharedResourceTotals := map[string]*ResourceTotals{}
+	totalUnmountedCost := 0.0
+
+	// TODO comment
```
michaelmdresser, line 447: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+
+		// External allocations get aggregated post-hoc (see step 6) and do
+		// not necessarily contain complete sets of properties, so they are
+		// moved to a separate AllocationSet.
+		if sa.IsExternal() {
+			delete(sas.externalKeys, sa.Name)
+			delete(sas.SummaryAllocations, sa.Name)
+			externalSet.Insert(sa)
+			continue
+		}
```
michaelmdresser, line 485: This is a 1:1 mapping from current `AllocationSet` behavior right?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+
+		// Track total unmounted cost because it must be taken out of total
+		// allocated costs for sharing coefficients.
+		if sa.IsUnmounted() {
+			totalUnmountedCost += sa.TotalCost()
+		}
+	}
+
+	// TODO do we need to handle case where len(SummaryAllocations) == 0?
```
michaelmdresser, line 510: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+
+	// (#) Retrieve pre-computed allocation resource totals, which are to be
+	// used to compute idle coefficients, based on the ratio of any given
+	// allocation's per-resource cost to the allocated per-resource totals.
+	var allocTotals map[string]*ResourceTotals
+	if options.IdleByNode {
+		allocTotals = options.AllocationResourceTotalsStore.GetResourceTotalsByNode(*sas.Window.Start(), *sas.Window.End())
+		if allocTotals == nil {
+			// TODO
+			log.Warningf("SummaryAllocation: nil allocTotals by node for %s", sas.Window)
```
michaelmdresser, line 520: :eyes:

Also, should this log be deduped?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+		}
+	} else {
+		allocTotals = options.AllocationResourceTotalsStore.GetResourceTotalsByCluster(*sas.Window.Start(), *sas.Window.End())
+		if allocTotals == nil {
+			// TODO
+			log.Warningf("SummaryAllocation: nil allocTotals by cluster for %s", sas.Window)
```
michaelmdresser, line 526: Likewise!

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+			filteredTotals[key].RAMCost += rt.RAMCost
+		}
+	}
+
+	sharingCoeffs := map[string]float64{}
+
+	// (#) Distribute idle cost, if sharing
+	// (#) Distribute tenancy costs, if sharing
+	// (#) Aggregate
+	// TODO better comment
```
michaelmdresser, line 558: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	for _, sa := range sas.SummaryAllocations {
+		// Generate key to use for aggregation-by-key and allocation name
+		key := sa.generateKey(aggregateBy, options.LabelConfig)
+
+		// TODO comment
```
michaelmdresser, line 563: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+			// Distribute idle allocations by coefficient per-idleId, per-allocation
+			for _, idle := range idleSet.SummaryAllocations {
+				var key string
+
+				// Only share idle allocation with current allocation (sa) if
+				// the relevant property matches (i.e. Cluster or Node,
+				// depending on which idle sharing option is selected)
+				if options.IdleByNode {
+					if idle.Properties.Node != sa.Properties.Node {
+						continue
```
michaelmdresser, line 585: Any danger here if there are two clusters with identically-named nodes?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+		// which the allocation is inserted into the set.
+		sa.Name = key
+
+		// If merging unallocated allocations, rename all unallocated
+		// allocations as simply __unallocated__
+		if options.MergeUnallocated && sa.IsUnallocated() {
+			sa.Name = UnallocatedSuffix
+		}
+
+		// TODO do we need to be going off of "cluster/node" for "node" in all these places?
```
michaelmdresser, line 615: I think this is the same question as my comment just above. I would guess that the answer is yes, but it won't be a problem in _most_ situations because cloud providers tend to be pretty verbose and unique in the way they name cluster nodes. There is no guarantee, though, especially when thinking about the "long tail" of cluster configurations.

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+				sharingCoeffDenominator -= 1.0
+			}
+		} else {
+			sharingCoeffDenominator -= totalUnmountedCost
+		}
+
+		// Compute sharing coeffs by dividing the thus-far accumulated
+		// numerators by the now-finalized denominator.
+		for key := range sharingCoeffs {
+			sharingCoeffs[key] /= sharingCoeffDenominator
```
michaelmdresser, line 775: Would we encounter a problem if this division results in a NaN or other edge-case float?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	for _, sa := range externalSet.SummaryAllocations {
+		skip := false
+
+		// TODO deal with filters...
+		// for _, ff := range options.FilterFuncs {
+		// 	if !ff(sa) {
+		// 		skip = true
+		// 		break
+		// 	}
+		// }
```
michaelmdresser, line 815: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	if !options.SplitIdle {
+		for _, idleAlloc := range aggSet.IdleAllocations() {
+			aggSet.Delete(idleAlloc.Name)
+			idleAlloc.Name = IdleSuffix
+			aggSet.Insert(idleAlloc)
+		}
+	}
+
+	// TODO
+	// (#) Distribute remaining, undistributed idle
```
michaelmdresser, line 835: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	if sas == nil || len(sas.SummaryAllocations) == 0 {
+		return idles
+	}
+
+	sas.RLock()
+	defer sas.RUnlock()
+
+	for key := range sas.idleKeys {
+		if sa, ok := sas.SummaryAllocations[key]; ok {
+			idles[key] = sa // TODO Clone()?
```
michaelmdresser, line 943: :eyes:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	return idles
+}
+
+// Insert aggregates the current entry in the SummaryAllocationSet by the given Allocation,
+// but only if the Allocation is valid, i.e. matches the SummaryAllocationSet's window. If
+// there is no existing entry, one is created. Nil error response indicates success.
+func (sas *SummaryAllocationSet) Insert(sa *SummaryAllocation) error {
+	if sas == nil {
+		return fmt.Errorf("cannot insert into nil SummaryAllocationSet")
+	}
```
michaelmdresser, line 956: I think we should check `sa == nil` here, too.

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+
+	// If the given Allocation is an external one, record that
+	if sa.IsExternal() {
+		sas.externalKeys[sa.Name] = true
+	}
+
+	// If the given Allocation is an idle one, record that
+	if sa.IsIdle() {
+		sas.idleKeys[sa.Name] = true
+	}
```
michaelmdresser, line 989: Pathological case: non-external/idle allocation gets inserted with name `k`, then another allocation with name `k` is inserted that is marked as external/idle. What should happen?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+type SummaryAllocationSetRange struct {
+	sync.RWMutex
+	Step                  time.Duration           `json:"step"`
+	SummaryAllocationSets []*SummaryAllocationSet `json:"sets"`
+	Window                Window                  `json:"window"`
+}
+
+// NewSummaryAllocationSetRange instantiates a new range composed of the given
+// SummaryAllocationSets in the order provided.
+func NewSummaryAllocationSetRange(sass ...*SummaryAllocationSet) *SummaryAllocationSetRange {
```
michaelmdresser, line 1004: Should this function check to ensure that `sass` does not contain `SummaryAllocationSet`s with overlapping or identical windows?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+// properties and options.
+func (sasr *SummaryAllocationSetRange) AggregateBy(aggregateBy []string, options *AllocationAggregationOptions) error {
+	sasr.Lock()
+	defer sasr.Unlock()
+
+	for _, sas := range sasr.SummaryAllocationSets {
+		err := sas.AggregateBy(aggregateBy, options)
+		if err != nil {
+			return err
+		}
```
michaelmdresser, line 1058: If an error occurs mid-aggregation, this won't "roll back" the partially complete aggregation. Is this acceptable behavior?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+func (sasr *SummaryAllocationSetRange) Append(sas *SummaryAllocationSet) error {
+	if sasr.Step != 0 && sas.Window.Duration() != sasr.Step {
+		return fmt.Errorf("cannot append set with duration %s to range of step %s", sas.Window.Duration(), sasr.Step)
+	}
+
+	sasr.Lock()
+	defer sasr.Unlock()
+
+	// Append to list of sets
+	sasr.SummaryAllocationSets = append(sasr.SummaryAllocationSets, sas)
```
michaelmdresser, line 1075: Same question about overlapping/identical windows.

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	if sasr == nil {
+		return
+	}
+
+	for i, as := range sasr.SummaryAllocationSets {
+		f(i, as)
+	}
+}
+
+// TODO this stinks. Can we do better with external cost so that we can remove?
```
michaelmdresser, line 1104: I hope so! :smile:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+		}
+		sas := sasr.SummaryAllocationSets[i]
+
+		// Insert each Allocation from the given set
+		thatAS.Each(func(k string, alloc *Allocation) {
+			externalSA := NewSummaryAllocation(alloc, true, true)
+			err = sas.Insert(externalSA)
+			if err != nil {
+				// TODO ?
+				log.Errorf("SummaryAllocation: error inserting %s: %s", k, err)
```
michaelmdresser, line 1144: :eyes: and should this be deduped?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+		})
+	})
+
+	// err might be nil
+	return err
+}
+
+// TODO Custom MarshalJSON and UnmarshalJSON for these?
+// - Step is uintelligible (microseconds??)
+// - TotalCost would be nice
```
michaelmdresser, line 1155: Not sure about the costs of doing so, but we could separate the types between `SummaryAllocation` (used internally to actually compute the summarized allocation) and `SummaryAllocationResponse` which is easy to serialize/deserialize based on struct tags. Then, instead of writing a custom `MarshalJSON`, you just have to write a method like 
```go
func (sa *SummaryAllocation) ToResponse() SummaryAllocationResponse
```

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/kubecost/cost-model/pkg/log"
+	"github.com/patrickmn/go-cache"
+)
+
+// TODO can we use ResourceTotals for all things or do we need specific structs
+// like AllocationResourceTotals, AssetResourceTotals, etc.
```
michaelmdresser, line 14: Any elaboration needed here for future work?

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+
+type ResourceTotals struct {
+	Start                 time.Time `json:"end"`
+	End                   time.Time `json:"start"`
+	Cluster               string    `json:"cluster,omitempty"`
+	Node                  string    `json:"node,omitempty"`
```
michaelmdresser, line 20: I think its worth briefly explaining why only these two fields are necessary to identify resource totals.

michaelmdresser: Also, why are these `omitempty`?

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+				GPUCost:              alloc.GPUTotalCost(),
+				LoadBalancerCost:     alloc.LBTotalCost(),
+				NetworkCost:          alloc.NetworkTotalCost(),
+				PersistentVolumeCost: alloc.PVCost(),
+				RAMCost:              alloc.RAMTotalCost(),
+			}
+		}
+	})
+
+	// TODO clean up
```
michaelmdresser, line 89: What needs cleaning up?

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	log.Infof("ResourceTotals: recorded %.4f over %d %ss for %s", total, len(rts), prop, as.Window)
+
+	return rts
+}
+
+func ComputeResourceTotalsFromAssets(as *AssetSet, prop AssetProperty) map[string]*ResourceTotals {
+	rts := map[string]*ResourceTotals{}
+
+	// TODO comment
```
michaelmdresser, line 102: :eyes:

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+					CPUCost: cpuCost,
+					RAMCost: ramCost,
+					GPUCost: gpuCost,
+				}
+			}
+		} else if disk, ok := asset.(*Disk); ok && prop == AssetNodeProp {
+			// Only record attached disks when prop is Node
+			disks[disk.Properties().Name] = disk
+		} else if cm, ok := asset.(*ClusterManagement); ok && prop == AssetClusterProp {
+			// TODO ?
```
michaelmdresser, line 176: ?

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+					Node:    name,
+				}
+			}
+
+			rts[key].Count++
+			rts[key].AttachedVolumeCost += disk.TotalCost()
+		}
+	}
+
+	// TODO clean up
```
michaelmdresser, line 222: Clean up what?

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+	if rts == nil {
+		return errors.New("cannot update resource totals from AllocationSet for nil ResourceTotalsStore")
+	}
+
+	if as == nil {
+		return errors.New("cannot update resource totals from AllocationSet for nil AllocationSet")
+	}
+
+	start := *as.Window.Start()
+	end := *as.Window.End()
```
michaelmdresser, line 286: Any danger of `nil` panic here?

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+	if rts == nil {
+		return errors.New("cannot update resource totals from AssetSet for nil ResourceTotalsStore")
+	}
+
+	if as == nil {
+		return errors.New("cannot update resource totals from AssetSet for nil AssetSet")
+	}
+
+	start := *as.Window.Start()
+	end := *as.Window.End()
```
michaelmdresser, line 309: Any danger of `nil` panic here?

pkg/kubecost/totals.go
```diff
@@ -0,0 +1,367 @@
...
+
+type MemoryResourceTotalsStore struct {
+	byCluster *cache.Cache
+	byNode    *cache.Cache
+}
+
+func NewResourceTotalsStore() *MemoryResourceTotalsStore {
+	return &MemoryResourceTotalsStore{
+		byCluster: cache.New(cache.NoExpiration, cache.NoExpiration),
+		byNode:    cache.New(cache.NoExpiration, cache.NoExpiration),
```
michaelmdresser, line 330: Should these really have no expiration? Perhaps it should be set to the maximum ETL coverage?

pkg/kubecost/window.go
```diff
@@ -433,8 +434,16 @@ func (w Window) IsOpen() bool {
...
+	if w.start != nil {
+		buffer.WriteString(fmt.Sprintf("\"start\":\"%s\",", w.start.Format(time.RFC3339)))
+	} else {
+		buffer.WriteString(fmt.Sprintf("\"start\":\"%s\",", "null"))
+	}
+	if w.end != nil {
+		buffer.WriteString(fmt.Sprintf("\"end\":\"%s\"", w.end.Format(time.RFC3339)))
+	} else {
+		buffer.WriteString(fmt.Sprintf("\"end\":\"%s\"", "null"))
+	}
```
michaelmdresser, line 446: Nice!

michaelmdresser (approved):
Functionally this looks good to me. I trust that the integrations tests you're working on here will include external cost, idle cost, etc. comparisons with the current Allocation API.

I have a few questions about comments, edge cases, and potential panics, but I don't think they are merge blocking.

I have some deeper questions about how design choices with APIs like this tie in with the external cost discussions we've been having and other ideas. These comments are mainly for future reference in the conversations we're continuing to have, feel free to leave them alone to get this PR across the line!

pkg/kubecost/allocation.go
```diff
@@ -743,49 +743,64 @@ func NewAllocationSet(start, end time.Time, allocs ...*Allocation) *AllocationSe
...
 	//     the given properties) then aggregate; otherwise... ignore them?
+	//
 	// 10. If the merge idle option is enabled, merge any remaining idle
 	//     allocations into a single idle allocation. If there was any idle
 	//	   whose costs were not distributed because there was no usage of a
 	//     specific resource type, re-add the idle to the aggregation with
 	//     only that type.
+	//
+	// 11. Distribute any undistributed idle, in the case that idle
+	//     coefficients end up being zero and some idle is not shared.
```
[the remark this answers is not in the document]
nikovacevic: Yep, GPUs in particular. Kaelan wrote this step in the Allocation API a while back, but I haven't ported it over yet.

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/kubecost/cost-model/pkg/log"
+)
+
+// SummaryAllocation summarizes Allocation, keeping only the fields necessary
+// for providing a high-level view of identifying the Allocation (Name) over a
+// defined period of time (Start, End) and inspecting per-resource costs
```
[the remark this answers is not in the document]
nikovacevic: I definitely see what you mean. I don't mention `Window` here because, unlike `Allocation`, there is no `Window` on this type at all. You can have a `SummaryAllocation` with no concept of `Window`, feasibly. That's a matter entirely left to the `SummaryAllocationSet` now, which actually feels nicer, but is a departure from the existing precedent.

With that in mind, do you still think it warrants mentioning? I can try to sharpen up the idea a bit, regardless.

pkg/kubecost/allocation.go
```diff
@@ -743,49 +743,64 @@ func NewAllocationSet(start, end time.Time, allocs ...*Allocation) *AllocationSe
...
-	FilterFuncs       []AllocationMatchFunc
-	IdleByNode        bool
-	LabelConfig       *LabelConfig
-	MergeUnallocated  bool
-	SharedHourlyCosts map[string]float64
-	ShareFuncs        []AllocationMatchFunc
-	ShareIdle         string
-	ShareSplit        string
-	SplitIdle         bool
+	AllocationResourceTotalsStore ResourceTotalsStore
```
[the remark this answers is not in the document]
nikovacevic: I wrestled with this one. Putting it on the query, itself, was a simple way to "inject" the functionality into the query. Nice for me, writing the feature; bad in almost every other way, probably. So as I re-read the code, I think this belongs elsewhere. But where?

We could add it as a parameter on creation of an `ETLStore`. That, however, has the downside of actually requiring more threading of the type through constructors (a couple places per implementation of `ETLStore`, which is now 3 and growing) as opposed to living in two places (`QueryOptions` and `AggregationOptions`). Perhaps we'd come up with more ideas, given the opportunity to refactor ETL into a more sensible, hierarchical pattern altogether. But that's my thought.

Good call-out.

pkg/kubecost/allocation.go
```diff
@@ -879,6 +894,8 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		}
 	}
 
+	// TODO shouldn't we just return all shared cost in this case below?
+
```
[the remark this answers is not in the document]
nikovacevic: Glad you're excited! I am, too! Let's kick of a product conversation soon.

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/kubecost/cost-model/pkg/log"
+)
+
+// SummaryAllocation summarizes Allocation, keeping only the fields necessary
+// for providing a high-level view of identifying the Allocation (Name) over a
```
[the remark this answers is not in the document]
nikovacevic: Anything is possible! This is very abstract, so perhaps we should discuss. If I'm reading between the lines accurately, you could definitely imagine normalizing a lot of our core data, then joining it at advantageous points (query time? cached? whatever).

As an example, we could store only usage and node-ID on `Allocation`, store cost and node-ID on `Node` (i.e. `Asset`) and join to come up with total cost. Reconciliation is only reconciliation if you start with an incorrect price haha. This would just be: compute the price. A very pure idea, indeed. Today, though, we probably survive off of the performance gains from denormalizing, and thus not having to join at query time.

Is that what you have in mind?

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	ExternalCost           float64               `json:"externalCost"`
+	Share                  bool                  `json:"-"`
+}
+
+func NewSummaryAllocation(alloc *Allocation, reconcile, reconcileNetwork bool) *SummaryAllocation {
+	if alloc == nil {
+		return nil
+	}
+
+	// TODO evaluate performance penalties for "cloning" here
```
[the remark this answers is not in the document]
nikovacevic: That is a good idea. I'm also just considering pprof-ing it.

(BTW this is why FileStore penalizes SummaryAllocation, relative to Allocation—FileStore allows Allocation to access the data directly via mmap.)

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	return sa
+}
+
+func (sa *SummaryAllocation) Add(that *SummaryAllocation) {
+	if sa == nil {
+		log.Warningf("SummaryAllocation.Add: trying to add a nil receiver")
+		return
+	}
+
+	// TODO do we need to maintain this?
```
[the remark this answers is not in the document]
nikovacevic: It is not :smile:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+	sa.PVCost += that.PVCost
+	sa.RAMCost += that.RAMCost
+	sa.SharedCost += that.SharedCost
+}
+
+// CPUEfficiency is the ratio of usage to request. If there is no request and
+// no usage or cost, then efficiency is zero. If there is no request, but there
+// is usage or cost, then efficiency is 100%.
+func (sa *SummaryAllocation) CPUEfficiency() float64 {
+	if sa.CPUCoreRequestAverage > 0 {
```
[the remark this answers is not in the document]
nikovacevic: Agree. Will add in the same place to `Allocation`, too.

pkg/kubecost/summaryallocation.go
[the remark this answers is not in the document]
nikovacevic: That's correct. And there's a slightly bigger story. Basically, when I designed `AllocationSet` the first time, I was sensitive to the idea that it might end up being a core type, which others (perhaps even outside the company) might want to use for a very wide range of purposes. So safety seemed high-priority. Now, with `SummaryAllocation`, speed is the priority, and we've learned lessons about serialization. (Yay, Go, indeed.)

So this might be a little controversial. But it's also still kind of fair to say "treat it like a threadsafe map" haha. Like—use the methods! Don't just access the map! If you do, you're mis-using the type. But also, don't make me write "Each" and "Get" because you can just RWLock and iterate or access by key. I dunno. Maybe with this freedom, future code writers will shoot themselves in the foot. I hope not :shrug:

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1155 @@
...
+// threadsafe map.
+type SummaryAllocationSet struct {
+	sync.RWMutex
+	externalKeys       map[string]bool
+	idleKeys           map[string]bool
+	SummaryAllocations map[string]*SummaryAllocation `json:"allocations"`
+	Window             Window                        `json:"window"`
+}
+
+func NewSummaryAllocationSet(as *AllocationSet, ffs, sfs []AllocationMatchFunc, reconcile, reconcileNetwork bool) *SummaryAllocationSet {
```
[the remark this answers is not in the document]
nikovacevic: Sure, will do. Probably missing many comments here haha. Lots got copied over.

pkg/kubecost/summaryallocation.go
nikovacevic: Yeah, setting `Share = true` is a shortcut for saying, "Don't aggregate this like normal, share it into the other aggregations." Not very pretty, but it works.

In general, we're actually well on our way to cleaning up `AggregateBy` in this work, so that's nice. We should port these lessons back to Allocation API. Whether or not you should have to call `AggregateBy` to share—I could go either way. Technically, sharing is aggregating, so I'm not too worried about it. When filtering happened only in `AggregateBy`, that was a little messed up.

pkg/kubecost/summaryallocation.go
nikovacevic: I welcome the conversation! You have a much more prescriptive view of what an "allocation" is than I do, and that's neither a bad nor a good thing, necessarily, but it's an interesting question. From my perspective, an "idle allocation" is a perfectly reasonable idea, and fits well into the `Allocation` type, but I could be convinced otherwise. External, I am also less annoyed by, but I do think it's a much less "true" fit.

pkg/kubecost/summaryallocation.go
nikovacevic: Today, unlikely, but better safe than sorry.

pkg/kubecost/summaryallocation.go
nikovacevic: The behavior should match, but I'm not sure that premise is the case. Will double check, but I think they're the same in both `Allocation` and `SummaryAllocation`. Possibly mistaken, of course.

pkg/kubecost/summaryallocation.go
nikovacevic: Yes, basically just ported exactly for now.

pkg/kubecost/summaryallocation.go
nikovacevic: Yes, and I believe there is a TODO comment later on somewhere about exactly that. Good question.

pkg/kubecost/summaryallocation.go
nikovacevic: Ha yep, this one.

pkg/kubecost/summaryallocation.go
nikovacevic: The assumption is that the case where `sharingCoeffDenominator` is `0.0` occurs only when the set it empty, and then you don't even get here. But that's unsafe, so I'll address it.

pkg/kubecost/summaryallocation.go
nikovacevic: Oi, that's right. Blergh!!!

pkg/kubecost/summaryallocation.go
nikovacevic: Just gonna make this function unexported and not worry about cloning. We only use it one place, and it's not necessary to clone there.

pkg/kubecost/summaryallocation.go
nikovacevic: Well, we intentionally try our best to avoid that by appending suffices onto those (`__external__` and `__idle__`). So, we'd ultimately end up adding the two together and calling it all idle. IMO that's fine. Please, people, don't name your containers "__idle__" ahah.

pkg/kubecost/summaryallocation.go
nikovacevic: I've decided not to make it a rule. Users must use wisely. Open to changing that, but then we probably want to check:
* step sizes are all equal
* everything is consecutive (i.e. chronological)
* no gaps
* no overlapping windows

Is there a specific case you'd want to catch, or just thinking about it from a generally purist/rule-based perspective?

pkg/kubecost/summaryallocation.go
nikovacevic: If an error occurs mid-Aggregation, it's game over. The error should simply percolate out. And if you don't check it, then sorry.

One thing we could do to make this "safer" in some sense, is to obliterate the instance before returning the error so that callers don't use a wacked out half-aggregated set. Like `Sets = empty map`, boom, nothing left to use. Check yr errors, folks. What do you think of that idea?

pkg/kubecost/summaryallocation.go
nikovacevic: Same answer

pkg/kubecost/summaryallocation.go
nikovacevic: Yep, will dedupe. I hate this function. I want to delete it.

pkg/kubecost/summaryallocation.go
nikovacevic: True. For now, I think we don't have an issue, but I'm curious to hear your vision for a `...Response` type. Would we make one for each thing we JSON-serialize, ideally? What are the pros/cons?

pkg/kubecost/totals.go
nikovacevic: Maybe I can elaborate a bit, but basically, it says it all—do we benefit from having one struct for each, so that allocation- or asset-specific things can be recorded. Seems like we're in the clear for now, but I might leave the comment, as an idea.

pkg/kubecost/totals.go
nikovacevic: Frankly, we don't need struct tags at all. I don't serialize these anywhere. I'll just remove.

And I'm not opposed to leaving a comment about Cluster and Node, but it may not ultimately be true that those are the only two that are necessary. It's just that for now, that's all we ever use (e.g. idle by cluster or idle by node). So I will include that in the comment, but we may end up adding more properties later. Who knows. It's a kind of "fluid" type.

pkg/kubecost/totals.go
nikovacevic: The log below.

pkg/kubecost/totals.go
nikovacevic: The log below

pkg/kubecost/totals.go
nikovacevic: :shrug:

pkg/kubecost/totals.go
nikovacevic: Happy to nil check

pkg/kubecost/totals.go
nikovacevic: Will nil check

pkg/kubecost/totals.go
nikovacevic: What's the pathological case where you're concerned about no expiration? It's not clear to me how long we'd have to wait before it was better to have no data than the last-updated data.

pkg/kubecost/window.go
nikovacevic: :smile:

nikovacevic: Got burned by that at some point in development haha

nikovacevic (commented):
Oops, half of my responses were "review comments" and the other half were just responses... So, in summary, I've addressed many things and will add more (mostly comments, and one or two little TODO features) tomorrow before asking for a final review. Thank you, @michaelmdresser!

pkg/kubecost/allocation.go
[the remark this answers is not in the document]
michaelmdresser: I'm fairly happy with requiring these in `QueryOptions` because it makes a data dependency in the query explicit (side note for KCM, we could have a separate options struct for summary queries and just embed the regular query struct :man_shrugging:). Seeing the dependency in `AggregationOptions`, though, is what's bothering me and probably harks back to our "should aggregateby really be called aggregateby" discussion we had months ago.

pkg/kubecost/summaryallocation.go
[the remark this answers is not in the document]
michaelmdresser: > That's a matter entirely left to the SummaryAllocationSet now, which actually feels nicer, but is a departure from the existing precedent.

I think it's worth mentioning this in a comment on the type, just because there's a lot of momentum behind the start end vs. window start end concept right now.

pkg/kubecost/summaryallocation.go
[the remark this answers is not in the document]
michaelmdresser: Yes, my data flow dream basically comes down to normalization and joining. That's a perfect example, and I definitely agree that the denormalization we have today is keeping us alive on performance. My hope would be that "joining it at advantageous points" (i.e. storing raw allocation/asset in durable storage but caching joined data at specified granularity(s)) would counteract the performance hits.

pkg/kubecost/summaryallocation.go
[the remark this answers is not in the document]
michaelmdresser: :+1: :+1: I think this is excellent reasoning. Abusers of performance-focused code, beware! :grin:

pkg/kubecost/summaryallocation.go
michaelmdresser: I think there's some name for what I'm trying to do, which is basically assert conditions about our data (there are cool type systems that can do this, but that's neither here nor there). I think its useful to write down assumptions we make about arguments, even if just in comments. Putting that list you've just written into a comment would be perfect!

pkg/kubecost/summaryallocation.go
michaelmdresser: Mm I think there's danger in both directions, but I'm most in favor of what you've suggested (empty map or nil).

pkg/kubecost/summaryallocation.go
michaelmdresser: My ideal world does include fully separating API response types (serialized) and internal data types. This relates to my earlier comment about (de)normalization.

The main con I see is overhead in converting between types. The main pro I see is greater flexibility in internal data representation while maintaining API contracts by explicitly fencing the two areas off in the type system.

pkg/kubecost/totals.go
michaelmdresser: Say we've got a user with a 90d ETL and the pod manages to stay alive for 180+ days. Does this cache grow unbounded, eventually causing an OOM crash?

nikovacevic: Ah, very fair point. This does save me from having to manually "roll off the back". I'm convinced.

pkg/kubecost/summaryallocation.go
nikovacevic: Added the `IsOpen` check above.

pkg/kubecost/summaryallocation.go
nikovacevic: I think I misunderstood your comment. Yes, the two APIs share behavior. In fact, the even share that precise code path now in `package main`'s `ParseAllocationRequest`

pkg/kubecost/summaryallocation.go
```diff
@@ -0,0 +1,1232 @@
...
+// required for unfortunate reasons to do with performance and legacy order-of-
+// operations details, as well as the fact that reconciliation has been
+// pushed down to the conversion step between Allocation and SummaryAllocation.
+func NewSummaryAllocationSet(as *AllocationSet, ffs, sfs []AllocationMatchFunc, reconcile, reconcileNetwork bool) *SummaryAllocationSet {
+	if as == nil {
+		return nil
+	}
+
+	// If we can know the exact size of the map, use it. If filters or sharing
+	// functions are present, we can't know the size, so we make a default map.
```
AjayTripathy, line 292: Not super sure I agree with this idea. Filters are guaranteed to be less than the size of the data right? I think a larger memory allocation ie always allocating len(as.allocations) is better because it avoids the resize which is expensive.

AjayTripathy: I think it also has the benefits of making the code a biiiit neater, though that's not the goal here necessarily.

pkg/kubecost/totals.go
```diff
@@ -18,22 +18,51 @@ import (
...
+
+// TotalGPUCost returns GPU cost with adjustment.
+func (art *AllocationTotals) TotalGPUCost() float64 {
+	return art.GPUCost + art.GPUCostAdjustment
+}
+
+// TotalRAMCost returns RAM cost with adjustment.
+func (art *AllocationTotals) TotalRAMCost() float64 {
+	return art.RAMCost + art.RAMCostAdjustment
+}
```
michaelmdresser, line 60: What do you think about calling these `Adjusted...Cost()`?

michaelmdresser: (same applies to `AssetTotals`)

michaelmdresser (approved):
Still lookin' good!

pkg/kubecost/totals.go
nikovacevic: I agree that the naming with adjustments is weird. My goal is to communicate that *this is the one to use* so that people don't unwittingly use the unadjusted costs, thinking that "adjusted costs" aren't what they need—they just need to basic CPUCost. I don't know, maybe that's a bad justification.

In `allocation.go` we use `CPUTotalCost()` instead of `TotalCPUCost()` so I probably should've just stuck to convention. In general, I think this is messy no matter what, unfortunately.

I'm leaning toward keeping this the way it is, for the sake of ease. But I agree that it's not perfect, and that makes me a little sad. If we decide we like `AdjustedCPUCost()` then maybe we can take a pass and replace them all as a clean-up effort.

PR #989, opened 2021-11-17
Add logic for custom pricing values in assets.

kaelanspatel:
## What does this PR change?
Fixes an issue with allocation by adding logic for prices set in custompricing to override Node/Disk costs in assets, such that reconciliation behavior would be correct. Also, adds a unit test that will hopefully catch a similar issue before it happens.

I had thought about creating a test for custompricing post-ETL run, but since there are already tests for the pricing as well as the reconciliation, I think that's redundant.

No.

Fixes an issue where custom pricing was not being added to assets.

- 
- 

## How was this PR tested?
Tested allocation/asset behavior before and after by setting custom pricing and comparing.

No.

### DIFF
--- a/pkg/costmodel/cluster_helpers.go
+++ b/pkg/costmodel/cluster_helpers.go
@@ -1,8 +1,10 @@
 package costmodel
 
 import (
-	"github.com/kubecost/cost-model/pkg/cloud"
+	"strconv"
 	"time"
+
+	"github.com/kubecost/cost-model/pkg/cloud"
 
 	"github.com/kubecost/cost-model/pkg/env"
 	"github.com/kubecost/cost-model/pkg/log"
@@ -28,6 +30,8 @@ func mergeTypeMaps(clusterAndNameToType1, clusterAndNameToType2 map[nodeIdentifi
 
 func buildCPUCostMap(
 	resNodeCPUCost []*prom.QueryResult,
+	cp cloud.Provider,
+	preemptible map[NodeIdentifier]bool,
 ) (
 	map[NodeIdentifier]float64,
 	map[nodeIdentifierNoProviderID]string,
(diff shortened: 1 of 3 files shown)


pkg/costmodel/cluster_helpers.go
```diff
@@ -63,8 +71,33 @@ func buildCPUCostMap(
...
+
+		} else {
+
+			cpuCost = result.Values[0].Value
+
+		}
+
 		clusterAndNameToType[keyNon] = nodeType
 
+		log.Infof("ASSET CLUSTERNODES NODE: %s WITH CPUPRICE: %f", key.Name, cpuCost)
```
michaelmdresser, line 99: Is this a log from your debugging?

pkg/costmodel/cluster_helpers.go
```diff
@@ -108,7 +147,33 @@ func buildRAMCostMap(
...
+
+		} else {
+
+			ramCost = result.Values[0].Value
+
+		}
+
 		clusterAndNameToType[keyNon] = nodeType
+
+		log.Infof("ASSET CLUSTERNODES NODE: %s WITH RAMPRICE: %f", key.Name, ramCost)
```
michaelmdresser, line 175: Likewise.

pkg/costmodel/cluster_helpers.go
```diff
@@ -162,6 +256,8 @@ func buildGPUCostMap(
 			gpuCostMap[key] = gpuCost
 		}
 
+		log.Infof("ASSET CLUSTERNODES NODE: %s WITH GPUPRICE: %f", key.Name, gpuCost)
```
michaelmdresser, line 259: Likewise.

michaelmdresser (commented):
- I'm happy to see a unit test for this!
- I see a few extraneous logs.
- Based on this code, have we decided that custom prices will not affect emitted metrics?

kaelanspatel:
Thanks for the review! Removed those leftover logs...

Also, in regards to custom pricing and metrics, custom pricing overrides metrics retroactively. So yes, AFAIK custom prices in no way effect the metrics themselves.

pkg/costmodel/cluster_helpers.go
kaelanspatel: Yes, removed.

pkg/costmodel/cluster_helpers.go
kaelanspatel: Removed.

pkg/costmodel/cluster_helpers.go
kaelanspatel: Removed.

michaelmdresser (approved):
LGTM! Good job dealing with the early ambiguity and discussing with the team.

PR #994, opened 2021-11-18
Detect the correct azure cloud api object to use by billing region

Sean-Holcomb:
## What does this PR change?
This PR adds a method that switches out the azure SDK environment object that is used based on the configured billing region. The environment object contains the correct endpoints for that particular cloud.

- 
- 

Users on US gov cloud and China cloud should be able to use the rate card api

- https://github.com/kubecost/cost-model/issues/990
- https://kubecost.zendesk.com/agent/tickets/995

## How was this PR tested?
This PR defaults to the original functionality, and continues to work on my clusters, but I will have to rely on user feedback to ensure that the gov cloud option is working.

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -929,9 +929,6 @@ func determineCloudByRegion(region string) azure.Environment{
 	if strings.Contains(lcRegion, "china") {
 		return azure.ChinaCloud
 	}
-	if strings.Contains(lcRegion, "germany") {
-		return azure.GermanCloud
-	}
 	if strings.Contains(lcRegion, "gov") || strings.Contains(lcRegion, "dod") {
 		return azure.USGovernmentCloud
 	}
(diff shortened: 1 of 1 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -923,6 +923,19 @@ func (az *Azure) DownloadPricingData() error {
...
 
+// determineCloudByRegion uses region name to pick the correct Cloud Environment for the azure provider to use
+func determineCloudByRegion(region string) azure.Environment{
+	lcRegion := strings.ToLower(region)
+	if strings.Contains(lcRegion, "china") {
+		return azure.ChinaCloud
+	}
+	if strings.Contains(lcRegion, "gov") || strings.Contains(lcRegion, "dod") {
+		return azure.USGovernmentCloud
+	}
```
michaelmdresser, line 934: Is there any possibility of `region` containing one of these substrings but not being what we expect? I'm not sure about the possible options for `region` -- maybe we could do `strings.StartsWith` if all China regions start with `china` for example. I can't seem to find a full list of Azure regions that includes China and govcloud, so I'm not sure how to evaluate this.

Sean-Holcomb: I could do starts with there, would be a little more robust.
This is the link you are looking for https://azure.microsoft.com/en-us/global-infrastructure/geographies/#geographies

Sean-Holcomb: and yeah was a little concerned about "dod" and "gov" but there are no collisions there

michaelmdresser (approved):

PR #996, opened 2021-11-24
Fix Azure norwayeast instance pricing not being parsed correctly

kaelanspatel:
## What does this PR change?
Support norwayeast node pricing. Technically a fix as data already exists in API response + pricing logic, it was just being filtered out incorrectly. 

Also, added logs for when long-running pricing API requests fail, which can be missed when only returned e.g. when navigating away from settings.html before change to settings actually happens, page will report success but query to pricing can still fail silently.

No.

Support node pricing in Azure norwayeast region.

- https://kubecost.zendesk.com/agent/tickets/995
- https://github.com/kubecost/cost-model/issues/984

## How was this PR tested?
Tested on kubecost running on fresh Azure norwayeast Standard_DS2_v2 on-demand node and observing pricing configurations.

No.

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -358,6 +358,9 @@ func (a *Accesses) RefreshPricingData(w http.ResponseWriter, r *http.Request, ps
 	w.Header().Set("Access-Control-Allow-Origin", "*")
 
 	err := a.CloudProvider.DownloadPricingData()
+	if err != nil {
+		klog.V(1).Infof("Error refreshing pricing data: %s", err.Error())
+	}
 
 	w.Write(WrapData(nil, err))
 }
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -59,6 +59,7 @@ var (
 		"uk": "uk",
 		"us": "us",
 		"za": "southafrica",
+		"no": "norway",
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/costmodel/router.go
```diff
@@ -358,6 +358,9 @@ func (a *Accesses) RefreshPricingData(w http.ResponseWriter, r *http.Request, ps
 	w.Header().Set("Access-Control-Allow-Origin", "*")
 
 	err := a.CloudProvider.DownloadPricingData()
+	if err != nil {
+		klog.V(1).Infof("Error refreshing pricing data: %s", err.Error())
```
michaelmdresser, line 362: Should this be a `Warningf` or `Errorf` instead of `Infof`?

michaelmdresser (approved):
LGTM! 

Any suggestions on how we can improve the "silently fail" frontend experience? Would you open a short issue in cost-analyzer-helm-chart explaining the problem?

pkg/costmodel/router.go
kaelanspatel: Maybe? In this case I was just following the convention that was already in place e.g. [here](https://github.com/kubecost/cost-model/blob/c900910117b48d5f6886125f3d63f4f348066718/pkg/costmodel/router.go#L1572-L1575).

The error should go to the frontend anyway, this is just as a "hey look this is happening" in the kubernetes log in case you click off that settings page.

PR #997, opened 2021-12-02
Implement Contract for Window (opposite of Expand)

nikovacevic:
## What does this PR change?
* Implements `Contract`, `ContractStart` and `ContractEnd` for type `Window`, which does the opposite of `Expand`

## How was this PR tested?
* Unit tests forthcoming. Being used for a closed-source PR at the moment.

### DIFF
--- a/pkg/kubecost/window.go
+++ b/pkg/kubecost/window.go
@@ -3,7 +3,6 @@ package kubecost
 import (
 	"bytes"
 	"fmt"
-	"github.com/kubecost/cost-model/pkg/util/timeutil"
 	"math"
 	"regexp"
 	"strconv"
@@ -11,6 +10,7 @@ import (
 
 	"github.com/kubecost/cost-model/pkg/env"
 	"github.com/kubecost/cost-model/pkg/thanos"
+	"github.com/kubecost/cost-model/pkg/util/timeutil"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


AjayTripathy (approved):

pkg/kubecost/window.go
```diff
@@ -410,6 +410,32 @@ func (w Window) Expand(that Window) Window {
...
+}
+
+func (w Window) Contract(that Window) Window {
+	if that.start != nil {
+		w = w.ContractStart(*that.start)
+	}
+
+	if that.end != nil {
+		w = w.ContractEnd(*that.end)
+	}
```
michaelmdresser, line 434: Why are we avoiding contracting if start/end is `nil`? `nil` implies an open (infinite?) window, right? I think the `ContractStart` and `ContractEnd` already handle this well.

nikovacevic: The idea is that if `that.start` or `that.end` is `nil` then there's no value of `w` that  could possibly be contracted (`nil` is, in that sense, the "max") so we don't have any actions to take. Is that fair to say? I want to write some unit tests for this when I have a chance!

michaelmdresser: Oh! I read this wrong, I thought we were checking if `w.start != nil`, whoops! Makes total sense!

PR #1000, opened 2021-12-06
Fix shared overhead on Summary Allocation API

nikovacevic:
## What does this PR change?
* Fixes a bug where, due to order of operations, shared overhead would not be shared unless other resources were shared.

## How was this PR tested?
* Integration tests: https://github.com/kubecost/kubecost-integration-tests/pull/28
* Manual testing on `gcp-niko`

### DIFF
--- a/pkg/kubecost/summaryallocation.go
+++ b/pkg/kubecost/summaryallocation.go
@@ -671,6 +671,7 @@ func (sas *SummaryAllocationSet) AggregateBy(aggregateBy []string, options *Allo
 
 			shareSet.Insert(&SummaryAllocation{
 				Name:       fmt.Sprintf("%s/%s", name, SharedSuffix),
+				Properties: &AllocationProperties{},
 				Start:      *sas.Window.Start(),
 				End:        *sas.Window.End(),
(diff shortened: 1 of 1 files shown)


pkg/kubecost/summaryallocation.go
```diff
@@ -654,6 +655,29 @@ func (sas *SummaryAllocationSet) AggregateBy(aggregateBy []string, options *Allo
...
+			}
+
+			totalSharedCost := cost * hours
+
+			shareSet.Insert(&SummaryAllocation{
+				Name:       fmt.Sprintf("%s/%s", name, SharedSuffix),
+				Start:      *sas.Window.Start(),
+				End:        *sas.Window.End(),
+				SharedCost: totalSharedCost,
+			})
```
nikovacevic, line 677: If we don't insert this before line 686 (just below) then we skip computing `shareCoeffs` altogether and nothing is shared. (Which, we skip for performance reasons unless it's necessary.)

Sean-Holcomb (approved):
It makes sense that that block would have to run earlier

mbolt35 (approved):
LGTM

PR #1002, opened 2021-12-08
Check start and end in ComputeAllocation and nudge times to be within queried window

nikovacevic:
## What does this PR change?
- In `ComputeAllocation`, nudge start and end times to be within the query window. This is in response to cases where start is 1m ahead of the query window's start.

- No, but goes with:  https://github.com/kubecost/kubecost-cost-model/pull/594

## How was this PR tested?
Integration tests:

### Before
Screenshot from 2021-12-07 18-16-23
Screenshot from 2021-12-07 18-16-38
Screenshot from 2021-12-07 18-16-50

### After
Screenshot from 2021-12-07 18-27-43


mbolt35 (approved):
I think this is a good constraint, looking forward to testing a bit with the query time :)

PR #1006, opened 2021-12-09
Ajay tripathy pv test

AjayTripathy:
## What does this PR change?
Add test for using the CSV Provider pipeline with PVs

No

It's a unit test, so no impact, just better maintenance.

## How was this PR tested?
Unit test was run

No

### DIFF
--- a/test/cloud_test.go
+++ b/test/cloud_test.go
@@ -92,6 +92,32 @@ func TestNodeValueFromMapField(t *testing.T) {
 
 }
 
+func TestPVPriceFromCSV(t *testing.T) {
+	nameWant := "pvc-08e1f205-d7a9-4430-90fc-7b3965a18c4d"
+	pv := &v1.PersistentVolume{}
+	pv.Name = nameWant
+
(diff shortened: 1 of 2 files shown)


test/cloud_test.go
```diff
@@ -93,6 +93,32 @@ func TestNodeValueFromMapField(t *testing.T) {
 
 }
 
+func TestPVPriceFromCSV(t *testing.T) {
+	nameWant := "pvc-08e1f205-d7a9-4430-90fc-7b3965a18c4d"
+	pv := &v1.PersistentVolume{}
+	pv.Name = nameWant
+
```
mbolt35, line 100: You'll want to pull the latest develop, add this line:

```
confMan := config.NewConfigFileManager(&config.ConfigFileManagerOpts{
    LocalConfigPath: "./",
})
```
Then your CSVProvider will need to use the following for `Config`:
```
Config: cloud.NewProviderConfig(confMan, "../configs/default.json"),
```

AjayTripathy:
fixed in anotehr PR

PR #1009, opened 2021-12-09
Run Mode Selection via Command Line Arguments

mbolt35:
## What does this PR change?
* Introduces a single entry point that allows a command line selection of which mode to run cost-model in.
  * `cost-model`(default): Providing no arguments or `cost-model` will run the application exactly the same as before this change. 
  * `agent`: Using the `agent` command will run the cost-model in metrics-exporter/agent mode.

## How was this PR tested?
* Tested in an existing cluster without setting any command line argument (backwards compatibility testing)
* Tested in an existing cluster with `cost-model` argument passed in the pod spec (explicit argument test)
* Tested in an existing agent cluster with the `agent` argument passed in.


pkg/cmd/commands.go
```diff
@@ -0,0 +1,118 @@
...
+		Use:   CommandCostModel,
+		Short: "Cost-Model metric exporter and data aggregator.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return costmodel.Execute(opts)
+		},
+	}
+
+	// TODO: We could introduce a way of mapping input command-line parameters to a configuration
+	// TODO: object, and pass that object to the agent Execute()
+	// cmCmd.Flags().<Type>VarP(&opts.<Property>, "<flag>", "<short>", <default>, "<usage>")
```
michaelmdresser, line 88: There's a pretty neat CLI library for Go called [kong](https://github.com/alecthomas/kong) that lets you do this with struct tags, might be worth a look!

cmd/costmodel/main.go
```diff
@@ -1,29 +1,22 @@
...
-func Healthz(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
-	w.WriteHeader(200)
-	w.Header().Set("Content-Length", "0")
-	w.Header().Set("Content-Type", "text/plain")
-}
-
 func main() {
-	a := costmodel.Initialize()
+	klog.InitFlags(nil)
+	flag.Set("v", "3")
```
michaelmdresser, line 13: I know this is pulled from legacy, but I'm curious if hardcoding the log level is really a good idea.

michaelmdresser (approved):
Looks great! One idea about an alternative CLI library for consideration's sake, but I don't have any major problems with cobra!

The `go.sum` got a lot of additional deps and versions, is that because of cobra?

cmd/costmodel/main.go
mbolt35: Yeah, I really like the simplicity of the klog library, but there are some weird things happening with this. I'm not sure if passing `-v` to the application overrides this or not. If that's the case, I'm fine with leaving the default at 3 (it looks like the default inside klog is 0).

PR #1010, opened 2021-12-10
Remove external allocation

Sean-Holcomb:
## What does this PR change?
This PR removes significant portions of dead/unsupported/duplicate code from the cloud providers on this repo. Things that this PR removes:
- Cross cloud (This has been replace by the enterprise Multi-cloud feature)
- ExternalAllocations API (Replaced by ExternalAllocation process in the ETL)
- Duplicate GCP BigQuery Access
- Duplicate Azure Storage Access
- /outOfClusterCosts endpoint
These features are all either broken, have been replaced or have not been supported for a while.

Additionally I did a refactor on the Athena pricing source code, as it had significant duplicate code, the change was to match it to similar functions in the CS project. While I did not remove this Athena Integration code this time I would like to note that it is my opinion that this code should be removed from OS also. It is my position that Cloud Billing integration is a CS feature in general but a I have specific concerns about how this feature is used. We are querying the CUR to retrieve past savings plan and RI rates to use as the emission rate for the matching nodes in the future. The problem with this is that RI and saving plan coverage frequently jumps from node to node. While this may make Kubecost more "right" in the general sense, it will make predictions less accurate on the individual node in the long run. All that being said, any of these problems will be cleaned up by reconciliation within 48 hours.

- 
- 

- 
- 

## How was this PR tested?
Check image of this build on clusters of each provider, For the most part these changes removed the entire code path that they operated in so there is little room for error. The exception is the some refactoring to reduce duplicate code on the Athena pricing source in AWS, this feature continued to query and run without error on my AWS cluster.


mbolt35:
> Cross cloud (This has been replace by the enterprise Multi-cloud feature)
I assumed this would involve getting rid of this: https://github.com/kubecost/cost-model/blob/392d23e698b120bb7ae7f13842385888411f81d0/pkg/cloud/provider.go#L388-L409

and all of the code that uses it. No?

dwbrown2:
Nice! Do you feel we have the test coverage needed for you to feel
comfortable making this change? If not, should we add some tests?!

Sean-Holcomb:
@mbolt35 All code that uses it is gone, but looks like I forgot to pull it.

@dwbrown2 For the most part this PR deals in removing dead code paths. As for the part that I did change (Athena pricing source code) the refactor was to match the way that KCM accesses Athena. Since the only thing that changed is the method which accesses Athena it is harder to automate the tests.

pkg/env/costmodelenv.go
```diff
@@ -16,6 +16,7 @@ const (
 	AWSAccessKeySecretEnvVar = "AWS_SECRET_ACCESS_KEY"
 	AWSClusterIDEnvVar       = "AWS_CLUSTER_ID"
 
+	AzureStorageSubscriptionIDEnvVar       = "AZURE_SUBSCRIPTION_ID"
```
michaelmdresser, line 19: I think this could use a `go fmt`

michaelmdresser (approved):
Yes! Destroy the dead code! :bomb: :boom:

PR #1019, opened 2021-12-22
Ajay tripathy add rcs

AjayTripathy:
## What does this PR change?
Add replicacontroller information to the cluster enumeration

No

Internal-only API changes.


mbolt35 (approved):
LGTM

PR #1026, opened 2022-01-07
Add suport to ISO8601 duration in window parser

souenzzo:
## What does this PR change?

Add suport to ISO8601 duration in window parser

With https://github.com/golang/go/issues/32313, it would be possible to do this PR without add an external library

Allow users to use [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) in any `window` parameter in the API.

It should simplify the creation of programmatic access to the API's. 

- https://github.com/kubecost/docs/issues/156#issuecomment-1007005157

## How was this PR tested?

Unit test at `pkg/kubecost/window_test.go`

No

# Extra Comment's

- I don't use go for a while and I'm not sure if I added the external library in the right way.


AjayTripathy:
Hi @souenzzo thank you so much for the PR and we appreciate the contribution! We are trying not accept any more BSD-3 licenses into the project so would prefer to wait for https://github.com/golang/go/issues/32313 rather than adding an external library.

mattray:
This is being closed as it is out of date and may no longer be relevant. Feel free to update the PR if you feel this is incorrect.

PR #1028, opened 2022-01-11
Ajay tripathy add frontend interoperability

AjayTripathy:
## What does this PR change?
Add Summarizer code to open source code to interoperate the kubecost community edition frontend!

No

Users will be able to either use the community edition backend or the open core backend to power the /allocation.html view on kubecost.

## How was this PR tested?
Swap the image on the helm chart install to the open core version

No, but we should do one and note that this is the recommended path to install the open source.


AjayTripathy:
@michaelmdresser should be ready for another pass.

pkg/costmodel/router.go
```diff
@@ -1599,6 +1599,7 @@ func Initialize(additionalConfigWatchers ...*watcher.ConfigMapWatcher) *Accesses
 	a.Router.GET("/costDataModelRange", a.CostDataModelRange)
 	a.Router.GET("/aggregatedCostModel", a.AggregateCostModelHandler)
 	a.Router.GET("/allocation/compute", a.ComputeAllocationHandler)
+	a.Router.GET("/allocation/compute/summary", a.ComputeAllocationHandlerSummary)
```
michaelmdresser, line 1602: Does registering this endpoint here cause any problems with the closed source build?

michaelmdresser (approved):
Pending the one question, this LGTM!

pkg/costmodel/router.go
AjayTripathy: Nope! The closed source is /allocation/summary. If you registered _that_ endpoint here it would fail, but /allocation/summary should be elsewhere registered (/cmd/costmodel.go)

PR #1031, opened 2022-01-13
Nil pointer dereference fix

nealormsbee:
## What does this PR change?
- Fixes a comment typo.

- Causes GPU idle coefficient to be calculated with GPU costs instead of RAM costs.

- Formats a key composition for checking a map correctly. Was trying to
  access map using node name as a key when sharing idle by node, but
needs to access using \<cluster\>/\<node\>.

- Checks to make sure that the key is present in
  allocTotalsAfterFilters, as well as allocTotals. I cannot fathom what
would cause a key to be missing from this map, but that seems to be the
error reported, and I do know that a missing key likely means that the
subsequent coefficient we are attempting to compute should simply be 0.

No

Fixes a bug where requesting an allocation summary with `shareIdle=false&idleByNode=true` would crash the pod.
Also guards against another potential nil pointer dereference in the case where two maps are desynced.

N/A

## How was this PR tested?
An integration test is being written in our integration test suite.
Additionally, manually by querying `/allocation/summary` with different combinations of filters, shareIdle, and idleByNode options. Notably:

- All combinations of `shareIdle` and `idleByNode` true/false values
- Filtering by namespace, cluster, and node

No

### DIFF
--- a/pkg/kubecost/summaryallocation.go
+++ b/pkg/kubecost/summaryallocation.go
@@ -640,7 +640,7 @@ func (sas *SummaryAllocationSet) AggregateBy(aggregateBy []string, options *Allo
 	}
 
 	// If we're recording allocTotalsAfterFilters and there are shared costs,
-	// then record those resource totals here so that idle for thpse shared
+	// then record those resource totals here so that idle for those shared
 	// resources gets included.
 	if allocTotalsAfterFilters != nil {
 		for key, rt := range sharedResourceTotals {
@@ -822,7 +822,7 @@ func (sas *SummaryAllocationSet) AggregateBy(aggregateBy []string, options *Allo
 
 			var key string
 			if options.IdleByNode {
-				key = ia.Properties.Node
+				key = fmt.Sprintf("%s/%s", ia.Properties.Cluster, ia.Properties.Node)
 			} else {
 				key = ia.Properties.Cluster
 			}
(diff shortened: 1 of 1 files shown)


pkg/kubecost/summaryallocation.go
```diff
@@ -822,7 +822,7 @@ func (sas *SummaryAllocationSet) AggregateBy(aggregateBy []string, options *Allo
 
 			var key string
 			if options.IdleByNode {
-				key = ia.Properties.Node
+				key = fmt.Sprintf("%s/%s", ia.Properties.Cluster, ia.Properties.Node)
```
nealormsbee, line 825: I believe this was the cause of the reported error. `cluster/node` is used in all of the other key creation in this function if `idleByNode` is true. Here, it was just `node`.

pkg/kubecost/summaryallocation.go
```diff
@@ -831,18 +831,32 @@ func (sas *SummaryAllocationSet) AggregateBy(aggregateBy []string, options *Allo
 			// which equals the proportion of filtered-to-actual cost.
 			cpuFilterCoeff := 0.0
 			if allocTotals[key].CPUCost > 0.0 {
-				cpuFilterCoeff = allocTotalsAfterFilters[key].CPUCost / allocTotals[key].CPUCost
+				filteredAlloc, ok := allocTotalsAfterFilters[key]
+				if ok {
+					cpuFilterCoeff = filteredAlloc.CPUCost / allocTotals[key].CPUCost
+				} else {
+					cpuFilterCoeff = 0.0
+				}
```
nealormsbee, line 839: I was unable to create a situation where `allocTotals` had a key for an idle allocation that `allocTotalsAfterFilters` did not. Planning to ask Niko about it when he gets back, because maybe it's possible.

In any case, the initial stack trace we saw lead us to believe that `key` may be present in `allocTotals` but not `allocTotalsAfterFilters`. If that is the case, then I believe the allocation in question would have been filtered out, and the correct action is to set this ratio to 0

pkg/kubecost/summaryallocation.go
```diff
@@ -831,18 +831,32 @@ func (sas *SummaryAllocationSet) AggregateBy(aggregateBy []string, options *Allo
...
 			}
 
 			gpuFilterCoeff := 0.0
-			if allocTotals[key].RAMCost > 0.0 {
-				gpuFilterCoeff = allocTotalsAfterFilters[key].RAMCost / allocTotals[key].RAMCost
+			if allocTotals[key].GPUCost > 0.0 {
+				filteredAlloc, ok := allocTotalsAfterFilters[key]
```
nealormsbee, line 844: gpuCoeff was being determined using RAMKeys, which I think is a typo of a copy/paste miss.

nealormsbee (commented):
Left a few clarifying comments about the changes

michaelmdresser (approved):
LGTM! Thanks for the overview in review comments, it helped a lot.

nealormsbee:
Thanks for the review, @michaelmdresser !

@AjayTripathy this PR is against `develop`, but there's another one open against `master`. Just let me know whether you want to do a patch release and I'll merge one & close the other.

pkg/kubecost/summaryallocation.go
nikovacevic: Wow good catch! Thank you!

pkg/kubecost/summaryallocation.go
nikovacevic: Argh, yes, good catch.

pkg/kubecost/summaryallocation.go
nikovacevic: Happy to talk about this one if you want to. In general, the fix looks good :sunglasses:

PR #1040, opened 2022-01-25
Implement AllocationSetRange.AccumulateBy

alejandro-kubecost:
## What does this PR change?
A  new AccumulateBy function that sums AllocationSets based on the given time duration and it's interoperability with the resolution of the AllocationSets. A lockless window() function was also added in which the original Window() function now calls to offer backwards compatibility. Functionality should be the same

- This can stand alone but I imagine won't be accessible until exposed https://github.com/kubecost/kubecost-cost-model/pull/630

- The write up can be included in the final PR exposing the new function via it's parameter

- 
- 

## How was this PR tested?
- New tests were written in `allocation_test.go` testing nils, empty sets, happy paths, and combinations of the 3. Further testing can be done if desirable

- Soon to follow the PR merges

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2260,51 +2260,22 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
 	return allocSet, nil
 }
 
-// TODO accumulate into lower-resolution chunks of the given resolution
 func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*AllocationSetRange, error) {
 	allocSetRange := &AllocationSetRange{allocations: []*AllocationSet{}}
 	var allocSet *AllocationSet
 	var err error
 
+	//This uses a lock and can't be called after locking asr
+	asrWindow := asr.Window()
+
 	asr.Lock()
 	defer asr.Unlock()
 
-	// use window() with no lock
-	if asr.window().IsEmpty() {
+	if asrWindow.IsEmpty() {
 		return asr, nil
 	}
 
-	// Call total accumulate func if resolution is greater than total window duration
-	if resolution > asr.window().Duration() {
-		// unable to acquire lock here if old asr.accumulate is called
-		for _, as := range asr.allocations {
-			allocSet, err = allocSet.accumulate(as)
-			if err != nil {
-				return nil, err
-			}
-		}
-		allocSetRange.allocations = append(allocSetRange.allocations, allocSet)
-		return allocSetRange, nil
-	}
-
-	// two ways to get end of asr window
-	// 1. check if last set of asr
-	// 2. check if set window.end is asr.window.end
-	// both require no lock to be present
-	// option 1: 2 asr.window methods
-	// option 2: 2 window.length methods
-	asrWindow := asr.window()
-	asrResolution := asrWindow.Duration()
-
-	// if asrResolution < 1 hour then there is not enough data to accumulate
-	// if as.allocations(end - start) > resolution then no need, it's already accumulated. Requery to make more granular in the future?
-	// check this again?
-	if asrResolution <= time.Hour || asrResolution < resolution {
-		return asr, nil
-	}
-
-	// do not compound
-	// check as.window and accumulate till windowSum == time.duration wanted -> add accumulated set to allocSetRange
+	// check as.window and accumulate sets until sum of window durations == time.duration wanted -> add accumulated set to allocSetRange
 	var currAccumulatedSum time.Duration
 	for _, as := range asr.allocations {
 		allocSet, err = allocSet.accumulate(as)
@@ -2495,26 +2466,12 @@ func (asr *AllocationSetRange) UTCOffset() time.Duration {
 // Window returns the full window that the AllocationSetRange spans, from the
 // start of the first AllocationSet to the end of the last one.
 func (asr *AllocationSetRange) Window() Window {
-	asr.Lock()
-	defer asr.Unlock()
-	return asr.window()
-}
-
-func (asr *AllocationSetRange) window() Window {
-	var length int
-
-	if asr == nil || asr.allocations == nil {
-		length = 0
-	} else {
-		length = len(asr.allocations)
-	}
-
-	if asr == nil || length == 0 {
+	if asr == nil || asr.Length() == 0 {
 		return NewWindow(nil, nil)
 	}
 
 	start := asr.allocations[0].Start()
-	end := asr.allocations[length-1].End()
+	end := asr.allocations[asr.Length()-1].End()
 
 	return NewWindow(&start, &end)
 }
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2269,9 +2269,6 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 	asr.Lock()
 	defer asr.Unlock()
 
-	fmt.Printf("asr window duration: %v\n", asr.window().Duration())
-	fmt.Printf("resolution: %v\n", resolution)
-
 	// use window() with no lock
 	if asr.window().IsEmpty() {
 		return asr, nil
@@ -2279,7 +2276,7 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 
 	// Call total accumulate func if resolution is greater than total window duration
 	if resolution > asr.window().Duration() {
-		// unable to acquire lock here if old accumulate is called
+		// unable to acquire lock here if old asr.accumulate is called
 		for _, as := range asr.allocations {
 			allocSet, err = allocSet.accumulate(as)
 			if err != nil {
@@ -2304,14 +2301,13 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 	// check as.window and accumulate till windowSum == time.duration wanted -> add accumulated set to allocSetRange
 	var currAccumulatedSum time.Duration
 	for _, as := range asr.allocations {
-		//What time group are we in? Is time group bigger than duration?
 		allocSet, err = allocSet.accumulate(as)
 		if err != nil {
 			return nil, err
 		}
 		currAccumulatedSum += allocSet.Window.Duration()
 
-		fmt.Printf("asr end %v as end %v\n", asrWindow.end, allocSet.Window.end)
+		// fmt.Printf("asr end %v as end %v\n", asrWindow.end, allocSet.Window.end)
 		// two ways to get end of asr window
 		// 1. check if last set of asr
 		// 2. check if set window.end is asr.window.end
@@ -2519,8 +2515,6 @@ func (asr *AllocationSetRange) window() Window {
 	start := asr.allocations[0].Start()
 	end := asr.allocations[length-1].End()
 
-	fmt.Printf("start %v\n", start)
-	fmt.Printf("end %v\n", end)
 	return NewWindow(&start, &end)
 }
 
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2273,7 +2273,6 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 
 	asr.Lock()
 	defer asr.Unlock()
-
 	for i, as := range asr.allocations {
 		allocSet, err = allocSet.accumulate(as)
 		if err != nil {
@@ -2281,7 +2280,7 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 		}
 
 		if allocSet != nil {
-			currAccumulatedSum += allocSet.Window.Duration()
+			currAccumulatedSum = allocSet.Window.Duration()
 
 			// check if end of asr to sum the final set
 			// If total asr accumulated sum <= resolution return 1 accumulated set
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2261,10 +2261,12 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
 }
 
 // AccumulateBy sums AllocationSets based on the resolution given. The resolution given is subject to the scale used for the AllocationSets.
+// Resolutions not evenly divisible by the AllocationSetRange window durations accumulate sets until a sum greater than or equal to the resolution is met,
+// at which point AccumulateBy will start summing from 0 until the requested resolution is met again.
 // If the requested resolution is smaller than the window of an AllocationSet then the resolution will default to the duration of a set.
 // Resolutions larger than the duration of the entire AllocationSetRange will default to the duration of the range.
 func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*AllocationSetRange, error) {
-	allocSetRange := &AllocationSetRange{allocations: []*AllocationSet{}}
+	allocSetRange := NewAllocationSetRange()
 	var allocSet *AllocationSet
 	var currAccumulatedSum time.Duration
 	var err error
@@ -2282,10 +2284,10 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 			currAccumulatedSum += allocSet.Window.Duration()
 
 			// check if end of asr to sum the final set
-			// If accumulated sum !>= resolution return 1 accumulated set
+			// If total asr accumulated sum <= resolution return 1 accumulated set
 			if currAccumulatedSum >= resolution || i == len(asr.allocations)-1 {
 				allocSetRange.allocations = append(allocSetRange.allocations, allocSet)
-				allocSet = &AllocationSet{allocations: map[string]*Allocation{}}
+				allocSet = NewAllocationSet(time.Time{}, time.Time{})
 				currAccumulatedSum = 0
 			}
 		}
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2272,15 +2272,8 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 	asr.Lock()
 	defer asr.Unlock()
 
-	// // use window() with no lock
-	if asr.window().IsEmpty() {
-		return asr, nil
-	}
-
-	asrWindow := asr.window()
-
 	// check as.window and accumulate sets until sum of window durations == time.duration wanted -> add accumulated set to allocSetRange
-	for _, as := range asr.allocations {
+	for i, as := range asr.allocations {
 		allocSet, err = allocSet.accumulate(as)
 		if err != nil {
 			return nil, err
@@ -2289,8 +2282,9 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
 		if allocSet != nil {
 			currAccumulatedSum += allocSet.Window.Duration()
 
-			// check if end of asr to sum the final set too
-			if currAccumulatedSum >= resolution || NewWindow(nil, asrWindow.end).Equal(NewWindow(nil, allocSet.Window.end)) {
+			// check if end of asr to sum the final set
+			// If accumulated sum never reaches the desired resolution then all sets are returned as 1 accumulated set
+			if currAccumulatedSum >= resolution || i == len(asr.allocations)-1 {
 				allocSetRange.allocations = append(allocSetRange.allocations, allocSet)
 				allocSet = &AllocationSet{allocations: map[string]*Allocation{}}
 				currAccumulatedSum = 0
@@ -2469,20 +2463,12 @@ func (asr *AllocationSetRange) UTCOffset() time.Duration {
 // Window returns the full window that the AllocationSetRange spans, from the
 // start of the first AllocationSet to the end of the last one.
 func (asr *AllocationSetRange) Window() Window {
-	asr.Lock()
-	defer asr.Unlock()
-	return asr.window()
-}
-
-// (*AllocationSetRange).window() is not safe for concurrent use as it does not acquire a lock.
-// For the safe method please use (*AllocationSetRange).Window()
-func (asr *AllocationSetRange) window() Window {
-	if asr == nil || len(asr.allocations) == 0 {
+	if asr == nil || asr.Length() == 0 {
 		return NewWindow(nil, nil)
 	}
 
 	start := asr.allocations[0].Start()
-	end := asr.allocations[len(asr.allocations)-1].End()
+	end := asr.allocations[asr.Length()-1].End()
 
 	return NewWindow(&start, &end)
 }
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2263,20 +2263,20 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
 func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*AllocationSetRange, error) {
 	allocSetRange := &AllocationSetRange{allocations: []*AllocationSet{}}
 	var allocSet *AllocationSet
+	var currAccumulatedSum time.Duration
 	var err error
-
-	//This uses a lock and can't be called after locking asr
-	asrWindow := asr.Window()
 
 	asr.Lock()
 	defer asr.Unlock()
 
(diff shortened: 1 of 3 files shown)


pkg/kubecost/allocation.go
```diff
@@ -2261,7 +2261,70 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+	if resolution > asr.window().Duration() {
+		// unable to acquire lock here if old asr.accumulate is called
+		for _, as := range asr.allocations {
+			allocSet, err = allocSet.accumulate(as)
+			if err != nil {
+				return nil, err
+			}
+		}
+		allocSetRange.allocations = append(allocSetRange.allocations, allocSet)
+		return allocSetRange, nil
```
michaelmdresser, line 2287: This is basically a reimplementation of `(*AllocationSetRange).Accumulate()`, right? I think this is okay, but does the standard logic you've written for `AccumulateBy` not work in the `resolution > asr.window().Duration()` case?

pkg/kubecost/allocation.go
```diff
@@ -2261,7 +2261,70 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+
+	asrWindow := asr.window()
+	asrResolution := asrWindow.Duration()
+
+	// if asrResolution < 1 hour then there is not enough data to accumulate
+	// if as.allocations(end - start) > resolution then no need, it's already accumulated. Requery to make more granular in the future?
+	// check this again?
+	if asrResolution <= time.Hour || asrResolution < resolution {
+		return asr, nil
+	}
```
michaelmdresser, line 2304: I think the `asrResolution <= time.Hour` check shouldn't be here. This is ETL implementation-/configuration-specific information that shouldn't matter for a generic `AccumulateBy()` on a data type.

`asrResolution < resolution` looks like a possibly-unnecessary performance improvement to avoid copying. Does the standard logic not handle this case?

pkg/kubecost/allocation.go
```diff
@@ -2261,7 +2261,70 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+
+	// do not compound
```
michaelmdresser, line 2306: What does "do not compound" mean?

pkg/kubecost/allocation.go
```diff
@@ -2261,7 +2261,70 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+		// fmt.Printf("asr end %v as end %v\n", asrWindow.end, allocSet.Window.end)
+		// two ways to get end of asr window
+		// 1. check if last set of asr
+		// 2. check if set window.end is asr.window.end
+		// both require no lock to be present
+		// option 1: 2 asr.window methods
+		// option 2: 2 window.length methods
+
+		// check if end of asr to sum the final set too
+		if currAccumulatedSum >= resolution || NewWindow(nil, asrWindow.end).Equal(NewWindow(nil, allocSet.Window.end)) {
```
michaelmdresser: If you want to know if you have the final `AllocationSet` in `asr.allocations`, you can actually change your loop to include the index in the slice:

```go
for i, as := range asr.allocations {
    // ...

    if i == len(asr.allocations) - 1 {
        // do something because this is the last AllocationSet in asr
    } else if currAccumulatedSum >= resolution {
        // do something
    }
}
```

pkg/kubecost/allocation.go
```diff
@@ -2431,12 +2494,26 @@ func (asr *AllocationSetRange) UTCOffset() time.Duration {
 // Window returns the full window that the AllocationSetRange spans, from the
 // start of the first AllocationSet to the end of the last one.
 func (asr *AllocationSetRange) Window() Window {
-	if asr == nil || asr.Length() == 0 {
+	asr.Lock()
+	defer asr.Unlock()
+	return asr.window()
+}
+
+func (asr *AllocationSetRange) window() Window {
```
michaelmdresser, line 2503: I'd like to see a comment here explaining that `.window()` isn't safe for concurrent access unless the lock has already been acquired.

pkg/kubecost/allocation.go
```diff
@@ -2431,12 +2494,26 @@ func (asr *AllocationSetRange) UTCOffset() time.Duration {
...
+	var length int
+
+	if asr == nil || asr.allocations == nil {
+		length = 0
+	} else {
+		length = len(asr.allocations)
+	}
+
+	if asr == nil || length == 0 {
```
michaelmdresser, line 2512: Go convention generally asks you to return early rather than do this sort of if statement followed by a later check and return. I think the logic that was present beforehand adhered to this, like so:

```go
if asr == nil || asr.allocations == nil {
    return NewWindow(nil, nil)
}
```

There's also decent amount of repetition here that I'm not a huge fan of. Why not just leave the original logic as it was?

pkg/kubecost/allocation_test.go
```diff
@@ -2068,8 +2068,254 @@ func TestAllocationSetRange_Accumulate(t *testing.T) {
...
+	tomorrow := time.Now().UTC().Truncate(day).Add(day)
+	dur := time.Hour * 24 * 2
+
+	// Accumulating any combination of nil and/or empty set should result in empty set
+	// test 1 nil set
+	result, err := NewAllocationSetRange(nil).AccumulateBy(dur)
+	for _, as := range result.allocations {
+		if err != nil {
+			t.Fatalf("unexpected error accumulating nil AllocationSetRange: %s", err)
+		}
```
michaelmdresser, line 2085: This error check should be pulled out of the loop

michaelmdresser: I'd also recommend using `Errorf` over `Fatalf` if later test logic can function even if this specific check fails. `Fatalf` causes an immediate exit while `Errorf` will continue running the test code but return this error. Because you've chosen to run multiple tests in one test function, I'd rewrite most of the `Fatalf`s. Then, if someone breaks multiple things, they get all the errors at once instead of one at a time :grin: 

https://pkg.go.dev/testing#T.Fatalf

pkg/kubecost/allocation.go
```diff
@@ -2261,7 +2261,71 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+			currAccumulatedSum += allocSet.Window.Duration()
+
+			// check if end of asr to sum the final set too
+			if currAccumulatedSum >= resolution || NewWindow(nil, asrWindow.end).Equal(NewWindow(nil, allocSet.Window.end)) {
+				allocSetRange.allocations = append(allocSetRange.allocations, allocSet)
+				allocSet = &AllocationSet{allocations: map[string]*Allocation{}}
+				currAccumulatedSum = 0
+			}
+		}
+	}
```
michaelmdresser, line 2325: If you want to know if you have the final `AllocationSet` in `asr.allocations`, you can actually change your loop to include the index in the slice:

```go
for i, as := range asr.allocations {
    // ...

    if i == len(asr.allocations) - 1 {
        // do something because this is the last AllocationSet in asr
    } else if currAccumulatedSum >= resolution {
        // do something
    }
}
```

pkg/kubecost/allocation.go
```diff
@@ -2261,7 +2261,71 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+		allocSet, err = allocSet.accumulate(as)
+		if err != nil {
+			return nil, err
+		}
+
+		if allocSet != nil {
+			currAccumulatedSum += allocSet.Window.Duration()
+
+			// check if end of asr to sum the final set too
+			if currAccumulatedSum >= resolution || NewWindow(nil, asrWindow.end).Equal(NewWindow(nil, allocSet.Window.end)) {
```
michaelmdresser, line 2319: What happens if I `AccumulateBy` a `resolution` that doesn't divide nicely? For example:

```
asr = [
    AS1 // 1d Window
    AS2 // 1d Window
    AS3 // 1d Window
]
```

What happens if I AccumulateBy 36 hours? I'm pretty sure the code will treat this as equivalent to calling AccumulateBy with a 2d duration because of the sum and >= logic here.

I think this is worth discussing (unless you already have!) and then documenting whatever is decided upon in a comment on `AccumulateBy` so IDEs can pick it up and show it to package users. Then we can also add a test for it.

pkg/kubecost/allocation_test.go
```diff
@@ -2068,8 +2068,261 @@ func TestAllocationSetRange_Accumulate(t *testing.T) {
...
+	// test 2 nil sets 2 1d sets
+	result, err = NewAllocationSetRange(nil, NewAllocationSet(ago2d, yesterday), nil, NewAllocationSet(today, tomorrow), nil).AccumulateBy(dur)
+	for _, as := range result.allocations {
+		if err != nil {
+			t.Fatalf("unexpected error accumulating nil AllocationSetRange: %s", err)
+		}
+		if !as.IsEmpty() {
+			t.Fatalf("accumulating nil AllocationSetRange: expected empty; actual %s", result)
+		}
+	}
```
michaelmdresser, line 2133: I'd recommend moving all of these similar-looking tests to a separate unit test that loops over cases. See https://github.com/kubecost/cost-model/blob/9d9af5583c5bfea20052dcc71bd3606a37b7202a/pkg/kubecost/window_test.go#L769-L781 for an example of the cases format.

So maybe these tests would move to `TestAllocationSetRange_AccumulateBy_Nils(t *testing.T)` or something similar, and the rest of the code in this test can consider happier path logic.

michaelmdresser (commented):
This is a great start! I'm a huge fan of the testing. I'm sorry you had to run into concurrency and locking so early in your time with Go :sweat_smile: 

As ever with code reviews, feel free to push back if you think I'm asking for something that doesn't make sense!

pkg/kubecost/allocation.go
alejandro-kubecost: nothing now :)
removed

pkg/kubecost/allocation.go
alejandro-kubecost: yep, agreed this was missed after some changes. Removed

pkg/kubecost/allocation.go
alejandro-kubecost: Great question, this was partially discussed with @nikovacevic in regards to how you would handle the (in my opinion) similar situation of asking for a resolution more granular than what is available. 

E.g. Requesting 30min resolutions on sets that are split into 1h windows:
To achieve such a granularity (at least accurately) would mean having to do a new query that is suitable for such a resolution, something I don't think is worth implementing given the required changes. I think it would be a nice to have in the future though

pkg/kubecost/allocation.go
alejandro-kubecost: I wasn't sure on what our stance was for performance gains vs extra logic. Given that this isn't a function reliant on performance so much I agree. Removed

pkg/kubecost/allocation.go
alejandro-kubecost: Yeah this was the section I mentioned that I really didn't like either. I did manage to find a good work around by calling asr.Window() before locking asr and can remove all of this. Let me know if this work around has any inherent issues

pkg/kubecost/allocation_test.go
alejandro-kubecost: cool, thanks for the tip!

pkg/kubecost/allocation.go
alejandro-kubecost: Specifically this line:

https://github.com/kubecost/cost-model/pull/1040/files#diff-7c339daaa6b130f0e74be88b76e4c35d3941edf78f886fae793772b8f566c58cR2269

michaelmdresser: > However does this have the potential of asr data changing in between unlocking and locking again even if in succession?

Yes, I prefer your initial private, non-locking, non-safe `.window()` over trying to lock twice in `AccumulateBy`.

pkg/kubecost/allocation_test.go
```diff
@@ -2068,8 +2068,254 @@ func TestAllocationSetRange_Accumulate(t *testing.T) {
...
+	if alloc.RAMCost != 2.0 {
+		t.Fatalf("accumulating AllocationSetRange: expected 2.0; actual %f", alloc.RAMCost)
+	}
+	if alloc.RAMEfficiency() != 1.0 {
+		t.Fatalf("accumulating AllocationSetRange: expected 1.0; actual %f", alloc.RAMEfficiency())
+	}
+	// if alloc.TotalCost() != 12.0 {
+	// 	t.Fatalf("accumulating AllocationSetRange: expected 12.0; actual %f", alloc.TotalCost())
+	// }
```
michaelmdresser: Should this be commented out?

pkg/kubecost/allocation_test.go
```diff
@@ -2068,8 +2068,254 @@ func TestAllocationSetRange_Accumulate(t *testing.T) {
...
+	// result, err = NewAllocationSetRange(nil, todayAS, nil).Accumulate()
+	// if err != nil {
+	// 	t.Fatalf("unexpected error accumulating AllocationSetRange of length 1: %s", err)
+	// }
+	// if result == nil {
+	// 	t.Fatalf("accumulating AllocationSetRange: expected AllocationSet; actual %s", result)
+	// }
+	// if result.TotalCost() != 6.0 {
+	// 	t.Fatalf("accumulating AllocationSetRange: expected total cost 6.0; actual %f", result.TotalCost())
+	// }
```
michaelmdresser: Are these test cases you want to keep?

pkg/kubecost/allocation.go
michaelmdresser: > in regards to how you would handle the (in my opinion) similar situation of asking for a resolution more granular than what is available.
> E.g. Requesting 30min resolutions on sets that are split into 1h windows

Totally. I think you've made a good choice here. AFAICT the only other option is to just throw an error.

> To achieve such a granularity (at least accurately) would mean having to build a new query with parameters from the old that is suitable for such a resolution, something I don't think is worth implementing given the required changes. I think it would be a nice-to-have in the future though

That would be super cool, but probably

pkg/kubecost/allocation_test.go
alejandro-kubecost: the tests were being currently worked on last push, these will be kept and are currently working. The last step of wrapping up the tests will be rewriting them in a similar format to the window expand tests

pkg/kubecost/allocation.go
alejandro-kubecost: No worries and this makes sense to me, went ahead and reverted and improved. Thanks!

pkg/kubecost/allocation_test.go
alejandro-kubecost: good call, improved!

alejandro-kubecost: I noticed when looping through the test cases it was difficult to tell which case was responsible for the failure so I added a testId for each of them. Let me know what you think!

pkg/kubecost/allocation.go
```diff
@@ -2260,8 +2260,46 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
```
michaelmdresser, line 2293: I think you could replace the `NewWIndow().Equal(NewWindow())` check with just `asrWindow.end == allocSet.Window.end`

michaelmdresser (approved):
Looks solid! Thanks for writing the tests. I have one suggestion; feel free to ignore it.

pkg/kubecost/allocation.go
```diff
@@ -2260,8 +2260,39 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
 	return allocSet, nil
 }
 
-// TODO accumulate into lower-resolution chunks of the given resolution
-// func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) *AllocationSetRange
+// AccumulateBy sums AllocationSets based on the resolution given. The resolution given is subject to the scale used for the AllocationSets.
+// If the requested resolution is smaller than the window of an AllocationSet then the resolution will default to the duration of a set.
+// Resolutions larger than the duration of the entire AllocationSetRange will default to the duration of the range.
+func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*AllocationSetRange, error) {
+	allocSetRange := &AllocationSetRange{allocations: []*AllocationSet{}}
```
nikovacevic, line 2267: Small suggestion: we could use `NewAllocationSetRange` here.

Today, this matches that implementation, but imagine that the type changes and something important is added to the constructor. (Like, for instance, something could be added to that type that would nil panic if not properly instantiated.) We'd want that extra logic to come along.

pkg/kubecost/allocation.go
```diff
@@ -2260,8 +2260,39 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+		allocSet, err = allocSet.accumulate(as)
+		if err != nil {
+			return nil, err
+		}
+
+		if allocSet != nil {
+			currAccumulatedSum += allocSet.Window.Duration()
+
+			// check if end of asr to sum the final set
+			// If accumulated sum !>= resolution return 1 accumulated set
```
nikovacevic, line 2285: Maybe I'm misreading, but what would `!>=` mean here? (Or maybe it's just supposed to be `>=`?)

nikovacevic: Just making sure I'm with you here. These are the conditions for closing the current set and opening a new one:
* We've accumulated at least the given resolution. More is also fine -- meaning that even divisibility is not a requirement here, and we round up. (That would be a great detail to document!)
* We've reached the end of the range, so even if we're not at a set of full resolution, close it.

That square with your specification?

pkg/kubecost/allocation.go
```diff
@@ -2260,8 +2260,39 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
...
+			if currAccumulatedSum >= resolution || i == len(asr.allocations)-1 {
+				allocSetRange.allocations = append(allocSetRange.allocations, allocSet)
+				allocSet = &AllocationSet{allocations: map[string]*Allocation{}}
```
nikovacevic, line 2288: Same, RE using the existing constructor function. And here, there's actually a very good reason! There are other structures that `NewAllocationSet` instantiates, which we'd need to handle. Otherwise, this might nil panic.

pkg/kubecost/allocation_test.go
```diff
@@ -2067,9 +2067,332 @@ func TestAllocationSetRange_Accumulate(t *testing.T) {
 		t.Fatalf("accumulating AllocationSetRange: expected %f minutes; actual %f", 2880.0, alloc.Minutes())
 	}
 }
+func TestAllocationSetRange_AccumulateBy_Nils(t *testing.T) {
```
nikovacevic, line 2070: Thank you for writing unit tests! I hope that was useful and instructive.

Gotta run for now, so I'll have a look at all of these tomorrow -- but not terribly worried about having every last case covered. This is great.

nikovacevic (changes requested):
Requesting changes primarily for those two constructors (and a little extra detail in the documentation). But the function looks great otherwise. Thank you for all your hard work! First PR is looking good!

nikovacevic:
@alejandro-kubecost I want to also acknowledge that you were likely following convention in using `&AllocationSet` and `&AllocationSetRange`. Just realized that now! Must admit -- I don't know why we don't use them elsewhere, but it's possible that much of the code simply predates the constructors, themselves.

Do you have a strong feeling on the right move here?

pkg/kubecost/allocation.go
alejandro-kubecost: Fixed the typo on the comment and yep that sums it up pretty well!

The round up will be specific to the resolution of the AllocationSet window durations returned. So essentially will accumulate to the nearest ceiling of the requested resolution. 

E.g. User requests 2.5h resolution and each AllocationSet has a 1h window. The first accumulated set added to the AllocationSetRange will be the sum of the first three set's  for a total window duration of 3h

pkg/kubecost/allocation.go
alejandro-kubecost: Makes sense. The initial concern was that initiating a new AllocationSet would result in an incorrect accumulation based on the window start/end parameters we're forced to include in the call.  However I verified that AllocationSet's IsEmpty() will still return empty with non-nil window durations when accumulating so fixing this shouldn't cause issues

Fixed!

pkg/kubecost/allocation.go
alejandro-kubecost: Makes sense. The initial concern was that initiating a new AllocationSet would result in an incorrect accumulation based on the window start/end parameters we're forced to include in the call. However I verified that AllocationSet's IsEmpty() will still return empty with non-nil window durations when accumulating so fixing this shouldn't cause issues

Fixed!

pkg/kubecost/allocation.go
alejandro-kubecost: Initial concern was resolved so I'm in agreement, fixed!

nikovacevic (approved):
Thank you @alejandro-kubecost looks great! Congratulations on your first PR.

(Please wait to merge until after today's release goes out.)

pkg/kubecost/allocation.go
```diff
@@ -2273,15 +2273,14 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
...
-
 	for i, as := range asr.allocations {
 		allocSet, err = allocSet.accumulate(as)
 		if err != nil {
 			return nil, err
 		}
 
 		if allocSet != nil {
-			currAccumulatedSum += allocSet.Window.Duration()
+			currAccumulatedSum = allocSet.Window.Duration()
```
michaelmdresser, line 2283: This doesn't seem right. Why shouldn't the addition happen?

pkg/kubecost/allocation_test.go
```diff
@@ -2303,12 +2306,21 @@ func TestAllocationSetRange_AccumulateBy(t *testing.T) {
...
+
+			testId: "AccumulateBy Test 11",
+		},
 	}
 
 	for _, c := range cases {
 		result, err = c.asr.AccumulateBy(c.resolution)
 		sumCost := 0.0
-
+		fmt.Println(c.testId)
```
michaelmdresser: If you want to keep this, I'd recommend using `t.Logf` instead of `fmt.Println`. `t.Logf` messages won't show unless:
- the test fails OR
- you use `go test -v` instead of just regular `go test`

So, when debugging tests, you can use `go test -v` to get a more detailed picture of what's happening.

pkg/kubecost/allocation.go
michaelmdresser: For example, let's say I have the following AllocationSetRange:
```
asr = [
  as1 // window 1d
  as2 // window 1d
  as3 // window 1d
]
```

I try to aggregate by `2d`. I'm pretty sure that because this line is continually resetting `currAccumulatedSum`, the resulting ASR will be:

```
[
  as // window 3d
]
```

Because the `currAccumulatedSum >= resolution` check will never be true, so we end up in the final case of `i == len(asr.allocations)-1` and accumulate all three allocation sets.

I could be wrong! I'm not sure if this situation is addressed in your unit tests.

pkg/kubecost/allocation_test.go
michaelmdresser: Ah, I realize you removed this line. No need to add it back; this is just some nice info about `go test` 😄

alejandro-kubecost: that's still useful, thanks!

pkg/kubecost/allocation.go
alejandro-kubecost: That was initially how I thought it worked too, and hence the bug, but when accumulating the starting window doesn't get changed and remains the same until reset by `allocSet = NewAllocationSet(time.Time{}, time.Time{})`

After each accumulation the duration (`currAccumulatedSum`) will be the window.start of the first set passed and the window.end will be the end of the last set passed, so resetting is ok here. I tested this on gke and it looks to be working correctly now.

The newest unit test was the only test to fail and catch the bug, it wasn't caught previously due to not mocking enough allocation sets (4+)

alejandro-kubecost: I simplified the code further since `currAccumulateSum` is no longer needed with that bug fix and is now more readable

michaelmdresser (approved):
I like the update you made. Nice catch on the bug.

pkg/kubecost/allocation.go
michaelmdresser: Very nice catch, I like the removal of `currAccumulatedSum`.

PR #1050, opened 2022-02-04
Record adjustments for all AllocationTotals fields

nikovacevic:
## What does this PR change?
* Record adjustments for all AllocationTotals fields and reflect that in methods. Theoretically, the fact that we were not taking PV adjustments into account properly meant that sharing coefficients could become skewed. (Previously, that datum was thought to be irrelevant, but this is safer, even if it's not causing a tremendous amount of trouble.)

- Pairs with https://github.com/kubecost/kubecost-cost-model/pull/634

## How was this PR tested?
* Manually, using an auditing tool that I hope to integrate into the product soon

### DIFF
--- a/pkg/kubecost/totals.go
+++ b/pkg/kubecost/totals.go
@@ -54,6 +54,21 @@ func (art *AllocationTotals) TotalGPUCost() float64 {
 	return art.GPUCost + art.GPUCostAdjustment
 }
 
+// TotalLoadBalancerCost returns LoadBalancer cost with adjustment.
+func (art *AllocationTotals) TotalLoadBalancerCost() float64 {
+	return art.LoadBalancerCost + art.LoadBalancerCostAdjustment
+}
+
+// TotalNetworkCost returns Network cost with adjustment.
+func (art *AllocationTotals) TotalNetworkCost() float64 {
+	return art.NetworkCost + art.NetworkCostAdjustment
+}
+
+// TotalPersistentVolumeCost returns PersistentVolume cost with adjustment.
+func (art *AllocationTotals) TotalPersistentVolumeCost() float64 {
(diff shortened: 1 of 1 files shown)


pkg/kubecost/totals.go
```diff
@@ -108,9 +123,12 @@ func ComputeAllocationTotals(as *AllocationSet, prop string) map[string]*Allocat
 		arts[key].CPUCostAdjustment += alloc.CPUCostAdjustment
 		arts[key].GPUCost += alloc.GPUCost
 		arts[key].GPUCostAdjustment += alloc.GPUCostAdjustment
-		arts[key].LoadBalancerCost += alloc.LBTotalCost()
-		arts[key].NetworkCost += alloc.NetworkTotalCost()
+		arts[key].LoadBalancerCost += alloc.LoadBalancerCost
+		arts[key].LoadBalancerCostAdjustment += alloc.LoadBalancerCostAdjustment
+		arts[key].NetworkCost += alloc.NetworkCost
+		arts[key].NetworkCostAdjustment += alloc.NetworkCostAdjustment
 		arts[key].PersistentVolumeCost += alloc.PVCost()
```
AjayTripathy, line 130: Hm, again not blocking but it feels a little inconsistent to use PVCost() here-- is the difference here that no adjustment is taken into account on that one? If this is working though, no need to block.

AjayTripathy (approved):

pkg/kubecost/totals.go
nikovacevic: Correct. Kind of rough naming conventions, but the counterpart that includes adjustments is `PVTotalCost()`. These adjustments cause some major readability/naming issues, at least as-written.

PR #1051, opened 2022-02-04
Handle Rate Limit Responses in Prom Client

mbolt35:
## What does this PR change?
* Adds a `PrometheusClientConfig` instance for grouping configuration options passed to instantiation.
* Adds a `RetryOnRateLimit` option (defaulted to true) which will process a rate limit response, determine wait time from the response headers (if available), wait, then retry. 
* Reorganize the way we create the rate limited clients to remove `env.Get` from implementations, further decoupling of client functionality to allow for better unit tests 
* Added `RateLimitedResponseError` which is returned if retries are exhausted for rate limited responses 
* Handle Amazon Services strange `400` Status Code with `ThrottlingException` body, and treat as rate limited response. 

* Prometheus and Thanos Clients now handle Rate Limited Responses. A response is recognized as rate limited/throttled if: 
  * Response has 429 Status Code (Header contains `Retry-After` or not)
  * Response has 400 Status Code *and* Response body contains `"ThrottleException"`. This addresses cases where Amazon's Managed Prometheus will use a 400 instead of 429 for throttling. 

## How was this PR tested?
Unit Tests covering all rate limit conditions:
```
=== RUN   TestRateLimitedOnceAndSuccess
=== PAUSE TestRateLimitedOnceAndSuccess
=== RUN   TestRateLimitedOnceAndFail
=== PAUSE TestRateLimitedOnceAndFail
=== RUN   TestRateLimitedResponses
=== PAUSE TestRateLimitedResponses
=== RUN   TestConcurrentRateLimiting
=== PAUSE TestConcurrentRateLimiting
=== CONT  TestRateLimitedOnceAndSuccess
=== CONT  TestRateLimitedResponses
=== CONT  TestConcurrentRateLimiting
=== CONT  TestRateLimitedOnceAndFail
--- PASS: TestRateLimitedOnceAndSuccess (2.50s)
--- PASS: TestRateLimitedOnceAndFail (2.50s)
=== CONT  TestConcurrentRateLimiting
    /home/mbolt/kubecost/kubecost-cost-model/cost-model/pkg/prom/ratelimitedclient_test.go:348: Request was Rate Limited and Retries Exhausted:
         * Wait Time: 1.00 seconds, Retries Remaining: 4
         * Wait Time: 1.00 seconds, Retries Remaining: 3
         * Wait Time: 2.00 seconds, Retries Remaining: 2
         * Wait Time: 1.00 seconds, Retries Remaining: 1
         * Wait Time: 0.00 seconds, Retries Remaining: 0

=== CONT  TestRateLimitedResponses
    /home/mbolt/kubecost/kubecost-cost-model/cost-model/pkg/prom/ratelimitedclient_test.go:246: Request was Rate Limited and Retries Exhausted:
         * Wait Time: 2.00 seconds, Retries Remaining: 4
         * Wait Time: 1.93 seconds, Retries Remaining: 3
         * Wait Time: 1.00 seconds, Retries Remaining: 2
         * Wait Time: 1.00 seconds, Retries Remaining: 1
         * Wait Time: 3.00 seconds, Retries Remaining: 0

--- PASS: TestRateLimitedResponses (10.44s)
=== CONT  TestConcurrentRateLimiting
    /home/mbolt/kubecost/kubecost-cost-model/cost-model/pkg/prom/ratelimitedclient_test.go:348: Request was Rate Limited and Retries Exhausted:
         * Wait Time: 2.00 seconds, Retries Remaining: 4
         * Wait Time: 3.00 seconds, Retries Remaining: 3
         * Wait Time: 3.00 seconds, Retries Remaining: 2
         * Wait Time: 0.00 seconds, Retries Remaining: 1
         * Wait Time: 1.00 seconds, Retries Remaining: 0

    /home/mbolt/kubecost/kubecost-cost-model/cost-model/pkg/prom/ratelimitedclient_test.go:348: Request was Rate Limited and Retries Exhausted:
         * Wait Time: 4.18 seconds, Retries Remaining: 4
         * Wait Time: 0.00 seconds, Retries Remaining: 3
         * Wait Time: 1.00 seconds, Retries Remaining: 2
         * Wait Time: 2.00 seconds, Retries Remaining: 1
         * Wait Time: 2.00 seconds, Retries Remaining: 0

    /home/mbolt/kubecost/kubecost-cost-model/cost-model/pkg/prom/ratelimitedclient_test.go:348: Request was Rate Limited and Retries Exhausted:
         * Wait Time: 1.00 seconds, Retries Remaining: 4
         * Wait Time: 3.00 seconds, Retries Remaining: 3
         * Wait Time: 0.00 seconds, Retries Remaining: 2
         * Wait Time: 1.00 seconds, Retries Remaining: 1
         * Wait Time: 1.00 seconds, Retries Remaining: 0

    /home/mbolt/kubecost/kubecost-cost-model/cost-model/pkg/prom/ratelimitedclient_test.go:348: Request was Rate Limited and Retries Exhausted:
         * Wait Time: 2.00 seconds, Retries Remaining: 4
         * Wait Time: 3.00 seconds, Retries Remaining: 3
         * Wait Time: 0.00 seconds, Retries Remaining: 2
         * Wait Time: 1.00 seconds, Retries Remaining: 1
         * Wait Time: 1.00 seconds, Retries Remaining: 0

--- PASS: TestConcurrentRateLimiting (19.01s)
PASS
ok  	github.com/kubecost/cost-model/pkg/prom	19.018s
```

### DIFF
--- a/pkg/prom/prom.go
+++ b/pkg/prom/prom.go
@@ -3,14 +3,15 @@ package prom
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"net"
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/kubecost/cost-model/pkg/collections"
-	"github.com/kubecost/cost-model/pkg/env"
 	"github.com/kubecost/cost-model/pkg/log"
 	"github.com/kubecost/cost-model/pkg/util/atomic"
 	"github.com/kubecost/cost-model/pkg/util/fileutil"
@@ -57,19 +58,56 @@ func (auth *ClientAuth) Apply(req *http.Request) {
 }
 
 //--------------------------------------------------------------------------
+//  Rate Limited Error
+//--------------------------------------------------------------------------
+
+type RateLimitResponseStatus struct {
+	RetriesRemaining int
+	WaitTime         time.Duration
+}
+
+// String creates a string representation of the rate limit status
+func (rtrs *RateLimitResponseStatus) String() string {
+	return fmt.Sprintf("Wait Time: %.2f seconds, Retries Remaining: %d", rtrs.WaitTime.Seconds(), rtrs.RetriesRemaining)
+}
+
+// RateLimitedError contains a list of retry statuses that occurred during
+// retries on a rate limited response
+type RateLimitedResponseError struct {
+	RateLimitStatus []*RateLimitResponseStatus
+}
+
+// Error returns a string representation of the error, including the rate limit
+// status reports
+func (rlre *RateLimitedResponseError) Error() string {
+	var sb strings.Builder
+
+	sb.WriteString("Request was Rate Limited and Retries Exhausted:\n")
+
+	for _, rls := range rlre.RateLimitStatus {
+		sb.WriteString(" * ")
+		sb.WriteString(rls.String())
+		sb.WriteString("\n")
+	}
+
+	return sb.String()
+}
+
+//--------------------------------------------------------------------------
 //  RateLimitedPrometheusClient
 //--------------------------------------------------------------------------
 
 // RateLimitedPrometheusClient is a prometheus client which limits the total number of
 // concurrent outbound requests allowed at a given moment.
 type RateLimitedPrometheusClient struct {
-	id         string
-	client     prometheus.Client
-	auth       *ClientAuth
-	queue      collections.BlockingQueue
-	decorator  QueryParamsDecorator
-	outbound   *atomic.AtomicInt32
-	fileLogger *golog.Logger
+	id               string
+	client           prometheus.Client
+	auth             *ClientAuth
+	queue            collections.BlockingQueue
+	decorator        QueryParamsDecorator
+	retryOnRateLimit bool
+	outbound         *atomic.AtomicInt32
+	fileLogger       *golog.Logger
 }
 
 // requestCounter is used to determine if the prometheus client keeps track of
@@ -81,11 +119,14 @@ type requestCounter interface {
 
 // NewRateLimitedClient creates a prometheus client which limits the number of concurrent outbound
 // prometheus requests.
-func NewRateLimitedClient(id string, config prometheus.Config, maxConcurrency int, auth *ClientAuth, decorator QueryParamsDecorator, queryLogFile string) (prometheus.Client, error) {
-	c, err := prometheus.NewClient(config)
-	if err != nil {
-		return nil, err
-	}
+func NewRateLimitedClient(
+	id string,
+	client prometheus.Client,
+	maxConcurrency int,
+	auth *ClientAuth,
+	decorator QueryParamsDecorator,
+	retryOnRateLimit bool,
+	queryLogFile string) (prometheus.Client, error) {
 
 	queue := collections.NewBlockingQueue()
 	outbound := atomic.NewAtomicInt32(0)
@@ -105,14 +146,24 @@ func NewRateLimitedClient(id string, config prometheus.Config, maxConcurrency in
 		}
 	}
 
+	// default authentication
+	if auth == nil {
+		auth = &ClientAuth{
+			Username:    "",
+			Password:    "",
+			BearerToken: "",
+		}
+	}
+
 	rlpc := &RateLimitedPrometheusClient{
-		id:         id,
-		client:     c,
-		queue:      queue,
-		decorator:  decorator,
-		outbound:   outbound,
-		auth:       auth,
-		fileLogger: logger,
+		id:               id,
+		client:           client,
+		queue:            queue,
+		decorator:        decorator,
+		retryOnRateLimit: retryOnRateLimit,
+		outbound:         outbound,
+		auth:             auth,
+		fileLogger:       logger,
 	}
 
 	// Start concurrent request processing
@@ -168,6 +219,8 @@ type workResponse struct {
 
 // worker is used as a consumer goroutine to pull workRequest from the blocking queue and execute them
 func (rlpc *RateLimitedPrometheusClient) worker() {
+	retryRateLimit := rlpc.retryOnRateLimit
+
 	for {
 		// blocks until there is an item available
 		item := rlpc.queue.Dequeue()
@@ -197,6 +250,34 @@ func (rlpc *RateLimitedPrometheusClient) worker() {
 			// Execute Request
 			roundTripStart := time.Now()
 			res, body, warnings, err := rlpc.client.Do(ctx, req)
+
+			// If retries on rate limited response is enabled:
+			// * Check for a 429 StatusCode OR 400 StatusCode and message containing "ThrottlingException"
+			// * Attempt to parse a Retry-After from response headers (common on 429)
+			// * If we couldn't determine how long to wait for a retry, use 1 second by default
+			if retryRateLimit {
+				var status []*RateLimitResponseStatus
+				var retries int = 5
+
+				for httputil.IsRateLimited(res, body) && retries > 0 {
+					retries--
+
+					// calculate amount of time to wait before retry, default to 1s
+					retryAfter := httputil.RateLimitedRetryFor(res, time.Second)
+					status = append(status, &RateLimitResponseStatus{RetriesRemaining: retries, WaitTime: retryAfter})
+					log.DedupedInfof(50, "Rate Limited Prometheus Request. Waiting for: %.2f seconds. Retries Remaining: %d", retryAfter.Seconds(), retries)
+
+					// execute wait and retry
+					time.Sleep(retryAfter)
+					res, body, warnings, err = rlpc.client.Do(ctx, req)
+				}
+
+				// if we've broken out of our retry loop and the resp is still rate limited,
+				// then let's generate a meaningful error to pass back
+				if retries == 0 && httputil.IsRateLimited(res, body) {
+					err = &RateLimitedResponseError{RateLimitStatus: status}
+				}
+			}
 
 			// Decrement outbound counter
 			rlpc.outbound.Decrement()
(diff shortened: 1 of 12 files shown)


pkg/prom/ratelimitedclient_test.go
```diff
@@ -0,0 +1,350 @@
...
+// and response bodies
+type ResponseAndBody struct {
+	Response *http.Response
+	Body     []byte
+}
+
+// MockPromClient accepts a slice of responses and bodies to return on requests made.
+// It will cycle these responses linearly, then reset back to the first.
+// Also works with concurrent requests.
+type MockPromClient struct {
```
michaelmdresser, line 28: Perhaps we shouldn't export these test types?

pkg/prom/ratelimitedclient_test.go
```diff
@@ -0,0 +1,350 @@
...
+			},
+			Body: io.NopCloser(bytes.NewReader(body)),
+		},
+		Body: body,
+	}
+}
+
+// creates a ResponseAndBody representing some amazon services ThrottlingException 400 status
+func newHackyAmazonRateLimitedResponse() *ResponseAndBody {
+	body := []byte("<ThrottlingException>\n  <Message>Rate exceeded</Message>\n</ThrottlingException>\n")
```
michaelmdresser, line 107: 👀 Is Amazon using XML for their error messages?

pkg/util/httputil/httputil.go
```diff
@@ -90,6 +92,61 @@ func SetQuery(r *http.Request, query string) *http.Request {
...
+// IsRateLimited accepts a response and body to determine if either indicate
+// a rate limited return
+func IsRateLimited(resp *http.Response, body []byte) bool {
+	return IsRateLimitedResponse(resp) || IsRateLimitedBody(resp, body)
+}
+
+// RateLimitedRetryFor returns the parsed Retry-After header relative to the
+// current time. If the Retry-After header does not exist, the defaultWait parameter
+// is returned.
+func RateLimitedRetryFor(resp *http.Response, defaultWait time.Duration) time.Duration {
```
michaelmdresser, line 104: I'm not a huge fan of this name. The comment helps a ton. I'd suggest `ExtractRetryAfterOrDefault` or something similar.

pkg/util/httputil/httputil.go
```diff
@@ -90,6 +92,61 @@ func SetQuery(r *http.Request, query string) *http.Request {
...
+		// return 0 if the datetime has already elapsed
+		result := t.Sub(time.Now())
+		if result < 0 {
+			return 0
+		}
+		return result
+	}
+
+	// failed to parse datetime, return default
+	return defaultWait
```
michaelmdresser, line 131: This situation seems worthy of a `log.Debugf` but perhaps not a `log.Infof`.

We hardcode the klog log level right now, making it essentially impossible to use `log.Debugf` for anything useful. Do you think it'd be worthwhile to make the log level configurable? In a separate PR, naturally; I think this is a broader question about our logging methodology.

pkg/util/httputil/httputil.go
```diff
@@ -90,6 +92,61 @@ func SetQuery(r *http.Request, query string) *http.Request {
...
+}
+
+// IsRateLimitedResponse returns true if the status code is a 429 (TooManyRequests)
+func IsRateLimitedResponse(resp *http.Response) bool {
+	return resp.StatusCode == http.StatusTooManyRequests
+}
```
michaelmdresser, line 137: Any particular reason this check deserves its own function?

pkg/util/httputil/httputil.go
```diff
@@ -90,6 +92,61 @@ func SetQuery(r *http.Request, query string) *http.Request {
...
+
+// IsRateLimitedBody attempts to determine if a response body indicates throttling
+// has occurred. This function is a result of some API providers (AWS) returning
+// a 400 status code instead of 429 for rate limit exceptions.
+func IsRateLimitedBody(resp *http.Response, body []byte) bool {
+	// ignore non-400 status
+	if resp.StatusCode < http.StatusBadRequest || resp.StatusCode >= http.StatusInternalServerError {
+		return false
+	}
```
michaelmdresser, line 146: Why not `if resp.StatusCode != http.StatusBadRequest`?

michaelmdresser (approved):
Looks great! Unit testing and clear error messages/types are always wonderful. My comments are all nits, so feel free to ignore them.

pkg/prom/ratelimitedclient_test.go
mbolt35: Why not? If it's test code, there's not really anything problematic about doing this, except giving other test packages the possibility to re-use the exported types.

pkg/prom/ratelimitedclient_test.go
mbolt35: I don't fully understand the 4XX `<ThrottleException>` messages to be honest. I'm going off of what I've collected over the internet and what a customer has reported here.

pkg/util/httputil/httputil.go
mbolt35: I'm not fully sold on these return status codes always being 400 on `ThrottlingException`. There was a reference I found (that may or may not be outdated). The specific 400 is coming from a customer, I found references to 4XX status responses as well via some searching. Just making sure we inspect any 4XX status.

pkg/prom/prom.go
```diff
@@ -198,6 +256,41 @@ func (rlpc *RateLimitedPrometheusClient) worker() {
...
+			// * Attempt to parse a Retry-After from response headers (common on 429)
+			// * If we couldn't determine how long to wait for a retry, use 1 second by default
+			if retryRateLimit {
+				var status []*RateLimitResponseStatus
+				var retries int = 5
+
+				for httputil.IsRateLimited(res, body) && retries > 0 {
+					retries--
+
+					// calculate amount of time to wait before retry, default to 1s
```
valdisrigdon, line 270: It would be great to expose the initial defaults and max retry attempts. In our testing, we've set the initial retry at 100ms and backed off exponentially after that.

pkg/prom/prom.go
```diff
@@ -198,6 +256,41 @@ func (rlpc *RateLimitedPrometheusClient) worker() {
...
+					retryAfter := httputil.RateLimitedRetryFor(res, time.Second)
+					status = append(status, &RateLimitResponseStatus{RetriesRemaining: retries, WaitTime: retryAfter})
+					log.DedupedInfof(50, "Rate Limited Prometheus Request. Waiting for: %.2f seconds. Retries Remaining: %d", retryAfter.Seconds(), retries)
+
+					// To prevent total starvation of request threads, hard limit wait time to 10s. We also want quota limits/throttles
```
valdisrigdon, line 275: At least for AMP, the quotas are on a per second basis.

pkg/prom/prom.go
```diff
@@ -198,6 +251,34 @@ func (rlpc *RateLimitedPrometheusClient) worker() {
...
+					retries--
+
+					// calculate amount of time to wait before retry, default to 1s
+					retryAfter := httputil.RateLimitedRetryFor(res, time.Second)
+					status = append(status, &RateLimitResponseStatus{RetriesRemaining: retries, WaitTime: retryAfter})
+					log.DedupedInfof(50, "Rate Limited Prometheus Request. Waiting for: %.2f seconds. Retries Remaining: %d", retryAfter.Seconds(), retries)
+
+					// execute wait and retry
+					time.Sleep(retryAfter)
+					res, body, warnings, err = rlpc.client.Do(ctx, req)
```
dahamr, line 284: To make this a bit more fault tolerant, there are other situations that would benefit from a retry. In our testing, we used https://github.com/hashicorp/go-retryablehttp defaults and added the weird ThrottleException check. See https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L435 as an example of what you might want to add to the retry.

pkg/prom/ratelimitedclient_test.go
dahamr: Yes, this is what we're seeing and I agree that it's weird. I have no idea if this is *always* the message or if it could have other words in it. Guess we'll find out. We're asking Amazon to look into this.

pkg/util/httputil/httputil.go
michaelmdresser: :+1: :+1: SGTM

pkg/prom/prom.go
mbolt35: @dahamr Thanks for the added insight here. This HTTP client we're wrapping with the concurrent throttle is an  older version of the prometheus client (which, to be honest, I am not a fan of at all). I like the idea of a base retry policy - thanks for the reference. I can easily add the retry default timeout option, backoff, and max. However, I may hold off on the policy until we have some time to test further -- I'd really like to replace the prometheus client with a standard http client anyways - I've starred the retryable http client for later :)

pkg/prom/prom.go
mbolt35: @valdisrigdon Not a problem - Can add those options for you guys.

mbolt35:
@dahamr @valdisrigdon Thanks for the input here. I've added the following environment variables: 
* `PROMETHEUS_RETRY_ON_RATE_LIMIT`: [Defaults to `true`] Setting to `false` will disable the response inspection and retries. 
* `PROMETHEUS_RETRY_ON_RATE_LIMIT_MAX_RETRIES`: [Defaults to `5`] The maximum retries attempted per request (where the response is rate limited).
* `PROMETHEUS_RETRY_ON_RATE_LIMIT_DEFAULT_WAIT`: [Defaults to "100ms"] The default wait time used if a `Retry-After` header doesn't exist in the response. The value is used to apply an exponential backoff based on the retry attempt as well. So, wait times of: `100ms`, `200ms`, `400ms`, `800ms`, `1600ms` for the default retries of `5`. 

Hope this will at least get you guys past the weird AWS throttling. Let me or @AjayTripathy if you need anymore help here!

pkg/util/timeutil/timeutil.go
```diff
@@ -86,35 +87,129 @@ func FormatStoreResolution(dur time.Duration) string {
...
-		unit = 24.0 * time.Hour
-	default:
-		return 0, fmt.Errorf("error parsing duration: %s did not match expected format [0-9+](s|m|d|h)", duration)
-	}
-
-	amountStr := duration[:len(duration)-1]
-	amount, err := strconv.ParseInt(amountStr, 10, 64)
-	if err != nil {
-		return 0, fmt.Errorf("error parsing duration: %s did not match expected format [0-9+](s|m|d|h)", duration)
+	return goParseDuration(duration)
```
mbolt35, line 97: A note about all the additions below: We had code that did a very miserable job of parsing duration strings. The reasoning behind our own parse method was to support the `"d"` unit for days, which is not supported in the `time` package of the go std lib. However, to add a new unit to the go std library version isn't difficult, it just requires lifting a handful of code out of that standard library and making a single addition to the `unitMap` below. 

Moving to this gives us more reliability around parsing duration units, and continues to support all existing use cases of the API. All existing unit tests continued to pass as well as some newer tests that check for `"d"` unit compatibility.

PR #1052, opened 2022-02-06
Sean/cloud etl

Sean-Holcomb:
## What does this PR change?
This PR adds the Cloud Usage type which is used by the Cloud ETL to store Cloud Provider data. As it stands this type is an alias of the Cloud Asset type, however this branch contains a commit which has the full implementation of this class. Additionally there were some smaller support functions added which are used by the Cloud ETL

- https://github.com/kubecost/cost-analyzer-helm-chart/pull/1238
- https://github.com/kubecost/kubecost-cost-model/pull/637

- 
-


pkg/kubecost/cloudusage.go
```diff
@@ -0,0 +1,16 @@
...
+type CloudUsageSet = AssetSet
+
+// CloudUsageSetRange is temporarily aliased as the AssetSetRange until further infrastructure and pages can be built to support its usage
+type CloudUsageSetRange = AssetSetRange
+
+// CloudUsageAggregationOptions is temporarily aliased as the AssetAggregationOptions until further infrastructure and pages can be built to support its usage
+type CloudUsageAggregationOptions = AssetAggregationOptions
+
+// CloudUsageMatchFunc is temporarily aliased as the AssetMatchFunc until further infrastructure and pages can be built to support its usage
+type CloudUsageMatchFunc = AssetMatchFunc
```
nikovacevic: What, precisely, blocks us from implementing the new types?

It feels like we might want good types from the start. I'm just imagining that it will potentially be a huge pain to convert from this state of affairs to a whole new type system after we build in this direction.

nikovacevic (approved):
This side of the Cloud ETL equation looks good to me -- just want to have a quick conversation about the types so that we don't accidentally box ourselves into an uncomfortable place.

PR #1054, opened 2022-02-08
WIP audit

nikovacevic:


pkg/kubecost/totals.go
```diff
@@ -214,17 +266,34 @@ func ComputeAssetTotals(as *AssetSet, prop AssetProperty) map[string]*AssetTotal
...
-			totalRAMCost := node.RAMCost * (1.0 - node.Discount)
-			ramCost := totalRAMCost * adjustmentRate
-			ramCostAdjustment := totalRAMCost - ramCost
+			discountedGPUCost := node.GPUCost * (1.0 - node.Discount)
+			adjustedGPUCost := discountedGPUCost * adjustmentRate
+			gpuCostAdjustment := discountedGPUCost - adjustedGPUCost
+
+			discountedRAMCost := node.RAMCost * (1.0 - node.Discount)
+			adjustedRAMCost := discountedRAMCost * adjustmentRate
+			ramCostAdjustment := adjustedRAMCost - discountedRAMCost
```
nikovacevic, line 296: This is a big one. The subtraction that determines adjustment was backwards (so adjustments for pre-computed totals were all negative) and the total cost (adjustment included) was being used for the pre-adjustment cost below.

Now, the math is correct, the names are updated, and there is an example explaining all of this.

PR #1055, opened 2022-02-08
Fix refresh of DownloadPricingData

AjayTripathy:
## What does this PR change?
DownloadPricingData() is supposed to be called in a loop, but here on error it will not.

No

Fix CSV pricing refresh on errors.

Closes https://github.com/kubecost/cost-model/issues/1048

## How was this PR tested?
Forced an error with no CSV, noted that DownloadPricingData() was re-run

Not necessary


AjayTripathy:
Merged to main branch for patch then sync'd back-- closing.

PR #1058, opened 2022-02-08
KLog Initialization Cobra Fix

mbolt35:
## What does this PR change?
* KLog flag usage can break due to conflicts between Cobra + pflag and go std library "flag" package. 
* We can pipe go flags into pflag so they can be used appropriately.

### DIFF
--- a/cmd/costmodel/main.go
+++ b/cmd/costmodel/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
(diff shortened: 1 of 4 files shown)


Sean-Holcomb (approved):
Logging is working correctly now

Sean-Holcomb (approved):
logging is working correctly with this PR

PR #1059, opened 2022-02-08
KLog Initialization Fix

mbolt35:
## What does this PR change?
* KLog flag usage can break due to conflicts between Cobra + pflag and go std library "flag" package. 
* We can pipe go flags into pflag so they can be used appropriately.


Sean-Holcomb (approved):
lgmt

PR #1060, opened 2022-02-08
Add fine-grained control for disabling metrics

kaelanspatel:
## What does this PR change?
Adds a metrics configmap and the associated logic for watching/writing contents to a file. Also adds the ability to disable any metric by specifying it in the helm chart (see ). 

Disabling metrics takes the form of the `metricsConfigs.disabledMetrics` field in the helm chart:
```
kubecostProductConfigs:
  ...
  metricsConfigs:
    disabledMetrics:
      - <metric-to-be-disabled>
      - <metric-to-be-disabled>
      etc.
```
This should allow for emission of any metric to be disabled, as this PR adds logic to simply exclude initialization of that metric in the metrics emitter.

- https://github.com/kubecost/cost-analyzer-helm-chart/pull/1243
- 

Adds support for disabling emission of specific metrics in the helm chart

- fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1200
- 

## How was this PR tested?
Manually by disabling and reenabling metrics, and examining prometheus.

No.

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -40,23 +40,27 @@ type ClusterInfoCollector struct {
 // collected by this Collector.
 func (cic ClusterInfoCollector) Describe(ch chan<- *prometheus.Desc) {
 	disabledMetrics := cic.metricsConfig.GetDisabledMetricsMap()
-
-	if _, ok := disabledMetrics["kube_pod_annotations"]; !ok {
-		ch <- prometheus.NewDesc("kubecost_cluster_info", "Kubecost Cluster Info", []string{}, nil)
+	if _, disabled := disabledMetrics["kube_pod_annotations"]; disabled {
+		return
 	}
+
+	ch <- prometheus.NewDesc("kubecost_cluster_info", "Kubecost Cluster Info", []string{}, nil)
+
 }
 
 // Collect is called by the Prometheus registry when collecting metrics.
 func (cic ClusterInfoCollector) Collect(ch chan<- prometheus.Metric) {
 	disabledMetrics := cic.metricsConfig.GetDisabledMetricsMap()
-
-	if _, ok := disabledMetrics["kube_pod_annotations"]; !ok {
-		clusterInfo := cic.ClusterInfo.GetClusterInfo()
-		labels := prom.MapToLabels(clusterInfo)
-
-		m := newClusterInfoMetric("kubecost_cluster_info", labels)
-		ch <- m
+	if _, disabled := disabledMetrics["kube_pod_annotations"]; disabled {
+		return
 	}
+
+	clusterInfo := cic.ClusterInfo.GetClusterInfo()
+	labels := prom.MapToLabels(clusterInfo)
+
+	m := newClusterInfoMetric("kubecost_cluster_info", labels)
+	ch <- m
+
 }
 
 //--------------------------------------------------------------------------
@@ -138,11 +142,11 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud
 
 	disabledMetrics := metricsConfig.GetDisabledMetricsMap()
 	var toRegisterGV []*prometheus.GaugeVec
-	var toRegisterGage []prometheus.Gauge
+	var toRegisterGauge []prometheus.Gauge
 
 	metricsInit.Do(func() {
 
-		if _, ok := disabledMetrics["node_cpu_hourly_cost"]; !ok {
+		if _, disabled := disabledMetrics["node_cpu_hourly_cost"]; !disabled {
 			cpuGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 				Name: "node_cpu_hourly_cost",
 				Help: "node_cpu_hourly_cost hourly cost for each cpu on this node",
@@ -150,7 +154,7 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud
 			toRegisterGV = append(toRegisterGV, cpuGv)
 		}
 
-		if _, ok := disabledMetrics["node_ram_hourly_cost"]; !ok {
+		if _, disabled := disabledMetrics["node_ram_hourly_cost"]; !disabled {
 			ramGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 				Name: "node_ram_hourly_cost",
 				Help: "node_ram_hourly_cost hourly cost for each gb of ram on this node",
@@ -158,7 +162,7 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud
 			toRegisterGV = append(toRegisterGV, ramGv)
 		}
 
-		if _, ok := disabledMetrics["node_gpu_hourly_cost"]; !ok {
+		if _, disabled := disabledMetrics["node_gpu_hourly_cost"]; !disabled {
 			gpuGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 				Name: "node_gpu_hourly_cost",
 				Help: "node_gpu_hourly_cost hourly cost for each gpu on this node",
@@ -166,7 +170,7 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud
 			toRegisterGV = append(toRegisterGV, gpuGv)
 		}
 
-		if _, ok := disabledMetrics["node_gpu_count"]; !ok {
+		if _, disabled := disabledMetrics["node_gpu_count"]; !disabled {
 			gpuCountGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 				Name: "node_gpu_count",
 				Help: "node_gpu_count count of gpu on this node",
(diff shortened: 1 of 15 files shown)


dwbrown2:
*Note: I haven't reviewed the code*

I think this has some really tricky product implications, i.e. User1 user disables some metrics and User2 thinks Kubecost is broken for some unknown reason. Have we thought through diagnostics, UI messaging, or other mechanisms to help manage?

kaelanspatel:
I think that's a valid concern. I think a diagnostic/log that relays configuration might be good.

Not disregarding possible issues, but I do think we'd want to bill a feature like this (along with any features that effect metrics emissions) as "only do this if you know what you're doing", in the sense that allowing these changes to be made will open us up to problems regardless.

dwbrown2:
Good points. I personally think the following could be valuable: 

1) a comment in the helm chart to this effect
2) a link to a docs page from helm chart
3) an endpoint or diagnostics page where you can go to see current disable rules (referenced in docs page)

Should we also probably be highlight this in a bug report in some capacity for support team?

kaelanspatel:
> 1. a comment in the helm chart to this effect
> 2. a link to a docs page from helm chart

There is an (admittedly short) note of this in the helm chart PR. I'll create a section of documentation explaining/warning about this.

> 3. an endpoint or diagnostics page where you can go to see current disable rules (referenced in docs page)

I think this would make the values very clear, but I do wonder whether it's necessary. 

This feature can obviously egregiously break Kubecost, but if the criteria for that is "Kubecost is not behaving as I thought it would" then really any config (shared resources, custompricing) could do so too. Again, not to discount user issues, but short of tracking every metric we wouldn't be doing much other than regurgitating helm values.

I'm not opposed to adding a specific page or diagnostic, but I wonder if a log in metrics emission would be sufficient.

> Should we also probably be highlight this in a bug report in some capacity for support team?

Related to the above, it should also be easy to see if any metrics are disabled by looking at the helm values and/or the logs related to watching the metrics configmap, both of which are likely to be present in the a bug report.

pkg/costmodel/metrics.go
```diff
@@ -125,97 +134,154 @@ var (
 )
 
 // initCostModelMetrics uses a sync.Once to ensure that these metrics are only created once
-func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud.Provider, clusterInfo clusters.ClusterInfoProvider) {
+func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider cloud.Provider, clusterInfo clusters.ClusterInfoProvider, metricsConfig *metrics.MetricsConfig) {
+
+	disabledMetrics := metricsConfig.GetDisabledMetricsMap()
+	var toRegisterGV []*prometheus.GaugeVec
+	var toRegisterGage []prometheus.Gauge
```
michaelmdresser, line 141: "Gage" -> "Gauge"

nit, feel free to ignore

pkg/costmodel/metrics.go
```diff
@@ -125,97 +134,154 @@ var (
...
-		prometheus.MustRegister(cpuGv, ramGv, gpuGv, gpuCountGv, totalGv, pvGv, spotGv)
-		prometheus.MustRegister(ramAllocGv, cpuAllocGv, gpuAllocGv, pvAllocGv)
-		prometheus.MustRegister(networkZoneEgressCostG, networkRegionEgressCostG, networkInternetEgressCostG)
-		prometheus.MustRegister(clusterManagementCostGv, lbCostGv)
+		for i := range toRegisterGV {
+			prometheus.MustRegister(toRegisterGV[i])
+		}
+		for i := range toRegisterGage {
+			prometheus.MustRegister(toRegisterGage[i])
+		}
```
michaelmdresser, line 279: Since you have these slices of emitters, what if you checked the disabled condition in these loops, rather than hardcoding an if block for each one?

So, something like
```go
for _, gv := range toRegisterGV {
    if _, ok := disabledMetrics[gv.Name]; ok {
        continue
    }

    prometheus.MustRegister(gv)
}
```

pkg/metrics/jobmetrics.go
```diff
@@ -18,36 +18,45 @@ var (
...
-					reasonKnown := false
-					for _, reason := range jobFailureReasons {
-						reasonKnown = reasonKnown || failureReason(&condition, reason)
-
-						ch <- newKubeJobStatusFailedMetric(jobName, jobNS, "kube_job_status_failed", reason, boolFloat64(failureReason(&condition, reason)))
-					}
+	disabledMetrics := kjc.metricsConfig.GetDisabledMetricsMap()
+
+	if _, ok := disabledMetrics["kube_pod_annotations"]; !ok {
+		jobs := kjc.KubeClusterCache.GetAllJobs()
```
michaelmdresser, line 39: As a general point of Go style, aimed at avoiding excessive indentation, its recommended to do an early return, like so:

```suggestion

	if _, ok := disabledMetrics["kube_pod_annotations"]; ok {
	    return
	}
	jobs := kjc.KubeClusterCache.GetAllJobs()
```

pkg/metrics/metricsconfig.go
```diff
@@ -0,0 +1,104 @@
...
+
+		return mc, nil
+	} else if err != nil {
+		return mc, err
+	}
+
+	err = json.Unmarshal(body, mc)
+	if err != nil {
+		return mc, err
+	}
```
michaelmdresser, line 47: Please wrap these errors to help with potential future debugging.

pkg/metrics/jobmetrics.go
michaelmdresser: It has the added benefit of making the git diffs a little easier to parse, too.

pkg/metrics/nodemetrics.go
```diff
@@ -22,24 +22,45 @@ var (
...
-	ch <- prometheus.NewDesc("kube_node_status_allocatable", "The allocatable for different resources of a node that are available for scheduling.", []string{}, nil)
-	ch <- prometheus.NewDesc("kube_node_status_allocatable_cpu_cores", "The allocatable cpu cores.", []string{}, nil)
-	ch <- prometheus.NewDesc("kube_node_status_allocatable_memory_bytes", "The allocatable memory in bytes.", []string{}, nil)
-	ch <- prometheus.NewDesc("kube_node_labels", "all labels for each node prefixed with label_", []string{}, nil)
-	ch <- prometheus.NewDesc("kube_node_status_condition", "The condition of a cluster node.", []string{}, nil)
+	disabledMetrics := nsac.metricsConfig.GetDisabledMetricsMap()
+
+	if _, ok := disabledMetrics["kube_node_status_capacity"]; !ok {
+		ch <- prometheus.NewDesc("kube_node_status_capacity", "Node resource capacity.", []string{}, nil)
+	}
```
michaelmdresser, line 35: I think the idea to check disabled status inside a loop would clean this up a bit, too.

pkg/costmodel/metrics.go
```diff
@@ -32,22 +32,31 @@ import (
...
+	}
 }
 
 // Collect is called by the Prometheus registry when collecting metrics.
 func (cic ClusterInfoCollector) Collect(ch chan<- prometheus.Metric) {
-	clusterInfo := cic.ClusterInfo.GetClusterInfo()
-	labels := prom.MapToLabels(clusterInfo)
+	disabledMetrics := cic.metricsConfig.GetDisabledMetricsMap()
+
+	if _, ok := disabledMetrics["kube_pod_annotations"]; !ok {
```
michaelmdresser, line 53: This is a _total_ nit, feel free to ignore, but it reads a little better if you do this:
```suggestion
	if _, disabled := disabledMetrics["kube_pod_annotations"]; !disabled {
```

michaelmdresser (commented):
I haven't gone over it with a super fine-toothed comb, yet. I have a few suggestions about how this can be cleaned up, but overall the idea looks good. 

Out of curiosity, is it possible to disable these metrics at some higher level, like whatever manages metric emission? Based on the way `prometheus.MustRegister` looks to be some sort of global thing, I suspect not 😢

pkg/costmodel/metrics.go
mbolt35: Instead of doing a range loop, I also think you could just do something like:

```go
prometheus.MustRegister(toRegisterGV...)
```

pkg/metrics/jobmetrics.go
mbolt35: @michaelmdresser It can also reduce branches as well, which can make code way more performant. I don't know if it applies in this particular case, but it's a good reason to apply early returns/continues up and down the stack

pkg/costmodel/metrics.go
kaelanspatel: I tried both of these actually, but I didn't get either to work so I just opted to do it the dumb way.

Specifically, `MustRegister` apparently won't take `[]*Prometheus.GaugeVec` as an argument even though it'll take several `*Prometheus.GaugeVec`. Also, I don't think you can get the name out of the Gaugevec that easily.

michaelmdresser: Annoying that you can't get the name out of the Gaugevec. Bolt's suggestion should work, though. It's a standard Go method for passing a slice of things to a function that takes variadic args. See https://gobyexample.com/variadic-functions

kaelanspatel: I understand that, I'm saying that when I tried using `toRegisterGV...` it gives me an IncompatibleAssign.

michaelmdresser: Darn

pkg/metrics/jobmetrics.go
kaelanspatel: I've implemented this where I could. I know it's not super clean, but it's just the price we have to pay for this feature I guess.

pkg/costmodel/metrics.go
kaelanspatel: Done.

kaelanspatel:
@dwbrown2 in addition to the proposed changes, I've added a debug message of the format
```
log.go:47] [Info] Starting metrics init with disabled metrics: [<disabled metric 1> <disabled metric 2>...]
```
This should show up anytime metrics are initialized and also exist in the bug report logs.

michaelmdresser (approved):
LGTM! Thanks for dealing with the nitpicky review

PR #1061, opened 2022-02-09
Fix totals bugs: compute adjustments properly, use adjustments everywhere, and fix cpu/gpu typo

nikovacevic:
## What does this PR change?
* Compute adjustments properly in `ComputeAssetTotals`. (This one is tricky, or at least it was as it was written before -- the example should help.)
* Record and use adjustments everywhere. There used to be places we only needed total costs, so it seemed easier to not even record adjustments, but that was confusing so now everything has adjustments and we use the `TotalResourceCost()` functions everywhere, which fixed a few inaccuracies.
* Fix a "cpu" vs "gpu" typo in `ComputeIdleCoefficients` which was causing trouble for GPU idle coefficients.

- Pairs with https://github.com/kubecost/kubecost-cost-model/pull/644

- Fixes numerical inaccuracies

## How was this PR tested?
- New audit tool, which will be committed to code soon!  See attached files.

Before: everything without a "check" displays a table with the mismatched numbers.
[a33fe579-results-before.txt](https://github.com/kubecost/cost-model/files/8028747/a33fe579-results-before.txt)

After: everything matches to within $0.001.
[a33fe579-results.txt](https://github.com/kubecost/cost-model/files/8028746/a33fe579-results.txt)

### DIFF
--- a/pkg/kubecost/totals.go
+++ b/pkg/kubecost/totals.go
@@ -54,6 +54,21 @@ func (art *AllocationTotals) TotalGPUCost() float64 {
 	return art.GPUCost + art.GPUCostAdjustment
 }
 
+// TotalLoadBalancerCost returns LoadBalancer cost with adjustment.
+func (art *AllocationTotals) TotalLoadBalancerCost() float64 {
+	return art.LoadBalancerCost + art.LoadBalancerCostAdjustment
+}
+
+// TotalNetworkCost returns Network cost with adjustment.
+func (art *AllocationTotals) TotalNetworkCost() float64 {
+	return art.NetworkCost + art.NetworkCostAdjustment
+}
+
+// TotalPersistentVolumeCost returns PersistentVolume cost with adjustment.
+func (art *AllocationTotals) TotalPersistentVolumeCost() float64 {
+	return art.PersistentVolumeCost + art.PersistentVolumeCostAdjustment
+}
+
 // TotalRAMCost returns RAM cost with adjustment.
 func (art *AllocationTotals) TotalRAMCost() float64 {
 	return art.RAMCost + art.RAMCostAdjustment
@@ -61,8 +76,8 @@ func (art *AllocationTotals) TotalRAMCost() float64 {
 
 // TotalCost returns the sum of all costs.
 func (art *AllocationTotals) TotalCost() float64 {
-	return art.TotalCPUCost() + art.TotalGPUCost() + art.LoadBalancerCost +
-		art.NetworkCost + art.PersistentVolumeCost + art.TotalRAMCost()
+	return art.TotalCPUCost() + art.TotalGPUCost() + art.TotalLoadBalancerCost() +
+		art.TotalNetworkCost() + art.TotalPersistentVolumeCost() + art.TotalRAMCost()
 }
 
 // ComputeAllocationTotals totals the resource costs of the given AllocationSet
@@ -104,13 +119,22 @@ func ComputeAllocationTotals(as *AllocationSet, prop string) map[string]*Allocat
 		}
 
 		arts[key].Count++
+
 		arts[key].CPUCost += alloc.CPUCost
 		arts[key].CPUCostAdjustment += alloc.CPUCostAdjustment
+
 		arts[key].GPUCost += alloc.GPUCost
 		arts[key].GPUCostAdjustment += alloc.GPUCostAdjustment
-		arts[key].LoadBalancerCost += alloc.LBTotalCost()
-		arts[key].NetworkCost += alloc.NetworkTotalCost()
-		arts[key].PersistentVolumeCost += alloc.PVCost()
+
+		arts[key].LoadBalancerCost += alloc.LoadBalancerCost
+		arts[key].LoadBalancerCostAdjustment += alloc.LoadBalancerCostAdjustment
+
+		arts[key].NetworkCost += alloc.NetworkCost
+		arts[key].NetworkCostAdjustment += alloc.NetworkCostAdjustment
+
+		arts[key].PersistentVolumeCost += alloc.PVCost() // NOTE: PVCost() does not include adjustment
+		arts[key].PersistentVolumeCostAdjustment += alloc.PVCostAdjustment
+
 		arts[key].RAMCost += alloc.RAMCost
 		arts[key].RAMCostAdjustment += alloc.RAMCostAdjustment
 	})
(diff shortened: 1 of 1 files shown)


pkg/kubecost/totals.go
```diff
@@ -214,17 +268,34 @@ func ComputeAssetTotals(as *AssetSet, prop AssetProperty) map[string]*AssetTotal
...
+			discountedGPUCost := node.GPUCost * (1.0 - node.Discount)
+			adjustedGPUCost := discountedGPUCost * adjustmentRate
+			gpuCostAdjustment := discountedGPUCost - adjustedGPUCost
 
-			totalRAMCost := node.RAMCost * (1.0 - node.Discount)
-			ramCost := totalRAMCost * adjustmentRate
-			ramCostAdjustment := totalRAMCost - ramCost
+			discountedRAMCost := node.RAMCost * (1.0 - node.Discount)
+			adjustedRAMCost := discountedRAMCost * adjustmentRate
+			ramCostAdjustment := adjustedRAMCost - discountedRAMCost
```
nikovacevic, line 298: This is a big one. The subtraction that determines adjustment was backwards (so adjustments for pre-computed totals were all negative of their intended value) and the total cost (adjustment included) was being used for the pre-adjustment cost below.

Now, the math is correct, the names are updated, and there is an example explaining all of this.

pkg/kubecost/totals.go
```diff
@@ -332,7 +438,7 @@ func ComputeIdleCoefficients(shareSplit, key string, cpuCost, gpuCost, ramCost f
 	}
 
 	if allocationTotals[key].GPUCost > 0 {
-		gpuCoeff = cpuCost / allocationTotals[key].GPUCost
+		gpuCoeff = gpuCost / allocationTotals[key].GPUCost
```
nikovacevic, line 441: Here's the cpu/gpu typo

pkg/kubecost/totals.go
```diff
@@ -104,13 +119,22 @@ func ComputeAllocationTotals(as *AllocationSet, prop string) map[string]*Allocat
...
-		arts[key].PersistentVolumeCost += alloc.PVCost()
+
+		arts[key].LoadBalancerCost += alloc.LoadBalancerCost
+		arts[key].LoadBalancerCostAdjustment += alloc.LoadBalancerCostAdjustment
+
+		arts[key].NetworkCost += alloc.NetworkCost
+		arts[key].NetworkCostAdjustment += alloc.NetworkCostAdjustment
+
+		arts[key].PersistentVolumeCost += alloc.PVCost() // NOTE: PVCost() does not include adjustment
+		arts[key].PersistentVolumeCostAdjustment += alloc.PVCostAdjustment
```
nikovacevic, line 136: Before we weren't taking `PVCostAdjustment` into account, so that could skew idle/resource sharing coefficients that simply compare `alloc.TotalCost()` to `totals.TotalCost()`.

pkg/kubecost/totals.go
```diff
@@ -158,15 +201,26 @@ func (art *AssetTotals) TotalGPUCost() float64 {
...
 	return art.RAMCost + art.RAMCostAdjustment
 }
 
 // TotalCost returns the sum of all costs
 func (art *AssetTotals) TotalCost() float64 {
-	return art.AttachedVolumeCost + art.ClusterManagementCost + art.TotalCPUCost() +
-		art.TotalGPUCost() + art.PersistentVolumeCost + art.TotalRAMCost()
+	return art.TotalAttachedVolumeCost() + art.TotalClusterManagementCost() +
+		art.TotalCPUCost() + art.TotalGPUCost() + art.TotalLoadBalancerCost() +
+		art.TotalPersistentVolumeCost() + art.TotalRAMCost()
```
nikovacevic, line 223: Now, even if it doesn't matter per-se, we record adjustments for every field. That way there's no confusion or ambiguity about when to use the `TotalWhateverCost()` function -- we just always do.

pkg/kubecost/totals.go
```diff
@@ -328,15 +434,15 @@ func ComputeIdleCoefficients(shareSplit, key string, cpuCost, gpuCost, ramCost f
...
 	}
 
 	if allocationTotals[key].GPUCost > 0 {
-		gpuCoeff = cpuCost / allocationTotals[key].GPUCost
+		gpuCoeff = gpuCost / allocationTotals[key].TotalGPUCost()
 	}
 
 	if allocationTotals[key].RAMCost > 0 {
-		ramCoeff = ramCost / allocationTotals[key].RAMCost
+		ramCoeff = ramCost / allocationTotals[key].TotalRAMCost()
```
nikovacevic, line 445: Here we were not using adjustments, but now we are. This was contributing to idle coefficients being off as a function of the size of adjustments.

pkg/kubecost/totals.go
```diff
@@ -328,15 +434,15 @@ func ComputeIdleCoefficients(shareSplit, key string, cpuCost, gpuCost, ramCost f
 	}
 
 	if allocationTotals[key].CPUCost > 0 {
-		cpuCoeff = cpuCost / allocationTotals[key].CPUCost
+		cpuCoeff = cpuCost / allocationTotals[key].TotalCPUCost()
 	}
 
 	if allocationTotals[key].GPUCost > 0 {
-		gpuCoeff = cpuCost / allocationTotals[key].GPUCost
+		gpuCoeff = gpuCost / allocationTotals[key].TotalGPUCost()
```
nikovacevic, line 441: Here's the cpu/gpu typo

AjayTripathy (approved):

pkg/kubecost/totals.go
```diff
@@ -214,17 +268,34 @@ func ComputeAssetTotals(as *AssetSet, prop AssetProperty) map[string]*AssetTotal
...
+			// 4. Subtract (3 - 2) to get adjustment in doller-terms
+			// 5. Use (2 + 4) as total cost, so (2) is "cost" and (4) is "adjustment"
 
-			totalGPUCost := node.GPUCost * (1.0 - node.Discount)
-			gpuCost := totalGPUCost * adjustmentRate
-			gpuCostAdjustment := totalGPUCost - gpuCost
+			// Example:
+			// - node.CPUCost   = 10.00
+			// - node.Discount  =  0.20  // We assume a 20% discount
+			// - adjustmentRate =  0.75  // CUR says we need to reduce to 75% of our post-discount node cost
```
michaelmdresser, line 280: @nikovacevic This comment is superb. For my own education, I didn't realize that billing worked like this. Is node.Discount something like an RI discount and adjustmentRate is like an account- or project-wide discount negotiated with AWS?

michaelmdresser:
Does a rebuild have to happen to reflect these fixes in existing data? Or are totals built from ETL data and not serialized?

michaelmdresser:
Also, when do we plan to get these changes back into develop?

AjayTripathy:
These totals are built from ETL data, so no need to rebuild AFAIK. I can open a PR to merge these back into develop right now: https://github.com/kubecost/cost-model/pull/1062

pkg/kubecost/totals.go
nikovacevic: `node.Discount` is a combination of settings and special knowledge, encoded in:
```
cm.Provider.CombinedDiscountForNode(node.NodeType, node.Preemptible, discount, negotiatedDiscount)
```

`adjustmentRate` is just a way of spreading the node's single cost `adjustment` across it's constituent resources.

So, on GKE your discount might be an expected 30% for an n1 with sustained use, or 20% for an n2 sustained use, or 0% for an e2 sustained use. Then, after we reconcile with the CUR, a single adjustment is added to or subtracted from the node's cost, but we don't know how that applies to CPU, RAM and GPU costs. So we use adjustment rate to spread the adjustment across the resources. With discount _and_ adjustment applied in the correct order, we get a node with CPU, RAM, and GPU that sum to agree with the CUR.

michaelmdresser: Wheeeew. Alright, thanks for the explanation!

pkg/kubecost/totals.go
mbolt35: bitmoji

PR #1064, opened 2022-02-16
migrate to AWS-SDK-V2

Sean-Holcomb:
## What does this PR change?
This PR migrates the AWS cloud provider to use the aws-sdk-go-v2 which has been shown to have significant improvements in Athena memory usage. This change has been implemented on all integrations with AWS services and uses a unified configuration approach to ensure that different configurations of Kubecost are all functional.

- https://github.com/kubecost/kubecost-cost-model/pull/664
- 

AWS users will see memory performance increases in their cloud assets and reconciliation processes.

- 
- 

## How was this PR tested?
This PR tested by rebuilding assets with the new api and ensuring that reconciliation and cloud assets were correctly repopulated in addition to pricing sources continuing to function properly. This was tested on deployments with the following configuration. Service Key Secret, AWS multi-cloud, Azure multi-cloud, s3 CSV, Service Account Annotation, Master Payer ARN, Front-End with no spot bucket. Each of these configurations were able to repopulate cloud assets and repopulate and integrate with pricing sources where applicable.


pkg/cloud/awsprovider.go
```diff
@@ -161,20 +158,42 @@ type AWS struct {
...
+		// The AWS SDK v2 requires an object fulfilling the CredentialsProvider interface, which cloud.AWSAccessKey does
+		cfg, err = config.LoadDefaultConfig(context.TODO(), config.WithCredentialsProvider(accessKey), config.WithRegion(region))
+		if err != nil {
+			return cfg, fmt.Errorf("failed to initialize AWS SDK config for region %s: %s", region, err)
+		}
+	}
+
+	return cfg, nil
+}
+
```
Sean-Holcomb, line 196: Base config function which should be used by all non athena SDK v2 methods takes into account possibility of annotation configuration

pkg/cloud/awsprovider.go
```diff
@@ -370,6 +391,23 @@ type AwsAthenaInfo struct {
...
+	if aai.MasterPayerARN != "" {
+		// Create the credentials from AssumeRoleProvider to assume the role
+		// referenced by the roleARN.
+		stsSvc := sts.NewFromConfig(cfg)
+		creds := stscreds.NewAssumeRoleProvider(stsSvc, aai.MasterPayerARN)
+		cfg.Credentials = awsSDK.NewCredentialsCache(creds)
+	}
+	return cfg, nil
+}
+
```
Sean-Holcomb, line 410: Configuration method for Athena SDK v2, locks region to athenaRegion and also applies Master Payer ARN if applicable.

pkg/cloud/awsprovider.go
```diff
@@ -403,48 +441,98 @@ func (aws *AWS) GetConfig() (*CustomPricing, error) {
...
+	if config.ServiceKeySecret == "" {
+		config.ServiceKeySecret = env.GetAWSAccessKeySecret()
+	}
+
+	if config.ServiceKeyName == "" && config.ServiceKeySecret == "" {
+		log.DedupedInfof(1, "missing service key values for AWS cloud integration attempting to use service account integration")
+	}
+
+	return &AWSAccessKey{AccessKeyID: config.ServiceKeyName, SecretAccessKey: config.ServiceKeySecret}, nil
+}
```
Sean-Holcomb, line 468: Largely moved from KSM as it is useful here

pkg/cloud/awsprovider.go
```diff
@@ -1177,88 +1265,28 @@ func (awsProvider *AWS) ClusterInfo() (map[string]string, error) {
...
-		for _, tag := range inst.Tags {
-			tagKey := *tag.Key
-			for matchNum, group := range clusterIdRx.FindStringSubmatch(tagKey) {
-				if matchNum != 1 {
-					continue
-				}
-				return makeStructure(group)
-			}
-		}
-	}*/
```
Sean-Holcomb, line 1261: remove commented code

pkg/cloud/awsprovider.go
```diff
@@ -1407,18 +1418,10 @@ func (a *AWS) GetAddresses() ([]byte, error) {
...
-					default:
-						errorCh <- aerr
-					}
-					return
-				} else {
-					errorCh <- err
-					return
-				}
+				errorCh <- err
+				return
```
Sean-Holcomb, line 1424: awserr casting did not make a difference here

pkg/cloud/awsprovider.go
```diff
@@ -1537,107 +1529,91 @@ func (a *AWS) GetDisks() ([]byte, error) {
...
 	})
 }
 
-func generateAWSGroupBy(lastIdx int) string {
-	sequence := []string{}
-	for i := 1; i < lastIdx+1; i++ {
-		sequence = append(sequence, strconv.Itoa(i))
-	}
-	return strings.Join(sequence, ",")
-}
```
Sean-Holcomb, line 1570: remove dead code

pkg/cloud/awsprovider.go
```diff
@@ -1177,88 +1265,28 @@ func (awsProvider *AWS) ClusterInfo() (map[string]string, error) {
...
 		return m, nil
 	}
 
+	if c.ClusterName != "" {
+		return makeStructure(c.ClusterName)
+	}
+
 	maybeClusterId := env.GetAWSClusterID()
 	if len(maybeClusterId) != 0 {
+		klog.V(2).Infof("Returning \"%s\" as ClusterName", maybeClusterId)
```
Sean-Holcomb, line 1286: simplified this section as it was very confusing before

pkg/cloud/awsprovider.go
mbolt35: So, this config will always use annotation configuration values? Do you know the order in which it evaluates auth methods?

pkg/cloud/awsprovider.go
mbolt35: Yikes. Yeah, good fix!

pkg/cloud/awsprovider.go
```diff
@@ -1537,107 +1529,91 @@ func (a *AWS) GetDisks() ([]byte, error) {
...
+	isQueryStillRunning := true
+	for isQueryStillRunning {
+		qe, _ := client.GetQueryExecution(ctx, inp)
+		if qe.QueryExecution.Status.State == "SUCCEEDED" {
+			isQueryStillRunning = false
 		}
-		return nil
-	} else {
-		return fmt.Errorf("No results available for %s", query)
+		time.Sleep(10 * time.Second)
```
mbolt35: Do we really want to wait for 10 seconds?

mbolt35: I definitely think 10seconds is a long time to wait between polls. Was 2 seconds causing problems?

pkg/cloud/awsprovider.go
Sean-Holcomb: So this does prioritize set ENV variables, however the methods which retrieve the AccessKeyID and SecretAccessKey set those env variables so they should not be set. The list of priority is here. https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials. This does leave the door open to other config methods if the user should so choose, but I think it is fair to say that if you want to use the service account annotation configuration you should not have erroneous other configuration methods floating around in you container/deployment.

pkg/cloud/awsprovider.go
Sean-Holcomb: I lifted this directly from @AjayTripathy's branch, 10 seconds does seem a little long. I think 2 sounds more appropriate

Sean-Holcomb: updated to 2

mbolt35 (approved):
LGTM!

PR #1076, opened 2022-02-24
Don't specify architecture in Dockerfile go build

michaelmdresser:
Specifying the architecture forces the image to only support AMD64 even
if we're trying to build an ARM64 image in an ARM64 environment with
something like docker buildx. Making this change means the built binary
will support the architecture of the environment it was built in.

## How was this PR tested?

CI build on this repo.


mbolt35 (approved):
LGTM

michaelmdresser:
Picked into `v1.91` by https://github.com/kubecost/cost-model/pull/1079

PR #1080, opened 2022-02-24
Go-JSON Replacement 

mbolt35:
## What does this PR change?
* Replaces json-iter library with go-json: https://github.com/goccy/go-json
  * The json-iter library has a show-stopping bug which will degrade performance of the go runtime almost to a full stop. See this issue for details: https://github.com/json-iterator/go/issues/603
  * go-json seems to have a more active maintainer

* Fix for requests which don't receive responses. 

## How was this PR tested?
* Using several specific benchmarks that repro'd the bug case in json-iter
* All kubecost ETL structure json encode/decode unit tests are passing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,8 +21,8 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.14.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/getsentry/sentry-go v0.6.1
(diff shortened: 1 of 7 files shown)


AjayTripathy:
Were we able to test this on the scale cluster for possible performance regressions?

nikovacevic (approved):

AjayTripathy (approved):

PR #1092, opened 2022-03-02
Update S3 Storage Implementation

mbolt35:
## What does this PR change?
* Updates the S3 storage implementation to the latest thanos configuration specification. 

* Updates the configuration for our S3 storage implementation to match the latest Thanos Configuration: https://thanos.io/tip/thanos/storage.md/#s3

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1250

### DIFF
--- a/pkg/storage/s3storage.go
+++ b/pkg/storage/s3storage.go
@@ -15,6 +15,9 @@ import (
 
 	"github.com/kubecost/cost-model/pkg/log"
 
+	aws "github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+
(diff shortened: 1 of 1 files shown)


AjayTripathy:
Could you note how this was tested?

mbolt35:
@AjayTripathy This is a PR to bring the implementation to spec with Thanos. How would you propose we test that?

AjayTripathy:
The specific issue here was that eks annotations weren't working when the same key config was applied to Thanos as to kubecost cost-model right? So I would propose we set a cluster up configured that way and note that we can save the etl.

AjayTripathy:
We have a cluster set up now, so approving and we can look at the nightly.

AjayTripathy (approved):

PR #1093, opened 2022-03-02
Make service check thread safe

Sean-Holcomb:
## What does this PR change?
This PR is to address a concurrent map panic which appear to occur when two processes attempt to write to the serviceAccountChecks map of the AWS or Azure provider. To prevent this I have wrapped the map in a struct which also contains a Mutex and routed all reading and writing through methods which use the Mutex. Additionally I have added an initializer that runs in the initializers of the provider implementation, so there is no longer a need for the extra nil check initialization peppered around the code.

It is worth noting that in my testing I found that this object no longer appears to be in use. It was primarily used on the assets page to check if the OOC configuration banner needed to be displayed, however this functionality is now covered by the Cloud Status feature, and the API call to get the contents of this map is never actually made on Azure or AWS with standard cloud configuration. Rather than merging this update it may be better to remove this functionality entirely. This is the relevant function on the FE https://github.com/kubecost/cost-analyzer-frontend/blob/1fc3d2928d34ec2962f0b9c76fd82b5cb210ad5f/src/pages/Assets/index.tsx#L506

- 
- 

Users will no longer experience rare concurrent map write panic

- partially addresses [ZD1445](https://kubecost.zendesk.com/agent/tickets/1445)
- 

## How was this PR tested?
Tested on Azure and AWS cluster with standard cloud configurations

### DIFF
--- a/pkg/cloud/provider.go
+++ b/pkg/cloud/provider.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/kubecost/cost-model/pkg/util"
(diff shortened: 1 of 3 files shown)


pkg/cloud/provider.go
```diff
@@ -212,6 +213,39 @@ type ServiceAccountStatus struct {
 	Checks []*ServiceAccountCheck `json:"checks"`
 }
 
+// ServiceAccountChecks is a thread safe map for holding ServiceAccountCheck objects
+type ServiceAccountChecks struct {
+	serviceAccountChecks map[string]*ServiceAccountCheck
+	lock                 *sync.RWMutex
```
mbolt35: For what it's worth, I have used `sync.Mutex` pointers exclusively in our code base for a long time, and I actually just had to admit recently that I was wrong to @michaelmdresser as much as it pained me to do so 😄 . Using a pointer here isn't the end of the world, but the problematic case is on a dereferenced copy (you'd get a copy of the resource and a copy of the pointer to the Mutex). This sounds really bad, and it is, but dereferencing pointer types in general will typically not end well (especially using any sort of complex type). For instance, even if you were to change the mutex here, if someone deref copied a ServiceAccountChecks instance, the map would not be shared between two instances as well. 

So while I have generally been a preference based advocate for using `new(sync.Mutex)` in the past, I think the better/safer choice here is to use either `lock sync.RWMutex` or embed `sync.RWMutex`. I will follow suit in my future code writing as well to ensure we don't end up with any nasty deref copied shared locks :(

mbolt35:
This looks good to me, you can decide yourself on the `sync.Mutex` case, but I think going with the no-pointer version is safer _generally_. Although in the context of this changeset, both the pointer version and non-pointer version are completely ok.

mbolt35 (approved):
LGTM

pkg/cloud/provider.go
Sean-Holcomb: thanks for letting me know, I think the embedded makes the most sense here.

PR #1094, opened 2022-03-02
duplicate iterated item before taking address

Sean-Holcomb:
## What does this PR change?
Credit to @mbolt35 for solving this.
This PR fixes a bug with the `projectAddresses` and `projectDisks` endpoints where the results would be the correct length but each item in their respective arrays were duplicates.

- 
- 

Users can now use these endpoints again.

- 
- 

## How was this PR tested?
This PR was tested manually on an AWS cluster, were the endpoints are now returning unique results.


nikovacevic (approved):
Nice

PR #1104, opened 2022-03-08
Changes to support ETL restructuring

nikovacevic:
See https://github.com/kubecost/kubecost-cost-model/pull/690


Sean-Holcomb (approved):
Looks good, excited for the new interfaces

PR #1129, opened 2022-03-24
Assign default GKE cluster name if none set

biancaburtoiu:
## What does this PR change?
This PR assigns a default cluster name "GKE Cluster # 1" at the moment of building the cluster metadata in the case where none had been previously set. 

No

GKE cluster names will now have a fallback value.

opened due to https://github.com/kubecost/cost-model/issues/1125 but will not constitute a fix for the user's issue

## How was this PR tested?
N/A

N/A

### DIFF
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -296,7 +296,7 @@ func (gcp *GCP) UpdateConfig(r io.Reader, updateType string) (*CustomPricing, er
 	})
 }
 
-// ClusterName returns the name of a GKE cluster, as provided by metadata.
+// ClusterInfo returns information on the GKE cluster, as provided by metadata.
 func (gcp *GCP) ClusterInfo() (map[string]string, error) {
 	remoteEnabled := env.IsRemoteEnabled()
 
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -318,6 +318,11 @@ func (gcp *GCP) ClusterInfo() (map[string]string, error) {
 		attribute = c.ClusterName
 	}
 
+	// Use a default name if none has been set until this point
+	if attribute == "" {
+		attribute = "GKE Cluster #1"
+	}
+
 	m := make(map[string]string)
 	m["name"] = attribute
 	m["provider"] = "GCP"


pkg/cloud/gcpprovider.go
```diff
@@ -318,6 +318,11 @@ func (gcp *GCP) ClusterInfo() (map[string]string, error) {
 		attribute = c.ClusterName
 	}
 
+	// Use a default name if none has been set until this point
+	if attribute == "" {
+		attribute = "GKE Cluster #1"
```
michaelmdresser, line 323: Is this fallback cluster name distinct enough that it can be recognized as the fallback value when doing debugging with a user?

michaelmdresser (commented):
Do you know if there will be any impact in multi-cluster setups where the metadata isn't available for 2+ clusters? I suppose they're probably already broken, so maybe this is a non-issue.

pkg/cloud/gcpprovider.go
biancaburtoiu: The `GetClusterInfo()` is implemented for all providers; in both the [aws](https://github.com/kubecost/cost-model/blob/5a227f20f69d37df06c46734839f27b52f4ccbaf/pkg/cloud/awsprovider.go#L1278) and [azure](https://github.com/kubecost/cost-model/blob/4b3f80128a0f42e8d6216482baa4ae28792e7c00/pkg/cloud/azureprovider.go#L1151) implementations the default cluster name is following the scheme `<PROVIDER> Cluster #1` so I set the same for consistency!

biancaburtoiu: I'm not sure about the multi-cluster edge case, I'll look into it :)

michaelmdresser: Thanks for the references, that's what I was hoping! Sorry for the slow response here.

michaelmdresser (approved):

michaelmdresser:
cc @AdamStack18 this may have slipped your gaze, not sure where this PR falls under release priorities (though it should be easy enough to merge regardless!)

dwbrown2:
What's the plan here? I'm guessing we missed this release?

Adam-Stack-PM:
@dwbrown2 I will update the PR template to ensure the right labels are added in the future.  I will leave it up to @nealormsbee to make the call if this needs to be pushed to v1.94.

nealormsbee:
It's straightforward and a nice thing to have. I'm just going to merge it.

PR #1131, opened 2022-03-25
Implement UID id of pods to fix time running inaccuracy

kaelanspatel:
## What does this PR change?
Adds a flag `.Values.kubecostProductConfigs.ingestPodUID` which toggles appending pod UID to pod names in the costmodel, with associated functionality.

A short summary of how this works if the flag is enabled:
- UID is queried from `kube_pod_container_status_running` (this requires either Kubecost's replicated KSM metrics, or KSM v2.1.0+)
- UID is appended to `podKey` in `buildPodMap()`
- A map is created that tracks the "default" `podKey` to the edited `podKey`. Because pods are no longer only IDed by name with this flag, there now can be multiple pods where there was once one pod. Other metrics don't have UIDs and therefore cannot generate the edited `podKey`(s) to match metrics to pods in the `podMap`. Therefore, we track every "default" `podKey` and apply metrics that generate that key to all the edited `podKeys` that match up with it.

This allows us to catch uncontrolled pods that start/stop under the same name and register them as different allocations.

- https://github.com/kubecost/cost-analyzer-helm-chart/pull/1327
- 

Add an option to track pods by UID as well as name.

- https://github.com/kubecost/cost-model/issues/991
- 

## How was this PR tested?
Tested for proper behavior by creating pods with varying pods with the same name and viewing behavior between develop/flag false Allocation endpoint queries vs flag true queries. Query results on the left are normal behavior, query results on the right are with flag true.

With three pods all called `samename':
image

Querying `/model/allocation/summary?window=2022-03-22T18%3A00%3A00Z%2C2022-03-22T22%3A00%3A00Z&aggregate=pod&external=true&accumulate=true&shareIdle=false&idleByNode=false&shareTenancyCosts=true&shareSplit=weighted` gives:
[Result](https://user-images.githubusercontent.com/32113845/160028332-5395972e-8747-4937-9c47-1ff64779800a.png)

With two pods both called `samename_deluxe` with labels/cpu/PVC:
image

Querying `/model/allocation?window=2022-03-23T23%3A00%3A00Z%2C2022-03-24T00%3A00%3A00Z&aggregate=pod&accumulate=true` gives:
[Result](https://user-images.githubusercontent.com/32113845/160027678-45dd897a-37ef-4966-b16e-f94d9cedad30.png)

Also, as a baseline to ensure data is not being changed, looking at an un-duplicated controlled Kubecost instance by querying `/model/allocation?window=yesterday&aggregate=pod` shows no significant change:
[Result](https://user-images.githubusercontent.com/32113845/160028755-f2e2f347-5449-4a9b-a802-e20bfe89e939.png)

Tested several aggregations with/without UID against develop and saw no significant change of values.
Tested flag behavior when unset/true/false.

No.


pkg/costmodel/allocation.go
```diff
@@ -110,7 +111,26 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
...
+
+	// This is because ingesting UID allows us to catch uncontrolled pods
+	// with the same names. However, this will lead to a many-to-one metric
+	// to podKey relation, so this map allows us to map the metric's
+	// "<pod_name>" key to the edited "<pod_name> <pod_uid>" keys in podMap.
+	ingestPodUID := env.IsIngestingPodUID()
+	podUIDKeyMap := make(map[podKey][]podKey)
+
+	if ingestPodUID {
+		log.Infof("CostModel.ComputeAllocation: ingesting UID data from KSM metrics...")
```
michaelmdresser: This log will spam because of how often `ComputeAllocation` is called. I recommend either making it a `log.Debugf` or moving it to a single check on process start.

pkg/costmodel/allocation.go
```diff
@@ -575,7 +604,26 @@ func (cm *CostModel) buildPodMap(window kubecost.Window, resolution, maxBatchSiz
 			return err
 		}
 
-		applyPodResults(window, resolution, podMap, clusterStart, clusterEnd, resPods)
+		// queryFmtPodsUID will return both UID-containing results, and non-UID-containing results,
+		// so filter out the non-containing results so we don't duplicate pods.
```
michaelmdresser: If `ingestPodUID` is true, why would any of the results not have a UID?

pkg/costmodel/allocation.go
```diff
@@ -606,6 +654,24 @@ func applyPodResults(window kubecost.Window, resolution time.Duration, podMap ma
...
+		// If pod UIDs are being used to ID pods, append them to the pod name in
+		// the podKey.
+		if ingestPodUID {
+
+			uid, err := res.GetString("uid")
+			if err != nil {
+				log.Warningf("CostModel.ComputeAllocation: UID ingestion enabled, but query result missing field: %s", err)
+			} else {
+
+				newKey := newPodKey(cluster, namespace, pod+" "+uid)
```
michaelmdresser, line 835: I'm not a huge fan of using a space to delimit here, but I understand that its hard to pick something clear here. Did you give any thought to using another delimeter? Just want to make sure we think it through, because I suspect we'll be stuck with our choice for a long time.

pkg/costmodel/allocation.go
```diff
@@ -720,302 +786,422 @@ func applyPodResults(window kubecost.Window, resolution time.Duration, podMap ma
...
 
-		if _, ok := pod.Allocations[container]; !ok {
-			pod.AppendContainer(container)
-		}
+		for _, pod := range pods {
+			container, err := res.GetString("container")
+			if err != nil {
+				log.DedupedWarningf(10, "CostModel.ComputeAllocation: CPU request query result missing 'container': %s", key)
+				continue
+			}
```
michaelmdresser: It looks like this logic should not be run in this new loop, but rather outside of it, so we that the `res` is skipped just once, instead of repeating the check+skip `len(pods)` number of times.

pkg/costmodel/allocation.go
```diff
@@ -720,302 +786,422 @@ func applyPodResults(window kubecost.Window, resolution time.Duration, podMap ma
...
 
-		pod.Allocations[container].CPUCoreRequestAverage = res.Values[0].Value
+			if _, ok := pod.Allocations[container]; !ok {
+				pod.AppendContainer(container)
+			}
```
michaelmdresser, line 1050: Say we have the following situation: two pods, with the same name, in the same window, but with different containers running. Like so:

same-pod-different-container-cpu-request

What kind of `Allocation`s do we end up with?

My suspicion is that we get 4:
1) Pod name: "A 1" +
    Container name "C1" + 
    CPUCoreRequestAverage: 1 +
    Start: 0
    End: halfway through window
    Minutes: ~half of window
2) Pod name: "A 1" +
    Container name "C2" + 
    CPUCoreRequestAverage: 3 +
    Start: 0
    End: halfway through window
    Minutes: ~half of window
3) Pod name: "A 2" +
    Container name "C1" + 
    CPUCoreRequestAverage: 1 +
    Start: halfway through window
    End: end of window
    Minutes: ~half of window
4) Pod name: "A 2" +
    Container name "C2" + 
    CPUCoreRequestAverage: 3 +
    Start: halfway through window
    End: end of window
    Minutes: ~half of window

In this situation, do we double-count the cost of "UID 1 C1" and "UID 2 C2"?

(I'm picking this line of code merely for convenience, this is a general design question about applying metric information to _all_ pods with the same name if the UID is not available for a given metric)

michaelmdresser (commented):
A few comments to start. My one big question right now is about a slightly pathological edge case and possible double counting. I'm not entirely done with review, but I'd like to ask the big question up front before I go much further.

pkg/costmodel/allocation.go
michaelmdresser: Kaelan and I discussed. One thing I'd like to note is that according to https://github.com/kubecost/cost-model/blob/485851c6fbd8f1d23656bb279cce38006ffaa0a8/pkg/costmodel/allocation.go#L2522-L2538, the container start and end come from the POD start and end. So we're already wrong in this case right now.

pkg/costmodel/allocation.go
kaelanspatel: Changed to `log.Debugf`.

pkg/costmodel/allocation.go
kaelanspatel: Discussed this w/ Michael. This is because Kubecost has two `kube_pod_container_status_running` metrics by default (replicated and included KSM versions). Querying avg by UID will return both, hence we filter. Updated the comment to clarify this.

pkg/costmodel/allocation.go
kaelanspatel: Discussed this w/ Michael. It's not ideal, but seeing as this is already an artificial manipulation of the pod name, I think whatever we choose will be mostly the same.

pkg/costmodel/allocation.go
kaelanspatel: Good catch, yes. Changed this to occur outside the `pod` loop so that it skips the entire thing.

pkg/costmodel/allocation.go
kaelanspatel: As Michael says, this is extant behavior. In the above case currently (or with these UID changes off), the pod will simply show one pod named "A" with two containers "C1" and "C2" running for its entire runtime.

This is effectively the same double-counting as which Michael explains above: two pods "A 1" and "A 2" having two containers each running for each pod's entire runtime is, added up, approximately the same runtime as the one pod "A" with two containers, just more "visibly wrong" because we can see the different pods but the containers are incorrect.

Discussed this with Ajay. There's really no way to mitigate this, the only link between how we get the containers/allocations and the pods is the name. We can't see which container belongs to which pod specifically if 1. the pod names are the same, and 2. the container names are different (and therefore a distinct entry in pod.Allocations), so this is just an unfortunate edge case.

pkg/costmodel/allocation.go
michaelmdresser: Just for posterity, K8s labels and IDs don't use spaces, so this should be safe. Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

pkg/costmodel/allocation.go
```diff
@@ -1026,41 +1230,72 @@ func applyGPUsAllocated(podMap map[podKey]*Pod, resGPUsRequested []*prom.QueryRe
...
+		} else {
+			pods = []*Pod{pod}
 		}
 
-		for _, alloc := range pod.Allocations {
-			alloc.NetworkTransferBytes = res.Values[0].Value / float64(len(pod.Allocations))
+		for _, pod := range pods {
+			for _, alloc := range pod.Allocations {
+				alloc.NetworkTransferBytes = res.Values[0].Value / float64(len(pod.Allocations))
```
michaelmdresser: Should this also be divided by `float64(len(pods))`?

pkg/costmodel/allocation.go
```diff
@@ -1070,18 +1305,33 @@ func applyNetworkTotals(podMap map[podKey]*Pod, resNetworkTransferBytes []*prom.
...
+			}
+		} else {
+			pods = []*Pod{pod}
 		}
 
-		for _, alloc := range pod.Allocations {
-			alloc.NetworkReceiveBytes = res.Values[0].Value / float64(len(pod.Allocations))
+		for _, pod := range pods {
+			for _, alloc := range pod.Allocations {
+				alloc.NetworkReceiveBytes = res.Values[0].Value / float64(len(pod.Allocations))
```
michaelmdresser: Same question.

pkg/costmodel/allocation.go
```diff
@@ -1100,15 +1350,30 @@ func applyNetworkAllocation(podMap map[podKey]*Pod, resNetworkGiB []*prom.QueryR
...
 
-		for _, alloc := range pod.Allocations {
-			gib := res.Values[0].Value / float64(len(pod.Allocations))
-			costPerGiB := costPerGiBByCluster[podKey.Cluster]
-			alloc.NetworkCost = gib * costPerGiB
+		for _, pod := range pods {
+			for _, alloc := range pod.Allocations {
+				gib := res.Values[0].Value / float64(len(pod.Allocations))
+				costPerGiB := costPerGiBByCluster[podKey.Cluster]
+				alloc.NetworkCost = gib * costPerGiB
```
michaelmdresser: Same question.

pkg/costmodel/allocation.go
kaelanspatel: That would be "more correct" I think, as the query value is a summation. Good catch. Added.

pkg/costmodel/allocation.go
kaelanspatel: Yes, added.

pkg/costmodel/allocation.go
kaelanspatel: Yes, added.

michaelmdresser (approved):
Awesome work! I know it was challenging to deal with the complexity here. Hopefully we don't have to think too much about this edge-case use in the future thanks to this work :grin:

michaelmdresser:
Just a heads up, looks like there's a merge conflict now. Hopefully it isn't too hairy to deal with.

kaelanspatel:
Yep, saw it last night. Thought I merged this...

Will fix and merge.

PR #1135, opened 2022-03-31
Quick solution for ETL Allocation hourly build/rebuild

### LINKED ISSUE #1317: Rebuild daily ETL via hourly
Support ETL rebuilding in chunks. There are certain architectures of PromQL that are resilient to a high QPS but not queries that run over long periods of time. This can cause our Daily ETL rebuild pipelines to fail while the hourly pipelines continue to succeed. We should make this easier by breaking down Daily ETL rebuilds into smaller timerange rebuilds. It may be easiest to just support a 1h resolution but ideally we could have any divisor of 24 supported as a chunk size to run a rebuiild on.

nikovacevic:
## What does this PR change?
- Spike solution of ETL Allocation hourly build/rebuild with accumulation;
- Convert Allocation queries to use QueryAtTime;
- Guard ClusterNodes, Disks, and LoadBalancers from bad input.

- https://github.com/kubecost/kubecost-cost-model/pull/707
- https://github.com/kubecost/cost-analyzer-helm-chart/pull/1341

## Does this close any issues?
- Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1317

ComputeAllocation (which drives Allocation ETL) will query for durations no longer than the configured maximum Prometheus query duration, which defaults to 1h.

## How was this PR tested?
Tested manually using logs on `gke-niko` (logs attached).

First I installed using helm:
```
helm upgrade -i -f gke-niko-values.yaml kubecost ./cost-analyzer \
  --namespace kubecost \
  --set kubecostFrontend.fullImageName=null \
  --set kubecostModel.fullImageName=nikovacevic/kubecost-cost-model:hourly12
```
As the logs bear out, when the default setting was left unchanged, Prometheus queries proceeded as usual (no [Hourly] logs in `hourly-logs-max-1d`) and there is a mixture of `[1d]` and `[1h]` queries in `prom-logs-max-1d`.
[hourly-logs-max-1d.txt](https://github.com/kubecost/cost-model/files/8392863/hourly-logs-max-1d.txt)
[prom-logs-max-1d.txt](https://github.com/kubecost/cost-model/files/8392864/prom-logs-max-1d.txt)

Then I used the new helm flag to set the max minutes for a Prometheus query to 60:
```
helm upgrade -i -f gke-niko-values.yaml kubecost ./cost-analyzer \
  --namespace kubecost \
  --set kubecostFrontend.fullImageName=null \
  --set kubecostModel.fullImageName=nikovacevic/kubecost-cost-model:hourly12 \
  --set kubecostModel.maxPrometheusQueryDurationMinutes=60
```
And, as the subsequent logs bear out, there are now indications that daily builds are proceeding using incremental build-ups of 1h builds. And the Prometheus logs have only instances of `[1h]` window durations.

[hourly-logs-max-1h.txt](https://github.com/kubecost/cost-model/files/8392885/hourly-logs-max-1h.txt)
[prom-logs-max-1h.txt](https://github.com/kubecost/cost-model/files/8392886/prom-logs-max-1h.txt)

As a side-effect, builds on my small cluster seem to be about 3-4x slower, simply due to the sheer number of queries being run, but that's a rather unscientific take from a small cluster. (e.g. a daily build takes ~6-8s instead of ~1-2s.)

Leaving undocumented for now. Up to @AjayTripathy how to proceed there.

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -117,18 +117,12 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 			return kubecost.NewAllocationSet(start, end), fmt.Errorf("error computing allocation for %s: %s", kubecost.NewClosedWindow(s, e), err)
 		}
 
-		// TODO remove log
-		log.Infof("[Hourly] computeAllocation => AllocationSet{ %s window, %d allocations, %.5f total cost }", as.Window, as.Length(), as.TotalCost())
-
 		// Append to the range
 		asr.Append(as)
 
 		// Set s equal to e to set up the next query, if one exists.
 		s = e
 	}
-
-	// TODO remove log
-	log.Infof("[Hourly] ComputeAllocation(%s) => AllocationSetRange{ %s window, %d sets, %.5f total cost }", kubecost.NewClosedWindow(start, end), asr.Window(), asr.Length(), asr.TotalCost())
 
 	// Populate annotations, labels, and services on each Allocation. This is
 	// necessary because Properties.Intersection does not propagate any values
@@ -137,6 +131,10 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	allocationAnnotations := map[string]map[string]string{}
 	allocationLabels := map[string]map[string]string{}
 	allocationServices := map[string]map[string]bool{}
+
+	// Also record errors and warnings, then append them to the results later.
+	errors := []string{}
+	warnings := []string{}
 
 	asr.Each(func(i int, as *kubecost.AllocationSet) {
 		as.Each(func(k string, a *kubecost.Allocation) {
@@ -167,6 +165,9 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 				}
 			}
 		})
+
+		errors = append(errors, as.Errors...)
+		warnings = append(warnings, as.Warnings...)
 	})
 
 	// Accumulate to yield the result AllocationSet. After this step, we will
@@ -204,9 +205,6 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	// Maintain RAM and CPU max usage values by iterating over the range,
 	// computing maximums on a rolling basis, and setting on the result set.
 	asr.Each(func(i int, as *kubecost.AllocationSet) {
-
-		// TODO pass through warnings and errors
-
 		as.Each(func(key string, alloc *kubecost.Allocation) {
 			resultAlloc := result.Get(key)
 			if resultAlloc == nil {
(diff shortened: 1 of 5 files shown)


nikovacevic:
Found some bugs... this is not ready yet.

pkg/costmodel/allocation.go
```diff
@@ -83,6 +83,169 @@ func (cm *CostModel) Name() string {
...
+		}
+
+		// Set start and end parameters (s, e) for next individual computation.
+		e = s.Add(duration)
+
+		// Compute the individual AllocationSet for just (s, e)
+		as, err := cm.computeAllocation(s, e, resolution)
+		if err != nil {
+			return kubecost.NewAllocationSet(start, end), fmt.Errorf("error computing allocation for %s: %s", kubecost.NewClosedWindow(s, e), err)
+		}
```
nikovacevic, line 118: @AjayTripathy this is one we might discuss -- should this fail, or should it continue with an `Error` in the `AllocationSet`? And should we retry, or save that for a P1 implementation?

pkg/costmodel/allocation.go
```diff
@@ -110,142 +273,141 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	clusterStart := map[string]time.Time{}
 	clusterEnd := map[string]time.Time{}
 
-	cm.buildPodMap(window, resolution, env.GetETLMaxBatchDuration(), podMap, clusterStart, clusterEnd)
+	// TODO:CLEANUP remove "max batch" idea and clusterStart/End
```
nikovacevic, line 275: Note: this is for a separate clean up effort (been using that tag `TODO:CLEANUP` to plant flags for ideas) and not for the current PR. It's an outstanding thing I don't want to touch now.

AjayTripathy (approved):

AjayTripathy:
This LGTM. I deployed and tested this noting that a rebuild of 24h data in 1h chunks was consistent.

pkg/env/costmodelenv.go
```diff
@@ -429,14 +429,10 @@ func IsETLEnabled() bool {
...
-// timeout for large windows; e.g. if a 24h query is expected to timeout, but
-// a 6h query is expected to complete in 1m, then 6h could be a good value.
-func GetETLMaxBatchDuration() time.Duration {
-	// Default to 6h
-	hrs := time.Duration(GetInt64(ETLMaxBatchHours, 6))
-	return hrs * time.Hour
+func GetETLMaxPrometheusQueryDuration() time.Duration {
+	dayMins := 60 * 24
+	mins := time.Duration(GetInt64(ETLMaxPrometheusQueryDurationMinutes, int64(dayMins)))
+	return mins * time.Minute
```
michaelmdresser, line 435: Do you think we should force this number of minutes to be a divisor of `60 * 24` or `60`? I'm wondering what will happen if we try to build in prime number windows, for example.

nikovacevic:
Here's some before/after testing for accuracy:

[before-controller.txt](https://github.com/kubecost/cost-model/files/8523788/before-controller.txt)
[after-controller.txt](https://github.com/kubecost/cost-model/files/8523786/after-controller.txt)
[before-namespace.txt](https://github.com/kubecost/cost-model/files/8523789/before-namespace.txt)
[after-namespace.txt](https://github.com/kubecost/cost-model/files/8523787/after-namespace.txt)

In general, I'm seeing error margins between 0% and 0.05%, which seems perfectly good to me.

@AjayTripathy are we good to merge, then? Anything else you want to test?

pkg/env/costmodelenv.go
nikovacevic: It should be fine. The code just advances in steps of `max(this value, remaining interval)`, so providing a number that is relatively prime with the duration querying will just result in a final segment of a slightly smaller window. But good question!

pkg/costmodel/allocation.go
AjayTripathy: I think it should continue with an error? We can save retries for p1

AjayTripathy:
> 



> 

I think we're good to merge!

PR #1136, opened 2022-03-31
Query spot data under new aws.SpotRefreshAllowed condition

### LINKED ISSUE #1128: Spot Data Feed integration error
I open this as suggested by @AjayTripathy in #1111 

The diagnostics page shows integration errors for `Spot Data Feed`, specifically `Config awsSpotDataBucket is missing` and
`operation error S3: ListObjects, serialization failed: serialization failed: input member Bucket must not be empty`.

Screenshot 2022-03-23 at 09 28 22

We do not have spot instances.
We have never used them and we have no plans to do so in the near future.
We have reserved EC2 instances and ES clusters. 

I tried looking for documentation to properly configure the software but couldn't find it.

Our config (for the Helm Chart):
```yaml
kubecostProductConfigs:
  clusterProfile: high-availability
  athenaProjectID: [REDACTED]
  athenaBucketName: "s3://[REDACTED]"
  athenaRegion: eu-west-1
  athenaDatabase: [REDACTED]
  athenaTable: "[REDACTED]"
#      awsSpotDataRegion: eu-west-1
  masterPayerARN: "arn:aws:iam::[REDACTED]:role/[REDACTED]"
  projectID: "[REDACTED]"
```

We used and then commented `awsSpotDataRegion` value in an attempt to resolve the issue.

gz#1557

biancaburtoiu:
## What does this PR change?
This PR introduces a new boolean `aws.SpotRefreshEnabled` which decides whether the spot data refresh should take place or not. This condition is determined by whether at least one of the fields `aws.SpotDataBucket`, `awsSpotDataRegion`, `projectID` has been defined by the user (i.e. _is a nonempty string_). In short: 
- the success case requires all three fields to be populated; then we'll query spot feed 
- if 1 <= x < 3 of these fields are populated; we won't query and the frontend should produce a warning
- if 0 of these fields are populated; we won't query and we'll assume the user's current setup doesn't make use of AWS Spot so the frontend show a greyed-out "Spot feed not set up" message

Frontend PR: https://github.com/kubecost/cost-analyzer-frontend/pull/594

Upon the changes made in this PR, the frontend will be able to render a clearer, more informative message regarding spot data feed.

- fixes https://github.com/kubecost/cost-model/issues/1128

## How was this PR tested?
Using temporary logs, I verified that the spot query executes or does not based on the conditions described above.

N/A


pkg/cloud/awsprovider.go
```diff
@@ -1011,24 +1016,26 @@ func (aws *AWS) DownloadPricingData() error {
...
 
-				// Reoccurring refresh checks update times
-				aws.refreshSpotPricing(false)
-			}
-		}()
+					// Reoccurring refresh checks update times
+					aws.refreshSpotPricing(false)
+				}
+			}()
+		}
```
biancaburtoiu, line 1038: This is just wrapping the existing logic under an `if {}` block. 
I considered writing it defensively 
```golang
if !aws.SpotRefreshAllowed {
    return nil
} 

// continue execution as normal (i.e. the query)
```
but I thought that would be less readable. Open to suggestions though!

dwbrown2:
Nice! Do you know if this would also make it so that this would be removed from warnings on other pages when spot data feed isn't set? If so that would be cool.

image

I'm actually not sure what API request these are using today. @Sean-Holcomb is the expert here I'm pretty sure!

biancaburtoiu:
@dwbrown2 thanks for pointing this out! I'll look into all the places spot data checks happen and stop warnings from appearing unless the setup actually warrants them. 

It might be possible that this same problem (warnings coming in for data we didn't set up _on purpose_) happens in other cases as well. I'll investigate and raise a tech debt issue if we need to!

pkg/cloud/awsprovider.go
```diff
@@ -57,22 +57,31 @@ func (aws *AWS) PricingSourceStatus() map[string]*PricingSource {
 
 	sps := &PricingSource{
 		Name: SpotPricingSource,
+		Enabled: true,
 	}
-	sps.Error = ""
```
biancaburtoiu, line 61: View this diff side-to-side; what happens here is I'm wrapping the existing logic into the `else` block of the `if !aws.SpotRefreshEnabled` statement. This makes it so if the flag is false we simply set `available:false, enabled:false` and a hardcoded message for the `error` field.

pkg/cloud/awsprovider.go
```diff
@@ -774,6 +784,13 @@ func (aws *AWS) DownloadPricingData() error {
 
 	aws.ConfigureAuthWith(c) // load aws authentication from configuration or secret
 
+	// Need a valid value for at least one of these fields to consider spot pricing as enabled
+	if len(aws.SpotDataBucket) != 0 || len(aws.SpotDataRegion) != 0 || len(aws.ProjectID) != 0 {
+		aws.SpotRefreshEnabled = true
+	} else {
+		aws.SpotRefreshEnabled = false
+	}
```
nikovacevic: One question here about the lifecycle of this new value: what happens when a user starts with all empty fields and then fills in one or more fields. Do we update `aws.SpotRefreshEnabled` to go from `false` to `true`?

One alternative approach to consider, which would eliminate the need to update would be to make `aws.SpotRefreshEnabled` a function, like:
```go
func (aws) SpotRefreshEnabled() bool {
    return len(aws.SpotDataBucket) != 0 || len(aws.SpotDataRegion) != 0 || len(aws.ProjectID) != 0
}
```

Just a thought!

nikovacevic (approved):
This makes sense to me. My only feedback is that one comment/question, so if you feel that's been adequately addressed, then this looks good to me. Do you feel that you've been able to do adequate testing?

pkg/cloud/awsprovider.go
biancaburtoiu: To answer your question - modifying the spot configs in the UI and then clicking 'Save changes' in the Settings page should trigger the `DownloadPricingData()` method which recalculates the value of `aws.SpotRefreshEnabled` depending on the newly-modified configs.

Thanks for suggesting the function! This way I don't need to declare a new field in the aws struct, so I'll use that 👍

biancaburtoiu: Re: testing, I've added some logs and traced them in the container while trying out different config combinations in the frontend. The `pricingSourceStatus` endpoint returns the expected results in all scenarios. If the user declares these configs through their values.yaml file, they will also be accounted for when initialising the pods. Do you think there's any other edge case to look at?

PR #1137, opened 2022-04-01
JSON-serialization NaN fix

nealormsbee:
## What does this PR change?

Adds a test for TestComputeIdleCoefficients to ensure that coefficients aren't Inf when total CPU/GPU/RAM costs are zeroed out by adjustment.

Also checks .Total\*Cost() instead of .Total\* to make the test pass.

No

The summary allocation endpoint will no longer break when adjustments zero out total costs for CPU/GPU/RAM

- Fixes https://kubecost.zendesk.com/agent/tickets/1566

## How was this PR tested?
A unit test is supplied in the PR. I have not 100% confirmed that this is the source of the error in the linked ticket. I think the fastest way to confirm will be to get the user a build with this fix and see whether it resolves their issue. Alternatively, we can try to deduce whether their GPU adjustment is zeroing-out their total GPU costs.

An additional PR to merge into the RC is forthcoming. Thanks to @Sean-Holcomb for looking at this with me.

Nope.

### DIFF
--- a/pkg/kubecost/totals.go
+++ b/pkg/kubecost/totals.go
@@ -484,15 +484,15 @@ func ComputeIdleCoefficients(shareSplit, key string, cpuCost, gpuCost, ramCost f
 		return coeff, coeff, coeff
 	}
 
-	if allocationTotals[key].CPUCost > 0 {
+	if allocationTotals[key].TotalCPUCost() > 0 {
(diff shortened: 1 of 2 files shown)


Sean-Holcomb (approved):
This seems like a strong candidate for this bug, and needs fixing anyways

PR #1140, opened 2022-04-01
[WIP] feat: add k8s apiserver to no proxy

### LINKED ISSUE #1139: Proxies when deploying kubecost
Hello,

We set [proxies](https://github.com/kubecost/cost-analyzer-helm-chart/blob/c98f4856bc96d50c5a8a775547ac8ce539e7037b/cost-analyzer/values.yaml#L179) on our kubecost deployment so it can access our central Prometheus running outside of our clusters. 

What we quickly realized is, if we rebuild the cluster and the internal IP of the API server changes, our deployment doesn't work. 

Could we set the no_proxy to include the APIServer IP by default in the code? We can do something similar [as the rest clients](https://github.com/kubernetes/kubernetes/blob/9c97ae286c4dbce137a35fdb3dd79dd2866e71f8/staging/src/k8s.io/client-go/rest/config.go#L517). Happy to submit a PR.

Or is there a better way of doing this?

rumstead:
## What does this PR change?

closes #1139 

Nope

It should make their lives easier when setting proxies on the containers as env variables 

- #1139 

## How was this PR tested?
Unit test and locally (still to come)

No but I can


kirbsauce:
thanks @rumstead !

dramich:
Closing as the issue is now closed.

PR #1147, opened 2022-04-07
Go Version 1.18 Upgrade 

mbolt35:
## What does this PR change?
* Updates the language to require go 1.18, which adds fuzz testing and generics.

* Should be no impact. 

## How was this PR tested?
* Manually, go runtime is expected to be 100% backwards compatible.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,9 +8,6 @@ require (
 	github.com/Azure/azure-sdk-for-go v61.6.0+incompatible
 	github.com/Azure/go-autorest/autorest v0.11.17
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.6
-	github.com/Azure/go-autorest/autorest/to v0.4.0 // indirect
-	github.com/Azure/go-autorest/autorest/validation v0.3.1 // indirect
-	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -53,4 +46,81 @@ require (
...
+	github.com/googleapis/gnostic v0.4.1 // indirect
+	github.com/gorilla/css v1.0.0 // indirect
+	github.com/hashicorp/golang-lru v0.5.1 // indirect
+	github.com/imdario/mergo v0.3.5 // indirect
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/jmespath/go-jmespath v0.4.0 // indirect
+	github.com/jstemmer/go-junit-report v0.9.1 // indirect
+	github.com/klauspost/compress v1.13.5 // indirect
+	github.com/klauspost/cpuid v1.3.1 // indirect
+	github.com/kubecost/events v0.0.1 // indirect
```
michaelmdresser, line 90: Are we already using the events library, or is this an artifact from development work?

michaelmdresser (approved):
:tada:

go.mod
mbolt35: I just added it with the assumption that it will be used, but clearly not required.

PR #1152, opened 2022-04-12
switch to using buffers in key() creation

AjayTripathy:
## What does this PR change?
* Tons of churn doing allocations and growslice, join operations on arrays in key(). Just use a buffer.
Screen Shot 2022-04-11 at 2 40 58 PM

One quick questions is: should we be precomputing the buffer size to avoid any buffer growth operations? How much efficiency would that gain? @mbolt35 

* No

* Hopefully a performance improvement

## How was this PR tested?
* This is a pretty simple function that is string manipulation. A unit test is probably the right call but it is TODO... needs scaleteesting as well.

* No

* Yes...this should be released ASAP. Perhaps even pre next release as part of a set of performance improvements.

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -1,6 +1,7 @@
 package kubecost
 
 import (
+	"bytes"
 	"encoding"
 	"fmt"
 	"strings"
@@ -246,7 +247,7 @@ func AssetToExternalAllocation(asset Asset, aggregateBy []string, labelConfig *L
 // Valid values of `aggregateBy` elements are strings which are an `AssetProperty`, and strings prefixed
 // with `"label:"`.
 func key(a Asset, aggregateBy []string) (string, error) {
-	keys := []string{}
+	var buffer bytes.Buffer
 
 	if aggregateBy == nil {
 		aggregateBy = []string{
@@ -262,7 +263,7 @@ func key(a Asset, aggregateBy []string) (string, error) {
 		}
 	}
 
-	for _, s := range aggregateBy {
+	for i, s := range aggregateBy {
 		key := ""
 		switch true {
 		case s == string(AssetProviderProp):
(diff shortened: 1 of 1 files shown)


mbolt35 (approved):
Looks great. I do think this could be a meaningful change for larger CURs. Good find!

mbolt35:
> One quick questions is: should we be precomputing the buffer size to avoid any buffer growth operations? How much efficiency would that gain? @mbolt35

I'm unsure of exactly how much this would gain. `string.Builder` does appear to grow internally using a doubling:

```go
// grow copies the buffer to a new, larger buffer so that there are at least n
// bytes of capacity beyond len(b.buf).
func (b *Builder) grow(n int) {
	buf := make([]byte, len(b.buf), 2*cap(b.buf)+n)
	copy(buf, b.buf)
	b.buf = buf
}
```

There is some really nuanced code in the builder though that seems to avoid escaping in a few circumstances though, so I still think this is a much better approach than using `strings.Join`

mbolt35:
@AjayTripathy Ok, so `strings.Join` is not as bad as I thought it was, but I still think you're code is better, given that we can just on the fly append. `Join` uses a `strings.Builder` internally, and counts/grows the buffer ahead of time:

```go
// Join concatenates the elements of its first argument to create a single string. The separator
// string sep is placed between elements in the resulting string.
func Join(elems []string, sep string) string {
	switch len(elems) {
	case 0:
		return ""
	case 1:
		return elems[0]
	}
	n := len(sep) * (len(elems) - 1)
	for i := 0; i < len(elems); i++ {
		n += len(elems[i])
	}

	var b Builder
	b.Grow(n)
	b.WriteString(elems[0])
	for _, s := range elems[1:] {
		b.WriteString(sep)
		b.WriteString(s)
	}
	return b.String()
}
```

nikovacevic:
I'd be happy to approve and merge this, but we should run a simple benchmark before/after so that we understand how this changes things.

mbolt35:
@nikovacevic Yeah, I think we just want to add some basic benchmarks. @michaelmdresser agrees

AjayTripathy:
Yeah, strings.Join() is not that bad. I don't expect this PR to help that since you need to buf.Grow() under the hood just like strings.Join()...unless we precompute buffer size. What I expect is to remove the totally unnecessary process of reading this data into a slice in the first place, the "runtime.growslice" calls in the leftmost section.

AjayTripathy:
I'm just not seeing "key" be a meaningful amount of processing time on our scale clusters. Something needs to be done (perhaps number of labels?) to make our scale test clusters more reflective of end-user clusters. I'm going to go ahead and merge but we should revisit why this didn't make it into our scale test clusters.
Screen Shot 2022-04-13 at 10 48 52 PM

PR #1157, opened 2022-04-14
CacheGroup[T] Utility

mbolt35:
## What does this PR change?
* Add `CacheGroup[T]` utility and IntervalRunner helper which will leverage key based caching with single flighting. Default cache policies are based on # of keys and expiration.

* We will roll this utility implementation out to several endpoints which should 

## How was this PR tested?
* Unit Tests

* Not currently

* Yes, part of our plan for performance improvements.

### DIFF
--- a/pkg/util/cache/cachegroup.go
+++ b/pkg/util/cache/cachegroup.go
@@ -27,7 +27,7 @@ type CacheGroup[T comparable] struct {
 	max              int
 }
 
-// NewCacheGroup[T] creates a new cache group instance the max number of keys to cache at any given time.
+// NewCacheGroup[T] creates a new cache group instance given the max number of keys to cache. 
 // If a new cache entry is added that exceeds the maximum, the oldest entry is evicted.
 func NewCacheGroup[T comparable](max int) *CacheGroup[T] {
 	return &CacheGroup[T]{
--- a/pkg/util/cache/cachegroup.go
+++ b/pkg/util/cache/cachegroup.go
@@ -59,7 +59,7 @@ func (cg *CacheGroup[T]) Do(key string, factory func() (T, error)) (T, error) {
 
 		// assign cache once a result for the group key is returned
 		cg.lock.Lock()
-		cg.removeOldest()
+		cg.removeOldestBeyondCapacity()
 		cg.cache[key] = &cacheEntry[T]{
 			item: i,
 			ts:   time.Now(),
@@ -115,7 +115,7 @@ func (cg *CacheGroup[T]) DisableExpiration() {
 
 // locates the oldest entry and removes it from the map. caller should lock
 // prior to calling
-func (cg *CacheGroup[T]) removeOldest() {
+func (cg *CacheGroup[T]) removeOldestBeyondCapacity() {
 	// only remove the oldest entries if we're at max capacity
 	if len(cg.cache) < cg.max {
 		return
--- a/pkg/util/cache/cachegroup.go
+++ b/pkg/util/cache/cachegroup.go
@@ -1,0 +1,157 @@
+package cache
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/kubecost/cost-model/pkg/util/interval"
+	"golang.org/x/sync/singleflight"
+)
+
+// cacheEntry contains a T item and the time it was added to the cache
+type cacheEntry[T comparable] struct {
+	item T
+	ts   time.Time
+}
+
+// CacheGroup provides single flighting for grouping repeated calls for the same workload, as well
+// as a cache that extends the lifetime of the returned result by a specific duration.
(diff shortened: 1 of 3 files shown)


dwbrown2:
Nice, where did you land on default retention? Any different options you're thinking about?

mbolt35:
> Nice, where did you land on default retention? Any different options you're thinking about?

I haven't landed on default retention yet -- this is, so far, just the utility - I am working on some specific handlers right now. I'm thinking 30-60s to start off with.

pkg/util/cache/cachegroup.go
```diff
@@ -0,0 +1,157 @@
...
+	lock             sync.Mutex
+	cache            map[string]*cacheEntry[T]
+	group            singleflight.Group
+	expirationLock   sync.Mutex
+	expirationRunner *interval.IntervalRunner
+	expiry           time.Duration
+	max              int
+}
+
+// NewCacheGroup[T] creates a new cache group instance the max number of keys to cache at any given time.
```
nikovacevic: This sentence reads funny to me, but it could be me. Is it supposed to say, "... a new cache group instance [given] the max number of keys..."?

pkg/util/cache/cachegroup.go
```diff
@@ -0,0 +1,157 @@
...
+		cache: make(map[string]*cacheEntry[T]),
+		max:   max,
+	}
+}
+
+// Do accepts a group key and a factory function to execute a workload request. Any executions
+// of Do() using an identical key will wait on the originating request rather than executing a
+// new request, and the final result will be shared among any callers sharing the same key.
+// Additionally, once returned, the workload for that key will remained cached. An expiration
+// policy can be added for this cache by calling the WithExpiration method.
```
nikovacevic, line 43: Love the `WithWhatever()` chaining pattern

pkg/util/cache/cachegroup.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+}
+
+// locates the oldest entry and removes it from the map. caller should lock
+// prior to calling
+func (cg *CacheGroup[T]) removeOldest() {
+	// only remove the oldest entries if we're at max capacity
+	if len(cg.cache) < cg.max {
+		return
+	}
```
nikovacevic: Would you be amenable to calling this `removeOldestBeyondCapacity` or something that indicates this condition? Saw the call earlier and expected it to always remove the oldest, not to do so conditionally within the remove function.

pkg/util/cache/cachegroup.go
```diff
@@ -0,0 +1,157 @@
...
+		cg.cache[key] = &cacheEntry[T]{
+			item: i,
+			ts:   time.Now(),
+		}
+		cg.lock.Unlock()
+		return i, nil
+	})
+
+	if err != nil {
+		return defaultValue[T](), err
```
nikovacevic, line 72:
```go
return defaultValue[T](), err
```
... wild! Interesting pattern.

pkg/util/cache/cachegroup.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+}
+
+// default value helper function to returns the initialized value for a T instance
+// (ie: for value types, typically the 0 value. For pointer types, nil)
+func defaultValue[T any]() T {
+	var t T
+	return t
+}
```
nikovacevic, line 157: So interesting that this works. But... of course it does. Haha.

pkg/util/cache/cachegroup.go
```diff
@@ -0,0 +1,157 @@
...
+
+// CacheGroup provides single flighting for grouping repeated calls for the same workload, as well
+// as a cache that extends the lifetime of the returned result by a specific duration.
+type CacheGroup[T comparable] struct {
+	lock             sync.Mutex
+	cache            map[string]*cacheEntry[T]
+	group            singleflight.Group
+	expirationLock   sync.Mutex
+	expirationRunner *interval.IntervalRunner
```
nikovacevic, line 25: This is a cool pattern.

nikovacevic (approved):
Only comments are about comments and semantics. LGTM!

pkg/util/cache/cachegroup.go
mbolt35: Eh, I don't know how the Go team would feel about this, but it's currently the only way I know of to write a generic class that accepts value and pointer types, that is, if you're using `nil` for a default pointer type.

pkg/util/cache/cachegroup.go
mbolt35: If you come across a better way of doing this, let me know.

pkg/util/cache/cachegroup.go
mbolt35: It reads weird probably cause I was talking a mile a minute while writing it. I'll try to reframe it :)

PR #1166, opened 2022-04-19
Add append method for AssetLabels

biancaburtoiu:
## What does this PR change?
This PR adds an `append(labels map[string]string)` method to be used on objects of type `AssetLabels`. 

Yes, this PR must be merged before https://github.com/kubecost/kubecost-cost-model/pull/718 

Not directly

- addresses https://github.com/kubecost/cost-analyzer-helm-chart/issues/946 

## How was this PR tested?
N/A

No

Yes

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -360,15 +360,17 @@ func (al AssetLabels) Merge(that AssetLabels) AssetLabels {
 
 // Append joins AssetLabels with a given map of labels
 func (al AssetLabels) Append(newLabels map[string]string, overwrite bool) {
-	if len(newLabels) > 0 {
-		for label, value := range newLabels {
-			if _, ok := al[label]; ok {
-				if overwrite {
-					al[label] = value
-				}
-			} else {
+	if len(newLabels) == 0 {
+		return
+	}
+
+	for label, value := range newLabels {
+		if _, ok := al[label]; ok {
+			if overwrite {
 				al[label] = value
 			}
+		} else {
+			al[label] = value
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/kubecost/asset.go
```diff
@@ -358,6 +358,21 @@ func (al AssetLabels) Merge(that AssetLabels) AssetLabels {
 	return result
 }
 
+// Append joins AssetLabels with a given map of labels
+func (al AssetLabels) Append(newLabels map[string]string, overwrite bool) {
+	if len(newLabels) > 0 {
```
nikovacevic: One small nit here w/r/t Go style. Go style prefers early returns so that the critical path of code remains least indented if possible. In this case, that would be:
```go
if len(newLabels) == 0 {
    return
}

...
```

And, in fact, I think this condition is not precisely right. Would you agree that we want to append labels to an empty map? We only want to early return if the map is `nil` so that we don't panic. Let me know if you disagree!

nikovacevic (changes requested):

pkg/kubecost/asset.go
biancaburtoiu: Re: the defensive call -- I totally agree, I missed that. Fixing it. 

Recall that `newLabels` in this case represents the account tag labels we are planning to add; if that list is somehow empty, we'll want to exit without further action. 

I actually handled the case of the AssetLabels object being nil in the caller of this method, in the [other PR](https://github.com/kubecost/kubecost-cost-model/pull/718/files): 
image

WDYT?

nikovacevic: Oops! I misread and thought that was the existing labels. Thank you for correcting me :sweat_smile: 

LGTM!

nikovacevic (approved):

PR #1169, opened 2022-04-19
Use incoming args to set verbosity

dramich:
## What does this PR change?
* Respect the arguments being passed in for klog by parsing them after setting a default log level. Previously the args were not being parsed so any of the flags that were previously added would be ignored. 

* Yes, followup to: https://github.com/kubecost/cost-model/pull/1160

### DIFF
--- a/pkg/cmd/commands.go
+++ b/pkg/cmd/commands.go
@@ -44,10 +44,17 @@ func Execute(costModelCmd *cobra.Command) error {
 	// initialize klog and make cobra aware of all the go flags
 	klog.InitFlags(nil)
 
+	// Register all the klog flags with pflag
 	flag.CommandLine.VisitAll(func(f *flag.Flag) {
 		pflag.CommandLine.AddGoFlag(f)
 	})
+
+	// Set our default level so we don't use 0
 	pflag.CommandLine.Set("v", "3")
+	// Parse incoming args for pflag
+	pflag.CommandLine.Parse(os.Args[1:])
+
+	klog.Infof("Logging with verbosity: %v", pflag.CommandLine.Lookup("v").Value.String())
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/commands.go
```diff
@@ -44,10 +44,17 @@ func Execute(costModelCmd *cobra.Command) error {
...
 
+	// Register all the klog flags with pflag
 	flag.CommandLine.VisitAll(func(f *flag.Flag) {
 		pflag.CommandLine.AddGoFlag(f)
 	})
+
+	// Set our default level so we don't use 0
 	pflag.CommandLine.Set("v", "3")
+	// Parse incoming args for pflag
+	pflag.CommandLine.Parse(os.Args[1:])
```
mbolt35, line 55: I am a bit concerned about this line, as I'm pretty sure that the cobra library calls this. The cobra library is what leverages pflags (which I think obey posix, while go flags don't? Not 100% sure what the story is). However, the reason we have to do this strange "conversion" is because cobra tends to use pflags and any libraries that may use flags (klog in this case) aren't getting anything unless you explicitly pipe from pflags (your first change). 

Our application only really cares about two flags currently:
`-cost-model`
`-agent`

To support backwards compatibility, I also added a bit of hackery to ensure that cobra would default to the `-cost-model` command if nothing was provided. This code happens in the code below this line. 

While this `Parse()` call may be completely benign, I just want to make sure we're still able to handle the expected arguments. Sadly, `klog` here is becoming a real nuisance because of these CL requirements, where really we want to organize the application in a `-<mode> <mode args...>` way where the arguments after the command directive are piped to the specific mode command, but the requirement to support the other flags sort of pollutes the top level argument set. 

¯\_(ツ)_/¯

dramich:
Closing since klog is gone

PR #1172, opened 2022-04-21
Filters v2: (Allocation) in-Go representation + basic Matches() implementation

michaelmdresser:
## What does this PR change?
Introduces a structured representation for Allocation filters in the form of the `AllocationFilter` interface and 3 structs that implement that interface: `AllocationFilterCondition`, `AllocationFilterOr`, and `AllocationFilterAnd`. It supports representing all of our current `AllocationProperties` filters, OR, AND, and Not Equals. The code has detailed comments, check it out!

This PR _does not_ introduce anything user-facing or include any changes to existing filtering logic. It is the foundation for work that will introduce those changes.

### Why?
By moving from arbitrary `FilterFunc`s (which have served us well) to a structured representation, we can do many useful things:
- Provide a more featureful query frontend for our API users
- Introspect filters (answer: "what is this trying to filter?") for query-time optimization, like pre-agg ETL
- Translate filters to a backing store-specific language (e.g. SQL) for store-level filtering+optimization

### Important notes
The representation's goal is to support arbitrarily complex boolean expressions for maximum flexibility. This lets us experiment with query "frontends" (string languages, e.g. `namespace:"kubecost" AND cluster!:"testing"`) without worrying that the representation won't be able to support a given frontend.

This representation will be the unifying point for all filter work. It will be challenging to change once it has been merged and implementations are built on top of it, so please review carefully :pray: 

Once the design has settled, we can open a mirror(ish) PR for Asset filters. @mbolt35 do you see any way to use generics to unify the Asset and Allocation filter work?

### Backing design doc

You can find the full Filters v2 design doc [here (internal only)](https://docs.google.com/document/d/1HKkp2bv3mnvfQoBZlpHjfZwQ0FzDLOHKpnwV9gQ_KgU/edit#).

N/A

N/A (requires more work! see design doc!)

N/A

## How was this PR tested?
Unit tests in this PR

N/A. This PR is purely for devs and documentation is in docstrings.

No. This PR provides no functionality to the product. Future PRs that depend on this work may be marked as next release.

### DIFF
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -4,22 +4,32 @@ package kubecost
 // that can be filtered on (namespace, label, etc.)
 type FilterField int
 
+// If you add a FilterField, MAKE SURE TO UPDATE ALL FILTER IMPLEMENTATIONS! Go
+// does not enforce exhaustive pattern matching on "enum" types.
 const (
 	FilterClusterID FilterField = iota
+	FilterNode
 	FilterNamespace
-	// ControllerKind
-	// ControllerName
-	// Pod
-	// Container
+	FilterControllerKind
+	FilterControllerName
+	FilterPod
+	FilterContainer
+
+	// Filtering based on label aliases (team, department, etc.) should be a
+	// responsibility of the query handler. By the time it reaches this
+	// structured representation, we shouldn't have to be aware of what is
+	// aliased to what.
 
 	FilterLabel
-	// Annotation
+	FilterAnnotation
 )
 
 // FilterOp is an enum that represents operations that can be performed
 // when filtering (equality, inequality, etc.)
 type FilterOp int
 
+// If you add a FilterOp, MAKE SURE TO UPDATE ALL FILTER IMPLEMENTATIONS! Go
+// does not enforce exhaustive pattern matching on "enum" types.
 const (
 	FilterEquals FilterOp = iota
 
@@ -91,13 +101,32 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
 	switch filter.Field {
 	case FilterClusterID:
 		valueToCompare = a.Properties.Cluster
+	case FilterNode:
+		valueToCompare = a.Properties.Node
 	case FilterNamespace:
 		valueToCompare = a.Properties.Namespace
+	case FilterControllerKind:
+		valueToCompare = a.Properties.ControllerKind
+	case FilterControllerName:
+		valueToCompare = a.Properties.Controller
+	case FilterPod:
+		valueToCompare = a.Properties.Pod
+	case FilterContainer:
+		valueToCompare = a.Properties.Container
 	// Comes from GetAnnotation/LabelFilterFunc in KCM
 	case FilterLabel:
 		val, ok := a.Properties.Labels[filter.Key]
 
 		// TODO: What about label != ?
+		if !ok {
+			return false
+		}
+
+		valueToCompare = val
+	case FilterAnnotation:
+		val, ok := a.Properties.Annotations[filter.Key]
+
+		// TODO: What about annotation != ?
 		if !ok {
 			return false
 		}
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -96,6 +96,12 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
 	// The Allocation's value for the field to compare
 	var valueToCompare string
 
+	// toCompareMissing will be true if the value to be compared is missing in
+	// the Allocation. For example, if we're filtering based on the value of
+	// the "app" label, but the Allocation doesn't have an "app" label, this
+	// will become true. This lets us deal with != gracefully.
+	toCompareMissing := false
+
 	// This switch maps the filter.Field to the field to be compared in
 	// a.Properties and sets valueToCompare from the value in a.Properties.
 	switch filter.Field {
@@ -117,21 +123,19 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
 	case FilterLabel:
 		val, ok := a.Properties.Labels[filter.Key]
 
-		// TODO: What about label != ?
 		if !ok {
-			return false
+			toCompareMissing = true
+		} else {
+			valueToCompare = val
 		}
-
-		valueToCompare = val
 	case FilterAnnotation:
 		val, ok := a.Properties.Annotations[filter.Key]
 
-		// TODO: What about annotation != ?
 		if !ok {
-			return false
+			toCompareMissing = true
+		} else {
+			valueToCompare = val
 		}
-
-		valueToCompare = val
 	default:
 		// TODO: log an error here? this should never happen
 		return false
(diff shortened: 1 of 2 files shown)


pkg/kubecost/allocationfilter.go
```diff
@@ -0,0 +1,210 @@
+package kubecost
+
+import "github.com/kubecost/cost-model/pkg/log"
+
+// FilterField is an enum that represents Allocation-specific fields that can be
+// filtered on (namespace, label, etc.)
+type FilterField int
```
michaelmdresser: We could make this `type FilterField string` instead, which would give us a much better serialization/deserialization story for pretty printing etc. Thoughts?

The same goes for `FilterOp`.

michaelmdresser: At the cost of slight overhead in terms of string allocations, but I don't think we're going to have a meaningful number of filters allocated at any given time.

pkg/kubecost/allocationfilter.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
+	// TODO: For these nil cases, what about != filters?
+	if a == nil {
+		return false
+	}
+	if a.Properties == nil {
+		return false
+	}
```
michaelmdresser: I'd particularly like thoughts on this `TODO`

dwbrown2:
I wasn't able to hear all of the discussion in sprint planning today. Are we merging this as part of 1.93?

pkg/kubecost/allocationfilter.go
nikovacevic: Could go either way, but that might be nice!

michaelmdresser:
> I wasn't able to hear all of the discussion in sprint planning today. Are we merging this as part of 1.93?

This PR is a core dependency of _all_ upcoming filter work, but has no actual affect on Kubecost the product. So it doesn't have to be merged this release unless we're trying to get something from the P1/P2 categories in this release. This comes down to product priorities and dev team capacity; I'm not sure what the story is here. I'm committed to keeping up with code review on this PR but I can't commit to doing anything more pre-1.93.0-rc.0.

Adam-Stack-PM:
> > I wasn't able to hear all of the discussion in sprint planning today. Are we merging this as part of 1.93?
> 
> This PR is a core dependency of _all_ upcoming filter work, but has no actual affect on Kubecost the product. So it doesn't have to be merged this release unless we're trying to get something from the P1/P2 categories in this release. This comes down to product priorities and dev team capacity; I'm not sure what the story is here. I'm committed to keeping up with code review on this PR but I can't commit to doing anything more pre-1.93.0-rc.0.

Any downside to including it in v1.93?  Niko mentioned we may be able to get feedback from a small set of users.

michaelmdresser:
> Any downside to including it in v1.93? Niko mentioned we may be able to get feedback from a small set of users.

There's no downside -- this PR is effectively dead code until a query frontend is written.

I know its a little odd. This PR falls completely outside the scope of release/feature planning. Future PRs that depend on this one (like a query frontend) do fall in scope of release planning because those PRs will change the product. You can think of this PR like drafting a blueprint or creating a mold.

We can't get feedback until the functionality is exposed to users which requires a query frontend.

pkg/kubecost/allocationfilter.go
michaelmdresser: Updated Field and Op to be strings.

michaelmdresser:
Question for reviewers: Do you think Allocation filters (and therefore Asset filters) should get their own package instead of living in `pkg/kubecost`? Namespacing filter fields may become important -- we probably don't want someone to be filtering assets by Pod, right?

pkg/kubecost/allocationfilter.go
```diff
@@ -0,0 +1,210 @@
...
+// OR.
+type AllocationFilterOr struct {
+	Filters []AllocationFilter
+}
+
+// AllocationFilterOr is a set of filters that should be evaluated as a logical
+// AND.
+type AllocationFilterAnd struct {
+	Filters []AllocationFilter
+}
```
nikovacevic, line 86: Are these intended to accept more than 2 filters, or precisely 2?

pkg/kubecost/allocationfilter.go
```diff
@@ -0,0 +1,210 @@
...
+// - Querying backing data stores efficiently (e.g. translation to SQL)
+//
+// Custom implementations of this interface outside of this package should not
+// expect to receive these benefits. Passing a custom implementation to a
+// handler may in errors.
+type AllocationFilter interface {
+	// Matches is the canonical in-Go function for determing if an Allocation
+	// matches a filter.
+	Matches(a *Allocation) bool
+}
```
nikovacevic, line 56: I like the flexibility here. We'll just want to be careful with implementation -- i.e. if the `Query` functions take this as a type, they still won't be introspectable (is that a word?) without type casting.

pkg/kubecost/allocationfilter.go
nikovacevic: And did you consider combining these into one structure, with an `Op` for AND and OR? Not saying that's better, just thinking about alternatives.

pkg/kubecost/allocationfilter.go
nikovacevic: Do you have a sense for what type, precisely, we'd want to use for `QueryOptions`?

pkg/kubecost/allocationfilter.go
```diff
@@ -0,0 +1,210 @@
...
+		return false
+	}
+
+	return false
+}
+
+func (and AllocationFilterAnd) Matches(a *Allocation) bool {
+	filters := and.Filters
+	if len(filters) == 0 {
+		// TODO: Should an empty set of ANDs be true or false?
```
nikovacevic: I think `true` makes sense. As in "there are no conditions to meet" which is a condition that all Allocations meet.

nikovacevic (commented):
Will be happy to approve soon, but want to touch on a few of these topics first. Code looks good.

nikovacevic:
@dramich any input for an internal filter representation that has some structure (i.e. isn't just `func(thing) bool`)? (Asking because we've briefly touched on this project, and curious to hear if you have strong opinions.)

pkg/kubecost/allocationfilter.go
michaelmdresser: Intended to accept more than two so that users of the API can e.g. chain ANDs without making a really deep tree for no reason.

I did think about making an `Op` for AND and OR. I decided against it because I'm not trying to make a general language, so I don't need the concept of an arbitrary expression. I think explicitly having AND and OR at the top level of the representation will make using the API easier and more clear for both filter builders and filter consumers. I'm definitely open to discussing further -- there could be some benefit to representing AND/OR as an `Op` that I haven't recognized.

pkg/kubecost/allocationfilter.go
michaelmdresser: The need to type cast is unfortunate. I think its inevitable that `QueryOptions` will have a `Filter kubecost.AllocationFilter` field, though -- filters can be nested within ANDs and ORs and I can't think of a good way to represent that general idea without an interface.

The fallback plan is fairly straightforward. If whatever takes `QueryOptions` isn't able to introspect (due to an unknown implementation of `AllocationFilter` OR a bug in introspection) it can just fall through to calling `Matches()` at the end of the day.

I miss ML-style match statements :(

nikovacevic: > filters can be nested within ANDs and ORs and I can't think of a good way to represent that general idea without an interface

This was actually part of my reasoning about making it a single "composite" filter with an `Op` -- so that you could express multiple kinds of composite filters without an interface. But trust your design decisions! I can see it either way.

nikovacevic (approved):
Looks good, Michael. And remember -- we can always tweak it as we start using it, so it doesn't have to be perfect day one!

PR #1175, opened 2022-04-25
Add athena workgroup to athena config

### LINKED ISSUE #897: Add custom athena workgroup support
AWA Athena supports multiple workgroup configurations (with different parameters, specifically `Query result location`)
, but cost-model is uses `primary` one (which is created by default) and it's not possible to specify custom workgroup to be used by cost-model.

That would be great if it's possible to specify athena workgroup name as a parameter for kube-cost.

The alternative is to stick to a default `primary` group, but we are using it already for another purpose, so we can't change `Query result location` and it will be used for kubecost as well. But this will cause that query results from different services generated to the same bucket.

Users can specify custom athena workgroup in `values.yaml`.

dramich:
This allows setting the workgroup if required but continues to use the
default of primary if nothing is set.

* https://github.com/kubecost/cost-analyzer-helm-chart/pull/1386
* https://github.com/kubecost/kubecost-cost-model/pull/734

* This allows setting the workgroup if required but continues to use the
default of primary if nothing is set.

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/897

## How was this PR tested?
* Setting the workgroup, and setting an invalid workgroup which causes the error:

```
I0423 18:57:52.534675       1 awsprovider.go:864] Failed to lookup savings plan data: Error fetching Savings Plan Data: QueryAthenaPaginated: start query error: operation error Athena: StartQueryExecution, https response error StatusCode: 400, RequestID: xxx, InvalidRequestException: WorkGroup is not found.
``` 

* Change was added to the chart

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -56,7 +56,7 @@ func (aws *AWS) PricingSourceStatus() map[string]*PricingSource {
 	sources := make(map[string]*PricingSource)
 
 	sps := &PricingSource{
-		Name: SpotPricingSource,
+		Name:    SpotPricingSource,
 		Enabled: true,
 	}
 
@@ -80,7 +80,7 @@ func (aws *AWS) PricingSourceStatus() map[string]*PricingSource {
 	sources[SpotPricingSource] = sps
 
 	rps := &PricingSource{
-		Name: ReservedInstancePricingSource,
+		Name:    ReservedInstancePricingSource,
 		Enabled: true,
 	}
 	rps.Error = ""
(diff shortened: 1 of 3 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -1615,6 +1619,11 @@ func (aws *AWS) QueryAthenaPaginated(ctx context.Context, query string, fn func(
 		ResultConfiguration:   resultConfiguration,
 	}
 
+	// Only set if there is a value, the default intput is nil which defaults to the 'primary' workgroup
```
nikovacevic: intput => input

nikovacevic (approved):

pkg/cloud/awsprovider.go
dramich: Thanks, updated

PR #1178, opened 2022-04-26
Drop klog and migrate to logging package

dramich:
- Go mod update
- Update logging package and flag parsing
- Swap klog to log package

## What does this PR change?
Removed klog from all of our logging - it is still used in packages we depend on (k8s-client)
Moved all klog calls to our logging package which now uses zerolog under the hood
Defaulted to using a pretty style printing to keep the output similar to existing logging. This can be changed to json with a flag/envvar which should technically be faster. We don't fully take advantage of the structured part yet but it would still be better than nothing if a user wants to use it. 

New logging format:
```
2022-04-27T10:00:11.28658-07:00 ??? Log level set to info
2022-04-27T10:00:11.286854-07:00 INF Starting cost-model (git commit "1.91.0-rc.0")
2022-04-27T10:00:11.286939-07:00 FTL No address for prometheus set in $PROMETHEUS_SERVER_ENDPOINT. Aborting.
```

Old logging format:
```
I0427 09:35:51.473189   31474 router.go:1351] Starting cost-model (git commit "1.91.0-rc.0")
F0427 09:35:51.473367   31474 router.go:1370] No address for prometheus set in $PROMETHEUS_SERVER_ENDPOINT. Aborting.
```

Note that the file and line number has been dropped. If this is deemed required it can be added back or have a flag to disable/enable it

Not yet

The logging output will look different but similar. See above about JSON stuff. 

https://github.com/kubecost/cost-model/issues/1170

## How was this PR tested?
Swapped out logging and run kubecost

* 

*


pkg/services/clusterservice.go
```diff
@@ -22,13 +22,13 @@ func newClusterManager() *clusters.ClusterManager {
...
 		if err != nil {
-			klog.V(1).Infof("[Error] Failed to create costmodel.db: %s", err.Error())
+			log.Errorf("[Error] Failed to create costmodel.db: %s", err.Error())
 			return cm.NewConfiguredClusterManager(cm.NewMapDBClusterStorage(), clustersConfigFile)
 		}
 
 		store, err := clusters.NewBoltDBClusterStorage("clusters", db)
 		if err != nil {
-			klog.V(1).Infof("[Error] Failed to Create Cluster Storage: %s", err.Error())
+			kog.Errorf("[Error] Failed to Create Cluster Storage: %s", err.Error())
```
nikovacevic: kog => log

nikovacevic (commented):
Looks great. I'll want to do another pass, but as a quick skim, just caught one apparent typo.

pkg/services/clusterservice.go
dramich: Ah, yeah, it's inside a comment so building didn't catch it.

pkg/log/log.go
```diff
@@ -59,11 +84,19 @@ func DedupedInfof(logTypeLimit int, format string, a ...interface{}) {
 }
 
 func Profilef(format string, a ...interface{}) {
-	klog.V(3).Infof(fmt.Sprintf("[Profiler] %s", format), a...)
+	log.Info().Msgf(fmt.Sprintf("[Profiler] %s", format), a...)
+}
```
michaelmdresser, line 89: `Profilef` seems like a good candidate for a somewhat structured log, maybe:

```go
// "type" is such an overloaded word, I'm not sure
// what the best practice is for structured log keys
log.Info().Str("type", "profile").Msgf(...) 
```

What do you think?

I totally understand if you want to leave this kind of update for another time.

michaelmdresser (approved):
Looks great, nice touches adjusting a couple of the overly-verbose log levels!

pkg/log/log.go
dramich: Yeah, how and what to add is a good question. We could potentially turn the profile logger into a sub logger i.e. https://github.com/rs/zerolog#sub-loggers-let-you-chain-loggers-with-additional-context 
I don't have enough context yet to say where to go with the profile logger, perhaps it should end up behind a flag so it's not always printing logs since profiling is usually resource intensive and not always needed.

mbolt35:
This is beautiful - Viper looks helpful, maybe we can leverage it more in some other areas where we have config chaos.

PR #1179, opened 2022-04-27
Attribute batch/v1 CronJob controller correctly

michaelmdresser:
## What does this PR change?

As explained in the new comments, batch/v1beta1 CronJobs create Jobs with
a 10 character timestamp at the end. Our regex (and therefore Pod ->
Job -> CronJob controller attribution) works for this case. However,
batch/v1 CronJobs apparently create Jobs with 8 character timestamps.
This breaks our controller attribution, resulting in every Job being its
own controller.

This commit updates our regex (and adds a unit test) to support the
batch/v1 CronJob Job name format.

### Meta: this is just a stopgap

While this change fixes the immediate problem of v1 CronJob controller attribution, it doesn't address the more general case: Pods with multiple levels of ownership. Below is metadata from a live CronJob (all data from `kubectl get ... -o yaml`) which demonstrates this situation. Note how the Pod has an `ownerReference` to the Job and the Job has an `ownerReference` to the CronJob.

The costmodel's [`CostData` struct](https://github.com/kubecost/cost-model/blob/1ac8ea3a6230d71f1f8fc392ed80646f507751bd/pkg/costmodel/costmodel.go#L75-L100) only expects one level of ownership and hardcodes the type of controllers that can own a Pod: Deployments, Services, DaemonSets, StatefulSets, and Jobs (this is how we refer to CronJobs, with hacked support via the regex edited in this PR). Our limited representation prevents us from correctly attributing controllers for 3rd party deployment solutions like Argo Rollouts (https://github.com/kubecost/cost-model/issues/879) which use CRDs and custom controllers to manage K8s workloads.

The runtime complexity of discovering the true owner/controller of a Pod may be too high, but I want to mention it because this is a real problem.

#### Multi-level ownership example: CronJob
```yaml
apiVersion: batch/v1
kind: CronJob
metadata:
  creationTimestamp: "2022-04-27T13:27:23Z"
  name: cj-v1
  namespace: cj-test
  resourceVersion: "145572802"
  uid: b829078e-49c8-42ad-92a7-568cef370a41
```

```yaml          
apiVersion: batch/v1
kind: Job
metadata:
  creationTimestamp: "2022-04-27T14:25:00Z"
  labels:
    controller-uid: d1c02d48-1426-4917-8680-21f3eabb8e77
    job-name: cj-v1-27517825
  name: cj-v1-27517825
  namespace: cj-test
  ownerReferences:
  - apiVersion: batch/v1
    blockOwnerDeletion: true
    controller: true
    kind: CronJob
    name: cj-v1
    uid: b829078e-49c8-42ad-92a7-568cef370a41
  resourceVersion: "145572801"
  uid: d1c02d48-1426-4917-8680-21f3eabb8e77
```

```yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2022-04-27T14:25:00Z"
  generateName: cj-v1-27517825-
  labels:
    controller-uid: d1c02d48-1426-4917-8680-21f3eabb8e77
    job-name: cj-v1-27517825
  name: cj-v1-27517825-2p7dc
  namespace: cj-test
  ownerReferences:
  - apiVersion: batch/v1
    blockOwnerDeletion: true
    controller: true
    kind: Job
    name: cj-v1-27517825
    uid: d1c02d48-1426-4917-8680-21f3eabb8e77
  resourceVersion: "145572798"
  uid: 40f44d32-2b0d-4e42-8e8a-a7237140955b
```

N/A

batch/v1 CronJob data will have the correct controller. Before this fix, the Jobs created by a batch/v1 CronJob would each belong to their own controller with the name of the Job (according to Kubecost) instead of belonging to the parent CronJob.

This will help some users affected by https://github.com/kubecost/kubecost-cost-model/issues/651. At least one report (https://kubecost.zendesk.com/agent/tickets/1712) of request sizing over-estimation is caused by this CronJob controller mis-attribution and should be resolved by this ticket, though an ETL rebuilt _will_ be required. There is a more fundamental problem with KCM#651 that I will address separately.

### Unit tests

I pulled test Job names from live clusters, one with a batch/v1 CronJob
and another with a batch/v1beta1 CronJob. GKE wouldn't let me create a
new batch/v1beta1 CronJob -- it seems to auto-upgrade. Fortunately, there
already was a batch/v1beta1 CronJob in our integration test cluster.

### Live tests

Deployed this code to a cluster with batch/v1 CronJobs. Observed that ETL data built after this change had the correct controller name (the name of the cronjob) instead of the individual job (name of the cronjob + timestamp) which it had beforehand.

After a full rebuild, this PR also corrected request sizing's savings over-estimation for those CronJobs. Again, this doesn't fully resolve the request sizing problem, just for over-estimation of batch/v1 CronJobs on freshly-built ETL data.

N/A

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -44,7 +44,11 @@ const (
 )
 
 // isCron matches a CronJob name and captures the non-timestamp name
-var isCron = regexp.MustCompile(`^(.+)-\d{10}$`)
+//
+// We support either a 10 character timestamp OR an 8 character timestamp
+// because batch/v1beta1 CronJobs creates Jobs with 10 character timestamps
+// and batch/v1 CronJobs create Jobs with 8 character timestamps.
+var isCron = regexp.MustCompile(`^(.+)-(\d{10}|\d{8})$`)
 
 type CostModel struct {
 	Cache                      clustercache.ClusterCache
--- a/pkg/costmodel/costmodel_test.go
+++ b/pkg/costmodel/costmodel_test.go
@@ -1,0 +1,67 @@
+package costmodel
+
+import (
+	"testing"
+)
+
+func Test_CostData_GetController_CronJob(t *testing.T) {
+	cases := []struct {
+		name string
+		cd   CostData
+
+		expectedName          string
+		expectedKind          string
+		expectedHasController bool
+	}{
+		{
+			name: "batch/v1beta1 CronJob Job name",
+			cd: CostData{
+				// batch/v1beta1 CronJobs create Jobs with a 10 character
+				// timestamp appended to the end of the name.
+				//
+				// It looks like this:
+				// CronJob: cronjob-1
+				// Job: cronjob-1-1651057200
+				// Pod: cronjob-1-1651057200-mf5c9
+				Jobs: []string{"cronjob-1-1651057200"},
+			},
+
+			expectedName:          "cronjob-1",
+			expectedKind:          "job",
+			expectedHasController: true,
+		},
+		{
+			name: "batch/v1 CronJob Job name",
+			cd: CostData{
+				// batch/v1CronJobs create Jobs with an 8 character timestamp
+				// appended to the end of the name.
+				//
+				// It looks like this:
+				// CronJob: cj-v1
+				// Job: cj-v1-27517770
+				// Pod: cj-v1-27517770-xkrgn
+				Jobs: []string{"cj-v1-27517770"},
+			},
+
+			expectedName:          "cj-v1",
+			expectedKind:          "job",
+			expectedHasController: true,
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			name, kind, hasController := c.cd.GetController()
+
+			if name != c.expectedName {
+				t.Errorf("Name mismatch. Expected: %s. Got: %s", c.expectedName, name)
+			}
+			if kind != c.expectedKind {
+				t.Errorf("Kind mismatch. Expected: %s. Got: %s", c.expectedKind, kind)
+			}
+			if hasController != c.expectedHasController {
+				t.Errorf("HasController mismatch. Expected: %t. Got: %t", c.expectedHasController, hasController)
+			}
+		})
+	}
+}


mbolt35:
> The costmodel's [CostData struct](https://github.com/kubecost/cost-model/blob/1ac8ea3a6230d71f1f8fc392ed80646f507751bd/pkg/costmodel/costmodel.go#L75-L100) only expects one level of ownership and hardcodes the type of controllers that can own a Pod: Deployments, Services, DaemonSets, StatefulSets, and Jobs (this is how we refer to CronJobs, with hacked support via the regex edited in this PR). Our limited representation prevents us from correctly attributing controllers for 3rd party deployment solutions like Argo Rollouts (https://github.com/kubecost/cost-model/issues/879) which use CRDs and custom controllers to manage K8s workloads.

> The runtime complexity of discovering the true owner/controller of a Pod may be too high, but I want to mention it because this is a real problem.

I'm curious about the metrics we/KSM emits for jobs -- if we don't have access to the kubernetes cluster data, can we traverse more than one depth? Will we ever need to? Or does ETL handle this for us?

mbolt35 (approved):
LGTM - Great find. Added a question to the PR, if you know it off the top of your head. Otherwise, no worries.

nikovacevic (approved):

nikovacevic:
Merged for you, @michaelmdresser, in the interest of code freeze. Hope that's alright!

nikovacevic:
@mbolt35 @michaelmdresser the deeper problem here (multi-layer ownership) is an interesting problem for a few reasons. Primarily, it's not easy to represent those relationships using Prometheus metrics. (I'm guessing it's possible, but would have to think about it more.) So it goes beyond the representation we use. (That `CostData` struct isn't even relevant to ETL.)

Today, we use labels to line up allocations with basic controllers (deployments, etc.) but that obviously can't work for CRDs. So we'd have to start using owner reference IDs, which might not exist in any out-of-the-box metrics.

dwbrown2:
I personally don't see how we can reasonably support CRDs until there is some sort of common standard for declaring ownership/controller.

michaelmdresser:
Thanks for merging @nikovacevic!

---

> Today, we use labels to line up allocations with basic controllers (deployments, etc.) but that obviously can't work for CRDs. So we'd have to start using owner reference IDs, which might not exist in any out-of-the-box metrics.

It seems we do have a metric for `ownerReferences` called `kube_pod_owner`, but it looks like it only supports one layer of ownership because its just for a Pod's `ownerReferences`:

https://github.com/kubecost/cost-model/blob/cf97a87e1314ba5881fc7c1752ee5603df617aab/pkg/metrics/podlabelmetrics.go#L78-L83

---

> I personally don't see how we can reasonably support CRDs until there is some sort of common standard for declaring ownership/controller.

I'm fairly certain that `ownerReferences` is this common standard! See the K8s API specification for `ObjectMeta` (https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#ObjectMeta) -- pretty much all K8s objects have `ObjectMeta` attached to them, and `ObjectMeta` includes an `ownerReferences` field. Fortunately, owner information includes more than just UID.

If we eventually choose to pursue "correct" controller attribution through multiple layers, I think there are a few questions to answer:
1. What is the correct way to represent a fully-qualified controller (+ owner ref) in a metric? Perhaps `kubecost_controller_owner{namespace=?, controllerType=?, controllerKind=?, ownerNamespace=?, ownerType=?, ownerName=?}`.
2. What is the right way to extract the owner reference (if it exists) for a Custom Resource? We can't assume that we will have the CRD code, so we may have to pull a JSON representation of all custom resources and walk/parse the JSON to extract ownership information `ObjectMeta.ownerReferences`.
3. What is the definition of the "owner" or "main controller" of a given Pod? Is it _always_ the first controller/owner in the chain (i.e. the only parent-less/controller-less controller)?
4. What is the performance cost of supporting arbitrarily-long ownership paths? What is the correct data structure to represent these paths?

nikovacevic:
FWIW this feels like a "Kubecost 2.0" discussion. We'll have more leeway for making complex decisions regarding multiple layers of ownership if Prometheus is not mediating the data, and we can use the Kubernetes API directly. (Not saying it's impossible -- @michaelmdresser is pursuing the ideas in the right direction; just feels like it'd be a struggle.)

michaelmdresser:
Totally agree! I almost added "and should we even represent this data in a metric to begin with?"

PR #1185, opened 2022-04-28
Allow alloc/asset set clone of errors/warnings

kaelanspatel:
## What does this PR change?
Adds cloning of errors/warnings to AllocationSets and AssetSets.

* 

It may have slight performance ramifications, but I have discussed this w/ others and think it will be minimal.

## How was this PR tested?
Tested as part of the functionality of the linked issue.

No

Yes

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -1668,11 +1668,30 @@ func (as *AllocationSet) Clone() *AllocationSet {
 		idleKeys[k] = v
 	}
 
+	var errors []string
+	var warnings []string
+
+	if as.Errors != nil {
+		errors = make([]string, len(as.Errors))
+		copy(errors, as.Errors)
+	} else {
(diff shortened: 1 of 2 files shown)


mbolt35 (approved):
LGTM

pkg/kubecost/allocation.go
```diff
@@ -1668,11 +1668,30 @@ func (as *AllocationSet) Clone() *AllocationSet {
...
 
+	var errors []string
+	var warnings []string
+
+	if as.Errors != nil {
+		errors = make([]string, len(as.Errors))
+		copy(errors, as.Errors)
+	} else {
+		errors = nil
+	}
```
michaelmdresser, line 1679: Just for the future, `nil` is the default value of a slice like `[]string`, so when you declare:
```go
var errors []string
```

`errors` will be `nil` until you change it!

PR #1188, opened 2022-05-02
use custom cloud provider + no watchers for etl readonly mode

AjayTripathy:
## What does this PR change?
* Allows users to deploy kubecost to clusters that do not access the k8s api. The purpose is to still monitor federated clusters that _do_ connect to the k8s API.

* No

* Support running kubecost in 

## How was this PR tested?
* Manually, ran against a federated thanos and noted that the UI/ETL still functions. 

* The flag is documented. I would prefer to discourage running in this mode, so not calling it out further.

* I don't want to rush the team into getting this into the next release, but if it makes it great :)

### DIFF
--- a/pkg/cloud/provider.go
+++ b/pkg/cloud/provider.go
@@ -423,7 +423,11 @@ func ShareTenancyCosts(p Provider) bool {
 func NewProvider(cache clustercache.ClusterCache, apiKey string, config *config.ConfigFileManager) (Provider, error) {
 	nodes := cache.GetAllNodes()
 	if len(nodes) == 0 {
-		return nil, fmt.Errorf("Could not locate any nodes for cluster.")
+		log.Infof("Could not locate any nodes for cluster.") // valid in ETL readonly mode
+		return &CustomProvider{
+			Clientset: cache,
+			Config:    NewProviderConfig(config, "default.json"),
+		}, nil
 	}
 
 	cp := getClusterProperties(nodes[0])
--- a/pkg/clustercache/clustercache.go
+++ b/pkg/clustercache/clustercache.go
@@ -134,27 +134,30 @@ func NewKubernetesClusterCache(client kubernetes.Interface) ClusterCache {
 	}
 
 	// Wait for each caching watcher to initialize
-	var wg sync.WaitGroup
-	wg.Add(16)
-
 	cancel := make(chan struct{})
-
-	go initializeCache(kcc.namespaceWatch, &wg, cancel)
-	go initializeCache(kcc.nodeWatch, &wg, cancel)
-	go initializeCache(kcc.podWatch, &wg, cancel)
-	go initializeCache(kcc.kubecostConfigMapWatch, &wg, cancel)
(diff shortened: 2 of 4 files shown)


pkg/env/costmodelenv.go
```diff
@@ -84,8 +84,14 @@ const (
 	PrometheusRetryOnRateLimitDefaultWaitEnvVar = "PROMETHEUS_RETRY_ON_RATE_LIMIT_DEFAULT_WAIT"
 
 	IngestPodUIDEnvVar = "INGEST_POD_UID"
+
+	ETLReadOnlyMode = "ETL_READ_ONLY"
 )
 
+func GetETLReadOnlyMode() bool {
```
mbolt35: Just a small nit. We've been using `Is` prefixes for boolean values. `IsETLReadOnlyMode()`

pkg/cloud/provider.go
```diff
@@ -423,7 +423,11 @@ func ShareTenancyCosts(p Provider) bool {
 func NewProvider(cache clustercache.ClusterCache, apiKey string, config *config.ConfigFileManager) (Provider, error) {
 	nodes := cache.GetAllNodes()
 	if len(nodes) == 0 {
-		return nil, fmt.Errorf("Could not locate any nodes for cluster.")
+		log.Infof("Could not locate any nodes for cluster.") // valid in ETL readonly mode
+		return &CustomProvider{
+			Clientset: cache,
+			Config:    NewProviderConfig(config, "default.json"),
+		}, nil
```
mbolt35, line 430: We should really plan on allowing runtime changes to provider (or maybe only transformations from custom -> specific).

pkg/clustercache/clustercache.go
```diff
@@ -134,27 +134,30 @@ func NewKubernetesClusterCache(client kubernetes.Interface) ClusterCache {
...
+		go initializeCache(kcc.statefulsetWatch, &wg, cancel)
+		go initializeCache(kcc.replicasetWatch, &wg, cancel)
+		go initializeCache(kcc.pvWatch, &wg, cancel)
+		go initializeCache(kcc.pvcWatch, &wg, cancel)
+		go initializeCache(kcc.storageClassWatch, &wg, cancel)
+		go initializeCache(kcc.jobsWatch, &wg, cancel)
+		go initializeCache(kcc.hpaWatch, &wg, cancel)
+		go initializeCache(kcc.podWatch, &wg, cancel)
+		go initializeCache(kcc.replicationControllerWatch, &wg, cancel)
+	}
```
mbolt35, line 160: This is fine for now. The ConfigMap watchers should be moved out of this type anyway, as it doesn't meet the contract, and ultimately, I think a NoOpClusterCache or something would be preferable, but not possible with the configmap watcher dependency.

mbolt35 (approved):
To support a non-crashing version of our product with no k8s API access and no provider, I am ok with these changes. However, I think if we have to build on these changes, we should have a better plan on how to accomplish these requirements in a more fluid way.

PR #1191, opened 2022-05-04
Collect attached volume totals by cluster

nikovacevic:
## What does this PR change?
* Collects attached volume resource totals by cluster and node, instead of just by node

* Required for https://github.com/kubecost/kubecost-cost-model/pull/747

See https://github.com/kubecost/kubecost-cost-model/pull/747 for more.

### DIFF
--- a/pkg/kubecost/totals.go
+++ b/pkg/kubecost/totals.go
@@ -295,9 +295,9 @@ func ComputeAssetTotals(as *AssetSet, prop AssetProperty) map[string]*AssetTotal
 				key = fmt.Sprintf("%s/%s", node.Properties().Cluster, node.Properties().Name)
 			}
 
-			// Add node name to list of node names, but only if aggregating
-			// by node. (These are to be used later for attached volumes.)
-			nodeNames[key] = true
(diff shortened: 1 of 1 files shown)


michaelmdresser (approved):

pkg/kubecost/totals.go
```diff
@@ -295,9 +295,9 @@ func ComputeAssetTotals(as *AssetSet, prop AssetProperty) map[string]*AssetTotal
 				key = fmt.Sprintf("%s/%s", node.Properties().Cluster, node.Properties().Name)
 			}
```
mbolt35, line 296: Can we just remove the key variable and this conditional?

mbolt35 (approved):
LGTM. It looks like we can remove the `key` variable and the condition for setting the key, only thing that think is reasonable to change.

pkg/kubecost/totals.go
nikovacevic: No, we use it later on. It's just been replaced in that one line.

(Besides, Pike wouldn't let me keep it if I weren't using it :rofl:)

PR #1193, opened 2022-05-04
Don't add gpu pricing when no gpu

### LINKED ISSUE #1389: Allocation API has GPU costs when GPU count is 0
**Describe the bug**
When using a custom pricing model, GPU prices can be assigned to \_\_idle\_\_ costs, even though the GPU count for the node is 0. 

**To Reproduce**
1. Create a cluster with no GPU-enabled nodes.
2. Install Kubecost with `kubecostProductConfigs.customPricesEnabled = true` in the values.yaml. The default pricing values will work to demonstrate the bug. 
3. Navigate to the Cost Allocation page and see \_\_idle\_\_ GPU costs.

**Expected behavior**
GPU costs should be $0.

**Screenshots**
image

dramich:
## What does this PR change?
* When building node pricing, no longer add GPU costs if the node does not have an associated GPU

* Fix idle cost being associated with nodes that do not have a GPU when using custom pricing

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1389

## How was this PR tested?
* Running in a custom cluster before and after the change
* Running in an EKS cluster with two non-GPU nodes and a single GPU node. Tested before and after the fix to ensure idle time was not attributed to nodes without the GPU and GPU node showed idle time. 

* n/a

*

### DIFF
--- a/pkg/costmodel/cluster_helpers.go
+++ b/pkg/costmodel/cluster_helpers.go
@@ -246,10 +246,10 @@ func buildGPUCostMap(
 		clusterAndNameToType[keyNon] = nodeType
 
 		// If gpu count is available use it to multiply gpu cost
-		if value, ok := gpuCountMap[key]; ok && value != 0 {
+		if value, ok := gpuCountMap[key]; ok {
 			gpuCostMap[key] = gpuCost * value
 		} else {
-			gpuCostMap[key] = gpuCost
+			gpuCostMap[key] = 0
 		}
 
 	}
--- a/pkg/costmodel/cluster_helpers_test.go
+++ b/pkg/costmodel/cluster_helpers_test.go
@@ -772,7 +772,7 @@ func TestBuildGPUCostMap(t *testing.T) {
 					Cluster:    "cluster1",
 					Name:       "node1",
 					ProviderID: "provider1",
-				}: 2,
+				}: 0,
 			},
 		},
 		{
@@ -799,7 +799,7 @@ func TestBuildGPUCostMap(t *testing.T) {
 					Cluster:    "cluster1",
 					Name:       "node1",
 					ProviderID: "provider1",
-				}: 2,
+				}: 0,
 			},
 		},
 		{


pkg/costmodel/cluster_helpers.go
```diff
@@ -246,10 +246,10 @@ func buildGPUCostMap(
 		clusterAndNameToType[keyNon] = nodeType
 
 		// If gpu count is available use it to multiply gpu cost
-		if value, ok := gpuCountMap[key]; ok && value != 0 {
+		if value, ok := gpuCountMap[key]; ok {
 			gpuCostMap[key] = gpuCost * value
 		} else {
-			gpuCostMap[key] = gpuCost
+			gpuCostMap[key] = 0
```
michaelmdresser, line 252: Is it possible to be in a situation with custom pricing in which:
1. A node's GPU cost is (intentionally) set to something > 0
2. Our GPU count query doesn't pick up the node's GPU, possibly because the node metadata is set up incorrectly or our tracking of node metadata is incomplete

I think this change would cause the node's GPU cost to be 0 when it shouldn't be.

---

But perhaps that's okay and we should just be setting count correctly! If that's the case, then the code change you've made is perfect.

michaelmdresser, line 252: Is it possible to be in a situation with custom pricing in which:
1. A node's GPU cost is (intentionally) set to something > 0
2. Our GPU count query doesn't pick up the node's GPU, possibly because the node metadata is set up incorrectly or our tracking of node metadata is incomplete

I think this change would cause the node's GPU cost to be 0 when it shouldn't be.

---

But perhaps that's okay and we should just be setting count correctly! If that's the case, then the code change you've made is perfect.

dramich: That was one of the scenarios I was considering and I think that if we can't/didn't count the GPUs correctly pricing would be wrong anyways. At least if it shows up as a zero and you know the node has a GPU that is a better indication that something is wrong vs showing incorrect pricing on a node that has 10 GPUs and we show the cost of 1.

michaelmdresser: I'm on board with that reasoning.

michaelmdresser (approved):

pkg/costmodel/cluster_helpers.go
nikovacevic: This looks good to me. Let's just make sure that @AjayTripathy is aware.

nikovacevic (approved):

pkg/costmodel/cluster_helpers.go
AjayTripathy: I think this is fine but let's not snap this into the release since I'm not sure of all the implications and we don't have good GPU integration tests...that sounds good to everyone? Let's take a follow on task to add a GPU integration test for GCP and AWS, we'll have a month to get confident here.

AjayTripathy (approved):

PR #1197, opened 2022-05-10
Switch log to debug

nikovacevic:
## What does this PR change?
* Switches a spammy "update" log to `Debug`


AjayTripathy (approved):

AjayTripathy:
Just curious should we release v1.93 without this?

nikovacevic:
It's a small change, and it'd be better not to, so I'll advocate for a quick patch. Probably going to want a new RC anyways for https://github.com/kubecost/kubecost-cost-model/pull/765. (Or we can decide that both are so small that we can release without an RC after merging them.)

PR #1201, opened 2022-05-10
Drop color from logger output

dramich:
## What does this PR change?
* Drop color from logger output

* No

* Drops the color from the log outputs


teevans (approved):
LGTM

Adam-Stack-PM:
@dramich Should I add the v1.94 label?

PR #1203, opened 2022-05-11
Remove ANSI colors from podLogs endpoint and update logging init

dramich:
## What does this PR change?
* Add colors back to logs and remove the ANSI colors from the podLogs endpoint. This fixes the issue when a customer creates a bug report and we process logs from kubectl directly which adds the colors. 
* Move logging init to happen after cobra has processed the flags/args
and add a new flag to disable color output of the logs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 ui/.cache
 ui/dist
 ui/node_modules/
+cmd/costmodel/costmodel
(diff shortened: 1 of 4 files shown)


michaelmdresser (approved):

michaelmdresser (approved):

.gitignore
```diff
@@ -2,3 +2,4 @@
 ui/.cache
 ui/dist
 ui/node_modules/
+cmd/costmodel/costmodel
```
mbolt35, line 5: ?

dramich: git really wants me to check in the binary

PR #1208, opened 2022-05-16
Add AllocationFilter support for services contains

michaelmdresser:
Existing allocation filtering allows users to `filterServices=a,b,c` which
becomes a "slice contains" operation because `Properties.Services` is a
`[]string`. This commit adds `FilterServices` (Field) and `FilterContains` (Op)
to support this type of filter.

Unallocated behavior is based on existing `GetServiceFilterFunc` (in KCM).

## How was this PR tested?
New unit tests

### DIFF
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -24,6 +24,8 @@ const (
 
 	FilterLabel      = "label"
 	FilterAnnotation = "annotation"
+
+	FilterServices = "services"
 )
 
 // FilterOp is an enum that represents operations that can be performed
@@ -35,6 +37,7 @@ type FilterOp string
 const (
 	FilterEquals    FilterOp = "equals"
 	FilterNotEquals          = "notequals"
+	FilterContains           = "contains"
 )
 
 // AllocationFilter represents anything that can be used to filter an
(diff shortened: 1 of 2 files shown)


pkg/kubecost/allocationfilter.go
```diff
@@ -35,6 +37,7 @@ type FilterOp string
 const (
 	FilterEquals    FilterOp = "equals"
 	FilterNotEquals          = "notequals"
+	FilterContains           = "contains"
```
nikovacevic, line 40: Do you see `contains` being used for any other fields, or just "services"? (There is, of course, the idea of a string containing another string, for instance.)

pkg/kubecost/allocationfilter.go
```diff
@@ -169,6 +175,20 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
...
+			}
+
+			for _, s := range stringSlice {
+				if s == filter.Value {
+					return true
+				}
+			}
+		} else {
+			log.Warnf("Allocation Filter: invalid 'contains' call for non-list filter value")
+		}
```
nikovacevic, line 191: Ideally, we'll have a stronger mechanism for dealing with illegal filter expressions -- would you agree? Like an `Evaluate() error` function or similar, so that we can pre-empt these and return a 400?

nikovacevic (approved):

pkg/kubecost/allocationfilter.go
michaelmdresser: Yeah, definitely in favor of this. We could force it to be checked by parsing to a struct with unexported methods (e.g. `UncheckedFilter`) and then you have `UncheckedFilter.Check() (error, CheckedFilter)` to turn it into something usable.

pkg/kubecost/allocationfilter.go
michaelmdresser: I think I'd require (via Check/Evaluate) that `contains` only be used for array-like data. I'm not sure that we want to support queries that include complex string conditions -- if we did I'd consider CEL more.

PR #1210, opened 2022-05-16
Add support for translating current filters (filterNamespaces=) to the new AllocationFilter types

michaelmdresser:
## What does this PR change?
Adds a new package `pkg/util/filterutil` with support for translating the current HTTP filters (e.g. `filterNamespaces=kubecost,kube-system&filterLabels=app:foo`) into the new `AllocationFilter` types introduced in https://github.com/kubecost/cost-model/pull/1172.

> I would've put this under `httputil` but that causes a circular dependency.

The filter building is roughly based on the [ETL handler filter building in KCM (closed source)](https://github.com/kubecost/kubecost-cost-model/blob/693db781bf6c80471e5c8fdda34f68172bdb3e98/pkg/etl/handlers.go#L489-L677).

- ~This code does not yet support the wildcard syntax (e.g. `filterNamespaces=kube*`). That'll be a followup PR because it involves changes to the `AllocationFilter` types and I didn't want to overload this PR. See https://github.com/kubecost/cost-model/pull/1211 which adds this logic.~ The CM#1211 logic was merged into this branch.
- This logic is not yet hooked up to the HTTP handlers in CM or KCM. That requires the above wildcard support + plumbing work to swap out filter funcs throughout core query logic.

## How was this PR tested?
Unit tests in this PR.

### DIFF
--- a/pkg/util/filterutil/allocationfilters.go
+++ b/pkg/util/filterutil/allocationfilters.go
@@ -210,13 +210,16 @@ func AllocationFilterFromParamsV1(
 	}
 	for _, filterValue := range qp.GetList("filterServices", ",") {
 		// TODO: wildcard support
-		servicesFilter.Filters = append(servicesFilter.Filters,
-			kubecost.AllocationFilterCondition{
-				Field: kubecost.FilterServices,
-				Op:    kubecost.FilterContains,
-				Value: filterValue,
-			},
-		)
+		filterValue, wildcard := parseWildcardEnd(filterValue)
+		subFilter := kubecost.AllocationFilterCondition{
+			Field: kubecost.FilterServices,
+			Op:    kubecost.FilterContains,
+			Value: filterValue,
+		}
+		if wildcard {
+			subFilter.Op = kubecost.FilterStartsWith
+		}
+		servicesFilter.Filters = append(servicesFilter.Filters, subFilter)
 	}
 	filter.Filters = append(filter.Filters, servicesFilter)
 
--- a/pkg/util/filterutil/allocationfilters.go
+++ b/pkg/util/filterutil/allocationfilters.go
@@ -174,18 +174,19 @@ func AllocationFilterFromParamsV1(
 		filterV1DoubleValueFromList(qp.GetList("filterLabels", ","), kubecost.FilterLabel),
 	)
 
-	// TODO: filter service condition
-	// filterServices := qp.GetList("filterServices", ",")
-	// if len(filterServices) > 0 {
-	// 	subFilter := kubecost.AllocationFilterOr{
-	// 		Filters: []kubecost.AllocationFilter{},
-	// 	}
-
-	// 	for _, filter := range filterServices {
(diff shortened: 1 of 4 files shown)


Adam-Stack-PM:
@nikovacevic Could we get a review on this when you have time?

nikovacevic:
Yep!

pkg/util/filterutil/allocationfilters.go
```diff
@@ -0,0 +1,328 @@
...
+//
+// It takes an optional LabelConfig, which if provided enables "label-mapped"
+// filters like "filterDepartments".
+//
+// It takes an optional ClusterMap, which if provided enables cluster name
+// filtering. This turns all `filterClusters=foo` arguments into the equivalent
+// of `clusterID = "foo" OR clusterName = "foo"`.
+//
+// TODO: This does not yet support filters with wildcards (e.g.
+// filterNamespaces=kube*)
```
nikovacevic: Still true?

pkg/util/filterutil/allocationfilters.go
```diff
@@ -0,0 +1,328 @@
...
+				clusterNameToIDs[info.Name] = []string{info.ID}
+			}
+		}
+	}
+
+	filterClusters := qp.GetList("filterClusters", ",")
+	clustersOr := kubecost.AllocationFilterOr{
+		Filters: []kubecost.AllocationFilter{},
+	}
+	clustersOr.Filters = append(clustersOr.Filters, filterV1SingleValueFromList(filterClusters, kubecost.FilterClusterID))
```
nikovacevic: Should we go through this, even if nothing is provided to `filterClusters`?

nikovacevic (approved):
Holy moly, these functions are dense reading! I don't see anything wrong. Thank you for the tests. Anything you're concerned about?

pkg/util/filterutil/allocationfilters.go
nikovacevic: (Ditto all the others -- just trying to make sure we don't end up evaluating a ton of `true`s unnecessarily if we know there's no value in the filter.)

pkg/util/filterutil/allocationfilters.go
michaelmdresser: Removed!

pkg/util/filterutil/allocationfilters.go
michaelmdresser: That's a good call. While its sort of a premature optimization, I'm all for it. Updated all of these blocks to include a `> 0` guard based on `qp.GetList()`. Unit tests continue to pass.

michaelmdresser:
> Anything you're concerned about?

Mostly extensibility of this code. I tried to document as well as possible, but there are pitfalls around wildcard handling and the use of an "enum" to represent ops. Enum problems are just a feature of Go, not much we can do there other than be diligent and document well. Wildcards are a gotcha just because you have to remember them.

The benefit of this code is that tests are fairly straightforward to write, so it shouldn't be too bad.

PR #1213, opened 2022-05-17
Audit and Generic Store types

Sean-Holcomb:
## What does this PR change?
This PR implements types for Audit along with the `ETLSet` interface and the Generic `SetRange` which uses it.
The Audit types are meant to record the results of the different Audits. The general purpose is to record incorrect results along with values the are missing from one source or another. The `ETLSet` interface encompasses the functionality required by the Generic `StoreStrategy` and any Set which is modified to implement it can be moved to a generic `StoreStrategy`. `SetRange` is meant to cover the basic range functionality of managing the `ETLSets`. A `SetRange` of a specific `ETLSet` type can be inherited and then expanded to cover any specialized functionality such as `Accumulate`.

Other changes:
* Add adjustment to cluster management type
* fix issue with window.IsEmpty()
* fix issue with lb adjustment in total store

* https://github.com/kubecost/kubecost-cost-model/pull/783

* see https://github.com/kubecost/kubecost-cost-model/pull/783

* Closes ...

## How was this PR tested?
* see https://github.com/kubecost/kubecost-cost-model/pull/783

* 

*

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -914,7 +914,7 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	shouldAggregate := aggregateBy != nil
 	shouldFilter := len(options.FilterFuncs) > 0
 	shouldShare := len(options.SharedHourlyCosts) > 0 || len(options.ShareFuncs) > 0
-	if !shouldAggregate && !shouldFilter && !shouldShare {
+	if !shouldAggregate && !shouldFilter && !shouldShare && options.ShareIdle == ShareNone {
 		// There is nothing for AggregateBy to do, so simply return nil
 		return nil
 	}
--- a/pkg/kubecost/audit.go
+++ b/pkg/kubecost/audit.go
@@ -37,8 +37,8 @@ func ToAuditType(check string) AuditType {
 		return AuditAssetTotalStore
 	case string(AuditAssetAggStore):
 		return AuditAssetAggStore
-	case string(AuditClusterEquality):
-		return AuditClusterEquality
+	//case string(AuditClusterEquality):
+	//	return AuditClusterEquality
 	case string(AuditAll):
 		return AuditAll
 	default:
@@ -61,13 +61,14 @@ type AuditMissingValue struct {
 	Key         string
 }
 
-// AuditFloatResult structure for holding the results of a failed audit on a float value, Expected should be the Audit generated value
-// while actual is what is contained in the relevant store
+// AuditFloatResult structure for holding the results of a failed audit on a float value, Expected should be the value
+// calculated by the Audit func while Actual is what is contained in the relevant store.
 type AuditFloatResult struct {
 	Expected float64
 	Actual   float64
 }
 
+// Clone returns a deep copy of the caller
 func (afr *AuditFloatResult) Clone() *AuditFloatResult {
 	return &AuditFloatResult{
 		Expected: afr.Expected,
@@ -75,8 +76,8 @@ func (afr *AuditFloatResult) Clone() *AuditFloatResult {
 	}
 }
 
-// AllocationReconciliationAudit records the differences of between compute resource costs between allocations by nodes
-// and node assets keyed on node name and resource
+// AllocationReconciliationAudit records the differences of between compute resources (cpu, ram, gpu) costs between
+// allocations by nodes and node assets keyed on node name and compute resource
 type AllocationReconciliationAudit struct {
 	Status        AuditStatus
 	Description   string
@@ -85,6 +86,7 @@ type AllocationReconciliationAudit struct {
 	MissingValues []*AuditMissingValue
 }
 
+// Clone returns a deep copy of the caller
 func (ara *AllocationReconciliationAudit) Clone() *AllocationReconciliationAudit {
 	if ara == nil {
 		return nil
--- a/pkg/kubecost/audit.go
+++ b/pkg/kubecost/audit.go
@@ -1,0 +1,361 @@
+package kubecost
+
+import (
+	"golang.org/x/exp/slices"
+	"sync"
+	"time"
+)
+
+// AuditType the types of Audits, each of which should be contained in an AuditSet
+type AuditType string
+
+const (
+	AuditAllocationReconciliation AuditType = "AuditAllocationReconciliation"
+	AuditAllocationTotalStore     AuditType = "AuditAllocationTotalStore"
+	AuditAllocationAggStore       AuditType = "AuditAllocationAggStore"
+	AuditAssetReconciliation      AuditType = "AuditAssetReconciliation"
+	AuditAssetTotalStore          AuditType = "AuditAssetTotalStore"
+	AuditAssetAggStore            AuditType = "AuditAssetAggStore"
+	AuditClusterEquality          AuditType = "AuditClusterEquality"
+
+	AuditAll         AuditType = ""
+	AuditInvalidType AuditType = "InvalidType"
+)
+
+// ToAuditType converts a string to an Audit type
+func ToAuditType(check string) AuditType {
+	switch check {
+	case string(AuditAllocationReconciliation):
+		return AuditAllocationReconciliation
+	case string(AuditAllocationTotalStore):
+		return AuditAllocationTotalStore
+	case string(AuditAllocationAggStore):
+		return AuditAllocationAggStore
+	case string(AuditAssetReconciliation):
+		return AuditAssetReconciliation
+	case string(AuditAssetTotalStore):
+		return AuditAssetTotalStore
+	case string(AuditAssetAggStore):
+		return AuditAssetAggStore
+	case string(AuditClusterEquality):
+		return AuditClusterEquality
+	case string(AuditAll):
+		return AuditAll
+	default:
+		return AuditInvalidType
+	}
+}
+
+// AuditStatus are possible outcomes of an audit
+type AuditStatus string
+
+const (
+	FailedStatus  AuditStatus = "Failed"
+	WarningStatus             = "Warning"
+	PassedStatus              = "Passed"
+)
+
+// AuditMissingValue records when a value that should be present in a store or in the audit generated results are missing
+type AuditMissingValue struct {
+	Description string
+	Key         string
+}
+
+// AuditFloatResult structure for holding the results of a failed audit on a float value, Expected should be the Audit generated value
+// while actual is what is contained in the relevant store
+type AuditFloatResult struct {
+	Expected float64
+	Actual   float64
+}
+
+func (afr *AuditFloatResult) Clone() *AuditFloatResult {
+	return &AuditFloatResult{
+		Expected: afr.Expected,
+		Actual:   afr.Actual,
+	}
+}
+
+// AllocationReconciliationAudit records the differences of between compute resource costs between allocations by nodes
+// and node assets keyed on node name and resource
+type AllocationReconciliationAudit struct {
+	Status        AuditStatus
+	Description   string
+	LastRun       time.Time
+	Resources     map[string]map[string]*AuditFloatResult
+	MissingValues []*AuditMissingValue
+}
+
+func (ara *AllocationReconciliationAudit) Clone() *AllocationReconciliationAudit {
+	if ara == nil {
+		return nil
+	}
+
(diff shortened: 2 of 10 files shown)


pkg/kubecost/bingen.go
```diff
@@ -24,7 +24,7 @@ package kubecost
 // @bingen:generate:Window
 
 // Asset Version Set: Includes Asset pipeline specific resources
-// @bingen:set[name=Assets,version=15]
+// @bingen:set[name=Assets,version=16]
```
mbolt35: Let's not advance the versions here, create another set for the audit stuff.

pkg/kubecost/bingen.go
```diff
@@ -40,10 +40,15 @@ package kubecost
 // @bingen:generate:Network
 // @bingen:generate:Node
 // @bingen:generate:SharedAsset
+// @bingen:generate:AuditDetail
+// @bingen:generate:Audit
+// @bingen:generate:AuditType
+// @bingen:generate:AuditStatus
+// @bingen:generate[stringtable]:AuditSet
```
mbolt35: Remove these from Assets, and just add a section like this

```
// @bingen:set[name=Audit,version=1]
// @bingen:generate:AuditDetail
// @bingen:generate:Audit
// @bingen:generate:AuditType
// @bingen:generate:AuditStatus
// @bingen:generate[stringtable]:AuditSet
// @bingen:end

File: pkg/kubecost/bingen.go [hunk 3]
```diff
@@ -57,4 +62,4 @@ package kubecost
 // @bingen:generate:PVAllocation
 // @bingen:end
 
-//go:generate bingen -package=kubecost -version=15 -buffer=github.com/kubecost/cost-model/pkg/util
+//go:generate bingen -package=kubecost -version=16 -buffer=github.com/kubecost/cost-model/pkg/util
```
mbolt35: Revert back to 15

pkg/kubecost/bingen.go
```diff
@@ -40,10 +40,15 @@ package kubecost
...
+// @bingen:generate:AuditDetail
+// @bingen:generate:Audit
+// @bingen:generate:AuditType
+// @bingen:generate:AuditStatus
+// @bingen:generate[stringtable]:AuditSet
 // @bingen:end
 
 // Allocation Version Set: Includes Allocation pipeline specific resources
-// @bingen:set[name=Allocation,version=15]
+// @bingen:set[name=Allocation,version=16]
```
mbolt35: Revert back to 15

pkg/kubecost/set.go
```diff
@@ -0,0 +1,13 @@
+package kubecost
+
+import "encoding"
+
+type Set interface {
```
mbolt35: Probably want to be a bit more specific with a name here: `ETLSet`?

pkg/kubecost/set.go
```diff
@@ -0,0 +1,13 @@
...
+	CloneSet() Set
```
mbolt35: Just `Clone` here

pkg/kubecost/range.go
```diff
@@ -0,0 +1,73 @@
+package kubecost
+
+import (
+	"fmt"
+	"github.com/kubecost/cost-model/pkg/util/json"
+	"sync"
+)
+
+// Range is a generic implementation
+type Range[T Set] struct {
```
mbolt35: `ETLRange`? I'm more ok with just `Range` here than I am with `Set` so I'll leave it up to you.

pkg/kubecost/set.go
Sean-Holcomb: Yeah I like ETLSet

pkg/kubecost/set.go
Sean-Holcomb: The problem with Clone is that it that the signature is already taken all over the code base with the incorrect return

Adam-Stack-PM:
@Sean-Holcomb Is the goal to get this in v1.94?

pkg/kubecost/allocation.go
```diff
@@ -914,7 +914,7 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	shouldAggregate := aggregateBy != nil
 	shouldFilter := len(options.FilterFuncs) > 0
 	shouldShare := len(options.SharedHourlyCosts) > 0 || len(options.ShareFuncs) > 0
-	if !shouldAggregate && !shouldFilter && !shouldShare {
+	if !shouldAggregate && !shouldFilter && !shouldShare && options.ShareIdle == ShareNone {
```
michaelmdresser, line 917: Hey, just want to make sure I understand this change -- what's going on here?

pkg/kubecost/audit.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+}
+
+// AllocationReconciliationAudit records the differences of between compute resource costs between allocations by nodes
+// and node assets keyed on node name and resource
+type AllocationReconciliationAudit struct {
+	Status        AuditStatus
+	Description   string
+	LastRun       time.Time
+	Resources     map[string]map[string]*AuditFloatResult
```
michaelmdresser, line 85: Would you add a documentation comment on this field to explain what the map keys are? Something like:
```
// Resources is a ... keyed by X and Y
Resources map[string]map[string]....
```

pkg/kubecost/window.go
```diff
@@ -456,7 +456,7 @@ func (w Window) Hours() float64 {
 }
 
 func (w Window) IsEmpty() bool {
-	return !w.IsOpen() && w.end.Equal(*w.Start())
+	return w.start == nil && w.end == nil
```
michaelmdresser, line 460: I'd like to learn more about this change. Is there a reason why we want to start calling a `Window` with non-`nil` start and end not empty?

pkg/kubecost/audit.go
```diff
@@ -0,0 +1,361 @@
...
+	Description string
+	Key         string
+}
+
+// AuditFloatResult structure for holding the results of a failed audit on a float value, Expected should be the Audit generated value
+// while actual is what is contained in the relevant store
+type AuditFloatResult struct {
+	Expected float64
+	Actual   float64
+}
```
michaelmdresser: What is an Audit generated value? Just want to make sure these fields will make sense for users!

pkg/kubecost/etlset.go
```diff
@@ -0,0 +1,15 @@
+package kubecost
+
+import "encoding"
+
+// ETLSet is an interface which represents the basic data block of an ETL. It is keyed by its Window
+type ETLSet interface {
+	ConstructSet() ETLSet
```
michaelmdresser, line 7: Does `ConstructSet()` give us anything that just making the struct wouldn't?

pkg/kubecost/etlset.go
```diff
@@ -0,0 +1,15 @@
...
+	CloneSet() ETLSet
```
michaelmdresser, line 8: Any reason this shouldn't just be called `Clone()`?

pkg/kubecost/etlrange.go
```diff
@@ -0,0 +1,79 @@
...
+// is shared by the basic types but is meant to be extended by each implementation.
+type SetRange[T ETLSet] struct {
+	lock sync.RWMutex
+	sets []T
+}
+
+// Append attaches the given ETLSet to the end of the sets slice.
+// currently does not check that the window is correct.
+func (r *SetRange[T]) Append(that T) {
+	r.lock.Lock()
```
michaelmdresser: Would you add an `if r == nil` guard on this?

pkg/kubecost/etlrange.go
```diff
@@ -0,0 +1,79 @@
...
+// Each invokes the given function for each ETLSet in the SetRange
+func (r *SetRange[T]) Each(f func(int, T)) {
+	if r == nil {
+		return
+	}
+
+	for i, set := range r.sets {
+		f(i, set)
+	}
+}
```
michaelmdresser, line 36: @mbolt35 discussed pros and cons of Set(Range) locking a little while ago, but unfortunately I don't remember where our conversations landed -- should `Each` lock?

pkg/kubecost/etlrange.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Get retrieves the given index from the sets slice
+func (r *SetRange[T]) Get(i int) (T, error) {
+	if i < 0 || i >= len(r.sets) {
```
michaelmdresser: A `nil` guard here would also be great!

michaelmdresser (commented):
I love the Go-enums!

Some of my comments may be addressed by the KCM PR, I just haven't gotten to it yet and wanted to get you feedback earlier rather than later!

pkg/kubecost/etlset.go
Sean-Holcomb: In general ETLSet is going to be a pointer with a zero value of `nil`. In the case of GenericStoreStrategy when unmarshalling binary we need an empty set to load into which is not nil. The use case looks something like this.
```
func (fss *FileStoreStrategy[T]) load(key string) T {
nonNilSet := T.ConstructSet()
fss.loadFromFile(key, nonNilSet)
return nonNilSet
}
```

pkg/kubecost/etlset.go
Sean-Holcomb: Yes, the Clone() signature is taken across the codebase with the return value of the impl type and cannot easily be updated to return `ETLSet`. On the other hand `CloneSet()` can be quickly implemented for classes that already have `Clone()` by just returning the `Clone()` which would not require a massive refactor.

pkg/kubecost/window.go
Sean-Holcomb: My original take here was that the previous impl was a typo and should be `return w.IsOpen() && w.end.Equal(*w.Start())`. If felt that `return w.start == nil && w.end == nil` conveyed that more succinctly but reading it again it appears that the original author meant that a window that covers no time is empty. I feel that IsEmpty is a better idea for what I was trying to express. And a separate function like `window.HasDurration()` makes more sense.

pkg/kubecost/allocation.go
Sean-Holcomb: Its a bit of an edge case but the comment below is helpful
```
// There is nothing for AggregateBy to do, so simply return nil
```
This function, among too many other things, is responsible for sharing idle, so in the case where idle is supposed to be shared but none of the other functions were utilized, idle was not being shared.

pkg/kubecost/etlset.go
michaelmdresser: Got it! that makes sense.

pkg/kubecost/allocation.go
michaelmdresser: Aha! This is a great catch, then. Nice!

pkg/kubecost/etlset.go
michaelmdresser: Works for me!

pkg/kubecost/window.go
michaelmdresser: I'm still a bit worried about changing the definition of `IsEmpty()` here. I think the intent of a `nil` start or end is to represent an unbounded/infinite side of a window.

E.g.
- (start, end) = (now, nil) represents [now, ∞)
- (start, end) = (nil, now) represents (∞, now]
- (start, end) = (nil, nil) represents (∞, ∞)

I don't think it's quite right to say that (∞, ∞) is an "empty" window. It might be better to name the current `IsEmpty()` functionality as something more like `HasFiniteNonZeroDuration()`, but that's kind of weird, too.

If you need a method to check `w.start == nil && w.end == nil`, how about a new method like `func (w Window) IsFullyOpen() bool`?

pkg/kubecost/etlrange.go
mbolt35: This will continue to be an ongoing discussion, but where we eventually landed is that the ETL objects should not be locked at all. There are still remnants of locking in some of the objects before we were issuing copies from the stores. Since copy is the default for all ETL stores, we don't risk corruption (at least in context to the internal stores). If the receiver of the data shares these objects across go-routines, then it should be their responsibility to implement locking. 

We need to eventually clean up these objects to remove the old locking primitives, which should remove a lot of the complexity and actually speed up access (minimally).

mbolt35 (approved):
These changes look great! Thanks @Sean-Holcomb

PR #1214, opened 2022-05-18
Cost-Model Monitoring and Diagnostic Enhancements

mbolt35:
## What does this PR change?
* Adds http middleware that can monitor response times, sizes, hooked to the eventing system. 
* Adds prometheus metrics exporter that updates metrics based on events. 

* Adds the following prometheus metrics: 
  * `kubecost_http_requests_total` [Counter]: Total number of http requests serviced.
  * `kubecost_http_response_time_seconds` [Histogram]: Response time in seconds.
  * `kubecost_http_response_size_bytes` [Summary]: Response size in bytes. 

* Addresses a few of the requests on https://github.com/kubecost/cost-model/issues/264

## How was this PR tested?
* Live environment, observing `/model/metrics`

* Yes.


pkg/metrics/httpmetricmiddleware.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/kubecost/events"
+)
+
+// ResponseMetricMiddleware dispatches metric events for handles request and responses.
+func ResponseMetricMiddleware(handler http.Handler) http.Handler {
+	dispatcher := events.GlobalDispatcherFor[HttpHandlerMetricEvent]()
```
nikovacevic, line 13: Should I be using `GlobalDispatcherFor` in ETL, do you think?

nikovacevic: This is where my code review for you _really_ becomes your code review for me :laughing:

pkg/prom/diagnostics.go
```diff
@@ -66,75 +155,89 @@ func LogPrometheusClientState(client prometheus.Client) {
...
+	QueryFmt    string
+	Label       string
+	Description string
+	DocLink     string
+}
+
+// NewDiagnostic creates a new PrometheusDiagnostic instance using the provided definition data.
+func (pdd *diagnosticDefinition) NewDiagnostic(offset string) *PrometheusDiagnostic {
+	// FIXME: Any reasonable way to get the total number of replacements required in the query?
+	// FIXME: All of the other queries require a single offset replace, but CPUThrottle requires two.
```
nikovacevic, line 226: No easy ideas come to mind for this

nikovacevic (approved):

pkg/metrics/httpmetricmiddleware.go
mbolt35: I think it just depends -- I really don't care much for global singleton patterns, but I also don't like having to propagate the depths of the architecture to wire up eventing. 

It made sense to me to allow localized dispatchers for systems like ETL, but I did want to "normalize" a single per-type event dispatcher globally as well. In the case of metrics and the http middleware, these things are already "global" in nature, so appending the global dispatching was an easy decision. For ETL, it's really whether or not we want the events to be "exposed" - It's also perfectly reasonable to have a requirement like: If you want to subscribe to ETL events, you need to have a reference to the dispatcher created inside the ETL types. (Versus just allowing any part of the code to grab the global dispatcher and start spying 😄 )

nikovacevic: Switched to using the global. I like it more.

PR #1216, opened 2022-05-21
Add provider id for PV for GCP

Sean-Holcomb:
## What does this PR change?
Update to support GCP reconciliation. Adds provider Id's for GCP PVs and makes the general upgrade that all disks will have a provider id equal to their name if they do not already have one assigned. This second change will have an effect for all users that they will see disks that previously did not have a provider id as separate items at the time of the upgrade. This ID is determined at build time, so repairing or rebuilding the ETL will fix the issue.

* https://github.com/kubecost/kubecost-cost-model/pull/787

* Users will have provider ids on all disks. This will cause some disks to appear as separate items at the time of upgrade. Repairing/rebuilding older days will remove this discontinuity, applying the new provider id retroactively.  
* Closes ...

## How was this PR tested?
* see https://github.com/kubecost/kubecost-cost-model/pull/787

* https://github.com/kubecost/kubecost-cost-model/pull/787

*

### DIFF
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -1189,6 +1189,7 @@ func (gcp *GCP) getReservedInstances() ([]*GCPReservedInstance, error) {
 }
 
 type pvKey struct {
+	ProviderID             string
 	Labels                 map[string]string
 	StorageClass           string
 	StorageClassParameters map[string]string
@@ -1196,7 +1197,7 @@ type pvKey struct {
 }
 
 func (key *pvKey) ID() string {
-	return ""
+	return key.ProviderID
 }
 
 func (key *pvKey) GetStorageClass() string {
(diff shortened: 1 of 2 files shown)


nikovacevic (approved):
Seems reasonable to me. Let's just make sure that @AjayTripathy is aware for the sake of potential customer questions about the transition to PVs with provider IDs.

Adam-Stack-PM:
@AjayTripathy When you have time, would read through this PR to help address possible customer questions.  CC:@nikovacevic

Adam-Stack-PM:
@Sean-Holcomb Should this PR have an additional PR to update docs next week?

Sean-Holcomb:
Screen Shot 2022-05-27 at 9 48 00 AM
This is the discontinuity in action. Notice that their minutes total to 1440, which is the number of minutes in a day. The second item has one third the minutes and one third the cost. The only difference is that the second item has the new provider ID. It is worth noting that when reconciliation hits this day the new item will get the full price from the billing export for that day, making the previous item duplicate cost. Using Asset repair on this day will erase this discontinuity.

PR #1221, opened 2022-05-24
Storage Support for GCS and Azure

mbolt35:
## What does this PR change?
* Adds `storage.Storage` implementations for GCS and Azure 

* Adds support for shared configuration and backups over GCS and Azure.

## How was this PR tested?
* Live on GCS environment for GCS storage
* Azure Storage still needs testing... 

* Yes, will update Storage Documentation and ETL Backup Documentation.

### DIFF
--- a/pkg/storage/azurestorage.go
+++ b/pkg/storage/azurestorage.go
@@ -109,7 +109,7 @@ type AzureHTTPConfig struct {
 	TLSConfig TLSConfig `yaml:"tls_config"`
 }
 
-// AzureStorage implements the storeage.Storage interface against Azure APIs.
+// AzureStorage implements the storage.Storage interface against Azure APIs.
 type AzureStorage struct {
 	name         string
 	containerURL blob.ContainerURL
--- a/pkg/storage/azurestorage.go
+++ b/pkg/storage/azurestorage.go
@@ -241,6 +241,11 @@ func (as *AzureStorage) Name() string {
 	return as.name
 }
 
+// StorageType returns a string identifier for the type of storage used by the implementation.
+func (as *AzureStorage) StorageType() StorageType {
+	return StorageTypeBucketAzure
+}
+
 // FullPath returns the storage working path combined with the path provided
 func (as *AzureStorage) FullPath(name string) string {
 	name = trimLeading(name)
--- a/pkg/storage/azurestorage.go
+++ b/pkg/storage/azurestorage.go
@@ -277,7 +277,7 @@ func (b *AzureStorage) Read(name string) ([]byte, error) {
 	name = trimLeading(name)
 	ctx := context.Background()
 
-	log.Infof("AzureStorage::Read(%s)", name)
+	log.Debugf("AzureStorage::Read(%s)", name)
 
 	reader, err := b.getBlobReader(ctx, name, 0, blob.CountToEnd)
 	if err != nil {
@@ -298,7 +298,7 @@ func (b *AzureStorage) Write(name string, data []byte) error {
 	name = trimLeading(name)
 	ctx := context.Background()
 
-	log.Infof("AzureStorage::Write(%s)", name)
+	log.Debugf("AzureStorage::Write(%s)", name)
 
 	blobURL := getBlobURL(name, b.containerURL)
 	r := bytes.NewReader(data)
@@ -318,7 +318,7 @@ func (b *AzureStorage) Write(name string, data []byte) error {
 func (b *AzureStorage) Remove(name string) error {
 	name = trimLeading(name)
 
-	log.Infof("S3Storage::Remove(%s)", name)
+	log.Debugf("AzureStorage::Remove(%s)", name)
 	ctx := context.Background()
 
 	blobURL := getBlobURL(name, b.containerURL)
@@ -350,7 +350,7 @@ func (b *AzureStorage) Exists(name string) (bool, error) {
 func (b *AzureStorage) List(path string) ([]*StorageInfo, error) {
 	path = trimLeading(path)
 
-	log.Infof("S3Storage::List(%s)", path)
+	log.Debugf("AzureStorage::List(%s)", path)
 	ctx := context.Background()
 
 	// Ensure the object name actually ends with a dir suffix. Otherwise we'll just iterate the
--- a/pkg/storage/azurestorage.go
+++ b/pkg/storage/azurestorage.go
@@ -556,7 +556,7 @@ func getContainerURL(ctx context.Context, conf AzureConfig) (blob.ContainerURL, 
 
 	p := blob.NewPipeline(credentials, blob.PipelineOptions{
 		Retry:     retryOptions,
-		Telemetry: blob.TelemetryOptions{Value: "Thanos"},
+		Telemetry: blob.TelemetryOptions{Value: "Kubecost"},
 		RequestLog: blob.RequestLogOptions{
 			// Log a warning if an operation takes longer than the specified duration.
 			// (-1=no logging; 0=default 3s threshold)
--- a/pkg/storage/azurestorage.go
+++ b/pkg/storage/azurestorage.go
@@ -1,0 +1,642 @@
+package storage
+
+// Fork from Thanos S3 Bucket support to reuse configuration options
+// Licensed under the Apache License 2.0
+// https://github.com/thanos-io/thanos/blob/main/pkg/objstore/s3/s3.go
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/kubecost/cost-model/pkg/log"
+
+	"github.com/Azure/azure-pipeline-go/pipeline"
+	blob "github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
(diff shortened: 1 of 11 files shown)


pkg/storage/azurestorage.go
```diff
@@ -0,0 +1,642 @@
...
+		MaxRetryDelay: 0,
+	},
+	ReaderConfig: ReaderConfig{
+		MaxRetryRequests: 0,
+	},
+	HTTPConfig: AzureHTTPConfig{
+		IdleConnTimeout:       model.Duration(90 * time.Second),
+		ResponseHeaderTimeout: model.Duration(2 * time.Minute),
+		TLSHandshakeTimeout:   model.Duration(10 * time.Second),
+		ExpectContinueTimeout: model.Duration(1 * time.Second),
```
michaelmdresser, line 53: Why aren't these `time.Duration`?

pkg/storage/azurestorage.go
```diff
@@ -0,0 +1,642 @@
...
+
+func init() {
+	// Disable `ForceLog` in Azure storage module
+	// As the time of this patch, the logging function in the storage module isn't correctly
+	// detecting expected REST errors like 404 and so outputs them to syslog along with a stacktrace.
+	// https://github.com/Azure/azure-storage-blob-go/issues/214
+	//
+	// This needs to be done at startup because the underlying variable is not thread safe.
+	// https://github.com/Azure/azure-pipeline-go/blob/dc95902f1d32034f8f743ccc6c3f2eb36b84da27/pipeline/core.go#L276-L283
+	pipeline.SetForceLogEnabled(false)
```
michaelmdresser, line 69: :man_facepalming:

pkg/storage/azurestorage.go
```diff
@@ -0,0 +1,642 @@
...
+	ExpectContinueTimeout model.Duration `yaml:"expect_continue_timeout"`
+	MaxIdleConns          int            `yaml:"max_idle_conns"`
+	MaxIdleConnsPerHost   int            `yaml:"max_idle_conns_per_host"`
+	MaxConnsPerHost       int            `yaml:"max_conns_per_host"`
+	DisableCompression    bool           `yaml:"disable_compression"`
+
+	TLSConfig TLSConfig `yaml:"tls_config"`
+}
+
+// AzureStorage implements the storeage.Storage interface against Azure APIs.
```
michaelmdresser:
```suggestion
// AzureStorage implements the storage.Storage interface against Azure APIs.
```

pkg/storage/azurestorage.go
```diff
@@ -0,0 +1,642 @@
...
+}
+
+// Validate checks to see if any of the config options are set.
+func (conf *AzureConfig) validate() error {
+	var errMsg []string
+	if conf.MSIResource == "" {
+		if conf.UserAssignedID == "" {
+			if conf.StorageAccountName == "" ||
+				conf.StorageAccountKey == "" {
+				errMsg = append(errMsg, "invalid Azure storage configuration")
```
michaelmdresser, line 126: Could this be a little more descriptive?

pkg/storage/azurestorage.go
```diff
@@ -0,0 +1,642 @@
...
+	}, nil
+}
+
+// Read uses the relative path of the storage combined with the provided path to
+// read the contents.
+func (b *AzureStorage) Read(name string) ([]byte, error) {
+	name = trimLeading(name)
+	ctx := context.Background()
+
+	log.Infof("AzureStorage::Read(%s)", name)
```
michaelmdresser: Probably a `Debugf`? Same question for the other methods, like `Write`.

pkg/storage/azurestorage.go
```diff
@@ -0,0 +1,642 @@
...
+		return true
+	}
+
+	return false
+}
+
+func (b *AzureStorage) getBlobReader(ctx context.Context, name string, offset, length int64) (io.ReadCloser, error) {
+	log.Debugf("Getting blob: %s, offset: %d, length: %d", name, offset, length)
+	if name == "" {
+		return nil, errors.New("X-Ms-Error-Code: [EmptyContainerName]")
```
michaelmdresser, line 432: Is there a particular reason this string format is being used?

pkg/storage/azurestorage.go
```diff
@@ -0,0 +1,642 @@
...
+	}
+
+	if config.HTTPConfig.InsecureSkipVerify {
+		tlsConfig.InsecureSkipVerify = true
+	}
+	return &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
```
michaelmdresser, line 598: Are these long enough? Especially the Timeout.

pkg/storage/gcsstorage.go
```diff
@@ -0,0 +1,233 @@
+package storage
+
+// Fork from Thanos GCS Bucket support to reuse configuration options
+// Licensed under the Apache License 2.0.
+// https://github.com/thanos-io/thanos/blob/main/pkg/objstore/gcs/gcs.go
```
michaelmdresser, line 5: Is this all we need for attribution? Should we reference a commit it came from, or any specific authors?

michaelmdresser: Any specific changes you made to the fork worth pointing out, to help us poor reviewers?

michaelmdresser (commented):
By and large I think this looks good. This bucket interaction is pretty complex. I'm very close to approving, just a few questions.

Also, there are quite a few unwrapped errors that might make future debugging challenging.

pkg/storage/gcsstorage.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return data, nil
+}
+
+// Write uses the relative path of the storage combined with the provided path
+// to write a new file or overwrite an existing file.
+func (gs *GCSStorage) Write(name string, data []byte) error {
+	name = trimLeading(name)
+	log.Infof("GCSStorage::Write(%s)", name)
```
nikovacevic: Are we thinking that we want all of these as `Infof`? Could probably be `Debugf`, or even just deleted IMO.

nikovacevic (commented):
Good stuff! Sounds like we maybe want to test Azure still? Besides that, addressing logging, and addressing Michael's review, I'd be happy to approve.

pkg/storage/azurestorage.go
mbolt35: Possibly. I know you're not a fan of most of this configuration stuff, but it's _all_ 1-1 with the Thanos bucket storage config. The parsing code, how it handles auth, etc... The entire point is to leverage the same bucket storage configuration format that Thanos uses to avoid having to replicate buckets AND/OR auth schemes.

pkg/storage/azurestorage.go
mbolt35: I considered it, and probably should reduce the level at some point. It's only for Read/Write/Delete I think.

pkg/storage/azurestorage.go
mbolt35: No idea, thanos forked code :)

pkg/storage/azurestorage.go
mbolt35: I hope we're not requiring more than 30s to upload a single ETL file, BUT this is also just default thanos settings as well.

pkg/storage/gcsstorage.go
mbolt35: Hahaha! I see that this was the last thing you noticed as well. It's the same for all the bucket storage implementations. Based on what I have read concerning Apache 2, this is all that is required, but I would _love_ a confirmation there. 

Mainly, the fork was to ensure 1-1 with the Thanos Bucket Configuration for all of these bucket types. So, any functionality which could be configured in the Thanos format was forked and used. However, their "Bucket" API isn't the same as ours at all, it's designed around specific thanos use-cases (like recursively diving into sub-directories), so most of the implementation of the `storage.Storage` interface is "guided" by their implementation, but it's mostly different. The Azure implementations are probably the closest since the blob storage APIs for Azure are a complete disaster.

pkg/storage/gcsstorage.go
mbolt35: I could change these, but I also like spamming them... I'll flip a coin.

pkg/storage/azurestorage.go
mbolt35: Note: This is used to match the Thanos Bucket configuration format (Forked from thanos)

My guess is that they leverage the prometheus Durations which can parse a larger variety of duration strings (This is likely identical to the way we forked go duration code to add support for custom units). 

https://github.com/kubecost/cost-model/blob/868089f78585c40307bacb61d9aa54e39bc9170a/pkg/util/timeutil/timeutil.go#L95-L213

michaelmdresser: Works for me!

pkg/storage/azurestorage.go
michaelmdresser: Gotcha, I wasn't sure how much, if anything, was original authorship. I'll trust the robustness of the Thanos code, and we always have room to change in the future.

pkg/storage/azurestorage.go
michaelmdresser: Makes sense!

michaelmdresser (approved):

mbolt35:
Azure was tested on a live environment and monitoring reported greens for everything, so we're good to go! Pulling the trigger on both PRs.

Adam-Stack-PM:
@mbolt35 Great work.

PR #1225, opened 2022-05-28
[WIP] ETL Test Fixes and Better DataStructure Synchronization

mbolt35:
The entire purpose of this PR is to start a draft where we can all contribute to a discussion surrounding potential fixes with ETL tests being flakey, potential race conditions in our ETL data structures, and overall better practices to increase stability.

---

### Solutions
* There were a few methods on `AssetSetRange` and `AssetSet` that were using `self` funcs, which is annoyingly difficult to do effectively in go when it comes to making thread-safe data structures. This is mainly due to go not having recursive mutexes, so the lock holder has to relinquish a lock before and after each call. Since we want each public function on these structures to behave atomically (that is, any reads/writes within the func block should be behind one lock). 
* The `Each` iteration methods are concerning simply because of the lack of synchronization. However, we have been mostly careful concerning the context in which these iterator methods are used, so for the most part, our optimistic lack of locking has been beneficial. However, the fact that we can potentially deadlock if we add locking is concerning, so as one added layer of safety here, I'm adding soft copies to the collection `Each` iterates over. This will at least prevent the case where the collection at the moment the call is made is what you iterate over (the previous Each could have really nasty side effects if the collection was modified mid-iteration). We do _NOT_ want to deep copy this list, as that would be abysmal in terms of performance, either way, measurement is needed.
* The `Minutes()` func used two separate calls... one to Start(), and one to End(). These had nested `Each()` calls and required two separate lock/unlocks. I added a new func `StartAndEnd()` which will use a single loop, reader locks (it's clear the calls don't leave the body of the func or the loop for that matter). This should perform better and provide better safety .

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -2453,12 +2453,12 @@ func (as *AssetSet) AggregateBy(aggregateBy []string, opts *AssetAggregationOpti
 		opts = &AssetAggregationOptions{}
 	}
 
-	if as.IsEmpty() && len(opts.SharedHourlyCosts) == 0 {
-		return nil
-	}
-
 	as.Lock()
 	defer as.Unlock()
+
+	if len(as.assets) == 0 && len(opts.SharedHourlyCosts) == 0 {
+		return nil
+	}
 
 	aggSet := NewAssetSet(as.Start(), as.End())
 	aggSet.aggregateBy = aggregateBy
@@ -2556,6 +2556,7 @@ func (as *AssetSet) Clone() *AssetSet {
 	}
 }
 
+/*
 // Each invokes the given function for each Asset in the set
 func (as *AssetSet) Each(f func(string, Asset)) {
 	if as == nil {
@@ -2566,10 +2567,29 @@ func (as *AssetSet) Each(f func(string, Asset)) {
 		f(k, a)
 	}
 }
+*/
 
 // End returns the end time of the AssetSet's window
 func (as *AssetSet) End() time.Time {
 	return *as.Window.End()
+}
+
+// Each invokes the given function for each Asset in the set
+func (as *AssetSet) Each(f func(string, Asset)) {
+	if as == nil {
+		return
+	}
+
+	as.RLock()
+	assets := make(map[string]Asset, len(as.assets))
+	for k, v := range as.assets {
+		assets[k] = v
+	}
+	as.RUnlock()
+
+	for k, a := range assets {
+		f(k, a)
+	}
 }
 
 // FindMatch attempts to find a match in the AssetSet for the given Asset on
@@ -2703,13 +2723,13 @@ func (as *AssetSet) Insert(asset Asset) error {
 // IsEmpty returns true if the AssetSet is nil, or if it contains
 // zero assets.
 func (as *AssetSet) IsEmpty() bool {
-	if as == nil || len(as.assets) == 0 {
+	if as == nil {
 		return true
 	}
 
 	as.RLock()
 	defer as.RUnlock()
-	return as.assets == nil || len(as.assets) == 0
+	return len(as.assets) == 0
 }
 
 func (as *AssetSet) Length() int {
@@ -2737,14 +2757,12 @@ func (as *AssetSet) Resolution() time.Duration {
 }
 
 func (as *AssetSet) Set(asset Asset, aggregateBy []string) error {
-	if as.IsEmpty() {
-		as.Lock()
-		as.assets = map[string]Asset{}
-		as.Unlock()
-	}
-
 	as.Lock()
 	defer as.Unlock()
+
+	if len(as.assets) == 0 {
+		as.assets = map[string]Asset{}
+	}
 
 	// Expand the window to match the AssetSet, then set it
 	asset.ExpandWindow(as.Window)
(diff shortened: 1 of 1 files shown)


pkg/kubecost/asset.go
```diff
@@ -2453,13 +2453,13 @@ func (as *AssetSet) AggregateBy(aggregateBy []string, opts *AssetAggregationOpti
 		opts = &AssetAggregationOptions{}
 	}
 
-	if as.IsEmpty() && len(opts.SharedHourlyCosts) == 0 {
```
mbolt35, line 2456: This acquired the read lock to call `IsEmpty()`, relinquished the lock, then immediately acquired the write lock. In this case, contention can leak between these checks, and because go doesn't have upgradeable RW locks, it's better to do any cheap reads inline while holding the write lock.

pkg/kubecost/asset.go
```diff
@@ -2566,12 +2567,31 @@ func (as *AssetSet) Each(f func(string, Asset)) {
...
+	if as == nil {
+		return
+	}
+
+	as.RLock()
+	assets := make(map[string]Asset, len(as.assets))
+	for k, v := range as.assets {
+		assets[k] = v
+	}
+	as.RUnlock()
```
mbolt35, line 2588: While I'm not crazy about duplicating the collection here, it at least provides collection safety here. Since deep copy is out of the question due to performance, this is the next best thing.

pkg/kubecost/asset.go
```diff
@@ -2703,13 +2723,13 @@ func (as *AssetSet) Insert(asset Asset) error {
 // IsEmpty returns true if the AssetSet is nil, or if it contains
 // zero assets.
 func (as *AssetSet) IsEmpty() bool {
-	if as == nil || len(as.assets) == 0 {
```
mbolt35, line 2706: Reading the length of  the `assets` field outside of a lock.

pkg/kubecost/asset.go
```diff
@@ -2703,13 +2723,13 @@ func (as *AssetSet) Insert(asset Asset) error {
...
+	if as == nil {
 		return true
 	}
 
 	as.RLock()
 	defer as.RUnlock()
-	return as.assets == nil || len(as.assets) == 0
+	return len(as.assets) == 0
```
mbolt35, line 2732: `len(nil)` will also return 0

pkg/kubecost/asset.go
```diff
@@ -2737,15 +2757,13 @@ func (as *AssetSet) Resolution() time.Duration {
...
 
 func (as *AssetSet) Set(asset Asset, aggregateBy []string) error {
-	if as.IsEmpty() {
-		as.Lock()
-		as.assets = map[string]Asset{}
-		as.Unlock()
-	}
-
 	as.Lock()
 	defer as.Unlock()
```
mbolt35, line 2761: Instead of:
```
Read.Lock() // IsEmpty
Read.Unlock() // IsEmpty
Write.Lock() // If 
Write.Unlock() // If 
Write.Lock() // func
Write.Unlock() // func
```

Just put everything behind a single Write Lock.

pkg/kubecost/asset.go
```diff
@@ -2907,28 +2925,48 @@ func (asr *AssetSetRange) Append(that *AssetSet) {
...
+// Each invokes the given function for each AssetSet in the range
+func (asr *AssetSetRange) Each(f func(int, *AssetSet)) {
+	if asr == nil {
+		return
+	}
+
+	asr.RLock()
+	var assets = make([]*AssetSet, len(asr.assets))
+	copy(assets, asr.assets)
+	asr.RUnlock()
```
mbolt35, line 2950: While I'm not crazy about duplicating the collection, it at least provides collection safety. Since deep copy is out of the question due to performance, this is the next best thing. Otherwise, we are opening the door for the collection to be modified during the each iteration (potentially by the callback func itself).

pkg/kubecost/asset.go
```diff
@@ -2907,28 +2925,48 @@ func (asr *AssetSetRange) Append(that *AssetSet) {
...
 		return nil, fmt.Errorf("AssetSetRange: index out of range: %d", i)
 	}
 
-	asr.RLock()
-	defer asr.RUnlock()
 	return asr.assets[i], nil
 }
 
 func (asr *AssetSetRange) Length() int {
-	if asr == nil || asr.assets == nil {
```
mbolt35, line 2931: nil check on `asr.assets` outside of a lock

pkg/kubecost/asset.go
```diff
@@ -2996,9 +3034,10 @@ func (asr *AssetSetRange) InsertRange(that *AssetSetRange) error {
 
 // IsEmpty returns false if AssetSetRange contains a single AssetSet that is not empty
 func (asr *AssetSetRange) IsEmpty() bool {
-	if asr == nil || asr.Length() == 0 {
```
mbolt35, line 2999: asr.Length() call acquires and relinquishes RLock. Also, not needed, as a range loop will not iterate over 0 length collection (or nil).

pkg/kubecost/asset.go
```diff
@@ -3024,26 +3063,47 @@ type AssetSetRangeResponse struct {
...
+	}
+
+	startAs := asr.assets[0]
+	endAs := asr.assets[len(asr.assets)-1]
+	if startAs == nil || endAs == nil {
+		return NewWindow(nil, nil)
+	}
+
+	start := startAs.Start()
+	end := endAs.End()
```
mbolt35, line 3106: We were just YOLOing this before. Acquiring the read lock twice during each `Length()` call. Just need a single RLock here and direct access.

pkg/kubecost/asset.go
```diff
@@ -3053,17 +3113,22 @@ func (asr *AssetSetRange) Window() Window {
...
 			}
 			if a.Start().Before(start) {
 				start = a.Start()
 			}
-		})
-	})
+		}
+		as.RUnlock()
+	}
+	asr.RUnlock()
```
mbolt35, line 3131: Avoid `Each` if there exists access to the assets, since it now copies the collection. Additionally, we want to acquire read locks for the internal assets collection. While this change does add read locking for each Asset as well, I'm not sure it's required. Something to look into.

pkg/kubecost/asset.go
```diff
@@ -3096,14 +3166,57 @@ func (asr *AssetSetRange) End() (time.Time, error) {
 	return end, nil
 }
 
+// Each iterates over all AssetSets in the AssetSetRange and returns the start and end over
+// the entire range
+func (asr *AssetSetRange) StartAndEnd() (time.Time, time.Time, error) {
```
mbolt35, line 3171: The `Minutes()` function was calling `Start()` and `End()` back to back, which was requiring nested loops each call. There could also exist contention between each call. It made sense to combine these into a single loop, and add proper locking.

mbolt35:
Closed this and deleted branch. Data Structure Cleanup is underway here: https://github.com/opencost/opencost/pull/1354

PR #1229, opened 2022-06-01
Move files to use config path

dramich:
## What does this PR change?
* Hardcoded paths will now use the GetConfigPathWithDefault

* It shouldn't, the default is still the same and the helm chart still sets the previously hardcoded path. It also does not expose this env so a user can not set this through the helm chart. 
* 
## How was this PR tested?
* Running locally, building and running the image

* Added running locally docs


mbolt35 (approved):
LGTM

PR #1239, opened 2022-06-03
UI build fails with "Browser scripts cannot have imports or exports"

srpomeroy:
## What does this PR change?
* Adds module type to js import. Ref: https://github.com/parcel-bundler/parcel/discussions/6490

* No

* It won't

* No

## How was this PR tested?
* Build is now successful

* no

* Unsure if OSS UI is included in regular release cycle

### DIFF
--- a/ui/src/index.html
+++ b/ui/src/index.html
@@ -7,6 +7,6 @@
     </head>
     <body>
         <div id="app" class="page-container"></div>
-        <script src="./app.js"></script>
+        <script src="./app.js" type="module"></script>
     </body>
 </html>


Adam-Stack-PM:
@teevans is OSS UI included in the regular release cycle?  See @srpomeroy PR note.

dwbrown2:
@srpomeroy did you encounter with this install path or another? https://github.com/kubecost/opencost/blob/master/deploying-as-a-pod.md

srpomeroy:
@dwbrown2 its from the simple UI deployment here: https://github.com/kubecost/opencost/blob/develop/ui/README.md

the parcel step.

srpomeroy:
I meant to include this in the initial post.
image

nealormsbee (approved):
LGTM!

nealormsbee:
The force push is because I thought that, in addition to the `type="module"` update, the API query path needed to be prefixed with `/model`.

So I made that update and was going to approve, but then realized that the prefix is only required because I'm running this against a closed-source backend. The FP was removing the commit that I had added to make that adjustment.

dwbrown2:
Not sure I fully follow, more work needed here or should be all set?

nealormsbee:
We're all set! I was just explaining the force push.

srpomeroy:
Still not clear if this is part of any regular release cadence. Do I just merge and let @teevans figure it out?

nealormsbee:
You can go ahead and merge this. If it needs to go into the release, we'll cherry-pick the commit into the `v1.94` branch. I think that would be the ideal, but is also unlikely to make a big impact.

We build images for the release, including an opencost image, but that image does not run the UI. The UI is for users grabbing the repo and running it locally. It would be *ideal* for the UI code at the `1.94.0` git tag not to be broken, but I strongly suspect users that want to run the UI are just pulling `develop` and running that.

Final call up to @teevans. I can create the cherrypick branch for this if there's time and inclination to get this into 1.94.0

srpomeroy:
Thanks @nealormsbee !

PR #1248, opened 2022-06-09
Created diff for assets with test cases

nickcurie:
## What does this PR change?
* Adds diffing functionality to compare asset sets + a handful of test cases

* No

## How was this PR tested?
* Just unit tests at the moment

* Unsure

## Notes
* I plan on doing some additional changes to this functionality as well as testing a bit more broadly


AjayTripathy:
Congrats on the first PR @nickcurie !! Next up, let's try and put this on an API endpoint! I propose a really simple design where we just have an API that looks like
```[
  {nodeName1: 
     { 
      hourlyCost: 0.10
      state: "added|removed"
    }
  }
]
```

What do you think?

nikovacevic:
@AjayTripathy we're not there yet. @nickcurie please continue to just follow the design document.

pkg/kubecost/asset.go
```diff
@@ -2838,6 +2838,31 @@ func (as *AssetSet) accumulate(that *AssetSet) (*AssetSet, error) {
 	return acc, nil
 }
 
+type Diff struct {
+	Asset Asset
```
nikovacevic: What do you think about making this generic (`Diff[T]` with `entity T`) or general (`Diff` with `name string`)? Anything else we should include here, like change in cost or window before/after?

Be mindful that we may want to run diffs between large data sets, so there will be a tradeoff between having the entire Asset (or T) and having only the relevant fields.

pkg/kubecost/asset.go
```diff
@@ -2838,6 +2838,31 @@ func (as *AssetSet) accumulate(that *AssetSet) (*AssetSet, error) {
...
+	ChangedType string
+}
+
+func DiffAsset(before, after *AssetSet) []Diff{
+	changedAssets := []Diff{}
+
+	for assetKey1, asset1 := range before.assets {
+		if _, ok := after.assets[assetKey1]; !ok {
+			d := Diff{asset1, "removed"}
```
nikovacevic: What do you think about codifying these strings as constants?

pkg/kubecost/asset.go
```diff
@@ -2838,6 +2838,31 @@ func (as *AssetSet) accumulate(that *AssetSet) (*AssetSet, error) {
 	return acc, nil
 }
 
+type Diff struct {
+	Asset Asset
+	ChangedType string
+}
+
+func DiffAsset(before, after *AssetSet) []Diff{
```
nikovacevic: Are you familiar with the concept of a "receiver" in Go? If not, read up about it and consider whether this should be a simple function like this, or something where an `AssetSet` is a receiver, with the function `Diff(that *AssetSet)`.

pkg/kubecost/diff_test.go
```diff
@@ -0,0 +1,72 @@
...
+	window1 := NewWindow(&start, &end)
+
+	node1 := NewNode("node1", "cluster1", "123abc", start, end, window1)
+	node2 := NewNode("node2", "cluster1", "123abc", start, end, window1)
+	node3 := NewNode("node3", "cluster1", "123abc", start, end, window1)
+	node4 := NewNode("node4", "cluster1", "123abc", start, end, window1)
+	disk1 := NewDisk("disk1", "cluster1", "123abc", start, end, window1)
+	disk2 := NewDisk("disk2", "cluster1", "123abc", start, end, window1)
+
+	cases := map[string]struct {
```
nikovacevic, line 24: These tests are looking great!

pkg/kubecost/asset.go
```diff
@@ -2838,6 +2838,31 @@ func (as *AssetSet) accumulate(that *AssetSet) (*AssetSet, error) {
 	return acc, nil
 }
 
+type Diff struct {
```
nikovacevic: Whenever you're ready, let's make sure all the public structs and functions have idiomatic comments.

See https://go.dev/doc/effective_go#commentary and https://go.dev/doc/comment

nikovacevic (commented):
Good work so far! Respond to the feedback, then let me know if you have any questions.

nikovacevic:
@AjayTripathy design doc is here: https://docs.google.com/document/d/1TW5b9NmlEcucqcLh7ChQp1SxInU2K0NXXVOKm7gcrkI/edit#heading=h.xovpz6caff1j

If course, please feel welcome to propose fields!

pkg/kubecost/asset.go
nickcurie: This is definitely a good idea, I'll get this added.

pkg/kubecost/asset.go
nickcurie: I like the idea of making the Diff struct take in generic types. If performance becomes a concern here this can certainly be changed, but I would like to get an idea of how this feature is going to be used.

Regarding a change in cost, I plan to add that functionality in future iterations.

pkg/kubecost/asset.go
nickcurie: @michaelmdresser and I thought it would make more sense for this function to be simple since we can ensure that the first argument is the earlier AssetSet (or AllocationSet when that is implemented) and the second is the later one.

nikovacevic: Sounds good!

nikovacevic (approved):
Looks good! Feel free to merge.

Adam-Stack-PM:
@nickcurie Great work.

PR #1250, opened 2022-06-10
Lexer and Parser for v2 of the Allocation filter language

michaelmdresser:
## What does this PR change?

- Changes `pkg/util/filterutil` to `pkg/util/allocationfilterutil` -- I think this is more in line with our package naming strategy because the functions provided by the package are specific to Allocations
- Introduces `pkg/util/allocationfilterutil/v2` which supports parsing v2 of the Allocation "filter language" into the `kubecost.AllocationFilter` type. The biggest addition when compared with the current language (query params like `filterNamespaces=kubecost`) is support for `!=`. Check out example queries in `parser_test.go`. See also [the design doc](https://docs.google.com/document/d/1HKkp2bv3mnvfQoBZlpHjfZwQ0FzDLOHKpnwV9gQ_KgU/edit?pli=1#heading=h.cs9gvm9cb74e). The lexer and parser are adapted from Chapters 4 and 6 of Robert Nystrom's _Crafting Interpreters_. My copy is Copyright 2021. The book is also available online at http://craftinginterpreters.com/contents.html if you would like to check out the source and find explanations for why this code is the way it is.
- Added a new `kubecost.FilterOp`: `notcontains`. This let's us support "inequality" filter queries for services, which is an array/slice property.

- [x] There are a couple of TODOs in the code right now that I hope to remove before merge. I'll try to add a comment to each to explain why they are present and ask for some ideas.

### Remaining work for future PRs
- We have to wire up the Allocation ETL with the new struct, removing the `AllocationFilterFunc` type. This includes switching all query handlers to use `pkg/util/allocationfilterutil` (v1) to parse query params.
- We need to decide how to introduce this new filter language in the Allocation APIs. Should it be a new query parameter? A "v2" of the API itself? Should there be a deprecation plan for the old filters?
- (longer timeline) Once the filter language is available via API, we can add support on the frontend for `!=` queries. How to best expose the functionality remains to be seen!
- (optional, depending on product requirements) We can add support for "wildcard" queries in the language. I propose something like `namespace~:"kube"` for substring matching, but I think there are several options -- we could support prefixes, suffixes, substrings, or any combination of the above. The current API supports prefix matching with `filterNamespaces=kube*`.
- (optional, depending on customer needs) We could add support for even more complex queries in the future. Right now, it is impossible to express a query like `namespace:"x"|label[app]="foo"`. There is a note in the code of this PR about how we might add support for queries like that.

### Other notes
- If this language is successful, we can implement a sibling language for Asset filtering

* Builds on the foundational work in https://github.com/kubecost/opencost/pull/1210

N/A, requires other work

## How was this PR tested?
Unit tests

Not yet. As soon as we expose this via an API, we should document the grammar, example queries, limitations, etc. This PR can be used for reference.

### DIFF
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -138,7 +138,7 @@ type AllocationFilterAnd struct {
 }
 
 func (af AllocationFilterAnd) String() string {
-	s := "(or"
+	s := "(and"
 	for _, f := range af.Filters {
 		s += fmt.Sprintf(" %s", f)
 	}
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -1,6 +1,7 @@
 package kubecost
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/kubecost/opencost/pkg/log"
@@ -80,6 +81,8 @@ type AllocationFilter interface {
 	// Matches is the canonical in-Go function for determing if an Allocation
 	// matches a filter.
 	Matches(a *Allocation) bool
+
+	String() string
 }
 
 // AllocationFilterCondition is the lowest-level type of filter. It represents
@@ -100,16 +103,44 @@ type AllocationFilterCondition struct {
 	Value string
 }
 
+func (afc AllocationFilterCondition) String() string {
+	if afc.Key == "" {
+		return fmt.Sprintf(`(%s %s "%s")`, afc.Op, afc.Field, afc.Value)
+	}
+
+	return fmt.Sprintf(`(%s %s[%s] "%s")`, afc.Op, afc.Field, afc.Key, afc.Value)
+}
+
 // AllocationFilterOr is a set of filters that should be evaluated as a logical
 // OR.
 type AllocationFilterOr struct {
 	Filters []AllocationFilter
 }
 
+func (af AllocationFilterOr) String() string {
+	s := "(or"
+	for _, f := range af.Filters {
+		s += fmt.Sprintf(" %s", f)
+	}
+
+	s += ")"
+	return s
+}
+
 // AllocationFilterOr is a set of filters that should be evaluated as a logical
 // AND.
 type AllocationFilterAnd struct {
 	Filters []AllocationFilter
+}
+
+func (af AllocationFilterAnd) String() string {
+	s := "(or"
+	for _, f := range af.Filters {
+		s += fmt.Sprintf(" %s", f)
+	}
+
+	s += ")"
+	return s
 }
 
 func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -52,6 +52,10 @@ const (
 	// ["a", "b", "c"] FilterContains "a" = true
 	FilterContains = "contains"
 
+	// FilterNotContains is an array/slice non-membership operator
+	// ["a", "b", "c"] FilterNotContains "d" = true
+	FilterNotContains = "notcontains"
+
 	// FilterStartsWith matches strings with the given prefix.
 	// "kube-system" StartsWith "kube" = true
 	//
@@ -243,6 +247,26 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
 			}
 		} else {
 			log.Warnf("Allocation Filter: invalid 'contains' call for non-list filter value")
+		}
+	case FilterNotContains:
+		if stringSlice, ok := valueToCompare.([]string); ok {
+			// services!:"__unallocated__" should match
+			// len(a.Properties.Services) > 0
+			//
+			// TODO: is this true?
+			if filter.Value == UnallocatedSuffix {
+				return len(stringSlice) > 0
+			}
+
+			for _, s := range stringSlice {
+				if s == filter.Value {
+					return false
+				}
+			}
+
+			return true
+		} else {
+			log.Warnf("Allocation Filter: invalid 'notcontains' call for non-list filter value")
 		}
 	case FilterStartsWith:
 		if toCompareMissing {
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -208,12 +208,9 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
 
 	switch filter.Op {
 	case FilterEquals:
-		if toCompareMissing {
-			return false
-		}
-
 		// namespace:"__unallocated__" should match a.Properties.Namespace = ""
-		if valueToCompare == "" {
+		// label[app]:"__unallocated__" should match _, ok := Labels[app]; !ok
+		if toCompareMissing || valueToCompare == "" {
 			return filter.Value == UnallocatedSuffix
 		}
 
@@ -221,14 +218,18 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
 			return true
 		}
 	case FilterNotEquals:
-		if toCompareMissing {
-			return true
-		}
-
 		// namespace!:"__unallocated__" should match
 		// a.Properties.Namespace != ""
+		// label[app]!:"__unallocated__" should match _, ok := Labels[app]; ok
 		if filter.Value == UnallocatedSuffix {
+			if toCompareMissing {
+				return false
+			}
 			return valueToCompare != ""
+		}
+
+		if toCompareMissing {
+			return true
 		}
 
 		if valueToCompare != filter.Value {
--- a/pkg/util/allocationfilterutil/v2/lexer.go
+++ b/pkg/util/allocationfilterutil/v2/lexer.go
@@ -119,6 +119,9 @@ func (s scanner) atEnd() bool {
 
 // advance returns a byte because we only accept ASCII, which has to fit in a
 // byte
+//
+// TODO: If we add unicode support, advance() will probably have to return a
+// rune.
 func (s *scanner) advance() byte {
 	b := s.source[s.nextByte]
 	s.nextByte += 1
@@ -158,9 +161,6 @@ func (s *scanner) peek() byte {
 }
 
 func (s *scanner) scanToken() {
-	// TODO: DON'T ACCEPT NON-ASCII
-	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-
 	c := s.advance()
 	switch c {
 	case ':':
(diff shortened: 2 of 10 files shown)


pkg/kubecost/allocationfilter.go
```diff
@@ -213,6 +248,26 @@ func (filter AllocationFilterCondition) Matches(a *Allocation) bool {
...
 		}
+	case FilterNotContains:
+		if stringSlice, ok := valueToCompare.([]string); ok {
+			// services!:"__unallocated__" should match
+			// len(a.Properties.Services) > 0
+			//
+			// TODO: is this true?
+			if filter.Value == UnallocatedSuffix {
+				return len(stringSlice) > 0
+			}
```
michaelmdresser, line 260: I'm a little wary of `__unallocated__` queries and I don't know what the right answer is. What Allocations should `services!:"__unallocated__"` match? I will update the test cases accordingly once we come to consensus.

pkg/util/allocationfilterutil/v2/lexer.go
```diff
@@ -0,0 +1,272 @@
...
+	identifier   // K8s valid name + sanitized Prom: 'app', 'abc_label'
+
+	eof
+)
+
+// These maps serve a dual purpose. (1) to help the lexer identify special
+// strings that should become filterField1/2 instead of identifiers and (2) to
+// help the parser convert tokens into AllocationFilterConditions.
+//
+// TODO: Should the lexer _also_ attach the FilterField to help the parser out?
```
michaelmdresser: Essentially, the lexer could attach the appropriate `FilterField` to the lexed `token` (because it knows at lex time) instead of making the parser do an extra map lookup. I leaned away from this idea because it would complicate the `token` type, but I'm open to ideas!

pkg/util/allocationfilterutil/v2/lexer.go
```diff
@@ -0,0 +1,272 @@
...
+func (s *scanner) peek() byte {
+	if s.atEnd() {
+		return 0
+	}
+	return s.source[s.nextByte]
+}
+
+func (s *scanner) scanToken() {
+	// TODO: DON'T ACCEPT NON-ASCII
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
```
michaelmdresser: I'm almost certain that we only need to support ASCII characters in our filters because of K8s naming conventions, but it might be possible that Prometheus or cloud provider data will be encoded in UTF-8.

If we need to support UTF-8, I'll update the lexer appropriately. If we don't, I think I should add some guard that causes the lexer to fail if a UTF-8 character is found.

pkg/util/allocationfilterutil/v2/lexer.go
```diff
@@ -0,0 +1,272 @@
...
+	case '+':
+		s.addToken(plus)
+	case '!':
+		if s.match(':') {
+			s.addToken(bangColon)
+		} else {
+			s.errors = append(s.errors, fmt.Errorf("Position %d: Unexpected '!'", s.nextByte-1))
+		}
+	// strings
+	// TODO: reject whitespace chars
```
michaelmdresser: In a sense, this question is similar to the UTF-8 support question. We could:
1. Remove all whitespace pre-lex
2. Ignore all whitespace that isn't in a string
3. Allow whitespace to be passed in, but error out because it isn't supported

Does anyone have thoughts? I lean towards 2.

pkg/util/allocationfilterutil/v2/lexer.go
```diff
@@ -0,0 +1,272 @@
...
+		s.keyedAccess()
+	default:
+		// identifiers
+		// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
+		// We can keep it simple and not _force_ the first character to be a non-number
+		// because we don't need numbers in this language. If we need to extend the
+		// language to support numbers, this has to become just isAlpha() and then
+		// s.identifier() will keep using isAlphanumeric() in its main loop.
+		//
+		// TODO: does this match all character we support for cluster IDs?
```
michaelmdresser: This is just a more targeted version of the UTF-8 question. I hope we can address in that thread.

pkg/util/allocationfilterutil/v2/lexer.go
```diff
@@ -0,0 +1,272 @@
...
+		if isIdentifierChar(c) {
+			s.identifier()
+			break
+		}
+
+		// TODO: how to phrase for unicode?
+		// extra handling to check unicode? https://stackoverflow.com/questions/53069040/checking-a-string-contains-only-ascii-characters
```
michaelmdresser: Same thing, see above.

pkg/util/allocationfilterutil/v2/lexer_test.go
```diff
@@ -0,0 +1,89 @@
...
+			name:     "identifier pure alpha",
+			input:    "abc",
+			expected: []token{{kind: identifier, s: "abc"}, {kind: eof}},
+		},
+		{
+			name:     "label access",
+			input:    "app[kubecost]",
+			expected: []token{{kind: identifier, s: "app"}, {kind: keyedAccess, s: "kubecost"}, {kind: eof}},
+		},
+		// TODO: more cases
```
michaelmdresser: I will work on adding more complex test cases. Trying to open the PR early to get the review gears churning :smile:

pkg/util/allocationfilterutil/v2/parser_test.go
```diff
@@ -0,0 +1,298 @@
...
+		Properties: &props,
+	}
+
+	a.Name = a.Properties.String()
+	return a
+}
+
+func TestParse(t *testing.T) {
+	// TODO: unallocated cases
+	cases := []struct {
```
michaelmdresser: Once we make a decision about `services!:"__unallocated__"` I will add more test cases.

pkg/kubecost/allocationfilter.go
nikovacevic: What do you think about just matching the existing behavior?

```go
// special case: allocations that cannot get services or have no services
// pass the filter for `filterValue == "__unallocated__"`
if len(serviceList) == 0 {
	return (filterValue == "__unallocated__")
}
```

pkg/util/allocationfilterutil/v2/lexer.go
nikovacevic: Trust you to make the call here, but my gut says that it's fine how it is. No need to make a tradeoff that increases complexity unless we're afraid this will be too slow, which seems unlikely. Maybe others will have opinions.

pkg/util/allocationfilterutil/v2/lexer.go
nikovacevic: I think it's possible that CUR data will have whitespace within strings. So probably 2.

pkg/util/allocationfilterutil/v2/lexer.go
nikovacevic: How difficult would it be to support UTF-8? Probably okay for > 99% of use cases with just ASCII.

Adam-Stack-PM:
@michaelmdresser This is great. Thanks for moving the ball forward on filters.

pkg/util/allocationfilterutil/v2/parser_test.go
```diff
@@ -0,0 +1,298 @@
...
+						Field: kubecost.FilterLabel,
+						Key:   "foo",
+						Op:    kubecost.FilterEquals,
+						Value: "bar",
+					},
+				}},
+			}},
+		},
+		{
+			input: `namespace:"kubecost"+label[app]:"cost_analyzer"+annotation[a1]:"b2"+cluster:"cluster-one"+node!:"node-123","node-456"+controllerName:"kubecost-cost-analyzer","kubecost-prometheus-server"+controllerKind!:"daemonset","statefulset","job"+container!:"123-abc_foo"+pod!:"aaaaaaaaaaaaaaaaaaaaaaaaa"+services!:"abc123"`,
```
nikovacevic: :scream:

pkg/util/allocationfilterutil/v2/parser_test.go
```diff
@@ -0,0 +1,298 @@
...
+				allocGenerator(kubecost.AllocationProperties{Namespace: "kube-system"}),
+			},
+			shouldNotMatch: []kubecost.Allocation{
+				allocGenerator(kubecost.AllocationProperties{Namespace: "abc"}),
+			},
+		},
+		{
+			input: `label[app_abc]:"cost_analyzer"`,
+			expected: kubecost.AllocationFilterAnd{[]kubecost.AllocationFilter{
+				kubecost.AllocationFilterOr{[]kubecost.AllocationFilter{
```
nikovacevic, line 120: Why is it required to have these "Or" steps between? Is it to do with how we accept comma-separated values and implicitly call that an "or"?

nikovacevic (commented):
This is looking good to me so far. Great work. Besides answering your specific questions, how can I be helpful here?

pkg/util/allocationfilterutil/v2/lexer.go
michaelmdresser: Updated to support case 2.

pkg/kubecost/allocationfilter.go
michaelmdresser: As far as I know, we have never supported the idea of inequality with `"__unallocated__"` before. In the current implementation of `AllocationFilterCondition.Matches()`, we do something that looks like the code referenced in this review comment. See here:

https://github.com/kubecost/opencost/blob/040df8955b01af129898b9c980415c9553cd1ec2/pkg/kubecost/allocationfilter.go#L193-L197

The idea is that we have to change the `if` condition for `!=` queries because if its like this:
 ```go
if len(serviceList) > 0 { 
    return filterValue != "__unallocated__"
}
```

you'll incorrectly enter that block for non-unallocated queries.

This might be a little confusing, it's hard to express this properly. Let me know if this makes sense. Ideally we get this straight so the behavior is clear to others in the future.

pkg/util/allocationfilterutil/v2/parser_test.go
michaelmdresser: Yep, that's right! Between each `+` is an AND and between each `,` is an OR, so the parser just creates an `AND` for each top-level condition and an `OR` within each top-level condition. We could add something like `func Flattened() AllocationFilter` to the `AllocationFilter` interface that would eliminate unnecessary AND and OR blocks (single-element or empty). Do you think that would be valuable?

pkg/util/allocationfilterutil/v2/parser_test.go
michaelmdresser: Indeed! Now that we have whitespace support, I've reformatted this test to be much more readable. I'm a bit sad, because I feel like I've reinvented the wheel a bit -- it looks a lot like YAML, but of course YAML doesn't let you express an operator by changing `:` to `!:` or let you do access (I think?) with `label[key]`.

pkg/util/allocationfilterutil/v2/lexer.go
michaelmdresser: In the interest of making forward progress and keeping the size of this PR smaller, I think we shouldn't add UTF-8 support in this PR. I've updated and moved some comments to smooth that process if we choose to add support later.

pkg/kubecost/allocationfilter.go
nikovacevic: Resolved out of band -- we agree that the logic looks right as it stands.

nikovacevic (approved):

PR #1257, opened 2022-06-14
get reconciliation map

Sean-Holcomb:
## What does this PR change?
* Add method to AssetSet which builds a map of the assets by providerID and Category. If there are duplicated assets then us the one that has previously been reconciled or failing that use the one with a higher cost to minimize cost duplication. 

* https://github.com/kubecost/kubecost-cost-model/pull/829

* 

* Closes ...

## How was this PR tested?
* see https://github.com/kubecost/kubecost-cost-model/pull/829

* 

*

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -2652,10 +2652,19 @@ func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
 	return nil, false, fmt.Errorf("Asset not found to match %s", query)
 }
 
+// ReconciliationMatchMap returns a map of the calling AssetSet's Assets, by provider id and category. This data structure
+// allows for reconciliation matching to be done in constant time and prevents duplicate reconciliation.
 func (as *AssetSet) ReconciliationMatchMap() map[string]map[string]Asset {
 	matchMap := make(map[string]map[string]Asset)
 
+	if as == nil {
+		return matchMap
+	}
+
 	for _, asset := range as.assets {
+		if asset == nil {
+			continue
+		}
 		props := asset.Properties()
 		// Ignore cloud assets and assets that cannot be matched when looking for reconciliation matches
 		if props == nil || props.ProviderID == "" || asset.Type() == CloudAssetType {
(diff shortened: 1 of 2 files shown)


pkg/kubecost/asset.go
```diff
@@ -2652,6 +2652,40 @@ func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
 	return nil, false, fmt.Errorf("Asset not found to match %s", query)
 }
 
+func (as *AssetSet) ReconciliationMatchMap() map[string]map[string]Asset {
+	matchMap := make(map[string]map[string]Asset)
```
michaelmdresser, line 2658: A comment on this method to explain the structure of the return type would be very helpful for future users. https://go.dev/doc/comment

michaelmdresser: Also, we should probably have a `nil` guard on this function, in case `as == nil`.

pkg/kubecost/asset.go
```diff
@@ -2652,6 +2652,40 @@ func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
...
+			log.DedupedWarningf(5, "duplicate asset found when reconciling for %s", props.ProviderID)
+			// if one asset already has adjustment use that one
+			if duplicateAsset.Adjustment() == 0 && asset.Adjustment() != 0 {
+				matchMap[props.ProviderID][props.Category] = asset
+			} else if duplicateAsset.Adjustment() != 0 && asset.Adjustment() == 0 {
+				matchMap[props.ProviderID][props.Category] = duplicateAsset
+				// otherwise use the one with the higher cost
+			} else if duplicateAsset.TotalCost()-duplicateAsset.Adjustment() < asset.TotalCost()-asset.Adjustment() {
+				matchMap[props.ProviderID][props.Category] = asset
+			}
```
michaelmdresser: It's not intuitively obvious to me why we're checking the higher cost before adjustment, rather than after adjustment. Would you explain a little further in a comment?

michaelmdresser (commented):
Looks pretty good overall. This seems very unit-testable!

pkg/kubecost/asset.go
Sean-Holcomb: It covers an edge case that this PR should prevent from happening anyway. Which is that both assets are already reconciled. I will just remove it.

michaelmdresser (approved):
Thanks for addressing feedback!

PR #1260, opened 2022-06-15
Add endpoint for install information

dramich:
## What does this PR change?
* Adds a new endpoint `/installInfo` to produce some information on the kubecost installation along with some cluster info. 
Example output:

```
{
    "containers": [
        {
            "containerName": "cost-analyzer-frontend",
            "image": "gcr.io/kubecost1/frontend:prod-1.93.2",
            "imageID": "gcr.io/kubecost1/frontend@sha256:3e79e019a0fee917bccc81d020451f97f0c914b5222ec6f1a859502f771baee2",
            "startTime": "2022-06-10 13:44:16 -0600 MDT",
            "restarts": 0
        },
        {
            "containerName": "cost-model",
            "image": "gcr.io/kubecost1/cost-model:prod-1.93.2",
            "imageID": "gcr.io/kubecost1/cost-model@sha256:0a17c93f808745c9ae8bad15066b7bf6986a55fadc3effb5a49495f29560ab03",
            "startTime": "2022-06-10 13:44:16 -0600 MDT",
            "restarts": 0
        }
    ],
    "clusterInfo": {
        "nodeCount": "1",
        "podCount": "12"
    },
    "version": "dev (HEAD)"
}
```

## How was this PR tested?
* Run costmodel and ping the endpoint

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -21,6 +21,7 @@ import (
 	"github.com/kubecost/opencost/pkg/util/httputil"
 	"github.com/kubecost/opencost/pkg/util/timeutil"
 	"github.com/kubecost/opencost/pkg/util/watcher"
+	"github.com/kubecost/opencost/pkg/version"
 	"github.com/microcosm-cc/bluemonday"
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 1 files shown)


dwbrown2:
Nice, Dan!! Thanks for this contribution. 

Would love to get a unique build ID as part of this one day soon. Think it's worth opening a separate issue to track?

dramich:
> Would love to get a unique build ID as part of this one day soon. Think it's worth opening a separate issue to track?

Yes, the binary currently has no info in it, we should be adding version tag, commit hash and even build time.

pkg/costmodel/router.go
```diff
@@ -1156,6 +1156,66 @@ func (a *Accesses) GetInstallNamespace(w http.ResponseWriter, r *http.Request, _
...
+	Image         string `json:"image"`
+	StartTime     string `json:"startTime"`
+	Restarts      int32  `json:"restarts"`
+}
+
+func (a *Accesses) GetInstallInfo(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+
+	pods, err := a.KubeClientSet.CoreV1().Pods(env.GetKubecostNamespace()).List(context.Background(), metav1.ListOptions{
```
mbolt35, line 1178: Do we need to filter on pod status as well? Albeit rare, I've run into issues where there can be multiple cost-model pods (where some have a status of `Evicted`). For example:

To automatically run `kubectl logs` against the cost-model pod: 
```bash
$ kubectl get pods -n kubecost -l app=cost-analyzer -o jsonpath='{.items[0].metadata.name}' | xargs -I {} kubectl logs -f -n kubecost -c cost-model {}
```

where the result of: 
```bash
kubectl get pods -n kubecost -l app=cost-analyzer -o jsonpath='{.items[0].metadata.name}'
```
gives me one of the `Evicted` pods. 

It's not obvious whether or not `Limit: 1` via the API is different than just taking the first result.

dramich: Good question, even if I get back the evicted pod the info I am pulling from the containers and pods should still be relevant enough for the endpoint (high level install info). If we only go after active pods that could cause a problem if the pods haven't passed the readiness probe because they are new or are actually having an issue.

dramich: And if we get to the point where the container is dying then the UI will be useless as well 🤷

michaelmdresser: I know I do slightly funky things in my dev cluster, but I think I have an example for what Bolt is talking about. I have my Kubecost workloads scheduled on spot nodes. When the spot nodes die/are restarted, the `kubecost-cost-analyzer` pod ends in a "Completed" state. For example, this is my cluster right now:

```
→ k get pod -n kubecost -l app=cost-analyzer 
NAME                                      READY   STATUS      RESTARTS   AGE
kubecost-cost-analyzer-65c7d58669-kc4hz   0/2     Completed   0          16h
kubecost-cost-analyzer-6cc89cdc5-h9wn7    2/2     Running     0          17m
```

The 17m pod has been updated while I do dev work, which I normally do with the following:
```
kubectl set image deployment/kubecost-cost-analyzer cost-model=FOO --namespace kubecost
```

And this is what happens when I get the container images with some JSONPath. Note the differing container images:
```
→ k get pod -n kubecost -l app=cost-analyzer -o jsonpath='{range .items[*]}{.metadata.name}{range .spec.containers[*]}{"\n"}{"\t"}{.name}{"\t"}{.image}{"\n"}'

kubecost-cost-analyzer-65c7d58669-kc4hz
	cost-model	michaelkubecost/opencost:test-01

	cost-analyzer-frontend	gcr.io/kubecost1/frontend:prod-1.94.3

		
kubecost-cost-analyzer-6cc89cdc5-h9wn7
	cost-model	michaelkubecost/cost-model:log-01

	cost-analyzer-frontend	gcr.io/kubecost1/frontend:prod-1.94.3
```

pkg/costmodel/router.go
```diff
@@ -1156,6 +1156,66 @@ func (a *Accesses) GetInstallNamespace(w http.ResponseWriter, r *http.Request, _
...
 }
 
+type InstallInfo struct {
+	Containers  []ContainerInfo   `json:"containers"`
+	ClusterInfo map[string]string `json:"clusterInfo"`
+}
+
+type ContainerInfo struct {
+	ContainerName string `json:"containerName"`
+	Image         string `json:"image"`
```
michaelmdresser, line 1168: What do you think about including the `imageID` from [ContainerStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#containerstatus-v1-core)? I think it could be helpful in case we're looking at an image tagged `latest` and want to find the exact image.

Example live image and imageID in my cluster:
```
image: gcr.io/kubecost1/frontend:prod-1.94.3
imageID: gcr.io/kubecost1/frontend@sha256:2b39858fb8599f6789a3a565ab6648b7d5a3c1b333faa95ec15d5c8a510277da
```

pkg/costmodel/router.go
dramich: Added `FieldSelector: "status.phase=Running",`

pkg/costmodel/router.go
dramich: When looking at a pods status it will not say `latest` for the image, it will say the actual image including the tag tag that is being ran. ImageID is really useful when people push over the same image which we don't but I can add it anyways.

michaelmdresser (approved):

mbolt35 (approved):
LGTM

PR #1264, opened 2022-06-16
Add a version and commit to the binary

dramich:
## What does this PR change?
* Embed a version and git commit into the built binary. 

To see this in action
```
docker build -t cost-model --build-arg version=testing --build-arg commit=asdasd .
```

Which would give you a log like `2022-06-16T22:11:10.794033216Z INF Starting cost-model version testing (asdasd)`

* Kinda... https://github.com/kubecost/opencost/pull/1260 can use this

* Exposes more info that support can use to support them

## How was this PR tested?
* Build the image and output the results of `FriendlyVersion`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,9 @@ COPY go.sum .
 # golang:latest image does not contain GCC, while the AMD64 version does.
 ARG CGO_ENABLED=0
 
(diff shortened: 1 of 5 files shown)


mbolt35 (approved):
LGTM!

PR #1266, opened 2022-06-17
Fix gpu bug

Sean-Holcomb:
## What does this PR change?
* Fixes bug in the total store that can cause large idles in nodes with GPU cost. From what I can tell it does not affect nodes without GPU. This bug is caused by a swapping of direction on a subtraction when calculating GPU adjustment. Additionally GPU does not have discount applied to it anywhere else in the cost model so I removed it here.

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


nikovacevic (approved):

Adam-Stack-PM:
@Sean-Holcomb Can I add the v1.95 tag for tracking?

PR #1272, opened 2022-06-22
Export mapper.GoMap and add constructor

michaelmdresser:
## What does this PR change?
This is helpful for unit tests that need to construct query params. Using
a standard map[string]string is ideal for such situations.

Usage: `mapper.NewMapper(mapper.NewGoMap(map[string]string{"a": "b"}))`

* Forthcoming KCM PR that needs this change (or something like it)

## How was this PR tested?
Usage in unit tests of the forthcoming KCM PR.


mbolt35 (approved):
Looks good to me!

PR #1273, opened 2022-06-22
Add AllocationFilter.Flattened() for simplifying AllocationFilter expressions

michaelmdresser:
## What does this PR change?
Adds `Flattened() AllocationFilter` to the `AllocationFilter` interface. This enables AllocationFilters to be converted into a minimal representation (`nil` if no filter). This helps with situations where a parser might construct zero- or single-element AND or OR expressions that can be simplified, improving readability and making usage simpler.

While we have no `func (af AllocationFilter) Empty() bool` , an equal expression can be made with `af == nil || af.Flattened() == nil`.

We may wish to force the parser introduced in https://github.com/kubecost/opencost/pull/1250 to call `Flattened()` as a final step.

* A forthcoming OC PR will depend on this PR.
* Partially inspired by this discussion: https://github.com/kubecost/opencost/pull/1250#discussion_r902753694

## How was this PR tested?
* Unit tests in this PR
* Usage in forthcoming OC and KCM PRs

### DIFF
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -149,7 +149,8 @@ func (af AllocationFilterOr) String() string {
 	return s
 }
 
-func flatten(filters []AllocationFilter) []AllocationFilter {
+// flattened returns a new slice of filters after flattening.
+func flattened(filters []AllocationFilter) []AllocationFilter {
 	var flattenedFilters []AllocationFilter
 	for _, innerFilter := range filters {
 		if innerFilter == nil {
@@ -172,7 +173,7 @@ func flatten(filters []AllocationFilter) []AllocationFilter {
 // - the inner filter if filter contains one filter
 // - an equivalent AllocationFilterOr if filter contains more than one filter
 func (filter AllocationFilterOr) Flattened() AllocationFilter {
-	flattenedFilters := flatten(filter.Filters)
+	flattenedFilters := flattened(filter.Filters)
 	if len(flattenedFilters) == 0 {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/kubecost/allocationfilter.go
```diff
@@ -80,6 +80,19 @@ type AllocationFilter interface {
...
+	// conditions.
+	//
+	// It returns nil if the filter is filtering nothing.
+	//
+	// Example:
+	// (and (or (namespaceequals "kubecost")) (or)) ->
+	// (namespaceequals "kubecost")
+	//
+	// (and (or)) -> nil
+	Flattened() AllocationFilter
```
nikovacevic, line 100: If this returns a new `AllocationFilter` can we call the function `Flatten()`?

pkg/kubecost/allocationfilter.go
```diff
@@ -100,18 +113,79 @@ type AllocationFilterCondition struct {
...
+	return filter
+}
+
 // AllocationFilterOr is a set of filters that should be evaluated as a logical
 // OR.
 type AllocationFilterOr struct {
 	Filters []AllocationFilter
 }
 
+func flatten(filters []AllocationFilter) []AllocationFilter {
```
nikovacevic: Why `flatten` here, but `Flattened` as the public function?

nikovacevic (commented):
Some early feedback, but will take a second pass later!

pkg/kubecost/allocationfilter.go
michaelmdresser: I picked `Flattened()` from Python influence -- `[].sort()` modifies in place and `sorted([])` returns a new list. I'm happy to use either name; does go have any idiomatic naming for functions like this?

pkg/kubecost/allocationfilter.go
michaelmdresser: Updated to be consistent. Whatever the results of our above discussion, I'll change appropriately.

pkg/kubecost/allocationfilter.go
nikovacevic: I don't know that there is a specific convention in Go around verb tense. If you like `Flattened`, I won't argue! In general, we've tended to use present-tense, but it really doesn't matter too much.

pkg/kubecost/allocationfilter.go
nikovacevic: Sure, no problem either way.

nikovacevic (approved):

PR #1274, opened 2022-06-22
Replace FilterFunc core filters with AllocationFilter

michaelmdresser:
## What does this PR change?
Moves all core Allocation filtering to use the new `pkg/kubecost.AllocationFilter` type introduced in https://github.com/kubecost/opencost/pull/1172. Existing query parameters for filtering following the standard syntax (`filterNamespaces=...`) can use the utility functions in `pkg/filterutil` (https://github.com/kubecost/opencost/pull/1210) to convert those query params to the `AllocationFilter` type in a reliable way. A forthcoming KCM PR will provide an example.

This prepares us to hook up the new filter language (https://github.com/kubecost/opencost/pull/1250) to our APIs.

We keep FilterFuncs for Share/etc to minimize change surface. Eventually
those can be moved over to the new type, if desired.

* Requires https://github.com/kubecost/opencost/pull/1273
* Pairs with https://github.com/kubecost/kubecost-cost-model/pull/840

There should be no impact -- this is internal wiring and we expect the user-facing behavior to be identical.

## How was this PR tested?
WIP: Testing will be outlined in the KCM PR (https://github.com/kubecost/kubecost-cost-model/pull/840)

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -830,7 +830,7 @@ func NewAllocationSet(start, end time.Time, allocs ...*Allocation) *AllocationSe
 // simple flag for sharing idle resources.
 type AllocationAggregationOptions struct {
 	AllocationTotalsStore AllocationTotalsStore
-	FilterFuncs           []AllocationMatchFunc
+	Filter                AllocationFilter
 	IdleByNode            bool
 	LabelConfig           *LabelConfig
 	MergeUnallocated      bool
@@ -906,13 +906,19 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		options.ShareIdle = ShareNone
 	}
 
+	// Pre-flatten the filter so we can just check == nil to see if there are
+	// filters.
+	if options.Filter != nil {
+		options.Filter = options.Filter.Flattened()
+	}
+
 	var allocatedTotalsMap map[string]map[string]float64
 
 	// If aggregateBy is nil, we don't aggregate anything. On the other hand,
 	// an empty slice implies that we should aggregate everything. See
 	// generateKey for why that makes sense.
 	shouldAggregate := aggregateBy != nil
-	shouldFilter := len(options.FilterFuncs) > 0
+	shouldFilter := options.Filter != nil
 	shouldShare := len(options.SharedHourlyCosts) > 0 || len(options.ShareFuncs) > 0
 	if !shouldAggregate && !shouldFilter && !shouldShare && options.ShareIdle == ShareNone {
 		// There is nothing for AggregateBy to do, so simply return nil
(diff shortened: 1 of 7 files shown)


pkg/kubecost/allocation.go
```diff
@@ -1115,12 +1121,10 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
...
 
-		// (3) If any of the filter funcs fail, immediately skip the allocation.
-		for _, ff := range options.FilterFuncs {
-			if !ff(alloc) {
-				skip = true
-				break
-			}
+		// (3) If the allocation does not match the filter, immediately skip the
+		// allocation.
+		if options.Filter != nil {
```
Sean-Holcomb, line 1126: Do you want to use the `shouldFilter` var you initialized here?

pkg/kubecost/allocation.go
```diff
@@ -1305,11 +1309,8 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	// aggregate if an exact match is found.
 	for _, alloc := range externalSet.allocations {
 		skip := false
-		for _, ff := range options.FilterFuncs {
-			if !ff(alloc) {
-				skip = true
-				break
-			}
+		if options.Filter != nil {
```
Sean-Holcomb, line 1312: same here

pkg/kubecost/allocation.go
```diff
@@ -1342,11 +1343,8 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
...
-			for _, ff := range options.FilterFuncs {
-				if !ff(idleAlloc) {
-					skip = true
-					break
-				}
+			if options.Filter != nil {
+				skip = !options.Filter.Matches(idleAlloc)
 			}
 			if skip {
 				continue
```
Sean-Holcomb, line 1350: nit: seems like we could remove the `skip` var here

pkg/kubecost/allocation.go
```diff
@@ -1481,11 +1479,8 @@ func computeShareCoeffs(aggregateBy []string, options *AllocationAggregationOpti
...
-		for _, ff := range options.FilterFuncs {
-			if !ff(alloc) {
-				filtered = true
-				break
-			}
+		if options.Filter != nil {
+			filtered = !options.Filter.Matches(alloc)
 		}
 		if filtered {
 			name = "__filtered__"
```
Sean-Holcomb, line 1486: nit: same here with `filtered`

Sean-Holcomb (approved):
This all looks good, just a few nits

Sean-Holcomb (approved):
The new filter looks good, re-approving in case that helps with the issue you are having

PR #1276, opened 2022-06-23
Add StartsWith empty string unit test to filters

michaelmdresser:
## What does this PR change?
Adds unit tests for starts with empty (e.g. `filterNamespaces="*"`) behavior in AllocationFilterCondition.

## How was this PR tested?
It is a test

### DIFF
--- a/pkg/kubecost/allocationfilter_test.go
+++ b/pkg/kubecost/allocationfilter_test.go
@@ -59,6 +59,36 @@ func Test_AllocationFilterCondition_Matches(t *testing.T) {
 			expected: false,
 		},
 		{
+			name: "ClusterID empty StartsWith '' -> true",
+			a: &Allocation{
+				Properties: &AllocationProperties{
+					Cluster: "",
+				},
+			},
+			filter: AllocationFilterCondition{
+				Field: FilterClusterID,
+				Op:    FilterStartsWith,
+				Value: "",
+			},
+
+			expected: true,
+		},
+		{
+			name: "ClusterID nonempty StartsWith '' -> true",
+			a: &Allocation{
+				Properties: &AllocationProperties{
+					Cluster: "abc",
+				},
+			},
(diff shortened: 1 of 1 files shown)


Sean-Holcomb (approved):
LGTM

pkg/kubecost/allocationfilter_test.go
```diff
@@ -58,6 +58,36 @@ func Test_AllocationFilterCondition_Matches(t *testing.T) {
...
+			name: "ClusterID empty StartsWith '' -> true",
+			a: &Allocation{
+				Properties: &AllocationProperties{
+					Cluster: "",
+				},
+			},
+			filter: AllocationFilterCondition{
+				Field: FilterClusterID,
+				Op:    FilterStartsWith,
+				Value: "",
```
dramich, line 71: Are we saying empty matches all? Your comment says `filterNamespaces="*"` which is different.

michaelmdresser: For an `AllocationFilterCondition`, `Op: StartsWith` and `Value: ""` should match everything because all strings have a prefix of empty string. This is what `filterXYZ="*"` gets translated to by https://github.com/kubecost/opencost/pull/1211 and is the expected behavior.

Note that the `Op` isn't `FilterEquals`, but `FilterStartsWith`. Does that make sense?

dramich: Yeah, so we are talking about external vs internal representation. Do we support the case of doing a wildcard anywhere?

michaelmdresser: We do support wildcards right now in the external representation -- `filterNamespaces="*"` is a valid query in today's Allocation API, as is `filterNamespaces="kube-*"`. 

> Note: The v2 filter language does not yet support wildcards. It will support an equivalent query with something like `filter=namespace~:"kube-"`.

PR #1279, opened 2022-06-24
Support GCP Spot VMs

kamilkrampa:
## What does this PR change?
* Adds support for GCP Spot VMs

* No

* https://github.com/kubecost/opencost/issues/1278

## How was this PR tested?
* Unit test in this PR

* No

### DIFF
--- a/pkg/cloud/gcpprovider.go
+++ b/pkg/cloud/gcpprovider.go
@@ -3,7 +3,6 @@ package cloud
 import (
 	"context"
 	"fmt"
-	"github.com/kubecost/opencost/pkg/kubecost"
 	"io"
 	"io/ioutil"
 	"math"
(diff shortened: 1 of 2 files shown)


Sean-Holcomb (approved):
This looks good to me and is inline with the GCP documentation on the subject. Given that we use the term preemptible internally to mean both spot and preemptible, this change results in the correct assignment of spot node usageType. 

From the documentation:
> GKE automatically adds the cloud.google.com/gke-spot=true 

the spot check seems to be accurate.

My one nit is that given the limited scope of this function `func getUsageType(labels map[string]string) string` seems like it would be a little better.

kamilkrampa:
@Sean-Holcomb I applied your feedback. Thanks!

PR #1297, opened 2022-07-05
Move all regex.MustCompile calls to vars

dramich:
regexp.MustCompile should be instantiated in a var so the panic happens
at startup. This also has a lovely side effect of improving performance
since the regex is not compiled every time it needs to be used.

In window.go there is a 20x performance improvement with this change.

```
goos: darwin
goarch: arm64
pkg: github.com/kubecost/opencost/pkg/kubecost
BenchmarkParseWindowUTC
BenchmarkParseWindowUTC-10    	   57448	     20021 ns/op
PASS
ok  	github.com/kubecost/opencost/pkg/kubecost	1.557s
```

```
goos: darwin
goarch: arm64
pkg: github.com/kubecost/opencost/pkg/kubecost
BenchmarkParseWindowUTC
BenchmarkParseWindowUTC-10    	 1063864	      1088 ns/op
PASS
ok  	github.com/kubecost/opencost/pkg/kubecost	2.122s
```

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/csv"
 	"fmt"
-	"github.com/kubecost/opencost/pkg/kubecost"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -15,6 +14,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/kubecost/opencost/pkg/kubecost"
 
 	"github.com/kubecost/opencost/pkg/clustercache"
 	"github.com/kubecost/opencost/pkg/env"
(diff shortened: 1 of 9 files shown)


mbolt35 (approved):
Great change, thanks Dan. Casual 20x faster 💯

Adam-Stack-PM:
@dramich Should we push this into v1.95 or wait till v1.96?

dramich:
> @dramich Should we push this into v1.95 or wait till v1.96?

I don't think it needs to be backported to the 1.95 branch.

mbolt35 (approved):
LGTM! Thanks for benchmarking as well. Results look awesome :)

Adam-Stack-PM:
@michaelmdresser Will this go into v1.96 as we have already created the RC branch?

michaelmdresser:
> Will this go into v1.96

Yes, this will go into v1.96. This is a performance and (possibly) correctness improvement, but it isn't the kind of fix that warrants a cherry pick, in my opinion.

PR #1300, opened 2022-07-06
Fix panic on trying to marshal pointer to nil AllocationSetRange

kaelanspatel:
## What does this PR change?
Adds a check in MarshalJSON for AllocationSetRange that makes sure the set is not nil. If it is, it marshals an empty version of what is would marshal (e.g. AS -> empty map of strings to allocations, ASR -> empty slice of AS).

Also adds a test for marshaling AllocationSetRange.

No

Fixed a panic when querying allocations.

## How was this PR tested?
* Tested using written test + recreating scenario where error occurred and making sure no panic happens. 

No

Yes

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -1889,6 +1889,9 @@ func (as *AllocationSet) Map() map[string]*Allocation {
 
 // MarshalJSON JSON-encodes the AllocationSet
 func (as *AllocationSet) MarshalJSON() ([]byte, error) {
+	if as == nil {
+		return json.Marshal(map[string]*Allocation{})
+	}
 	as.RLock()
 	defer as.RUnlock()
 	return json.Marshal(as.allocations)
@@ -2224,6 +2227,10 @@ func (asr *AllocationSetRange) Length() int {
 
 // MarshalJSON JSON-encodes the range
 func (asr *AllocationSetRange) MarshalJSON() ([]byte, error) {
+	if asr == nil {
+		return json.Marshal([]*AllocationSet{})
+	}
+
 	asr.RLock()
 	defer asr.RUnlock()
 	return json.Marshal(asr.allocations)
--- a/pkg/kubecost/allocation_test.go
+++ b/pkg/kubecost/allocation_test.go
@@ -2368,8 +2368,60 @@ func TestAllocationSetRange_InsertRange(t *testing.T) {
 // TODO niko/etl
 // func TestAllocationSetRange_Length(t *testing.T) {}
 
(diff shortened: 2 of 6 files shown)


pkg/kubecost/allocation.go
```diff
@@ -2224,6 +2227,10 @@ func (asr *AllocationSetRange) Length() int {
 
 // MarshalJSON JSON-encodes the range
 func (asr *AllocationSetRange) MarshalJSON() ([]byte, error) {
+	if asr == nil {
+		return json.Marshal([]*AllocationSet{})
```
kaelanspatel, line 2231: I'm not sure if this the "correct" behavior for a nil marshal...

@mbolt35 do you maybe have an opinion here?

mbolt35: I think using an empty AllocationSet is probably fine, although I think `nil` should be supported, and it does depend on whatever decoder is receiving the JSON... There's an example of returning `"null"` in go:

https://github.com/golang/go/blob/bebe9aa42322f951fc3972c263648297bf9e04d4/src/math/big/intmarsh.go#L68-L74

Where if the receiver object is `nil` they just return `[]byte("null")`

Note: In the above example, `Unmarshal` provides the symmetrical: 
```
if string(text) == "null" {
    return nil
}
```
So if we're going the `"null"` route, we'll need to make sure our `Unmarshal` handles `"null"` as well, which, honestly, just feels like more than it's worth.

I am definitely ok with either of these solutions, maybe even partial to the empty array encode. As long as we know we are correctly processing the encoding on the frontend, we good 👍

mbolt35 (approved):
LGTM

pkg/kubecost/allocation.go
Sean-Holcomb: It seems reasonable to have an error here. In the past it seems that we have not allowed nil Sets and SetRanges, doing an error would continue to prohibit those without the panic, but we would still need to track down why we are getting nil asr now instead of empty ones.

Sean-Holcomb (changes requested):
I am going to push some changes with the same for assetset and assetsetrange tonight. Please do not merge yet

Sean-Holcomb (approved):
Added in assets and set range updates to match Kaelan's changes

michaelmdresser:
This is tagged 1.95. Does this change need to make it in to 1.95?

Sean-Holcomb:
Yes this is on the P0 list

michaelmdresser:
Excellent, thanks. Please tag me when a cherry pick is opened so I can make sure it gets into the next RC.

Issue #1291, opened 2022-06-29
Make `minsPerResolution` configurable in cluster.go


jcharcalla:
We should be reading `minsPerResolution` out of the `kubecostModel.etlResolutionSeconds` parameter instead of hardcoding the value. 

Here is the environment variable link.
https://github.com/kubecost/opencost/blob/e5dc00d4fb594f1f4b6dfc098414e09f6eec599f/pkg/env/costmodelenv.go#L68

And here is where it is hardcoded to 1 min.
https://github.com/kubecost/opencost/blob/6ddb2d8c4766a72c7303443487dcaa1be494b3f7/pkg/costmodel/cluster.go#L391

Adam-Stack-PM:
@teevans Looks like an open issue from last weeks support rotation. Could you assign a priority P0,1,2 and remove the needs triage label?

teevans:
@jcharcalla - What do you think is a good priority level for this?

AjayTripathy:
I think that this should be done before the release, it's a clear bug and a very small fix

Adam-Stack-PM:
@avrodrigues5, How is this item going?  Anything blocking getting a PR opened today?

avrodrigues5:
@Adam-Stack-PM , I will sort out some help for testing and open a PR by EOD. 

Just one question, issue mentions only one occurrence of minsPerResolution but there are 4 and 3 occurrence hardcoded to 1 (line No 139, 391, 547) while one occurrence (line No 673) hardcoded to 5 in file https://github.com/kubecost/opencost/blob/6ddb2d8c4766a72c7303443487dcaa1be494b3f7/pkg/costmodel/cluster.go#L391 . What should be the default value 1 or 5?

Adam-Stack-PM:
@avrodrigues5 Thanks for getting this out. Where you able to get answers to your questions?

avrodrigues5:
@Adam-Stack-PM 
More i looked at this issue more questions i had
1. So the function that's used to compute [ETLResolution](https://github.com/kubecost/opencost/blob/08a98fac21673210f8b52681395e89188b08b7a1/pkg/env/costmodelenv.go#L451-L461) seems to default to 1m for prometheus user and if thanos is enabled only computes it to anything other than 1m default to 5m. Also what happens if user passes less than 60 seconds or between 60-120 seconds should it round down or round up, because it is a possibility.
[...]

PR #1303, opened 2022-07-08
Making minsPerResolution configurable in cluster.go

avrodrigues5:
## What does this PR change?
*  Make the minsPerResolution Configurable with ETL_RESOLUTION_SECONDS env variable.

* No

* I have tried to keep this most least breaking change, to avoid impact on user 

* Closes ... #1291 

## How was this PR tested?
* Yet to device a test plan tomorrow 

* No

* Will discuss it tomorrow.

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -136,8 +136,12 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 	// minsPerResolution determines accuracy and resource use for the following
 	// queries. Smaller values (higher resolution) result in better accuracy,
 	// but more expensive queries, and vice-a-versa.
-	minsPerResolution := 1
-	resolution := time.Duration(minsPerResolution) * time.Minute
+	resolution := env.GetETLResolution()
+	//Ensuring if ETL_RESOLUTION_SECONDS is less than 60s default it to 1m
+	var minsPerResolution int
+	if minsPerResolution = int(resolution.Minutes()); int(resolution.Minutes()) == 0 {
+		minsPerResolution = 1
+	}
 
 	// hourlyToCumulative is a scaling factor that, when multiplied by an hourly
 	// value, converts it to a cumulative value; i.e.
@@ -388,8 +392,12 @@ func ClusterNodes(cp cloud.Provider, client prometheus.Client, start, end time.T
 	// minsPerResolution determines accuracy and resource use for the following
 	// queries. Smaller values (higher resolution) result in better accuracy,
 	// but more expensive queries, and vice-a-versa.
-	minsPerResolution := 1
-	resolution := time.Duration(minsPerResolution) * time.Minute
+	resolution := env.GetETLResolution()
+	//Ensuring if ETL_RESOLUTION_SECONDS is less than 60s default it to 1m
+	var minsPerResolution int
+	if minsPerResolution = int(resolution.Minutes()); int(resolution.Minutes()) == 0 {
+		minsPerResolution = 1
+	}
 
 	requiredCtx := prom.NewNamedContext(client, prom.ClusterContextName)
 	optionalCtx := prom.NewNamedContext(client, prom.ClusterOptionalContextName)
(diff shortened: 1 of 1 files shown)


avrodrigues5:
Hello @AjayTripathy ,

I kept the changes as minimal as possible but just to be sure the function in [costmodelenv](https://github.com/kubecost/opencost/blob/08a98fac21673210f8b52681395e89188b08b7a1/pkg/env/costmodelenv.go#L451-L461) still default to 1m for prometheus i hope the customers impacted have thanos enabled or this PR need more code change.

Not sure if  it in this scope or not but i changed all the hardcoded values of minPerResolution original issue only mentions line No [391](https://github.com/kubecost/opencost/blob/develop/pkg/costmodel/cluster.go#L391). One of [line](https://github.com/kubecost/opencost/blob/develop/pkg/costmodel/cluster.go#L673) the hardcoded value was 5 not sure how it will impact but will have to check with yall.

CC: @michaelmdresser

pkg/costmodel/cluster.go
```diff
@@ -136,8 +136,12 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end
...
 	// queries. Smaller values (higher resolution) result in better accuracy,
 	// but more expensive queries, and vice-a-versa.
-	minsPerResolution := 1
-	resolution := time.Duration(minsPerResolution) * time.Minute
+	resolution := env.GetETLResolution()
+	//Ensuring if ETL_RESOLUTION_SECONDS is less than 60s default it to 1m
+	var minsPerResolution int
+	if minsPerResolution = int(resolution.Minutes()); int(resolution.Minutes()) == 0 {
+		minsPerResolution = 1
+	}
```
michaelmdresser: If we ever silently change a config value that a user may have set, we should make a note of it. I recommend a `log.DedupedWarningf` or similar, e.g.:

```
log.DedupedWarningf(3, "ClusterDisks(): Configured ETL resolution (%d seconds) is below the 60 second threshold. Overriding with 1 minute.", resolution.Seconds())
```

michaelmdresser (commented):
The code here makes sense, but I have some reservations. I know you're trying to keep code line impact low, but I don't want to make changes to core data pipelines in haste in a way that will cause us problems later.

Here's my main concern:

I'm not sure that its a good idea to truncate a value that a user inputs in seconds (ETL resolution) to the next-lowest minute value (the `int()` conversion approach you're using). At other times when we have a configurable query resolution, like in `(*CostModel).computeAllocation()`, we use `timeutil.DurationString()` to get a Prometheus-compatible string:

https://github.com/kubecost/opencost/blob/6187d3d51dfcfcdf870ded6ae151e421c6a25d1f/pkg/costmodel/allocation.go#L306-L307

Which then gets formatted into queries like:

https://github.com/kubecost/opencost/blob/6187d3d51dfcfcdf870ded6ae151e421c6a25d1f/pkg/costmodel/allocation.go#L21

Like so:

https://github.com/kubecost/opencost/blob/6187d3d51dfcfcdf870ded6ae151e421c6a25d1f/pkg/costmodel/allocation.go#L759

We can compare that with how `ClusterDisks()` currently uses `minsPerResolution`. Notice how one of the `ClusterDisks()` Prom queries formats into `[%s:%dm]` (the second one is the resolution):

https://github.com/kubecost/opencost/blob/6187d3d51dfcfcdf870ded6ae151e421c6a25d1f/pkg/costmodel/cluster.go#L153

If we instead use `timeutil.DurationString()` you can change that to `[%s:%s]`.

---

More broadly, I'm still of the opinion that we should be striving for equality with how Allocation works as much as possible here, to avoid inconsistencies that may bite us later. Having a single source of truth for config values like this is something that I think is important, and that source of truth is currently `pkg/env`. For example, `(*CostModel).ComputeAllocation()` takes a resolution as an argument, which is [supplied by `env.GetETLResolution()`](https://github.com/kubecost/kubecost-cost-model/blob/5c95653539cd1b1e74f78ee863f6c50a829f30d1/pkg/allocation/etlstore.go#L1010). This feels like the right approach to me.

---

@mbolt35 do you have any thoughts here?

mbolt35 (approved):
These changes look fine to me. @michaelmdresser and I discussed some future direction with respect to config that we should likely take going forward, and this change is a step towards that goal. Thanks @avrodrigues5!

Oh! And moving the hard coded `5m` duration to the exact same resolution as the other queries is fine.

avrodrigues5:
Merging this code in next 30min,

Testing coducted:
Uploaded KCM image to test cluster and saw the minsPerResolution sets to default 1m 
Ran go test on opencost and KCM repos.

Added the log debug statement as asked in this PR review comments by @michaelmdresser (want to give a final look?)
CC: @mbolt35 

-Alan

avrodrigues5:
Tested the scenario to check assets over 7 days, merging the code !

michaelmdresser:
@avrodrigues5 Looks like you closed instead of merging. Reopening.

PR #1315, opened 2022-07-12
Add supported aggregations to the OpenCost spec

dwbrown2:
## What does this PR change?
* Specifically identify supported aggregation types 

* No

* Spec will be more specific

* No

## How was this PR tested?
* Reviewed markdown

* Done

* Can be merged at any time

### DIFF
--- a/spec/opencost-specv01.md
+++ b/spec/opencost-specv01.md
@@ -213,7 +213,7 @@ The following workload cost aggregations are supported in a complete implementat
 * deployment
 * statefulset
 * job
-* controller
+* controller name
 * controller kind
 * label
 * annotation
--- a/spec/opencost-specv01.md
+++ b/spec/opencost-specv01.md
@@ -158,8 +158,7 @@ Below are example inputs when measuring asset costs over a designated time windo
 
 ## Workload Costs
 
-
-Workloads are defined as entities to which Asset Costs are committed. Some resources solely have Usage Costs, but others have Allocation Costs independent of actual usage. Workload Costs should be understood as _max(request, usage)_ when Assets have Resource Allocation Costs, e.g. CPU or GPU. This formula effectively assigns costs that have been directly reserved or allocated by _kube-scheduler_. Workload Costs should be calculated at the lowest level possible, e.g. _container_ level[^2], and then they can be aggregated by any dimension, e.g. pod, deployment, statefulset, label, annotation, namespace, etc.
+Workloads are defined as entities to which Asset Costs are committed. Some resources solely have Usage Costs, but others have Allocation Costs independent of actual usage. Workload Costs should be understood as _max(request, usage)_ when Assets have Resource Allocation Costs, e.g. CPU or GPU. This formula effectively assigns costs that have been directly reserved or allocated by _kube-scheduler_. Workload Costs should be calculated at the lowest level possible, i.e. _container_ level[^2], and then they can be aggregated by any dimension.
 
 
 <table>
@@ -207,7 +206,19 @@ Workloads are defined as entities to which Asset Costs are committed. Some resou
   </tr>
 </table>
 
+The following workload cost aggregations are supported in a complete implementation in the OpenCost Spec: 
 
+* container
+* pod
+* deployment
+* statefulset
+* job
+* controller
+* controller kind
+* label
+* annotation
+* namespace
+* cluster
 
 ## Shared Costs
 


spec/opencost-specv01.md
```diff
@@ -207,7 +206,19 @@ Workloads are defined as entities to which Asset Costs are committed. Some resou
...
 
-
+The following workload cost aggregations are supported in a complete implementation in the OpenCost Spec: 
+
+* container
+* pod
+* deployment
+* statefulset
+* job
+* controller
```
jpelletier1: Small feedback: should we call this `controller name`?

spec/opencost-specv01.md
```diff
@@ -207,7 +206,19 @@ Workloads are defined as entities to which Asset Costs are committed. Some resou
...
+* controller kind
+* label
```
jpelletier1, line 218: Would this aggregate on:
1. the label's value (requiring a user to first select a key)
2. the label's key+value
3. the key only

I think it's #1, but wanted to double check

spec/opencost-specv01.md
```diff
@@ -158,8 +158,7 @@ Below are example inputs when measuring asset costs over a designated time windo
 
 ## Workload Costs
 
-
-Workloads are defined as entities to which Asset Costs are committed. Some resources solely have Usage Costs, but others have Allocation Costs independent of actual usage. Workload Costs should be understood as _max(request, usage)_ when Assets have Resource Allocation Costs, e.g. CPU or GPU. This formula effectively assigns costs that have been directly reserved or allocated by _kube-scheduler_. Workload Costs should be calculated at the lowest level possible, e.g. _container_ level[^2], and then they can be aggregated by any dimension, e.g. pod, deployment, statefulset, label, annotation, namespace, etc.
+Workloads are defined as entities to which Asset Costs are committed. Some resources solely have Usage Costs, but others have Allocation Costs independent of actual usage. Workload Costs should be understood as _max(request, usage)_ when Assets have Resource Allocation Costs, e.g. CPU or GPU. This formula effectively assigns costs that have been directly reserved or allocated by _kube-scheduler_. Workload Costs should be calculated at the lowest level possible, i.e. _container_ level[^2], and then they can be aggregated by any dimension.
```
jpelletier1, line 161: I think this makes sense, but a general question: does a Workload Cost only assume the cost of that Workload in a _single_ cluster? Or can you calculate a Workload Cost across _multiple_ clusters? Trying to think about scenarios where a business wants to understand the cost of a workload that runs in different regions, or across different environments.

jpelletier1 (commented):
Some minor feedback on the aggregation proposal

spec/opencost-specv01.md
jpelletier1: One thing I'd be curious about is whether users see value in aggregating by `deployment`/`statefulset`/`job` kinds. E.g., in my experience, I see most folks aggregating by `controller name`, `namespace`, or `label` — and _filtering_ by  kind=deployment, kind=job, etc.

Happy to keep it here if folks think it's important - no harm in making it part of the available aggregations. But I wonder if the concept of ‘filtering’ (in addition to aggregation) should also be introduced in the Spec?

I hold this loosely, so if folks disagree, happy to move it into a different issue or PR.

dwbrown2: +1 on `controller name` will get that updated

Aggregating on `deployment` allows users to quickly see all active instances over the measured/projected window while also seeing costs that is not allocated to a deployment, e.g. jobs. This essentially allows users to skip a step when they are focused on measuring or optimizing one piece of their workloads at a time. Let me know if that makes sense. 

I think you've definitely highlighted a need to discuss filtering. That's potentially another standalone section of the spec which is actually already supported in the OpenCost implementation.

spec/opencost-specv01.md
dwbrown2: This would aggregate based on value after a users has provided a key, e.g. environment in the example below. 

```
"environment" : "dev", "environment" : "qa", "environment" : "production"
```

Would you see a use case for supporting #2 or #3?

spec/opencost-specv01.md
dwbrown2: Everything in the spec has been focused on a single cluster to date, but all of these concepts hold when aggregating across multiple clusters, regions, etc.

spec/opencost-specv01.md
jpelletier1: That makes sense to me. I don't see an immediate use case for #2 or #3.

spec/opencost-specv01.md
jpelletier1: Thanks for clarifying! No additional comments from me.

spec/opencost-specv01.md
jpelletier1: Gotcha - thanks for the detail on aggregating on `deployment`. Definitely interesting, and perhaps achievable through filtering as well. Bottom line, I think some of this may just come down to UX implementation - as there could be a number of ways to get to what the user wants ;-)

Filtering may also be an aspect of UX; perhaps we can decide later if it makes sense to include in the Spec.

No additional comments from me!

AjayTripathy (approved):

PR #1321, opened 2022-07-14
Recognizes annotations when aggregating by owner

### LINKED ISSUE #1392: Can't view cost by annotation in Allocation UI
**Describe the bug**

After enabling namespace [annotation emissions](https://guide.kubecost.com/hc/en-us/articles/4407595918231-Enabling-Annotation-Emission) with `--set kubecostMetrics.emitNamespaceAnnotations=true` it is implied  by "Owner Label / Annotation" and the description of it on the settings page that annotations can be used for label mapping (See screen shot below). However when Using an Annotation and aggregating by owner on the "Cost allocation" page, my annotated Namespace assets do not appear. 

Using a Namespace label does function correctly. Additionally I'm aware that aggregating by annotation currently only works via the [API](https://guide.kubecost.com/hc/en-us/articles/4407595916823-Allocation-API#querying). This makes me wonder if the description "Owner Label / Annotation -  Name of k8s label or annotation to designate owner/contact." on the settings page of the UI is correct.

**To Reproduce**
1. Enable annotation emissions. `--set kubecostMetrics.emitNamespaceAnnotations=true`
2. Add an annotation to a namespace. EX. `Annotations:  app.test.com/primary-owner: vader`
3. Set the "Owner Label / Annotation" field on the "settings" page to the namespace annotation. I tried both `app.test.com/primary-owner` and `app_test_com_primary_owner`
4. On the "costs allocation" page aggregate by "owner", at this point only `__idle__`  and `__unallocated__` appear.

**Expected behavior**
Annotations appear in the aggregated by owner view of the "cost allocation" page if that is the intended behavior.

**Screenshots**
Annotated namespace:
```
$ kubectl describe ns kube-system
Name:         kube-system
Labels:       kubernetes.io/metadata.name=kube-system
Annotations:  app.test.com/primary-owner: vader
Status:       Active
```

Settings page:
Screen Shot 2022-04-26 at 3 58 38 PM

Costs Allocation page with no results:
Screen Shot 2022-04-27 at 2 37 24 PM

nickcurie:
## What does this PR change?
* When aggregating by owner, kubecost will generate keys for annotations

* No

* Users can now see cost breakdowns when using annotations and aggregating by owner 

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1392

## How was this PR tested?
* Unit testing and manual testing on a local cluster

* No

* Yes

### DIFF
--- a/pkg/kubecost/allocationprops_test.go
+++ b/pkg/kubecost/allocationprops_test.go
@@ -1,0 +1,44 @@
+package kubecost
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestGenerateKey(t *testing.T) {
+
+	cases := map[string]struct {
+		aggregate       []string
+		allocationProps *AllocationProperties
+		expected        string
+	}{
+		"agg by owner with labels": {
+			aggregate: []string{"owner"},
+			allocationProps: &AllocationProperties{
+				Labels:      map[string]string{"app": "cost-analyzer"},
+				Annotations: map[string]string{"owner": "test owner 123"},
+			},
+			expected: "test owner 123",
+		},
+		"agg by owner without labels": {
+			aggregate: []string{"owner"},
+			allocationProps: &AllocationProperties{
+				Annotations: map[string]string{"owner": "test owner 123"},
+			},
+			expected: "test owner 123",
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+
+			lc := NewLabelConfig()
+
+			result := tc.allocationProps.GenerateKey(tc.aggregate, lc)
+
+			if !reflect.DeepEqual(result, tc.expected) {
+				t.Fatalf("expected %+v; got %+v", tc.expected, result)
+			}
+		})
+	}
+}
--- a/pkg/kubecost/allocationprops.go
+++ b/pkg/kubecost/allocationprops.go
@@ -347,7 +347,8 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 			}
 		case agg == AllocationOwnerProp:
 			labels := p.Labels
-			if labels == nil {
+			annotations := p.Annotations
+			if labels == nil && annotations == nil {
 				names = append(names, UnallocatedSuffix)
 			} else {
 				labelNames := strings.Split(labelConfig.OwnerLabel, ",")
@@ -355,6 +356,8 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 					labelName = labelConfig.Sanitize(labelName)
 					if labelValue, ok := labels[labelName]; ok {
 						names = append(names, labelValue)
+					} else if annotationValue, ok := annotations[labelName]; ok {
+						names = append(names, annotationValue)
 					} else {
 						names = append(names, UnallocatedSuffix)
 					}


michaelmdresser (commented):
@dwbrown2 / @AjayTripathy I think we have a product-level question to answer about this PR. I'm sorry for not bringing this up sooner to help Nick out -- I was tied up with release work. If you'd like to move this discussion back to the issue, I'd be happy to do so.

The issue this is based on (https://github.com/kubecost/cost-analyzer-helm-chart/issues/1392) is only a problem because the Settings page promises that Owner, and only Owner, supports annotation mapping in addition to label mapping. The other mapping fields (Product, Env, etc.) do NOT support annotation mapping and have never promised to. Webb called this out in his [comment](https://github.com/kubecost/cost-analyzer-helm-chart/issues/1392#issuecomment-1116344593) on the issue: "I would for sure call the Settings description a bug."

My personal feeling is that its more consistent to just not support annotation mapping right now. If its sufficiently desired we should find a different way to do it properly instead of overloading the `LabelConfig`'s `OwnerLabel` [1] to somehow support annotations at the same time. Something like a `MappingConfig` to represent relationships between arbitrary groups (like Owner) and the underlying Kubernetes metadata [2] 

You can see the effect of the confusion in this PR -- we end up defaulting to a label and if we can't find a match we look for an annotation. This will result in unexpected behavior for users if they have annotations and labels with the same key.

What makes more sense to me is something like this:

- User picks if Owner (and other mappings like Product) should be mapped to a label OR an annotation (toggle switch)
- User picks what key should be used for the mapping

So you get a mapping like: "Owner maps to label `app`" OR "Owner maps to annotation `app`". 

Instead of what this PR implies, which is: "Owner maps to label `app`, but if label `app` does not exist it also maps to annotation `app`".

Also, closed-source filter func generation (and other locations where we handle `LabelConfig`) only supports _labels_ for the Owner mapping [3]. If we decide to go ahead with the approach in this PR, we will have to make additional changes everywhere Owner is handled.

--- 

@nickcurie this is good work and isn't meant to degrade that -- this is more of a product question we have to answer. As you get more familiar with the code, I think you'll be able to start spotting problems like this as well.

[1] https://github.com/kubecost/opencost/blob/2105a2b57a859c3716b11d8892ef2fb9ef67f50d/pkg/kubecost/config.go#L14
[2] I think this has the potential to be very powerful, actually! Imagine multicluster environments where teams use a different label key in each environment. We could find a way to map Team to the right label key in each cluster! I think this relates to [a separate discussion in another PR](https://github.com/kubecost/opencost/pull/1315#discussion_r919281872).
[3] (closed source) https://github.com/kubecost/kubecost-cost-model/blob/3166005fde13eb8eb0a49dbc702e54fdb30e4e6c/pkg/allocation/filters.go#L388-L396

teevans:
@michaelmdresser - This sounds like it should be moved to a different thread - I'd say if we've planned this work and executed it, let's go ahead and merge it in, and re-visit that experience.  Thoughts @dwbrown2?

michaelmdresser:
> let's go ahead and merge it in, and re-visit that experience

My only concern is that merging this locks us into a certain behavior pattern and fixing it could cause a breaking behavioral change for users if they start to depend on it. I defer to you and Webb here, so if you think this change is acceptable I'm on board.

AjayTripathy:
Why not just support both annotations and labels for all of owner/product/team/department and update the ui to match? That feels right to me if users are already trying that query. See the linked issue. Related but slightly different is we want to bring the ability to filter/aggregate by annotation to the UI, but I think that's beyond the scope of this PR.

michaelmdresser:
> Why not just support both annotations and labels for all of owner/product/team/department and update the ui to match? That feels right to me if users are already trying that query. See the linked issue.

I think that's a perfectly good idea! However, if we choose to do it that way I think we do it _right_ so we don't fall into another config trap that would lock us into confusing behavior. I attempted to explain the confusion above and offer an alternative, but I think I missed the mark. Here's an example that I think outlines this PR's behavior:

- Owner label/annotation set to: `foo`
- Deployment A with label `foo:x`
- Deployment B with label `foo:y` and annotation `foo:x`
- User filters `filterOwner=x` and gets Deployments A and B
- User filters `filterOwner=y` and gets Deployment B
- User aggregates by `owner` and Deployment A gets put under `owner:x` and Deployment B gets put under `owner:y`

Deployment B seems to have two owners, and that manifests itself differently depending on the query.

I'm still happy to approve this behavior if product understands the implications and is okay with them.

AjayTripathy:
I think that's rare enough of an edge case and easily explainable enough that I'm comfortable with the behavior. As long as one of label or annotation consistently wins which it does in the example I'm good with that.

Can we extend the behavior for all custom aggregations definitions like department/team/product etc and  make the UI change or file for a ui change on settings to make it clear that annotations are supported for all of these? cc   @nickcurie

dwbrown2:
So I agree with Michael's concerns and that we're making some meaningful product decisions pretty quickly here. A longer-term solution would likely be a more generic mapping stored on the backend between real-world concept, e.g. `team` and infrastructure concept, e.g. `label`, `annotation`, `namespace`, etc. That feels like a significant undertaking a not one we necessarily need to take on right now. 

@michaelmdresser one solution to the risk you highlight is to say that when the same key exists in both labels and annotations, we always favor labels. That could apply to both aggregating and filtering. That would address the double counting risk wouldn't it? If that sounds right to you all then I'm supportive of moving forward here!

michaelmdresser:
> A longer-term solution would likely be a more generic mapping stored on the backend between real-world concept... That feels like a significant undertaking a not one we necessarily need to take on right now.

Yes, I agree on both points.

> when the same key exists in both labels and annotations, we always favor labels.

Yep, that's the approach currently in this PR and I believe it does address any double-counting risk.

> If that sounds right to you all then I'm supportive of moving forward here!

Okay, SGTM! I'll give this a proper review and point out some next steps we need to take.

pkg/kubecost/allocationprops_test.go
```diff
@@ -0,0 +1,44 @@
...
+			expected: "test owner 123",
+		},
+		"agg by owner without labels": {
+			aggregate: []string{"owner"},
+			allocationProps: &AllocationProperties{
+				Annotations: map[string]string{"owner": "test owner 123"},
+			},
+			expected: "test owner 123",
+		},
+	}
```
michaelmdresser, line 30: Please add a test case for the situation where we have both an annotation and a label called `owner` (that have different values). E.g. a label `owner = foo` and an annotation `owner = bar`

michaelmdresser (commented):
This is great work, thanks for being patient with the back-and-forth. As Ajay suggested, let's extend this PR to support annotations in the other label-mapped fields (Product, Env, etc.) and update the text on the frontend's settings page to reflect the new support.

Soon (but not in this PR) we'll also have to update our filtering logic to support annotations for Owner (any any other label mappings we add annotation support for). Just making a note so we know where to look when we get there:
- The [owner filter func generator in KCM](https://github.com/kubecost/kubecost-cost-model/blob/dac77ec1359dfe653c9c6be62d151e33500f46ba/pkg/allocation/filters.go#L386-L411)
- The [filter v1 -> struct generation in OC](https://github.com/opencost/opencost/blob/56105676a8c54674233d54909879ebb1e4747923/pkg/util/allocationfilterutil/queryfilters.go#L202-L204)

nickcurie:
This is an old PR now, due to switching to opencost/opencost. The changes mentioned here are resolved and the functionality was extended to other aggregations in https://github.com/opencost/opencost/pull/1327 cc: @Adam-Stack-PM

PR #1322, opened 2022-07-14
Add Azure Storage Config values to provider config

Sean-Holcomb:
## What does this PR change?
* This PR adds fields to the Cloud Provider config for Azure Storage Configuration. This feature will allow the FE team to develop a UI for setting Azure configuration via the settings page in line with what exists for AWS and GCP. When looking for an Azure Storage Configuration the OC Azure Provider will first check the provider config, followed by the secret configuration method. If any of the required configuration fields are missing, the next configuration method will be checked.

* https://github.com/kubecost/kubecost-cost-model/pull/868

* 

* Related to https://github.com/kubecost/cost-analyzer-frontend/issues/802#issuecomment-1155587994

## How was this PR tested?
* This PR was tested on an Azure Cluster by switching back and forth between different Azure Storage configurations on the front end, repairing the last 7 days of Cloud Usage and ensure that connection was successfully made and the data updated. 

* 

*

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -37,6 +37,7 @@ const (
 	AzureDiskStandardStorageClass    = "standard_hdd"
 	defaultSpotLabel                 = "kubernetes.azure.com/scalesetpriority"
 	defaultSpotLabelValue            = "spot"
+	AzureStorageUpdateType           = "AzureStorage"
 )
 
 var (
@@ -509,7 +510,7 @@ func (ask *AzureServiceKey) IsValid() bool {
 }
 
 // Loads the azure authentication via configuration or a secret set at install time.
-func (az *Azure) getAzureAuth(forceReload bool, cp *CustomPricing) (subscriptionID, clientID, clientSecret, tenantID string) {
+func (az *Azure) getAzureRateCardAuth(forceReload bool, cp *CustomPricing) (subscriptionID, clientID, clientSecret, tenantID string) {
 	// 1. Check config values first (set from frontend UI)
 	if cp.AzureSubscriptionID != "" && cp.AzureClientID != "" && cp.AzureClientSecret != "" && cp.AzureTenantID != "" {
 		subscriptionID = cp.AzureSubscriptionID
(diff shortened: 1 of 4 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -1165,27 +1182,42 @@ func (az *Azure) UpdateConfigFromConfigMap(a map[string]string) (*CustomPricing,
...
-			return err
-		}
-		for k, v := range a {
-			kUpper := strings.Title(k) // Just so we consistently supply / receive the same values, uppercase the first letter.
-			vstr, ok := v.(string)
-			if ok {
-				err := SetCustomPricingField(c, kUpper, vstr)
-				if err != nil {
-					return err
+		if updateType == AzureStorageUpdateType {
```
michaelmdresser, line 1186: What is an Azure storage update? Mostly asking because I'm wondering why we only call `az.DownloadPricingData()` in the else block.

pkg/costmodel/allocation.go
```diff
@@ -293,8 +293,10 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
 	}
 
 	// TODO:CLEANUP remove "max batch" idea and clusterStart/End
-	cm.buildPodMap(window, resolution, env.GetETLMaxPrometheusQueryDuration(), podMap, clusterStart, clusterEnd, ingestPodUID, podUIDKeyMap)
-
+	err := cm.buildPodMap(window, resolution, env.GetETLMaxPrometheusQueryDuration(), podMap, clusterStart, clusterEnd, ingestPodUID, podUIDKeyMap)
+	if err != nil {
+		log.Errorf("CostModel.ComputeAllocation: failed to build pod map: %s", err.Error())
+	}
```
michaelmdresser, line 299: :eyes: nice catch

pkg/costmodel/router.go
```diff
@@ -615,6 +615,18 @@ func (a *Accesses) UpdateBigQueryInfoConfigs(w http.ResponseWriter, r *http.Requ
...
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+	data, err := a.CloudProvider.UpdateConfig(r.Body, cloud.AzureStorageUpdateType)
+	if err != nil {
+		w.Write(WrapData(data, err))
+		return
+	}
+	w.Write(WrapData(data, err))
+	return
+}
```
michaelmdresser, line 628: I'm not well-versed here -- should we really have this endpoint defined here AND in the closed-source PR? It feels like it should just be in KCM.

Sean-Holcomb: The ones for the other providers are present here just above, I am not sure why. I felt that the two options would be removing them all or just adding the new one, and I did not feel confident enough to remove the other ones

pkg/cloud/azureprovider.go
Sean-Holcomb: az.DownloadPricingData() needs to run when there is an update to the rate card configuration. The new endpoint doesn't deal with those config values, so it just holds up the return if it runs

pkg/costmodel/router.go
michaelmdresser: Got it. Thanks.

michaelmdresser (approved):

PR #1328, opened 2022-07-19
Diff asset tracks cost changes

nickcurie:
## What does this PR change?
* Diff asset function will keep track of total cost changes

* [Yes](https://github.com/kubecost/docs/pull/301)

* If an asset changed cost between two windows, it will be returned back to the user

* No

## How was this PR tested?
* Unit tests

* Yes

* No, don't have access

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -3,6 +3,7 @@ package kubecost
 import (
 	"encoding"
 	"fmt"
+	"math"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/kubecost/asset.go
```diff
@@ -2902,23 +2903,26 @@ type Diff[T any] struct {
...
 
 	for assetKey1, asset1 := range before.assets {
-		if _, ok := after.assets[assetKey1]; !ok {
+		if asset2, ok := after.assets[assetKey1]; !ok {
 			d := Diff[Asset]{asset1, DiffRemoved}
-			changedItems = append(changedItems, d)
+			changedItems[assetKey1] = d
+		} else if asset1.TotalCost() != asset2.TotalCost() {
+			d := Diff[Asset]{asset1, DiffChanged}
+			changedItems[assetKey1] = d
```
michaelmdresser: Since we have two assets to compare now, what do you think about the `Diff` type including a before and after for `T`, so users can see why the change happened? Like so:

```go
type Diff[T any] struct {
    Before T
    After T
    Kind DiffKind
}
```

We have this opportunity because we're messing with the API schema a bit anyway.

sundeepghuman:
Should we have a threshold for cost change here?

pkg/kubecost/asset.go
```diff
@@ -2902,27 +2904,34 @@ type Diff[T any] struct {
...
 
-// DiffAsset takes two AssetSets and returns a slice of Diffs by checking
-// the keys of each AssetSet. If a key is not found, a Diff is generated
-// and added to the slice.
-func DiffAsset(before, after *AssetSet) []Diff[Asset] {
-	changedItems := []Diff[Asset]{}
+// DiffAsset takes two AssetSets and returns a map of keys to Diffs by checking
+// the keys of each AssetSet. If a key is not found or is found with a different total cost,
+// a Diff is generated and added to the map.
+func DiffAsset(before, after *AssetSet, ratioCostChange float64) (map[string]Diff[Asset], error) {
```
michaelmdresser: Let's add a bit to the comment about the cost change. Something like "If an asset exists in before and after and the total cost after is greater than ratio*before total cost, the asset will be added to the diff"

michaelmdresser (commented):
Thanks for adding the cost change parameter as suggested by Sundeep.

The functionality looks solid to me, but now that we're truly comparing 2 assets, I think adding the Before and After field on the Diff type as I suggested earlier is a good idea here. That way, users can inspect why a cost change occurred. What do you think?

pkg/kubecost/asset.go
nickcurie: I agree that this is necessary, will work on getting this added

michaelmdresser (approved):

Sean-Holcomb (approved):
Looks good, nice unit tests

PR #1337, opened 2022-07-25
check secret for values on first load of rate card

### LINKED ISSUE #1331: Azure ratecard setting azureClientSecret does not update on changes to secret
I have found that changing the azureClientSecret in the secret does not change the value. This manifests in rate card continuing to fail when it is changed from an incorrect value to a correct value and not failing when changed from a correct value to an incorrect value. This is in contrast with other ratecard configuration values which appear to change when the secret is edited. 
Related to ZD#https://kubecost.zendesk.com/agent/tickets/2120
and
https://github.com/kubecost/kubecost-cost-model/issues/853

Sean-Holcomb:

## What does this PR change?
* This PR fixes a bug where changes to the rate card configuration secret or helm values for the `azureClientPassword` would not update. This issue has been difficult to detect because it is redacted when attempting to retrieve it from the api, and the other values from the secret do update. From what I can tell this stems from the fact that the other values are pulled into a config map which is populating the provider config. Because of this if a user had entered an incorrect `azureClientPassword`  in the past it would be kept permanently as the criteria for reloading the secret would never be met (one of the rate card config values == ""). To address this I have made it so the rate card api always favors the secret configuration. If a user wants to use API configuration (this has been removed from the front end) they will have to remove the secret. 

* 

* Fix issue with rate card api configuration

* Closes https://github.com/opencost/opencost/issues/1331

## How was this PR tested?
* Tested on azure cluster updating configurations

* 

*

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -511,23 +511,21 @@ func (ask *AzureServiceKey) IsValid() bool {
 
 // Loads the azure authentication via configuration or a secret set at install time.
 func (az *Azure) getAzureRateCardAuth(forceReload bool, cp *CustomPricing) (subscriptionID, clientID, clientSecret, tenantID string) {
-	// 1. Check config values first (set from frontend UI)
-	if cp.AzureSubscriptionID != "" && cp.AzureClientID != "" && cp.AzureClientSecret != "" && cp.AzureTenantID != "" {
-		subscriptionID = cp.AzureSubscriptionID
-		clientID = cp.AzureClientID
-		clientSecret = cp.AzureClientSecret
-		tenantID = cp.AzureTenantID
-
-		return
-	}
-
-	// 2. Check for secret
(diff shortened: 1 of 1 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -530,6 +520,14 @@ func (az *Azure) getAzureRateCardAuth(forceReload bool, cp *CustomPricing) (subs
 		tenantID = s.ServiceKey.Tenant
 		return
 	}
+	// 2. Check config values  (set though endpoint through endpoint)
```
biancaburtoiu: (nit) small typo here

mbolt35 (approved):
This looks good to me. Got me thinking, is there any reasonable way to test all of our infinite permutations of authentication mechanisms outside of our integration tests? It's a lot to keep track of, so we probably want to come up with a good way to unit test parts of this soon.

PR #1340, opened 2022-08-02
updated prometheus library

### LINKED ISSUE #1338: CVE-2022-21698 in dependency github.com/prometheus/client_golang v1.0.0
Dear Opencost team

github.com/prometheus/client_golang v1.0.0 is having High CVE CVE-2022-21698

Any patch to address it? thank you

```
===================================
Total: 1 (HIGH: 1, CRITICAL: 0)

+-------------------------------------+------------------+----------+-------------------+---------------+---------------------------------------+
|               LIBRARY               | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |                 TITLE                 |
+-------------------------------------+------------------+----------+-------------------+---------------+---------------------------------------+
| github.com/prometheus/client_golang | CVE-2022-21698   | HIGH     | v1.0.0            | 1.11.1        | prometheus/client_golang:             |
|                                     |                  |          |                   |               | Denial of service using               |
|                                     |                  |          |                   |               | InstrumentHandlerCounter              |
|                                     |                  |          |                   |               | -->avd.aquasec.com/nvd/cve-2022-21698 |
+-------------------------------------+------------------+----------+-------------------+---------------+---------------------------------------+


```

Regards,
Eric

nickcurie:

## What does this PR change?
* Updates prometheus library

* [Yes](https://github.com/kubecost/kubecost-cost-model/pull/900)

* 

* Closes https://github.com/opencost/opencost/issues/1338

## How was this PR tested?
* 

* No

* No, don't have access


pkg/prom/query.go
```diff
@@ -354,44 +342,17 @@ func (ctx *Context) RawQueryRange(query string, start, end time.Time, step time.
...
-	for _, w := range warnings {
-		// NoStoreAPIWarning is a warning that we would consider an error. It returns partial data relating only to the
-		// store apis which were reachable. In order to ensure integrity of data across all clusters, we'll need to identify
-		// this warning and convert it to an error.
-		if IsNoStoreAPIWarning(w) {
-			return nil, warnings, CommErrorf("Error: %s, Body: %s, Query: %s", w, body, query)
-		}
-
-		log.Warnf("fetching query '%s': %s", query, w)
-	}
```
michaelmdresser, line 379: @mbolt35 this is my main concern in this PR and I'm not sure what impact this might have. Pulling you in in the hope that you have some context. If it's worth talking about, I'm down to hop on a call with you and Nick to discuss.

Nick is currently working on the KCM changes because the prom API change is pretty wide-reaching.

michaelmdresser: Bolt is out, whoops. @AjayTripathy same question for you.

mbolt35: Yeah, how else do we get warnings? Even when the prom client wasn't returning them, we were still able to parse the warnings from the body. There were clear comments on why this was necessary. We also use this same path for Thanos. I don't think this is a desired change.

michaelmdresser: Thanks Bolt, that was my suspicion. 

We'll do some investigation to find out where warnings end up now (maybe in the standard response rather than a separate return). We have to make this change due to a security problem in the library (see linked issue) but we'll spend more time with it.

mbolt35: Even in the prom client we were using, the warnings always came back as nil. However, when warnings were actually sent back, they appeared in the body. That's why I had the "warningsFor" method which would parse any warnings that showed up on the body.

Note that the way we're using the prom client, we could just substitute a regular http client. My proposal is that we move to using an actual http client, OR move to using the API features on the client directly (assuming they're optimized). I think that leveraging the prom cli APIs might be a substantial task given the custom implementations we've written around query throttling and rate limit handling, but I'm supportive either way.

nickcurie: Michael and I went back and discussed what the changes to the prom API are and readded the warningsFrom method and its calls. Prom changed where they were sending the warnings in their API, so that has been resolved now.

We figured moving to a regular http client would be out of scope for the issue this PR was meant to resolve.

nickcurie:
Closing this PR in favor of https://github.com/opencost/opencost/pull/1344, since it made more sense to me to create a new branch and start over just in case something was missed in the transition.

PR #1342, opened 2022-08-02
allow the ability to disable writing to storage

Sean-Holcomb:
## What does this PR change?
* This PR adds a layer between the storage object and it creator which can block certain operation. In this case using an env variable ETL_STORE_READ_ONLY it will disable writes and deletes. This prevents an attached instance of kubecost from changing the store.

* https://github.com/kubecost/kubecost-cost-model/pull/901

* this feature is not meant for users

* Closes ...

## How was this PR tested?
* Tested by setting the Environmental variable variable to true and ensuring that only loads were occurring.

* Documentation to be included in the deployments repo and etl data generation repo

*

### DIFF
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -67,6 +67,7 @@ const (
 	ETLEnabledEnvVar                     = "ETL_ENABLED"
 	ETLMaxPrometheusQueryDurationMinutes = "ETL_MAX_PROMETHEUS_QUERY_DURATION_MINUTES"
 	ETLResolutionSeconds                 = "ETL_RESOLUTION_SECONDS"
+	ETLStoreReadOnlyMode                 = "ETL_STORE_READ_ONLY"
 	LegacyExternalAPIDisabledVar         = "LEGACY_EXTERNAL_API_DISABLED"
 
 	PromClusterIDLabelEnvVar = "PROM_CLUSTER_ID_LABEL"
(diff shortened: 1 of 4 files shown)


pkg/env/costmodelenv.go
```diff
@@ -67,6 +67,7 @@ const (
 	ETLEnabledEnvVar                     = "ETL_ENABLED"
 	ETLMaxPrometheusQueryDurationMinutes = "ETL_MAX_PROMETHEUS_QUERY_DURATION_MINUTES"
 	ETLResolutionSeconds                 = "ETL_RESOLUTION_SECONDS"
+	ETLStoreReadOnlyMode                 = "ETL_STORE_READ_ONLY"
```
nikovacevic, line 70: Agree with Bolt's comment that it's rough already having `ETL_READ_ONLY`... but this seems like making the best of a messy situation.

nikovacevic (approved):

mbolt35 (changes requested):
We definitely need some changes here. 

1. The `storage.Storage` APIs are not specific to ETL. They're used for shared config files and Provider Configuration storage as well. If you were to turn this on right now, then writing the updated Provider config would stop working. 
2. There's nothing in opencost that uses our ETL pipeline, so it's _really_ confusing to have ETL variables in our configuration. This seems to indicate we're leaking implementation we shouldn't be. 

My suggestion is that we make this change in our private repository, not at a core API level. You can just as easily wrap a _specific_ `storage.Storage` implementation within the ETL API (versus applying universal access rules to _all_ file storage based on an ETL specific configuration). 

Hope that makes sense, let me know if you have any questions.

nikovacevic:
Good catch, Bolt.

Sean-Holcomb:
Following @mbolt35 suggestions and moving changes to KCM

PR #1352, opened 2022-08-08
Making storage class available for persistent volume by pulling infor…

avrodrigues5:
…mation from Prometheus metrics and assigning local storage storageClass as Local

Signed-off-by: Alan Rodrigues <alanr5691@yahoo.com>

## What does this PR change?
*  Disk objects will have storage class, Storage class information is gather from custom Prometheus metrics `kubecost_pv_info` and local storage will have default storage class as `__local__` and unknown will have storage class `__unknown`.

*[ KCM PR](https://github.com/kubecost/kubecost-cost-model/pull/909)
*[cost-analyzer-helm-chart PR](https://github.com/kubecost/cost-analyzer-helm-chart/pull/1619)

*  Probably None from the Backend, it will have storage class information for Persistent volumes and Local storage.

* Closes ...[Issue Link](https://github.com/kubecost/cost-analyzer-helm-chart/issues/1575) 

## How was this PR tested?
*  KCM image pointing to respective Opencost branch was put in the Thanos backend cluster i have in my namespace and the information about gather from 'kubectl get pv' and local storage of external cluster was listed in [Rest Call ](http://34.170.146.238:9095/model/assets?window=1d&aggregate=&accumulate=true&filterTypes=Disk), Placed around with different setting like 1h, 7h and 1d. 

* No, not that I am aware, but if it needs i will do it.

* This was Bug for v1.96 but can go in v1.97 or nightly,upto Adam and others discretion

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -351,11 +351,6 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 		if disk.ProviderID == "" {
 			disk.ProviderID = disk.Name
 		}
-
-		// Explicitly specify unknown storage class for disk whose information is unavailable in prometheus metrics of kubecost_pv_info
-		if disk.StorageClass == "" {
-			disk.StorageClass = kubecost.UnknownStorageClass
-		}
 	}
 
 	return diskMap, nil
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -106,16 +106,17 @@ func NewClusterCostsFromCumulative(cpu, gpu, ram, storage float64, window, offse
 }
 
 type Disk struct {
-	Cluster    string
-	Name       string
-	ProviderID string
-	Cost       float64
-	Bytes      float64
-	Local      bool
-	Start      time.Time
-	End        time.Time
-	Minutes    float64
-	Breakdown  *ClusterCostsBreakdown
+	Cluster      string
+	Name         string
+	ProviderID   string
+	StorageClass string
+	Cost         float64
+	Bytes        float64
+	Local        bool
+	Start        time.Time
+	End          time.Time
+	Minutes      float64
+	Breakdown    *ClusterCostsBreakdown
 }
 
 type DiskIdentifier struct {
@@ -156,6 +157,7 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 	queryPVCost := fmt.Sprintf(`avg(avg_over_time(pv_hourly_cost[%s])) by (%s, persistentvolume,provider_id)`, durStr, env.GetPromClusterLabel())
 	queryPVSize := fmt.Sprintf(`avg(avg_over_time(kube_persistentvolume_capacity_bytes[%s])) by (%s, persistentvolume)`, durStr, env.GetPromClusterLabel())
 	queryActiveMins := fmt.Sprintf(`count(pv_hourly_cost) by (%s, persistentvolume)[%s:%dm]`, env.GetPromClusterLabel(), durStr, minsPerResolution)
+	queryPVStorageClass := fmt.Sprintf(`avg(avg_over_time(kube_persistentvolumeclaim_info[%s])) by (%s, volumename, storageclass)`, durStr, env.GetPromClusterLabel())
 
 	queryLocalStorageCost := fmt.Sprintf(`sum_over_time(sum(container_fs_limit_bytes{device!="tmpfs", id="/"}) by (instance, %s)[%s:%dm]) / 1024 / 1024 / 1024 * %f * %f`, env.GetPromClusterLabel(), durStr, minsPerResolution, hourlyToCumulative, costPerGBHr)
 	queryLocalStorageUsedCost := fmt.Sprintf(`sum_over_time(sum(container_fs_usage_bytes{device!="tmpfs", id="/"}) by (instance, %s)[%s:%dm]) / 1024 / 1024 / 1024 * %f * %f`, env.GetPromClusterLabel(), durStr, minsPerResolution, hourlyToCumulative, costPerGBHr)
@@ -165,6 +167,7 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 	resChPVCost := ctx.QueryAtTime(queryPVCost, t)
 	resChPVSize := ctx.QueryAtTime(queryPVSize, t)
 	resChActiveMins := ctx.QueryAtTime(queryActiveMins, t)
+	resChPVStorageClass := ctx.QueryAtTime(queryPVStorageClass, t)
 	resChLocalStorageCost := ctx.QueryAtTime(queryLocalStorageCost, t)
 	resChLocalStorageUsedCost := ctx.QueryAtTime(queryLocalStorageUsedCost, t)
 	resChLocalStorageBytes := ctx.QueryAtTime(queryLocalStorageBytes, t)
@@ -173,10 +176,12 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 	resPVCost, _ := resChPVCost.Await()
 	resPVSize, _ := resChPVSize.Await()
 	resActiveMins, _ := resChActiveMins.Await()
+	resPVStorageClass, _ := resChPVStorageClass.Await()
 	resLocalStorageCost, _ := resChLocalStorageCost.Await()
 	resLocalStorageUsedCost, _ := resChLocalStorageUsedCost.Await()
 	resLocalStorageBytes, _ := resChLocalStorageBytes.Await()
 	resLocalActiveMins, _ := resChLocalActiveMins.Await()
+
 	if ctx.HasErrors() {
 		return nil, ctx.ErrorCollection()
 	}
@@ -208,6 +213,9 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 			}
 		}
 		diskMap[key].Cost += cost
+
+		//Assigning explicitly the storage class of local storage to local
+		diskMap[key].StorageClass = "Local"
 	}
 
 	for _, result := range resLocalStorageUsedCost {
@@ -297,6 +305,34 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 		diskMap[key].Minutes = mins
 	}
 
+	//Iterating through Persistent Volume given by kube_persistentvolumeclaim_info and assign the storage class if known and unknown if not known.
+	for _, result := range resPVStorageClass {
+		cluster, err := result.GetString(env.GetPromClusterLabel())
+		if err != nil {
+			cluster = env.GetClusterID()
+		}
+
+		name, _ := result.GetString("volumename")
+
+		key := DiskIdentifier{cluster, name}
+		if _, ok := diskMap[key]; !ok {
+			log.DedupedWarningf(5, "ClusterDisks: Storage class information for unidentified disk or disk deleted from analysis")
+			continue
+		}
+
+		if len(result.Values) == 0 {
+			continue
+		}
+
+		storageClass, err := result.GetString("storageclass")
+
+		if err != nil {
+			diskMap[key].StorageClass = "Unknown"
+		} else {
+			diskMap[key].StorageClass = storageClass
+		}
+	}
+
 	for _, disk := range diskMap {
 		// Apply all remaining RAM to Idle
 		disk.Breakdown.Idle = 1.0 - (disk.Breakdown.System + disk.Breakdown.Other + disk.Breakdown.User)
@@ -304,6 +340,11 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 		// Set provider Id to the name for reconciliation
 		if disk.ProviderID == "" {
 			disk.ProviderID = disk.Name
+		}
+
+		// Explicitly specify unknown storage class for disk whose information is unavailable in prometheus metrics of kube_persistentvolumeclaim_info
+		if disk.StorageClass == "" {
+			disk.StorageClass = "Unknown"
 		}
 	}
 
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -305,7 +305,7 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 		diskMap[key].Minutes = mins
 	}
 
-	//Iterating through Persistent Volume given by kube_persistentvolumeclaim_info and assign the storage class if known and unknown if not known.
+	//Iterating through Persistent Volume given by custom metrics kubecost_pv_info and assign the storage class if known and __unknown__ if not populated.
 	for _, result := range resPVStorageClass {
 		cluster, err := result.GetString(env.GetPromClusterLabel())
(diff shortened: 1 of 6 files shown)


pkg/costmodel/cluster.go
```diff
@@ -156,6 +157,7 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end
 	queryPVCost := fmt.Sprintf(`avg(avg_over_time(pv_hourly_cost[%s])) by (%s, persistentvolume,provider_id)`, durStr, env.GetPromClusterLabel())
 	queryPVSize := fmt.Sprintf(`avg(avg_over_time(kube_persistentvolume_capacity_bytes[%s])) by (%s, persistentvolume)`, durStr, env.GetPromClusterLabel())
 	queryActiveMins := fmt.Sprintf(`count(pv_hourly_cost) by (%s, persistentvolume)[%s:%dm]`, env.GetPromClusterLabel(), durStr, minsPerResolution)
+	queryPVStorageClass := fmt.Sprintf(`avg(avg_over_time(kube_persistentvolumeclaim_info[%s])) by (%s, volumename, storageclass)`, durStr, env.GetPromClusterLabel())
```
michaelmdresser: I think we should be getting our information from all Persistent Volumes (PVs), not just Persistent Volume Claims (PVCs). What happens if the cluster has an unclaimed volume?

pkg/costmodel/cluster.go
```diff
@@ -208,6 +213,9 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end
 			}
 		}
 		diskMap[key].Cost += cost
+
+		//Assigning explicitly the storage class of local storage to local
+		diskMap[key].StorageClass = "Local"
```
michaelmdresser: Looking at the StorageClass definition [1] it looks like the `metadata.name` field can be an arbitrary string, so a user may be able to give a custom StorageClass the name `Local`. Is there a good string we could use to avoid this conflict? I know K8s maintains some limitations on strings, I just don't remember them off the top of my head.

Also, if we're going to be hardcoding strings for purposes like this, I think we should use a const, like we do for Allocation special strings: https://github.com/opencost/opencost/blob/cda202ddedb57d02bc307ea63052b29f1a504c97/pkg/kubecost/allocation.go#L21-L45

[1] https://kubernetes.io/docs/concepts/storage/storage-classes/#the-storageclass-resource

pkg/costmodel/cluster.go
```diff
@@ -297,6 +305,34 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end
...
+		}
+
+		if len(result.Values) == 0 {
+			continue
+		}
+
+		storageClass, err := result.GetString("storageclass")
+
+		if err != nil {
+			diskMap[key].StorageClass = "Unknown"
```
michaelmdresser: Similar question about this string.

pkg/kubecost/asset.go
```diff
@@ -1044,16 +1044,17 @@ func (cm *ClusterManagement) String() string {
...
+	properties   *AssetProperties
+	start        time.Time
+	end          time.Time
+	window       Window
+	adjustment   float64
+	Cost         float64
+	ByteHours    float64
+	Local        float64
+	Breakdown    *Breakdown
+	StorageClass string // @bingen:field[version=17]
```
michaelmdresser: Will Kubernetes allow 2 disks of the same `AssetProperties` but of different storage classes? If it does, we may have to think deeper about how this might impact this https://github.com/opencost/opencost/blob/0094cba41fc0827e1598ceba9f6901739aed6386/pkg/kubecost/asset.go#L2744

pkg/costmodel/cluster.go
michaelmdresser: We already emit the storage class for PVCs: https://github.com/opencost/opencost/blob/0094cba41fc0827e1598ceba9f6901739aed6386/pkg/metrics/pvcmetrics.go#L117

so it should be too hard to add that to our emission of data for PVs:
https://github.com/opencost/opencost/blob/0094cba41fc0827e1598ceba9f6901739aed6386/pkg/metrics/pvmetrics.go#L73

avrodrigues5: When I added `kube_persistentvolume_capacity_bytes`  to have a label called `storageclass` what happens is for instance in below image there are 4 entries of PV, 2 for each PV , i believe the first 2 are from KSM and the below 2 are written by us, correct me if i'm wrong here. When i added storageclass only the below two had `storageclass` label. Now for the query such as `avg(avg_over_time(kube_persistentvolume_capacity_bytes[<duration>])) by (<Cluster_id>, persistentvolume)` to get the capacity if i added storageClass in group by to make it active to use `GetString` i believe it will not take into account the first two entry of KSM, i don't know the ramification for this so i collect the PV volume information from the metric avg(avg_over_time(kube_persistentvolumeclaim_info[<duration>])) by (<cluster_id>, volumename, storageclass). 
Screen Shot 2022-08-08 at 9 19 33 AM

michaelmdresser: Hmm, I see how this could be a problem. It looks like regular KSM currently has a `kube_persistentvolume_info` metric [1], which contains exactly the kind of data we're looking for. However, we haven't required this metric in the past so we would have to write a new emitter for it if we wanted to use it.

I'd like to call in some better metric knowledge here. @mbolt35 could you perhaps weigh in? I'm currently leaning towards adding a metric emitter for `kube_pv_info` but I'm not certain if its a good idea.

[1] https://github.com/kubernetes/kube-state-metrics/blob/master/docs/persistentvolume-metrics.md

pkg/kubecost/asset.go
avrodrigues5: Kubernetes doesn't allow this. I can confirm that, ran the following test: 

I tried to create two storage classes `test-sc1 `and `test-sc2` and then created the same persistent volume name `pv-common` only changing the storage classes and i got the following error message 
`The PersistentVolumeClaim "pv-common" is invalid: spec: Forbidden: spec is immutable after creation except resources.requests for bound claims
  core.PersistentVolumeClaimSpec{
  	... // 2 identical fields
  	Resources:        {Requests: {s"storage": {i: {...}, s: "10Gi", Format: "BinarySI"}}},
  	VolumeName:       "",
- 	StorageClassName: &"test-sc2",
+ 	StorageClassName: &"test-sc1",
  	VolumeMode:       &"Filesystem",
  	DataSource:       nil,
  	DataSourceRef:    nil,
  }`


Which only makes sense as Persistent volume is not namespace bound but persistent volume claim is namespace bound. i.e 2 different namespace can have same pvc name. There's also one to one mapping between PV and PVC. For two disk asset property to be same PV name should be same as it is part of the [asset property struct](https://github.com/avrodrigues5/opencost/blob/72c7d197a7d7ce3f9283312db979d8eab25496d0/pkg/kubecost/assetprops.go#L134).

pkg/costmodel/cluster.go
avrodrigues5: To answer question on unclaimed persistent volume, I created an unclaimed persistent volume by creating a persistent volume claim and not assign it to any pod, attached is the image, i think we dont print any information about unclaimed persistent volume. The different metrics used to compute cost and capacity for pv volume `pv_hourly_cost`, `kube_persistentvolume_capacity_bytes` doesnt have information about unclaimed PVs. 

The interesting thing is the metrics `kube_persistentvolumeclaim_info` actually has the data but since we never create a disk object for unclaimed pv it was never used attached is the image.

Screen Shot 2022-08-08 at 3 53 56 PM

Screen Shot 2022-08-08 at 3 48 07 PM

nikovacevic: This seems like the key decision for this PR. Talked with Bolt, and we agree with you Michael -- we should probably emit a new metric (something like `kubecost_pv_info` or similar) that has `storageclass` (and whatever else we want, but can start there).

We'd recommend putting that logic here: https://github.com/opencost/opencost/blob/develop/pkg/metrics/pvmetrics.go

And for examples, you can look to the KSM implementation of `kube_persistentvolume_info` here: https://github.com/kubernetes/kube-state-metrics/blob/7e28ed2722a8edfade41238ee3f0e0117ff041b6/internal/store/persistentvolume.go#L161-L249

Please let me know if I can help with that at all!

Sound good to you, @michaelmdresser?

pkg/kubecost/asset.go
nikovacevic: This is a good question. Let's check on that -- I'm guessing it's fine as-is, but we'll want to be sure.

pkg/costmodel/cluster.go
nikovacevic: Agree, constants would be nice

nikovacevic (changes requested):
The big change we'll want to make is to emit that new metric. Let me know if you have any questions or concerns about that!

pkg/costmodel/cluster.go
michaelmdresser: That sounds excellent, thanks Niko! I agree that we should prefix it with `kubecost_` -- we're not thinking about a KSM compatibility thing like we were in the past.

pkg/costmodel/cluster.go
avrodrigues5: Is there a particular name you would like to differentiate local storage detected by kubecost since user can create a custom `Local` storage class? I think we can name it `__local__` since kubernetes resource name start with alphanumeric character and not a `- `so it can be unique and also as per the allocation type constants but let me know any suggestion on this. I will get all of the recommendation fixed and update this PR ASAP

michaelmdresser: `__local__` sounds perfect to me! Thanks for digging into the resource names, that gives me confidence.

AjayTripathy:
Is this adding new fields to the ETL? Did we carefully ensure that no rebuilds are triggered and that all schemas have been updated correctly?

mbolt35:
@AjayTripathy I can confirm that he set the versions correctly and rebuilt the bingen codecs. This is exactly what is required.

avrodrigues5:
Hello @nikovacevic , @michaelmdresser and @mbolt35 , could you please give any further review/approval so that i can get this PR merged?

-regards
Alan

pkg/costmodel/cluster.go
```diff
@@ -297,6 +305,34 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end
...
+			cluster = env.GetClusterID()
+		}
+
+		name, _ := result.GetString("persistentvolume")
+
+		key := DiskIdentifier{cluster, name}
+		if _, ok := diskMap[key]; !ok {
+			log.DedupedWarningf(5, "ClusterDisks: Storage class information for unidentified disk or disk deleted from analysis")
+			continue
+		}
```
michaelmdresser: If this is worth a log, let's also include the cluster/name combo in the log message!

pkg/kubecost/asset.go
```diff
@@ -1241,21 +1248,26 @@ func (d *Disk) add(that *Disk) {
 	d.Cost += that.Cost
 
 	d.ByteHours += that.ByteHours
+
+	if that.StorageClass != UnknownStorageClass {
+		d.StorageClass = that.StorageClass
+	}
```
michaelmdresser: I'm not entirely sure if this is the right way to handle StorageClass in `add()`. If I'm adding something with StorageClass `a` to something with StorageClass `b`, the StorageClass of the result is probably neither, right?

Check out how we do `d.Properties().Merge(that.Properties())` above -- I think it could be a useful guide.

michaelmdresser (commented):
This looks much better now that we have the new metric! Nice work with that. My only remaining blocking question is about `.add()`.

pkg/kubecost/asset.go
avrodrigues5: I might be wrong but add is called by Insert which is called by functions NewAssetSet, AggregatedBy, accumulate and InsertRange(called in KCM).

Believe the first two and last one should be okay with this if statement, but i think the accumulate might be impacted:
If a disk storage class is changed at any point of time the latest one should be listed not the __unknown__ . For example we are accumulating over 7d and first 3d the storageClass for a disk was was `standard` and the next 4 days storageClass was `standard-rwo`. It should show `standard-rwo` and not unknown in my opinion. Correct me if i'm wrong?

avrodrigues5: we can't change Storage class of a Persistent volume, I tired, you need to provision a new PV claim which results in new persistent volume name with another storage class.

michaelmdresser: Let's say I create a PV named `PV1` with storage class `SC1`. I use it for a day, then decide I need a different storage class. I delete `PV1`, then create a new PV named `PV1` with storage class `SC2`. Is this allowed? If so...

I'm pretty sure AggregateBy/Accumulate/Insert will treat these as the same "thing" and so we'll end up in this storage class case. Because we're combining the data from two same-named PVs with different storage classes, the storage class of the resulting Disk is ambiguous.

This is essentially a rephrasing of my question here: https://github.com/opencost/opencost/pull/1352#discussion_r940393985.

@nikovacevic a second opinion here would be helpful, because I think we're close to blurring the lines between AssetProperties and Disk-specific properties here in a way that might interact oddly with aggby/etc.

pkg/costmodel/cluster.go
michaelmdresser: In light of our log discussion last week, what do you think about storing this warning info and then logging a summary outside of the loop? I definitely won't block this PR on this, just an idea if you feel inspired :slightly_smiling_face:

pkg/kubecost/asset.go
nikovacevic: Yep, I think @michaelmdresser is correct -- we're going to want to keep the StorageClass only if they match. Otherwise, we'll want no StorageClass (i.e.`""`).

The simplest way to understand this is to ask:  what happens if you aggregate by "type"? You'll get results of `Node`, `Disk`, etc. And what will the storage class of that one, aggregated `Disk` row be? Well, if they all match, then you can persist that property through all the `add` calls. If more than one storage class is involved, then we'll need to wipe that field.

To be explicit, consider the two following scenarios:

Scenario 1
* Disk D1 has storage class SC1
* Disk D2 has storage class SC1
* Disk D3 has storage class SC1
* D1 + D2 + D3 = DT
* DT has storage class SC1

Scenario 2
* Disk D1 has storage class SC1
* Disk D2 has storage class SC1
* Disk D3 has storage class SC2
* D1 + D2 + D3 = DT
* DT has no storage class (i.e. "")

As written, I believe scenario 2 would result in DT.StorageClass = SC2, which would be incorrect.

Does that make sense? Let me know if I can help @avrodrigues5 !

pkg/kubecost/asset.go
```diff
@@ -1254,21 +1261,26 @@ func (d *Disk) add(that *Disk) {
...
```
nikovacevic: This is the major one we need to fix. See Michael's question, and my comment.

nikovacevic (changes requested):
I think we can probably just fix that one issue, then call this PR good. We'll also want to pull in Bolt's recent changes to these files and rebase.

nikovacevic (approved):
Thank you, Alan!

mbolt35 (approved):
In terms of the implementation itself, this all looks good to me. Most of the questions I had surrounding the addition of the new metric and how we'll handle the default empty storage class I think were answered in this PR. 

Awesome addition @avrodrigues5 !

PR #1354, opened 2022-08-10
ETL Data Structure Cleanup Part 1

mbolt35:
## What does this PR change?
* Updates all ETL Data Structures:
  * Removes Mutex, as locking hasn't been necessary for a while now. 
  * Changes access for fields to public
  * Remove the `Each` function, as regular map iteration is possible now. 
  * Remove the `Map` function, as it was really confusing and expensive.
  * Asset fields changed from ie: `Window()` to `GetWindow()`
  * Asset interface adds `SetWindow(window Window)`

##### Ignoring Fields
You can also ignore fields in the generation process. This is useful if you have a transient field you wish to be ignored during marshal and unmarshal. This is done by adding the `@bingen:field[ignore]` annotation directly after the field:

```go
type Person struct {
      FirstName string 
      LastName string
      FullName string //@bingen:field[ignore]
}
```

In this example, `FirstName` and `LastName` will both be marshalled and unmarshalled. `FullName` will be ignored.

##### Pre and Post Processing Types
You can apply pre and post processing hooks to any generated type. These hooks often work well with ignored/transient fields to ensure that any data being encoded can be populated from the transient data. Likewise, you can also ensure that any transient fields are populated on decode as well. These hooks are enabled via the `generate` options:

```go
// @bingen:generate[stringtable,preprocess,postprocess]:Person
```

Using the `Person` from the previous example, let's have the pre and post processor functions manage the `FullName` field. It's important to note that when you mark a type with `preprocess` and/or `postprocess`, you must also ensure the existence of two functions in the package you have targetting for generation: 

For the `preprocess`, the function name will be `preProcess<Type>(myType *<Type>)`. For our `Person` example, it would look like this:
```go
func preProcessPerson(p *Person) {
      // If the FullName field is set, update the FirstName and LastName fields
      if p.FullName != "" {
            firstAndLast := strings.Split(p.FullName, " ")
            // Make FullName contains a first and last name separated by space
            if len(firstAndLast) != 2 {
                  return
            }

            p.FirstName = firstAndLast[0]
            p.LastName = firstAndLast[1]
      }
}
```

For the `postprocess`, the function name will be `postProcess<Type>(myType *<Type>)`. For our `Person` example, it would look like this:
```go
func postProcessPerson(p *Person) {
      // Set the FullName field to the concatenation of FirstName and LastName separated by a space
      p.FullName = fmt.Sprintf("%s %s", p.FirstName, p.LastName)
}
```

## AssetSet Concrete Asset Type Projection
image

* It should not impact users - code maintenance.

## How was this PR tested?
* ETL Unit Tests, Live Environment Tests, Still needs more iteration on testing.

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/opencost/opencost/pkg/log"
@@ -109,6 +108,30 @@ type RawAllocationOnlyData struct {
 	RAMBytesUsageMax float64 `json:"ramByteUsageMax"`
 }
 
+// Clone returns a deep copy of the given RawAllocationOnlyData
+func (r *RawAllocationOnlyData) Clone() *RawAllocationOnlyData {
+	if r == nil {
+		return nil
+	}
+
+	return &RawAllocationOnlyData{
+		CPUCoreUsageMax:  r.CPUCoreUsageMax,
+		RAMBytesUsageMax: r.RAMBytesUsageMax,
+	}
+}
+
+// Equal returns true if the RawAllocationOnlyData is approximately equal
+func (r *RawAllocationOnlyData) Equal(that *RawAllocationOnlyData) bool {
+	if r == nil && that == nil {
+		return true
+	}
+	if r == nil || that == nil {
+		return false
+	}
+	return util.IsApproximately(r.CPUCoreUsageMax, that.CPUCoreUsageMax) &&
+		util.IsApproximately(r.RAMBytesUsageMax, that.RAMBytesUsageMax)
+}
+
 // PVAllocations is a map of Disk Asset Identifiers to the
 // usage of them by an Allocation as recorded in a PVAllocation
 type PVAllocations map[PVKey]*PVAllocation
@@ -148,6 +171,30 @@ func (pv PVAllocations) Add(that PVAllocations) PVAllocations {
 	return apv
 }
 
+// Equal returns true if the two PVAllocations are equal in length and contain the same keys
+// and values.
+func (this PVAllocations) Equal(that PVAllocations) bool {
+	if this == nil && that == nil {
+		return true
+	}
+	if this == nil || that == nil {
+		return false
+	}
+
+	if len(this) != len(that) {
+		return false
+	}
+
+	for k, pv := range this {
+		tv, ok := that[k]
+		if !ok || !pv.Equal(tv) {
+			return false
+		}
+	}
+
+	return true
+}
+
 // PVKey for identifying Disk type assets
 type PVKey struct {
 	Cluster string `json:"cluster"`
@@ -174,6 +221,19 @@ func (pvk *PVKey) FromString(key string) error {
 type PVAllocation struct {
 	ByteHours float64 `json:"byteHours"`
 	Cost      float64 `json:"cost"`
+}
+
+// Equal returns true if the two PVAllocation instances contain approximately the same
+// values.
+func (pva *PVAllocation) Equal(that *PVAllocation) bool {
+	if pva == nil && that == nil {
+		return true
+	}
+	if pva == nil || that == nil {
+		return false
+	}
+	return util.IsApproximately(pva.ByteHours, that.ByteHours) &&
+		util.IsApproximately(pva.Cost, that.Cost)
 }
 
 // AllocationMatchFunc is a function that can be used to match Allocations by
@@ -235,18 +295,6 @@ func (a *Allocation) Clone() *Allocation {
 		SharedCost:                 a.SharedCost,
 		ExternalCost:               a.ExternalCost,
 		RawAllocationOnly:          a.RawAllocationOnly.Clone(),
-	}
-}
-
-// Clone returns a deep copy of the given RawAllocationOnlyData
-func (r *RawAllocationOnlyData) Clone() *RawAllocationOnlyData {
-	if r == nil {
-		return nil
-	}
-
-	return &RawAllocationOnlyData{
-		CPUCoreUsageMax:  r.CPUCoreUsageMax,
-		RAMBytesUsageMax: r.RAMBytesUsageMax,
 	}
 }
 
@@ -329,34 +377,14 @@ func (a *Allocation) Equal(that *Allocation) bool {
 		return false
 	}
 
-	if a.RawAllocationOnly == nil && that.RawAllocationOnly != nil {
-		return false
-	}
-	if a.RawAllocationOnly != nil && that.RawAllocationOnly == nil {
+	if !a.RawAllocationOnly.Equal(that.RawAllocationOnly) {
 		return false
 	}
 
-	if a.RawAllocationOnly != nil && that.RawAllocationOnly != nil {
-		if !util.IsApproximately(a.RawAllocationOnly.CPUCoreUsageMax, that.RawAllocationOnly.CPUCoreUsageMax) {
-			return false
-		}
-		if !util.IsApproximately(a.RawAllocationOnly.RAMBytesUsageMax, that.RawAllocationOnly.RAMBytesUsageMax) {
-			return false
-		}
-	}
-
-	aPVs := a.PVs
-	thatPVs := that.PVs
-	if len(aPVs) == len(thatPVs) {
-		for k, pv := range aPVs {
-			tv, ok := thatPVs[k]
-			if !ok || *tv != *pv {
-				return false
-			}
-		}
-	} else {
+	if !a.PVs.Equal(that.PVs) {
 		return false
 	}
+
 	return true
 }
 
@@ -760,10 +788,9 @@ func (a *Allocation) add(that *Allocation) {
 // AllocationSet stores a set of Allocations, each with a unique name, that share
 // a window. An AllocationSet is mutable, so treat it like a threadsafe map.
 type AllocationSet struct {
-	sync.RWMutex
-	allocations  map[string]*Allocation
-	externalKeys map[string]bool
-	idleKeys     map[string]bool
+	Allocations  map[string]*Allocation
+	ExternalKeys map[string]bool
+	IdleKeys     map[string]bool
 	FromSource   string // stores the name of the source used to compute the data
 	Window       Window
 	Warnings     []string
@@ -774,9 +801,9 @@ type AllocationSet struct {
 // the given list of Allocations
 func NewAllocationSet(start, end time.Time, allocs ...*Allocation) *AllocationSet {
 	as := &AllocationSet{
-		allocations:  map[string]*Allocation{},
-		externalKeys: map[string]bool{},
-		idleKeys:     map[string]bool{},
+		Allocations:  map[string]*Allocation{},
+		ExternalKeys: map[string]bool{},
+		IdleKeys:     map[string]bool{},
 		Window:       NewWindow(&start, &end),
 	}
 
@@ -912,19 +939,16 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		Window: as.Window.Clone(),
 	}
 
-	as.Lock()
-	defer as.Unlock()
-
 	// (1) Loop and find all of the external, idle, and shared allocations. Add
 	// them to their respective sets, removing them from the set of allocations
 	// to aggregate.
-	for _, alloc := range as.allocations {
+	for _, alloc := range as.Allocations {
 		// External allocations get aggregated post-hoc (see step 6) and do
 		// not necessarily contain complete sets of properties, so they are
 		// moved to a separate AllocationSet.
 		if alloc.IsExternal() {
-			delete(as.externalKeys, alloc.Name)
-			delete(as.allocations, alloc.Name)
+			delete(as.ExternalKeys, alloc.Name)
+			delete(as.Allocations, alloc.Name)
 			externalSet.Insert(alloc)
 			continue
 		}
@@ -933,8 +957,8 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 		// shared later on. If they are not to be shared, then add them to the
 		// aggSet like any other allocation.
 		if alloc.IsIdle() {
-			delete(as.idleKeys, alloc.Name)
-			delete(as.allocations, alloc.Name)
+			delete(as.IdleKeys, alloc.Name)
+			delete(as.Allocations, alloc.Name)
 
 			if options.ShareIdle == ShareEven || options.ShareIdle == ShareWeighted {
 				idleSet.Insert(alloc)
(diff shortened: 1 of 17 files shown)


pkg/kubecost/allocation.go
```diff
@@ -1879,17 +1843,16 @@ func (as *AllocationSet) Map() map[string]*Allocation {
 		return map[string]*Allocation{}
 	}
 
-	return as.Clone().allocations
+	return as.Clone().Allocations
```
mbolt35: I'm not so sure this function is necessary. It's also really confusing and expensive.

mbolt35:
NOTE: These changes will not force a rebuild for ETL data. The field names are not part of the schema.

pkg/kubecost/allocation.go
nikovacevic: Let's nuke it. It shouldn't be necessary, now that the field is public.

nikovacevic:
> NOTE: These changes will not force a rebuild for ETL data. The field names are not part of the schema.

Wow, so cool.

Sean-Holcomb:
Should we also be looking at making the Asset types fields exported? I know they are mainly controlled via the interface. One notable missing control from the Assets interface is a SetWindow func, what do you think about putting that in?

michaelmdresser:
> NOTE: These changes will not force a rebuild for ETL data. The field names are not part of the schema.

Neat! Why is that? Does it just take field order into account?

pkg/kubecost/allocation.go
```diff
@@ -151,6 +174,23 @@ func (pv *PVAllocations) Add(that PVAllocations) PVAllocations {
 	return apv
 }
 
+// Equal returns true if the two PVAllocations are equal in length and contain the same keys
+// and values.
+func (aPVs *PVAllocations) Equal(thatPVs PVAllocations) bool {
+	if len(*aPVs) != len(thatPVs) {
+		return false
+	}
```
michaelmdresser: Should this `nil` check `aPVs`?

pkg/kubecost/allocation.go
```diff
@@ -2104,52 +2053,25 @@ func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*Allocati
...
-// Get retrieves the AllocationSet at the given index of the range.
-func (asr *AllocationSetRange) Get(i int) (*AllocationSet, error) {
-	if i < 0 || i >= len(asr.allocations) {
-		return nil, fmt.Errorf("AllocationSetRange: index out of range: %d", i)
-	}
-
-	asr.RLock()
-	defer asr.RUnlock()
-	return asr.allocations[i], nil
+	asr.Allocations = append(asr.Allocations, that)
```
michaelmdresser, line 2042: Perhaps we should nuke this function too?

pkg/kubecost/allocation.go
```diff
@@ -2331,33 +2261,74 @@ func (asr *AllocationSetRange) End() (time.Time, error) {
 	return end, nil
 }
 
+// Each iterates over all AssetSets in the AssetSetRange and returns the start and end over
+// the entire range
+func (asr *AllocationSetRange) StartAndEnd() (time.Time, time.Time, error) {
```
michaelmdresser:
```suggestion
// StartAndEnd iterates over all AssetSets in the AssetSetRange and returns the earliest start and
// latest end over the entire range
func (asr *AllocationSetRange) StartAndEnd() (time.Time, time.Time, error) {
```

michaelmdresser (approved):
This seems solid, and I think `continue` is the right replacement for `return` during an `Each()`. I agree with more testing, but I'm hopeful that the lack of need of locking will work out :pray: -- is that expected to be fine because we always clone out of Stores?

pkg/kubecost/allocation.go
mbolt35: Yeah, I think so! Good suggestion.

pkg/kubecost/allocation.go
mbolt35: Yeah, good call!

pkg/kubecost/allocation.go
mbolt35: > Neat! Why is that? Does it just take field order into account?

Yeah, basically, the fields have no names associated with the data in the format itself. It just deterministically parses the fields in order and pulls the field name from the AST data. 

> is that expected to be fine because we always clone out of Stores?
Yes, and the only real danger for races/corruption would've happened during iteration (via `Each`), and we didn't lock at all in `Each`, which I think is a fairly good hint at existing safety.

mbolt35:
@michaelmdresser 
> Neat! Why is that? Does it just take field order into account?

Yeah, basically, the fields have no names associated with the data in the format itself. It just deterministically parses the fields in order and pulls the field name from the AST data. 

> is that expected to be fine because we always clone out of Stores?

Yes, and the only real danger for races/corruption would've happened during iteration (via `Each`), and we didn't lock at all in `Each`, which I think is a fairly good hint at existing safety.

mbolt35:
@Sean-Holcomb 

> Should we also be looking at making the Asset types fields exported? I know they are mainly controlled via the interface. One notable missing control from the Assets interface is a SetWindow func, what do you think about putting that in?

Yeah, that's a great idea for the Asset fields, and we'll definitely need that when doing grpc stuff. We can also easily add SetWindow in the meantime!

pkg/kubecost/asset_unmarshal.go
```diff
@@ -635,13 +635,11 @@ func (lb *LoadBalancer) InterfaceToLoadBalancer(itf interface{}) error {
 func (sa *SharedAsset) MarshalJSON() ([]byte, error) {
 	buffer := bytes.NewBufferString("{")
 	jsonEncodeString(buffer, "type", sa.Type().String(), ",")
-	jsonEncode(buffer, "properties", sa.Properties(), ",")
-	jsonEncode(buffer, "labels", sa.Labels(), ",")
-	jsonEncode(buffer, "properties", sa.Properties(), ",")
```
mbolt35: Removed duplicate properties and labels encoding

pkg/kubecost/asset.go
```diff
@@ -23,25 +22,26 @@ type Asset interface {
...
 
 	// Temporal values
-	Start() time.Time
-	End() time.Time
+	GetStart() time.Time
+	GetEnd() time.Time
 	SetStartEnd(time.Time, time.Time)
-	Window() Window
+	GetWindow() Window
+	SetWindow(Window)
```
Sean-Holcomb, line 44: Nice!

pkg/kubecost/asset.go
```diff
@@ -2565,24 +2645,63 @@ func (as *AssetSet) Clone() *AssetSet {
...
+		SharedAssets:      sharedAssetsMap,
+		Errors:            errors,
+		Warnings:          warnings,
 	}
+
+	for k, v := range as.Assets {
+		newAsset := v.Clone()
+		assetSet.Assets[k] = newAsset
+		addToConcreteMap(assetSet, k, newAsset)
+	}
```
Sean-Holcomb: Should we just use Insert() here?

Sean-Holcomb (approved):
Great stuff, lets get this merged and get the merge conflicts started

pkg/kubecost/asset.go
mbolt35: I looked closely at this, and I don't think `Insert()` is necessary because we are cloning directly from another AssetSet, so we won't ever have the case where a new Asset should be added to an existing asset, or we will ever have to expand the window.

Although, I think I just realized something wrong with Insert() while looking into this a bit more.

pkg/kubecost/totals.go
```diff
@@ -303,139 +303,143 @@ func ComputeAssetTotals(as *AssetSet, prop AssetProperty) map[string]*AssetTotal
...
 	disks := map[string]*Disk{}
 
-	as.Each(func(name string, asset Asset) {
-		if node, ok := asset.(*Node); ok {
-			// Default to computing totals by Cluster, but allow override to use Node.
-			key := node.Properties().Cluster
-			if prop == AssetNodeProp {
-				key = fmt.Sprintf("%s/%s", node.Properties().Cluster, node.Properties().Name)
-			}
+	for _, node := range as.Nodes {
```
nikovacevic, line 306: This is wild. So awesome. `range as.Nodes` rules.

pkg/kubecost/totals.go
```diff
@@ -303,139 +303,143 @@ func ComputeAssetTotals(as *AssetSet, prop AssetProperty) map[string]*AssetTotal
...
-	})
+	}
+
+	// Record disks in an intermediate structure, which will be
+	// processed after all assets have been seen.
+	for _, disk := range as.Disks {
+		key := fmt.Sprintf("%s/%s", disk.Properties.Cluster, disk.Properties.Name)
+
+		disks[key] = disk
+	}
```
nikovacevic, line 442: Kinda suspect we could somehow drop or optimize here... but let's not think about that for now. (I remember thinking "crap I have to make a new map for disks, alone," which _might_ no longer be necessary.) But yeah, looks good as a direct translation.

pkg/kubecost/allocation.go
```diff
@@ -317,34 +365,14 @@ func (a *Allocation) Equal(that *Allocation) bool {
 		return false
 	}
 
-	if a.RawAllocationOnly == nil && that.RawAllocationOnly != nil {
+	if !a.RawAllocationOnly.Equal(that.RawAllocationOnly) {
```
nikovacevic, line 380: Nice, very clean

pkg/kubecost/allocation.go
```diff
@@ -795,10 +823,9 @@ func (a *Allocation) add(that *Allocation) {
 // AllocationSet stores a set of Allocations, each with a unique name, that share
 // a window. An AllocationSet is mutable, so treat it like a threadsafe map.
 type AllocationSet struct {
-	sync.RWMutex
-	allocations  map[string]*Allocation
-	externalKeys map[string]bool
-	idleKeys     map[string]bool
+	Allocations  map[string]*Allocation
+	ExternalKeys map[string]bool
+	IdleKeys     map[string]bool
```
nikovacevic, line 793: Here's to hoping we're all adult enough! Haha.

pkg/kubecost/allocation.go
```diff
@@ -1696,22 +1717,9 @@ func (as *AllocationSet) Delete(name string) {
...
-
-// Each invokes the given function for each Allocation in the set
-func (as *AllocationSet) Each(f func(string, *Allocation)) {
-	if as == nil {
-		return
-	}
-
-	for k, a := range as.allocations {
-		f(k, a)
-	}
```
nikovacevic, line 1679: :smiling_face_with_tear:

pkg/kubecost/asset.go
nikovacevic: :metal:

pkg/kubecost/asset.go
```diff
@@ -447,23 +447,23 @@ func (at AssetType) String() string {
...
-	start      time.Time
-	end        time.Time
-	window     Window
-	adjustment float64
+	Labels     AssetLabels
+	Properties *AssetProperties
+	Start      time.Time
+	End        time.Time
+	Window     Window
+	Adjustment float64
```
nikovacevic, line 455: I would be curious if, as a next step, it would be possible to drop the `Asset` interface altogether. (Yes, I wonder if it's truly _that bad_ that we would be better off without it...) In that case, we could rename `Any` to be `Asset`, as the general case. That way we have `pkg/kubecost.Asset` and we no longer have the altogether too-general `pkg/kubecost.Any`, which isn't even clearly an asset, by name alone.

Does that make sense?

nikovacevic: That would obviously primarily affect the various `Add` functions. Meh... idk. It's just a hard type taxonomy to convey in Go's type system.

pkg/kubecost/asset.go
```diff
@@ -2716,80 +2858,68 @@ func (as *AssetSet) ReconciliationMatchMap() map[string]map[string]Asset {
...
-		as.assets[k] = as.assets[k].Add(asset)
+		// possibly creates a new asset type, so we need to update the
+		// concrete type mappings
+		newAsset := as.Assets[k].Add(asset)
+		removeFromConcreteMap(as, k, as.Assets[k])
+
+		// overwrite the existing asset with the new one, and re-add the new
+		// asset to the concrete type mappings
+		as.Assets[k] = newAsset
+		addToConcreteMap(as, k, newAsset)
```
nikovacevic, line 2901: Okay, this is the crux. Let's break it down into two options:
1. Adding a Node to a set aggregated by e.g. `type`
2. Adding a Node to a set NOT aggregated by type -- e.g. `cluster`

### Situation 1
* if the key doesn't exist, then set into `Assets` and `addToConcreteMap` :heavy_check_mark: 
* if the key does exist, then we create a new `Node` from the result of `Add`ing to the existing entry for `"node"`; then we `removeFromConcreteMap` the existing entry; then we set the new sum into `Assets` and `addToConcreteMap`; I _think_ that's right :heavy_check_mark: 

### Situation 2 
* if the key doesn't exist, then set into `Assets` and `addToConcreteMap`; my only question here is... will we end up with a `Node` in the `nodeMap`? Should we just have an `Any`, because we're aggregating by e.g. `cluster` :question: 
* if the key does exist, then we create a new `Any` from the result of `Add`ing to the existing entry for the cluster; then we `removeFromConcreteMap` the existing entry; then we set the new sum into `Assets` and `addToConcreteMap`; I _think_ that's right :heavy_check_mark: 


One thought -- do you think it's worth writing some extra unit tests to confirm this for a few cases? That's something I'd be up for helping with. Just want to be careful here.

nikovacevic (approved):
I think this looks good. Just that _one_ cruxy spot in `Insert`, I want to double check with you. If you're happy with how you've addressed it, then I trust you. Just let me know in response to my review comment.

Lastly... THANK YOU FOR YOUR PATIENCE! I know this took a long time. Apologies.

pkg/kubecost/totals.go
mbolt35: Yeah, I figured this would open the door for lots of optimization potential, but I definitely wanted to get a 1-1 first.

pkg/kubecost/asset.go
mbolt35: I'll be honest, this seems reasonable by design, but there are a _lot_ of cases which just ask for the list of Assets. In that way, I feel like it's actually pretty useful, and the interface actually _does_ fit. However, I think when you're only given a list of interfaces (without any sort of visitor pattern implemented), it can be really hard to work without some type of `byType` -- I think we can definitely improve on this as we go....

One thing to note is that for backwards compatibility reasons currently, we _can't_ just axe the Asset map :(

pkg/kubecost/asset.go
mbolt35: > if the key doesn't exist, then set into Assets and addToConcreteMap; my only question here is... will we end up with a Node in the nodeMap? Should we just have an Any, because we're aggregating by e.g. cluster ❓

I'm not sure I follow. If the type we're inserting is a Node, it will definitely end up in the Node map. If you're saying there's a possibility to omit adding/removing the Node when we know that eventually we'll be moving that Node to an Any, I'd prefer the consistent route unless we can always guarantee a specific condition on the first add. 

> One thought -- do you think it's worth writing some extra unit tests to confirm this for a few cases? That's something I'd be up for helping with. Just want to be careful here.

Yeah, I think I have some tests in the bingen side of things that I can move over quickly....

PR #1359, opened 2022-08-15
Implement .Equals() for the AllocationFilter interface

michaelmdresser:
## What does this PR change?
Introduces `.Equals()` to the `type AllocationFilter interface`.

This is for logical equivalence between values of the
AllocationFilter type. This equivalence is for checking e.g. that two
ORs with filters in different orders are, at their core, filtering
the same thing: `(or a b c) = (or b c a)`

This is motivated by the need to check filter equivalence in unit tests
that don't have stable filter construction, like when constructing a
filter while iterating over a map.

* N/A

## How was this PR tested?
* Unit tests

### DIFF
--- a/pkg/kubecost/allocationfilter.go
+++ b/pkg/kubecost/allocationfilter.go
@@ -105,10 +105,6 @@ type AllocationFilter interface {
 	// Equals returns true if the two AllocationFilters are logically
 	// equivalent.
 	Equals(AllocationFilter) bool
-
-	// empty returns true if the filter isn't filtering anything, i.e. the
-	// filter is a no-op.
-	empty() bool
 }
 
 // AllocationFilterCondition is the lowest-level type of filter. It represents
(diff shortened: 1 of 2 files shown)


pkg/kubecost/allocationfilter.go
```diff
@@ -100,6 +101,10 @@ type AllocationFilter interface {
 	Flattened() AllocationFilter
 
 	String() string
+
+	// Equals returns true if the two AllocationFilters are logically
+	// equivalent.
+	Equals(AllocationFilter) bool
```
mbolt35, line 107: Does this really work for logical equivalence? ie: `!(a || b) == (!a && !b)` 

We may not have negation support yet, but just wanted to make sure.

michaelmdresser: I'm pretty sure this works for logical equivalence right now because we only support negation at the leaf level: `a != b` is supported but `not(a == b)` is not supported. This gets much thornier if/when we introduce higher-level negation. Does that address the question? I'm racking my brain for other gotchas of logical equivalence.

mbolt35: Yeah, no worries  - just being picky.

mbolt35 (approved):
LGTM

PR #1361, opened 2022-08-19
Fix PV alloc recon, alloc PV marshalling

### LINKED ISSUE #1621: Full Aggregation of Costs by Controller
Need a complete sum total of costs associated with a StatefulSet.

Cost Allocations page drill-down to K8S controllers currently only includes some portion of costs (CPU and Memory) for controllers. We'd like this to include all associated costs--network and storage.

N/A

This would allow us to track our customer resource usage cost in order to fairly pass those costs on-to customers.

gz#2356<br><br><i>(related to <a href='https://kubecost.zendesk.com/agent/tickets/2356'>Zendesk ticket #2356</a>)</i>

Sean-Holcomb:
## What does this PR change?
This PR fixes an issue with PV Allocation Reconciliation in which PV cluster was not being populated in the PVKey of the PVs map found in the Allocation object. Due to this missing value in the PVKey reconciliation matches were not occurring and between Allocations and the Asset PVs which had been adjusted. I have made two changes to the Allocation ingestion process to populate cluster when generating PVC objects and to use the Cluster of the PV rather than the PV when generating the PVKey for the allocation. Either of these changes alone would address the problem, however I have done both to avoid future issues. 

Additionally this PR includes implementations of PV json marshaling. Which was pivotal in being able to diagnose the issue, in the past we did not include this value when marshaling partially due to over the wire costs of sending this additional data, however with AllocationSummary this is much less of a concern. Another issue is the difficulty of implementing the JSON marshalling on objects with struct keys. Have included comments on some of the specific requirements that needed to be met. If we had this visibility earlier this issue would have been caught much sooner. 

* 

* Users should see Allocation PV adjustment which match their asset costs

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1621

## How was this PR tested?
* Tested on Multi-cloud cluster
Before:
Screen Shot 2022-08-19 at 9 16 58 AM
Screen Shot 2022-08-19 at 9 16 40 AM
After:
Screen Shot 2022-08-19 at 1 49 51 PM

* 

*


pkg/kubecost/allocation_json.go
```diff
@@ -0,0 +1,114 @@
...
+// this function is required to have a value caller for the actual values to be saved
+func (pvk PVKey) MarshalText() (text []byte, err error) {
+	return []byte(pvk.String()), nil
+}
+
+// UnmarshalText converts JSON key string to PVKey it compatible with JSON Unmarshaller from an Object key
+// this function is required to have a pointer caller for values to be pulled into marshalling struct
+func (pvk *PVKey) UnmarshalText(text []byte) error {
+	return pvk.FromString(string(text))
+}
```
michaelmdresser, line 93: Woah! I didn't know you could do this! Very cool.

michaelmdresser (approved):
LGTM! I learned something today!

I wonder a little bit about the overhead on the copies when marshaling, but I suspect that (as you've pointed out) Summary Allocations mostly obviate this concern. Increasingly, I see `/allocation` as a tool for inspecting the ETL and `/allocation/summary` as a tool for cost reporting.

kaelanspatel (approved):

PR #1362, opened 2022-08-19
Proto message definition for allocation summary 

avrodrigues5:
## What does this PR change?
Defines the message for allocation summary

This will be first of many PR

*  Probably None with just this proto definition

* 

## How was this PR tested?
*  NO immediate need to test the PR other than the build passing

* No

* Done


pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    rpc GetAllocationSummary (AllocationSummaryRequest) returns (stream AllocationSummaryResponse) {}
+}
+
+// AllocationSummaryRequest is the request send from client with a specify window
+// either absolute window with datetime or relative window with duration
+message AllocationSummaryRequest {
+    Window window =1;
+    repeated string aggregate =2;
+    bool accumulate=4;
+    FilterSelection filter = 5;
```
sundeepghuman: repeated FilterSelection filters = 5;

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    oneof window{
+        AbsoluteWindow absWindow=1;
+        RelativeWindow relWindow=2;
+    }
+}
+
+// Absolute Window represents the absolute window with start and end time
+// for 
+message AbsoluteWindow {
+    google.protobuf.Timestamp start = 1;
```
sundeepghuman: Use Date instead of timestamp (we only support dates in the UI)

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    repeated string filterClusters=1;
+    repeated string filterNodes = 2;
+    repeated string filterNamespaces = 3; 
+    repeated string filterControllerKinds = 4;
+    repeated string filterController = 5;
+    repeated string filterPods = 6;
+    repeated string filterAnnotations = 7;
+    repeated string filterLabels = 8;
+    repeated string filterServices = 9;
+}
```
sundeepghuman: message FilterSelection {
  enum Type {
    UNKNOWN_TYPE = 0;
    CLUSTER = 1;
    NODE = 2;
    ...
  }
  
  Type type = 1;
  string value = 2;
}

sundeepghuman: Sorry, that should be FILTER_TYPE_UNSPECIFIED = 0;

pkg/cmd/commands.go
```diff
@@ -39,7 +39,7 @@ func Execute(costModelCmd *cobra.Command) error {
 		return err
 	}
 
-	rootCmd := newRootCommand(costModelCmd)
+	rootCmd := newRootCommand(costModelCmd, additionalCommands)
```
sundeepghuman: Why is this file being modified in this PR?

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+// Allocation Summary Interface exported by the gRPC server.
+service AllocationSummary {
+    // An RPC for allocation.
+    //
+    // Obtains the allocation summary for a given window span.
+    
+    // Obtains the AllocationSummarySets satisfy the given AllocationSummaryRequest. Results are
```
sundeepghuman: Keep all lines no longer than 80 characters. This is protobuf style guide (not sure if it's published), but also makes it much easier to review without scrolling

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
+syntax = "proto3";
+
+package kubecost.service.allocation.allocationsummary;
+
+option go_package = "github.com/opencost/opencost/pkg/kubecost/service/allocation/allocationsummary";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+// Allocation Summary Interface exported by the gRPC server.
```
sundeepghuman: Allocation Summary Service

-the fact it's an interface and gRPC is redundant

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+service AllocationSummary {
+    // An RPC for allocation.
+    //
+    // Obtains the allocation summary for a given window span.
+    
+    // Obtains the AllocationSummarySets satisfy the given AllocationSummaryRequest. Results are
+    // streamed rather than returned at once.
```
sundeepghuman: Combine and clean up these comments (seems like a duplicate):

// Computes an AllocationSummary for the given time window and aggregation/filter requirements.

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    rpc GetAllocationSummary (AllocationSummaryRequest) returns (stream AllocationSummaryResponse) {}
+}
+
+// AllocationSummaryRequest is the request send from client with a specify window
+// either absolute window with datetime or relative window with duration
+message AllocationSummaryRequest {
+    Window window =1;
+    repeated string aggregate =2;
```
sundeepghuman: This should be an enum, as we only support predefined aggregations

sundeepghuman: be consistent with spacing... window = 1, aggregate = 2;

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    bool accumulate=4;
```
sundeepghuman: accumulate = 4;

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    
+    // Obtains the AllocationSummarySets satisfy the given AllocationSummaryRequest. Results are
+    // streamed rather than returned at once.
+    rpc GetAllocationSummary (AllocationSummaryRequest) returns (stream AllocationSummaryResponse) {}
+}
+
+// AllocationSummaryRequest is the request send from client with a specify window
+// either absolute window with datetime or relative window with duration
+message AllocationSummaryRequest {
+    Window window =1;
```
sundeepghuman: Would this be better labeled "Date Range" - or at least DateWindow. If we are only supporting dates for right now, but if there is a case we may want more in the future, can call it TimeRange or TimeWindow

pkg/kubecost/service/allocation/allocationsummary.proto
sundeepghuman: (repeated enum field)

pkg/kubecost/service/allocation/allocationsummary.proto
sundeepghuman: accumulate doesn't seem to ever not be set, remove it from the API until there is a use case

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    repeated string aggregate =2;
+    bool accumulate=4;
+    FilterSelection filter = 5;
+    SharedSelection shared = 6;
+    IdleSelection idle = 7;
+    bool external =8;
```
sundeepghuman: what is external?

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+}
```
sundeepghuman, line 15: What bout Chart and Cost metric?

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    repeated string filterControllerKinds = 4;
+    repeated string filterController = 5;
+    repeated string filterPods = 6;
+    repeated string filterAnnotations = 7;
+    repeated string filterLabels = 8;
+    repeated string filterServices = 9;
+}
+
+// Different selection of idle from the User Interface 
+enum IdleSelection {
```
sundeepghuman: all enums should always have ENUM_NAME_UNSPECIFIED = 0; as their first value. Fix here and elsewhere.

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    oneof data{
+        AllocationProgressUpdate status_update = 1;
+        CompleteAllocationResponse full_response = 2;
+    }
+}
+
+// message for progress updates
+message AllocationProgressUpdate {
+    float percent_complete = 1;
+    repeated AllocationGraphItem graphItem= 2;
```
sundeepghuman: all field names should be lowercase with underscore. graph_item - fix here and elsewhere.

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    double total_cost = 3; 
+}
+ 
+// message for allocation graph item listing the day and repeated entry of each graph data
+message AllocationGraphItem {
+    oneof span{
+        int32 day =1;
```
sundeepghuman: just a reminder to repeat comments above and fix all spacing

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+enum IdleSelection {
+    IDLE_COST_HIDE = 0;
+    IDLE_COST_SEPERATE = 1;
+    IDLE_COST_SHARE_BY_CLUSTER = 2;
+    IDLE_COST_SHARE_BY_NODE = 3;
+}
+
+// Different shared selection available for the user on the User Interface
+message SharedSelection {
+    bool shareTenancyCosts = 1;
```
sundeepghuman: all of these fields should be commented with what they do - here and elsewhere.

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    string shareNamespaces= 2;
```
sundeepghuman: repeated string shared_namespaces = 2;

sundeepghuman: Do we want separate fields for default_shared_namespaces and custom_shared_namespaces? Do we need to specify default at all? Ask @nikovacevic

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    string shareLabels = 3;
```
sundeepghuman: custom_shared_labels

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    double shareCost = 4;
```
sundeepghuman: custom_shared_overheard_cost - details like 'shared vs share' matter, be consistent and correct with grammar

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+// Enum for the shared split which can either be weighted or even
+enum ShareSplit {
+    SHARESPLIT_WEIGHTED = 0;
+    SHARESPLIT_EVEN = 1;
+}
+
+// message for the allocationsummary response can either be partial response with each graphdata
+// or complete allocation response having all the details
+message AllocationSummaryResponse {
+    oneof data{
```
sundeepghuman: data should almost never be used, it's generic and lacks meaning. oneof response

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+        AllocationProgressUpdate status_update = 1;
+        CompleteAllocationResponse full_response = 2;
```
sundeepghuman: Should this be "FinalAllocationResponse" - as it's not the complete picture (since it doesn't have the bargraph data)

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    }
+}
+
+// message for progress updates
+message AllocationProgressUpdate {
```
sundeepghuman: This is more than just a progress update, it also contains the bar graph.

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    float percent_complete = 1;
+    repeated AllocationGraphItem graphItem= 2;
+    double total_cost = 3; 
+}
+ 
+// message for allocation graph item listing the day and repeated entry of each graph data
+message AllocationGraphItem {
+    oneof span{
```
sundeepghuman: oneof time_window

pkg/kubecost/service/allocation/allocationsummary.proto
sundeepghuman: Use Date protobuf, not time

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+        int32 day =1;
+        int32 hour =2;
```
sundeepghuman: Not sure we want to use int32 for hour here, are we assuming the client will infer the date? Maybe this would be better put as bucket_start and "bucket_width" (either hour or date), but this needs input from the FrontEnd team @Neal

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    }
+    repeated AllocationData allocationData = 3; 
+}
```
sundeepghuman, line 190: Should you put total_cost here? Or can the client simply sum the items?

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+
+// message for each day important data of cost and name of each allocation
+message AllocationData {
```
sundeepghuman: AllocationData is misleading and vague, this is actually a "GraphItemEntry" or something like that.

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+ 
+// message for complete allocation data having repeated allocationSet of summary
+message CompleteAllocationResponse {
+    repeated AllocationSet allocationSet = 2;
+}
+   
+// message for each allocation set consists of allocation items and window the allocation items data belongs to
+message AllocationSet {
+    repeated AllocationItem allocations = 1;
+    Window window =2;
```
sundeepghuman: why is Window mentioned here?

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// message for each allocation Item 
+message AllocationItem {
+    Window window =1;
```
sundeepghuman: Why is window mentioned here?

pkg/kubecost/service/allocation/allocationsummary.proto
sundeepghuman, line 15: add a 

// The maximum number of items to return in the table. If unspecified/set to 0, we will return all items.
int32 max_table_items = 0;

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+}
+   
+// message for each allocation set consists of allocation items and window the allocation items data belongs to
+message AllocationSet {
+    repeated AllocationItem allocations = 1;
+    Window window =2;
+}
+
+// message for each allocation Item 
+message AllocationItem {
```
sundeepghuman: SummaryTableItem

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    Window window =1;
+    string name = 2;
+    double cpuCoreRequestAverage = 3;
```
sundeepghuman: I don't see these fields being use

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    double cpuCoreUsageAverage = 4;
```
sundeepghuman: This one either

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,147 @@
...
+    double cpuCost = 5;
```
sundeepghuman: fix all field name sytnax here and elsewhere: cpu_cost

pkg/kubecost/service/allocation/allocationsummary.proto
sundeepghuman: I believe you want to list the 'efficiency' column, but efficiency is a misnomer, the field should be called average_cpu_utilization

sundeepghuman (changes requested):
Should the generated files be showing up here? I'm used to not committing the proto files, and instead having the build targets depend on generated proto libraries - @mbolt35 does that make sense in our setup?

Also, please comment every message and every field.

pkg/kubecost/service/allocation/allocationsummary.proto
avrodrigues5: Hello @sundeepghuman , incoporate all the changes except this , Date is not available in google.protobuf package https://developers.google.com/protocol-buffers/docs/reference/google.protobuf. Closest type is timestamp itself.

pkg/kubecost/service/allocation/allocationsummary.proto
avrodrigues5: Minor clarification , the 0 after the name is not possible as field number https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers. I'm assuming u only meant default to 0 here.

pkg/kubecost/service/allocation/allocationsummary.proto
avrodrigues5: I fixed all of them with style guide https://developers.google.com/protocol-buffers/docs/style

pkg/kubecost/service/allocation/allocationsummary.proto
mbolt35: @michaelmdresser Can you take a look at this. How well does this line up with allocation v2 filters?

mbolt35:
> Should the generated files be showing up here? I'm used to not committing the proto files, and instead having the build targets depend on generated proto libraries - @mbolt35 does that make sense in our setup?

Yeah, it's normal to commit both the proto files and the generated messages in go.  I personally like the idea of separating the messages into a separate directory off the root, but that might just be a hassle. Thanos uses a generation script: https://github.com/thanos-io/thanos/blob/13a2eb8cecf33dc18246c62de8d6d46553c81f94/scripts/genproto.sh

to find all `.proto` files and run the generation. I can look around a bit more to see if I can locate anything else that looks more conventional.

pkg/kubecost/service/allocation/allocationsummary.pb.go
```diff
@@ -0,0 +1,1653 @@
...
+	// Verify that runtime/protoimpl is sufficiently up-to-date.
+	_ = protoimpl.EnforceVersion(protoimpl.MaxVersion - 20)
+)
+
+// Different selection of idle from the User Interface
+type IdleSelection int32
+
+const (
+	IdleSelection_IDLE_COST_HIDE             IdleSelection = 0
+	IdleSelection_IDLE_COST_SEPERATE         IdleSelection = 1
```
nikovacevic: SEPARATE

pkg/kubecost/service/allocation/allocationsummary.pb.go
```diff
@@ -0,0 +1,1653 @@
...
+	IdleSelection_IDLE_COST_SHARE_BY_CLUSTER IdleSelection = 2
+	IdleSelection_IDLE_COST_SHARE_BY_NODE    IdleSelection = 3
+)
+
+// Enum value maps for IdleSelection.
+var (
+	IdleSelection_name = map[int32]string{
+		0: "IDLE_COST_HIDE",
+		1: "IDLE_COST_SEPERATE",
```
nikovacevic: SEPARATE

pkg/kubecost/service/allocation/allocationsummary.pb.go
```diff
@@ -0,0 +1,1653 @@
...
+		2: "IDLE_COST_SHARE_BY_CLUSTER",
+		3: "IDLE_COST_SHARE_BY_NODE",
+	}
+	IdleSelection_value = map[string]int32{
+		"IDLE_COST_HIDE":             0,
+		"IDLE_COST_SEPERATE":         1,
```
nikovacevic: SEPARATE

pkg/kubecost/service/allocation/allocationsummary.proto
```diff
@@ -0,0 +1,150 @@
...
+        FILTER_TYPE_SERVICES = 9;
+    }
+    FilterType type=1;
+    repeated string value = 2;
+}
+
+// Different selection of idle from the User Interface 
+enum IdleSelection {
+    IDLE_COST_HIDE = 0;
+    IDLE_COST_SEPERATE = 1;
```
nikovacevic: SEPARATE

nikovacevic (changes requested):
Quick pass -- just pointing out some typos of "SEPARATE"

pkg/kubecost/service/allocation/allocationsummary.proto
michaelmdresser: I think v1 filters have a pretty natural proto enum representation because each field is a separate query param with comma-separated filter values (e.g. `filterNamespace=kubecost,kube-system` in HTTP API). I think the representation Alan has right now is pretty solid for that.

However, I think V2 filters (e.g. `filter=namespace:"kubecost","kube-system"+label[app]:"cost-analyzer"` in HTTP API) should stay as a string parameter because we maintain a parser for this filter language already.

It would be cool if we could enforce using either v1 filters or v2 filter but not both.

pkg/kubecost/service/allocation/allocationsummary.proto
avrodrigues5: As discussed yesterday only listing percentage completion in the Progress Update

pkg/kubecost/service/allocation/allocationsummary.proto
avrodrigues5: confirmed with Niko, the UI overwrite the custom shared resources with default resources. So one should be fine

pkg/kubecost/service/allocation/allocationsummary.proto
avrodrigues5: commented every single filed in every single message

pkg/kubecost/service/allocation/allocationsummary.proto
avrodrigues5: as discussed yesterday this message will only have percentage.

avrodrigues5:
closing this PR as theres a corresponding PR open

PR #1363, opened 2022-08-22
Update PV and LB to use Unmounted

Sean-Holcomb:
## What does this PR change?
This PR fixes a few issues with PV Ingestion from Prom data with the goal of achieving Cluster Equality. Cluster Equality in this context means that queries on Allocations and Assets grouped by Cluster would have totals that are equal. Cluster Equality does depend on post ingestion processes such as Reconciliation and shared tenancy costs, however the way certain values were being ingestion had down stream effects those processes. The main focus here has been ensuring that PV and LB costs that were not being set to Allocations are now being applied to an Unmounted Allocation for a cluster. This Unmount Allocation now has its PVs and Services populated which allows it to be properly reconciled. 

This PR fixes a bug with the interval calculations on PV ingestion. The intervals are a great tool for determining the proportion of a PV's cost should go to pods which are attached to it. This is especially true in Windows when pods do not exist for the entire interval. The bug is that we were determining the time coef (precent of time that an interval) was being calculate off the pods window rather than off the PV window. For example given:
```
PV:     |----|
POD1:   |--|
POD2:      |--|
```
In this example the PV runs for an hour and Pod1 runs for the first half hour and Pod2 for the second half hour.
With the bug these coefs:
```
POD1: {
  Time: 1
  Proportion: 1
},
POD2: {
  Time: 1
  Proportion: 1
}
```
This says that 100% of the time POD1 was running it can be attributed 100% of PV's cost, however this coefficient is multiplied with the total cost of the PVC which means 200% of the cost is distributed. To address this we calculate Time with the PVC window rather tan the POD window. 

This Issue ties in with the larger issue of Unmounted resources because there was no handling of the case where no Pods were running at the same time a PVC.

* 

* 

* Closes ...

## How was this PR tested?
* Unit tests for Interval have been updated, but there is a need for more Unit testing on other helper funcs in costmodel/allocation.go

* 

*


pkg/kubecost/allocation.go
```diff
@@ -30,7 +30,7 @@ const SharedSuffix = "__shared__"
 // UnallocatedSuffix indicates an unallocated allocation property
 const UnallocatedSuffix = "__unallocated__"
 
-// UnmountedSuffix indicated allocation to an unmounted PV
```
Sean-Holcomb, line 30: One of the big changes in this PR is that an Unmounted allocation can now have LB cost for LBs for which no pod has their selector for the given window

pkg/costmodel/allocation.go
```diff
@@ -2724,7 +2708,6 @@ func (p Pod) AppendContainer(container string) {
 // TODO:CLEANUP add PersistentVolumeClaims field to type Allocation?
 type PVC struct {
 	Bytes     float64   `json:"bytes"`
-	Count     int       `json:"count"`
```
Sean-Holcomb, line 2728: count is replaced by the coefficient calculated with intervals and the number of allocations in a pod

pkg/costmodel/allocation.go
```diff
@@ -549,164 +549,151 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
...
+			}
+		}
+	}
 
-					// Scale PV cost by PVC sharing coefficient.
-					if coeffComponents, ok := sharedPVCCostCoefficientMap[pvcKey][podKey]; ok {
-						cost *= getCoefficientFromComponents(coeffComponents)
-					} else {
-						log.Warnf("CostModel.ComputeAllocation: allocation %s and PVC %s have relation but no coeff", alloc.Name, pvc.Name)
-					}
```
Sean-Holcomb: This is where the total PVC cost is being multiplied by Coefficients which could sum to greater than 100%

pkg/costmodel/allocation.go
```diff
@@ -549,164 +549,151 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
...
-					// record the amount of total PVBytes Hours attributable to a given PV
-					if alloc.PVs == nil {
-						alloc.PVs = kubecost.PVAllocations{}
-					}
-					pvKey := kubecost.PVKey{
-						Cluster: pvc.Cluster,
-						Name:    pvc.Volume.Name,
-					}
-					alloc.PVs[pvKey] = &kubecost.PVAllocation{
-						ByteHours: pvc.Bytes * hrs / count,
```
Sean-Holcomb, line 696: Not multiplying by coef here can result in reconciliation giving more cost to this allocation then it should actually get.

pkg/kubecost/audit.go
```diff
@@ -38,8 +38,8 @@ func ToAuditType(check string) AuditType {
 		return AuditAssetTotalStore
 	case string(AuditAssetAggStore):
 		return AuditAssetAggStore
-	//case string(AuditClusterEquality):
-	//	return AuditClusterEquality
+	case string(AuditClusterEquality):
+		return AuditClusterEquality
```
nikovacevic, line 42: It's nice to have this to rely on -- I'm assuming that uncommenting this means that it passes now?

pkg/costmodel/cluster.go
```diff
@@ -155,7 +155,7 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end
 	ctx := prom.NewNamedContext(client, prom.ClusterContextName)
 	queryPVCost := fmt.Sprintf(`avg(avg_over_time(pv_hourly_cost[%s])) by (%s, persistentvolume,provider_id)`, durStr, env.GetPromClusterLabel())
 	queryPVSize := fmt.Sprintf(`avg(avg_over_time(kube_persistentvolume_capacity_bytes[%s])) by (%s, persistentvolume)`, durStr, env.GetPromClusterLabel())
-	queryActiveMins := fmt.Sprintf(`count(pv_hourly_cost) by (%s, persistentvolume)[%s:%dm]`, env.GetPromClusterLabel(), durStr, minsPerResolution)
+	queryActiveMins := fmt.Sprintf(`avg(kube_persistentvolume_capacity_bytes) by (%s, persistentvolume)[%s:%dm]`, env.GetPromClusterLabel(), durStr, minsPerResolution)
```
nikovacevic, line 160: Any risks to swapping this? And what exactly are the benefits? I can't think of any, in either direction, but want to check in the interest of being super careful.

pkg/costmodel/allocation.go
```diff
@@ -531,2253 +527,202 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
...
 
 	// (3) Build out AllocationSet from Pod map
-
 	for _, pod := range podMap {
 		for _, alloc := range pod.Allocations {
 			cluster := alloc.Properties.Cluster
 			nodeName := alloc.Properties.Node
 			namespace := alloc.Properties.Namespace
-			pod := alloc.Properties.Pod
+			podName := alloc.Properties.Pod
```
nikovacevic, line 582: Nice

pkg/costmodel/allocation.go
```diff
@@ -531,2253 +527,202 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
...
-				log.Profile(queryProfile, fmt.Sprintf("CostModel.ComputeAllocation: pod query %d try %d failed: %s", numQuery, numTries, queryPods))
-				resPods = nil
-			}
-		}
-
-		if err != nil {
-			return err
-		}
+// LB describes the start and end time of a Load Balancer along with cost
+type LB struct {
```
nikovacevic: Nit, but we might not want to export this. Seems reasonable for internal use.

nikovacevic (approved):
I think this looks good, though I will admit that this is a challenging review. Is there any way I can help with additional testing to make sure that this is all rock solid?

pkg/kubecost/audit.go
Sean-Holcomb: It passes our tests, and on my cluster, given everything that I have seen in the Costmodel in the last couple weeks I don't think it will pass in all situations but I think it is a good audit have in that it lets us know our short comings on this invariant.

pkg/costmodel/cluster.go
Sean-Holcomb: This is in regards to https://github.com/kubecost/kubecost-cost-model/pull/927 I suspect that the use of pv_hourly_cost for our active mins query is causing this log when kubecost is down and not emitting the metric but KSM still is. I might be totally off base on this, but I think this change is low risk and may address the problem

michaelmdresser (approved):
Okay, I think I'm on board with the idea here: allocating PV cost to Allocations based on the amount of time the Allocation attached the PV via PVC. I sort of bias towards Niko's (?) idea of allocating based on weighted cost, but at the end of the day I think having proper aggregation equality at the end of the day is more important.

I like the refactoring, too. Hopefully it'll be easier to navigate `computeAllocation()` logic in the future!

Sean-Holcomb:
I too would have advised against the interval approach if I knew about it in the start, it seemed like a waste to throw it out at this point rather than fix it. The accuracy to complexity trade off probably isn't worth it though.

michaelmdresser:
Yep, totally makes sense!

PR #1376, opened 2022-09-07
allow full matches on cloud type assets

Sean-Holcomb:
## What does this PR change?
This PR address a bug in which Cloud type assets are not considered for full match when preforming reconciliation. Because of this each time reconciliation is run the full cost of the Cloud type asset is added to itself because partial matches are put into the asset set via `as.insert()`. This fix allows cloud type assets to be considered for full match, which on subsequent runs will occur and if there has been a change in cost it will be reflected in the adjustment. Considering Cloud Assets here is not expected to be a common occurrence because at this time they can only be added into the Asset store via the process described  here.
 
Adding Cloud type assets during reconciliation is important because in case of a providerID match without a category match, this generally indicates a failure on the part of our categorization. This graceful failure allows for these to be identified, while also keeping data accurate as Cloud Usages are filtered out exclusively by providerID.

* 

* 

* Closes ...

## How was this PR tested?
* This fix was tested by making categorization for storage category assets coming from BigQuery to fail, in a similar way to what a user was experiencing. Once the bug had been reproduced by rerunning reconciliation multiple times.
Screen Shot 2022-09-06 at 4 51 02 PM
Screen Shot 2022-09-06 at 4 48 51 PM
Applying the fix and running reconciliation again allows for a full match, which reduces the value to the correct amount.
Screen Shot 2022-09-06 at 4 48 29 PM
An Asset rebuild clears the the incorrect data, and subsequent reconciliation runs no longer reproduce the bug

* 

*

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -2842,8 +2842,8 @@ func (as *AssetSet) ReconciliationMatchMap() map[string]map[string]Asset {
 			continue
 		}
 		props := asset.GetProperties()
-		// Ignore cloud assets and assets that cannot be matched when looking for reconciliation matches
-		if props == nil || props.ProviderID == "" || asset.Type() == CloudAssetType {
+		// Ignore assets that cannot be matched when looking for reconciliation matches
+		if props == nil || props.ProviderID == "" {
 			continue
 		}
 


AjayTripathy (approved):

dwbrown2:
@Sean-Holcomb random comment, but highlighting a known bug in your screenshot just in case it's easy to address while in there: 

image

https://github.com/kubecost/cost-analyzer-helm-chart/issues/1675

Sean-Holcomb:
Yeah that one doesn't look quick at all. From a UX perspective we might not want to show chips on aggregate items. But as someone pointed out in the ticket this can happen on unaggregated items

PR #1381, opened 2022-09-09
functions in SummaryAllocationSets to compute CPUEfficiency, RAMEfficiency and Efficiency 

avrodrigues5:
Signed-off-by: Alan Rodrigues <alanr5691@yahoo.com>

## What does this PR change?
* exposes functions needed to calculate efficiencies over summaryAllocationSets. This would avoid computations of the efficiencies on Frontend and would be leveraged in GRPC streaming service to present the data to frontend.

* [KCM PR](https://github.com/kubecost/kubecost-cost-model/pull/926)

* Probably None

* Closes ...

## How was this PR tested?
* Have to cross check if the logic is right 

* None

* v1.98

### DIFF
--- a/pkg/kubecost/summaryallocation.go
+++ b/pkg/kubecost/summaryallocation.go
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
 	return tc
 }
 
+// func to calculate average RAM efficiency over SummaryAllocationSet
+func (sas *SummaryAllocationSet) RAMEfficiency() float64 {
+	if sas == nil {
+		return 0.0
+	}
+
+	sas.RLock()
+	defer sas.RUnlock()
+
+	totalRAMBytesUsage := 0.0
+	totalRAMBytesRequest := 0.0
+	for _, sa := range sas.SummaryAllocations {
+		if sa.RAMBytesRequestAverage > 0 {
+			totalRAMBytesUsage += sa.RAMBytesUsageAverage
+			totalRAMBytesRequest += sa.RAMBytesRequestAverage
+		} else {
+			if sa.RAMBytesUsageAverage == 0.0 && sa.RAMCost == 0.0 {
+				totalRAMBytesUsage += 1.0
+				totalRAMBytesRequest += 1.0
+			}
+		}
+	}
+
+	if totalRAMBytesUsage > 0 {
+		return totalRAMBytesUsage / totalRAMBytesRequest
+	}
+
+	if totalRAMBytesUsage == 0.0 || totalRAMBytesRequest == 0.0 {
+		return 0.0
+	}
+	return 1.0
+}
+
+// func to calculate average CPU efficiency over SummaryAllocationSet
+func (sas *SummaryAllocationSet) CPUEfficiency() float64 {
+	if sas == nil {
+		return 0.0
+	}
+
+	sas.RLock()
+	defer sas.RUnlock()
+
+	totalCPUCoreUsage := 0.0
+	totalCPUCoreRequest := 0.0
+	for _, sa := range sas.SummaryAllocations {
+		if sa.CPUCoreRequestAverage > 0 {
+			totalCPUCoreUsage += sa.CPUCoreUsageAverage
+			totalCPUCoreRequest += sa.CPUCoreRequestAverage
+		} else {
+			if sa.RAMBytesUsageAverage == 0.0 && sa.RAMCost == 0.0 {
+				totalCPUCoreUsage += 1.0
+				totalCPUCoreRequest += 1.0
+			}
+		}
+	}
+
+	if totalCPUCoreUsage > 0 {
+		return totalCPUCoreUsage / totalCPUCoreRequest
+	}
+
+	if totalCPUCoreUsage == 0.0 || totalCPUCoreRequest == 0.0 {
+		return 0.0
+	}
+	return 1.0
+}
+
+// func to calculate average Total efficiency over SummaryAllocationSet
+func (sas *SummaryAllocationSet) TotalEfficiency() float64 {
+	if sas == nil {
+		return 0.0
+	}
+
+	sas.RLock()
+	defer sas.RUnlock()
+
+	tc := 0.0
+	totalRAMCostEff := 0.0
+	totalCPUCostEff := 0.0
+	totalRAMCost := 0.0
+	totalCPUCost := 0.0
+	for _, sa := range sas.SummaryAllocations {
+		if sa.RAMCost+sa.CPUCost > 0 {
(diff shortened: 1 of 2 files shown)


avrodrigues5:
@nikovacevic , please let me know if the computation used here is correct.

nikovacevic:
On it, thanks @avrodrigues5

pkg/kubecost/summaryallocation.go
```diff
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
...
+	totalRAMBytesRequest := 0.0
+	for _, sa := range sas.SummaryAllocations {
+		if sa.RAMBytesRequestAverage > 0 {
+			totalRAMBytesUsage += sa.RAMBytesUsageAverage
+			totalRAMBytesRequest += sa.RAMBytesRequestAverage
+		} else {
+			if sa.RAMBytesUsageAverage == 0.0 && sa.RAMCost == 0.0 {
+				totalRAMBytesUsage += 1.0
+				totalRAMBytesRequest += 1.0
+			}
```
nikovacevic: I'm not understanding this bit. Can you help with an explanation?

In general, my hope is that we can just add all the requests and usages together, simply; as in:
```go
for _, sa := range sas.SummaryAllocations {
    totalRAMBytesUsage += sa.RAMBytesUsageAverage
    totalRAMBytesRequest += sa.RAMBytesRequestAverage
}
```

pkg/kubecost/summaryallocation.go
```diff
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
...
+		}
+	}
+
+	if totalRAMBytesUsage > 0 {
+		return totalRAMBytesUsage / totalRAMBytesRequest
+	}
```
nikovacevic: I believe this should actually be:
```go
if totalRAMBytesRequest > 0 {
```

As a divide-by-zero check.

pkg/kubecost/summaryallocation.go
```diff
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
...
+
+	if totalRAMBytesUsage == 0.0 || totalRAMBytesRequest == 0.0 {
+		return 0.0
+	}
```
nikovacevic: And this should be:
```go
if totalRAMBytesUsage == 0.0 || totalRAMCost == 0.0
```

Which means something like "If there was no usage of any kind, then there is also no efficiency."

pkg/kubecost/summaryallocation.go
```diff
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
...
+	return 1.0
```
nikovacevic, line 1199: And if we get here, that means that `usage > 0` and `request == 0` so we call that 100% efficiency.. it's really more like "infinite" efficiency, but it's a special case where a user is saying "let this use all it needs, intentionally" -- so we call that 100%.

pkg/kubecost/summaryallocation.go
```diff
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
...
+	sas.RLock()
+	defer sas.RUnlock()
+
+	tc := 0.0
+	totalRAMCostEff := 0.0
+	totalCPUCostEff := 0.0
+	totalRAMCost := 0.0
+	totalCPUCost := 0.0
+	for _, sa := range sas.SummaryAllocations {
+		if sa.RAMCost+sa.CPUCost > 0 {
```
nikovacevic: I don't think we need this check here, but let me know if I'm missing something!

pkg/kubecost/summaryallocation.go
```diff
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
...
+			totalRAMCostEff += sa.RAMEfficiency() * sa.RAMCost
+			totalCPUCostEff += sa.CPUEfficiency() * sa.CPUCost
+			totalRAMCost += sa.RAMCost
+			totalCPUCost += sa.CPUCost
+		}
+	}
+
+	if totalRAMCost+totalCPUCost > 0 {
```
nikovacevic, line 1251: This is the only `cost > 0` check we need -- again, to guard against divide by zero.

pkg/kubecost/summaryallocation.go
```diff
@@ -1170,6 +1170,101 @@ func (sas *SummaryAllocationSet) TotalCost() float64 {
...
+		return (totalRAMCostEff + totalCPUCostEff) / (totalRAMCost + totalCPUCost)
+	}
+	return tc
```
nikovacevic: Let's discuss what we should return here. As far as I can tell, this value is just always `0.0`. If so, I don't think we need the named variable. We can just return `0.0` for clarity.

nikovacevic (changes requested):
Let me know if you have any questions about this one! As discussed, we're also going to want unit tests, so we can develop those when you're ready. Here to help.

pkg/kubecost/summaryallocation.go
avrodrigues5: I wasnt sure if i need to maintain exact numbers or add 1.0 for usage and request.

pkg/kubecost/summaryallocation.go
avrodrigues5: I checked this and edited both RAM and CPU efficiency.

pkg/kubecost/summaryallocation.go
avrodrigues5: I have to compute totalRAMCost which I did and edited the numbers for both RAM and CPU efficiency functions

pkg/kubecost/summaryallocation.go
avrodrigues5: we dont need it

pkg/kubecost/summaryallocation.go
avrodrigues5: No modification done here as this is guard against divided by 0.0

pkg/kubecost/summaryallocation.go
avrodrigues5: Missed this , i should copied the original total cost function.

nikovacevic (commented):
This LGTM Alan! Thank you! Let's just get it tested, and then I will be happy to approve.

nikovacevic (approved):

nikovacevic:
Thank you @avrodrigues5 !

avrodrigues5:
I dont have permission to merge the code, if you could also merge it for me @nikovacevic

AjayTripathy:
Merged! Thanks Alan!

PR #1401, opened 2022-09-23
Bug fix: ParseWindow function parses window with timestamp to a location and not UTC time

avrodrigues5:
## What does this PR change?
* For local KCM, API request with window timestamp of 1662163200,1662681600 gets parsed to 2022-09-02T17:00:00-0700, 2022-09-08T17:00:00-0700 because my location is PDT and doesn't return any ETL results. This change ensures that parsewindow functions returns window only in UTC(which i think ETL understands) when window request are timestamp and developer is testing functions locally with KCM.

* No

* No

* Closes ... None

## How was this PR tested?
* Local run failed to yield the window I wanted result for but actual system was working. 

* No

* Because i don't have permission


pkg/kubecost/window.go
```diff
@@ -278,8 +278,8 @@ func parseWindow(window string, now time.Time) (Window, error) {
 	if match != nil {
 		s, _ := strconv.ParseInt(match[1], 10, 64)
 		e, _ := strconv.ParseInt(match[2], 10, 64)
-		start := time.Unix(s, 0)
-		end := time.Unix(e, 0)
+		start := time.Unix(s, 0).UTC()
+		end := time.Unix(e, 0).UTC()
```
michaelmdresser, line 282: I think it might be more accurate to set the location to `now.Location()`, rather than strictly UTC. I believe that would make this code path consistent with the other paths -- the other paths derive `start` and `end` from `now` without modifying the location.

michaelmdresser: Maybe we should just enforce UTC like we do elsewhere in the product, consider also the code path for time.RFC3339 parsing which ignores `now`'s location completely. @nikovacevic do you have any thoughts? I'm having a hard time making a definitive call.

nikovacevic: I would ultimately like to use UTC everywhere, now that we're not supporting timezone. But we wrote this code before we decided to go all-UTC, so I'm only a bit hesitant to make changes in core logic that is working.

That said... I don't see any reason that Alan's change here shouldn't work, and it's relatively uncommon for users to use timestamps AFAIK. The FE doesn't, at least. So I say we go for it!

nikovacevic (approved):

PR #1403, opened 2022-09-26
PV usage metrics to ETL

nickcurie:
## What does this PR change?
* Pulls in PV usage metrics into ETL

* [KCM#963](https://github.com/kubecost/kubecost-cost-model/pull/963)
* [helm-chart#1713](https://github.com/kubecost/cost-analyzer-helm-chart/pull/1713)

* Users can see average byte usage and max byte usage for PVs

* 

## How was this PR tested?
* Locally

* Yes

* v1.98


pkg/costmodel/cluster.go
```diff
@@ -1291,4 +1341,120 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
...
+		claimNamespace, err := result.GetString("namespace")
+		if err != nil {
+			log.Warnf("ClusterDisks: pv usage data missing namespace")
+			continue
+		}
+
+		var volumeName string
+
+		for _, thatRes := range resPVCInfo {
+
```
michaelmdresser, line 1421: We glossed over the inner loop here while working on functionality. Now that we've got functionality working, do you see a way to avoid this inner loop? Maybe by constructing an intermediate `map` that gives you the data mapping you need?

pkg/costmodel/cluster.go
```diff
@@ -1291,4 +1341,120 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
...
+				log.Warnf("ClusterDisks: pv claim data missing volumename")
+			}
+			thatClaimName, err := thatRes.GetString("persistentvolumeclaim")
+			if err != nil {
+				log.Warnf("ClusterDisks: pv claim data missing persistentvolumeclaim")
+			}
+			thatClaimNamespace, err := thatRes.GetString("namespace")
+			if err != nil {
+				log.Warnf("ClusterDisks: pv claim data missing namespace")
+			}
```
michaelmdresser: In these error cases, should you have a `continue` after the `log.Warnf`?

pkg/costmodel/cluster.go
```diff
@@ -1291,4 +1341,120 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
...
+		claimNamespace, err := result.GetString("namespace")
+		if err != nil {
+			log.Warnf("ClusterDisks: pv usage data missing namespace")
+			continue
+		}
+
+		var volumeName string
+
+		for _, thatRes := range resPVCInfo {
+
```
michaelmdresser, line 1482: Same q

pkg/costmodel/cluster.go
```diff
@@ -1291,4 +1341,120 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
...
+				log.Warnf("ClusterDisks: pv claim data missing volumename")
+			}
+			thatClaimName, err := thatRes.GetString("persistentvolumeclaim")
+			if err != nil {
+				log.Warnf("ClusterDisks: pv claim data missing persistentvolumeclaim")
+			}
+			thatClaimNamespace, err := thatRes.GetString("namespace")
+			if err != nil {
+				log.Warnf("ClusterDisks: pv claim data missing namespace")
+			}
```
michaelmdresser: Same q

pkg/costmodel/cluster.go
```diff
@@ -107,17 +107,22 @@ func NewClusterCostsFromCumulative(cpu, gpu, ram, storage float64, window, offse
...
+	Name           string
+	ProviderID     string
+	StorageClass   string
+	VolumeName     string
+	ClaimName      string
+	ClaimNamespace string
+	Cost           float64
+	Bytes          float64
+	BytesUsedAvg   float64
+	BytesUsedMax   float64
```
michaelmdresser, line 120: I think you had a question about these because they aren't getting filled for local disks, just for PVs. Is that still right? What approaches have you considered?

pkg/kubecost/asset_json.go
```diff
@@ -332,6 +338,17 @@ func (d *Disk) InterfaceToDisk(itf interface{}) error {
 	if ByteHours, err := getTypedVal(fmap["byteHours"]); err == nil {
 		d.ByteHours = ByteHours.(float64)
 	}
+	if ByteHoursUsed, err := getTypedVal(fmap["byteHoursUsed"]); err == nil {
+		d.ByteHoursUsed = ByteHoursUsed.(float64)
+	}
+	if ByteUsageMax, err := getTypedVal(fmap["byteUsageMax"]); err == nil {
+		if ByteUsageMax == nil {
+			d.ByteUsageMax = nil
```
michaelmdresser, line 349: Would you add a case to `TestDisk_Unmarshal` that tests this null unmarshal?

pkg/kubecost/bingen.go
```diff
@@ -71,4 +71,4 @@ package kubecost
 // @bingen:generate:AuditSetRange
 // @bingen:end
 
-//go:generate bingen -package=kubecost -version=15 -buffer=github.com/opencost/opencost/pkg/util
+//go:generate bingen -package=kubecost -version=18 -buffer=github.com/opencost/opencost/pkg/util
```
michaelmdresser: I'm pretty sure you don't have to change the version here -- this only changes the version for things that don't specify their own version, like `Window`.

pkg/costmodel/cluster.go
nickcurie: We have metrics in prom that track container file system usage. I've included these prom queries in the latest commit, so local disks have avg and max usage in ETL.

pkg/costmodel/cluster.go
nickcurie: I think its worth coming back to this, since this is an efficiency ask. I think its valuable to get this PR out so that users can begin to interact with it.

mbolt35 (approved):
Looks good to me! Great work

PR #1404, opened 2022-09-27
ETL Data Structures: NewAccumulation()

mbolt35:
## What does this PR change?
* Adds a `NewAccumulation()` method on the `SetRange` ETL data structures which provides an Accumulate method which does not mutate. 

* Fixes a caching bug 

## How was this PR tested?
* Unit Tests and via endpoint

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -93,7 +93,8 @@ type Allocation struct {
 // A2 Using 2 CPU      ----      -----      ----
 // A3 Using 1 CPU         ---       --
 // _______________________________________________
-//                   Time ---->
+//
+//	Time ---->
 //
 // The logical maximum CPU usage is 5, but this cannot be calculated iteratively,
 // which is how we calculate aggregations and accumulations of Allocations currently.
@@ -1979,6 +1980,27 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
 	var err error
 
 	for _, as := range asr.Allocations {
+		allocSet, err = allocSet.accumulate(as)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return allocSet, nil
+}
+
+// AccumulateImmutable clones the first available AllocationSet to use as the data structure to
+// accumulate the remaining data. This leaves the original AllocationSetRange intact.
+func (asr *AllocationSetRange) AccumulateImmutable() (*AllocationSet, error) {
+	var allocSet *AllocationSet
+	var err error
+
+	for _, as := range asr.Allocations {
+		if allocSet == nil {
+			allocSet = as.Clone()
+			continue
+		}
+
 		allocSet, err = allocSet.accumulate(as)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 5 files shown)


pkg/kubecost/asset.go
```diff
@@ -68,63 +68,65 @@ type Asset interface {
...
+//     => Allocation{Name: "monitoring/__unallocated__", ExternalCost: 10.00, TotalCost: 10.00}, nil
 //
-//   4) no match
-//   aggregateBy = ["cluster"]
-//   allocationPropertyLabels = {"namespace":"kubernetes_namespace"}
-//   => nil, err
+//  4. no match
+//     aggregateBy = ["cluster"]
+//     allocationPropertyLabels = {"namespace":"kubernetes_namespace"}
+//     => nil, err
```
mbolt35, line 129: Apparently I have a VSCode extension that organizes comments (I sort of like it, although I didn't expect these changes).

michaelmdresser (approved):
LGTM! Commit history looks a little funky, though.

pkg/kubecost/allocation.go
```diff
@@ -1988,6 +1989,36 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
 	return allocSet, nil
 }
 
+// NewAccumulation clones the first available AllocationSet to use as the data structure to
+// accumulate the remaining data. This leaves the original AllocationSetRange intact.
+func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
```
michaelmdresser, line 1994: Now that I've thought about it some more, is there any reason this shouldn't be the logic for the standard `Accumulate()`?

mbolt35:
> LGTM! Commit history looks a little funky, though.

Yeah, it's because I forgot to commit a sig, then accidentally rebased my sig 1 commit too far in the past. Just going to squash on merge.

pkg/kubecost/allocation.go
mbolt35: So, for regular Allocations, from my testing, `NewAccumulation()` almost works exactly like `Accumulate()`, but this is not the case for `AssetSetRange` or `SummaryAllocationSetRange`. For API consistency, the `New___` variation seemed more explicit in that it created copies of the inner data before squashing. I _did_ explore updating `AllocationSetRange::Accumulate()` to match `AssetSetRange` and `SummaryAllocationSetRange`, but there are a large handful of tests which fail after that change, which seems to indicate an expected behavior there that's different from the latter two. 

For the most part, `Accumulate()` is fine to run given that the final SetRange object won't be used post accumulation. However, in the case with caching, we want the raw SetRange to remain immutable, so we need an immutable method _if_ accumulate=true.

michaelmdresser: Sounds perfect, thanks for the thorough response. `New___` is perfect here, I'm a big fan.

pkg/kubecost/asset.go
nikovacevic: Hahaha

nikovacevic (approved):
Thanks for doing this Bolt. Dig it. We should probably universalize Accumulate around the way we like it, I agree; but this is wonderful.

PR #1406, opened 2022-09-29
implement generic filters, implement window filter

Sean-Holcomb:

## What does this PR change?
* This PR adds generic filters based off of @michaelmdresser filterv2 design for allocation. By creating a new base filter for a type such as `AllocationCondition` the logical operators from filterv2 become available for use. Additionally this PR introduces the Window conditional which can be used as a filter on any type that has a window which is made accessible with `GetWindow() kubecost.Window`. 

This PR does not replace any of uses of the original filterv2 code, however it should be replaced in the future. 

* 

* 

* Closes ...

## How was this PR tested?
* This PR migrates the filter v2 tests which tested allocations and logical operator filters. New unit tests have been writen for `WindowConditional`

* 

*


mbolt35 (approved):
So far this looks awesome to me. Definitely pertains to my work, so a second reviewer would be great :)

pkg/filter/filter.go
```diff
@@ -0,0 +1,74 @@
...
+	// Example:
+	// (and (or (namespaceequals "kubecost")) (or)) ->
+	// (namespaceequals "kubecost")
+	//
+	// (and (or)) -> nil
+	Flattened() Filter[T]
+
+	// Equals returns true if the two AllocationFilters are logically
+	// equivalent.
+	Equals(Filter[T]) bool
```
michaelmdresser: I'm hesitant to enforce these two methods on the broad generic filter interface, but its been a while since I've been in this neighborhood. Have you seen a need for them on the interface?

pkg/filter/window.go
```diff
@@ -0,0 +1,60 @@
...
+
+type windowed interface {
+	GetWindow() kubecost.Window
+}
+
+// WindowOperation are operations that can be performed on types that have windows
+type WindowOperation string
+
+const (
+	WindowOpContains WindowOperation = "contains"
```
michaelmdresser: This overlaps with the existing `FilterContains` op. Would you add a doc comment to this const to explain exactly what behavior the new op will exhibit to avoid confusion?

Ideally the string representation would also be different from `FilterContains` (assuming the behavior differs), but that's a nit; up to you.

pkg/filter/window.go
```diff
@@ -0,0 +1,60 @@
...
+
+// WindowCondition is a filter can be used on any type that has a window and implements GetWindow()
+type WindowCondition[T windowed] struct {
+	Window kubecost.Window
+	Op     WindowOperation
+}
+
+func (wc WindowCondition[T]) String() string {
+	return fmt.Sprintf(`(window %s by "%s")`, wc.Op, wc.Window.String())
+}
```
michaelmdresser: The syntax I chose for stringifying Allocation filters was based on lisp (sorry, Bolt) mostly because it's one of the easiest ways to serialize an AST. That  looks like `(op arg1 arg2 arg3 ...)` e.g. `(+ 1 3 5)`. 

If you want to stick with that convention (you don't have to!) I think I would write this as:
```go
return fmt.Sprintf(`(%s window "%s")`, wc.Op, wc.Window.String())
```

That makes it clear what the op is, and what order its args are, which should help a user understand the directionality of the logic: is it GetWindow() within wc.Window? Or is it wc.Window within GetWindow()?

michaelmdresser: We'd end up with something like `(contains window "[t1-t2]")` which I think means "GetWindow() contains `[t1-t2]`" or "`[t1-t2]` falls within GetWindow()". This appears to be inconsistent with your implementation in `Matches()`, so I think there's some possible confusion  here.

michaelmdresser (commented):
This is cool! I'm particularly a fan of the new And and Or, they really highlight the usefulness of generics.

I have a few questions in PR comments, and one more broadly below. I wouldn't call any of them strictly merge-blocking, so if you need to merge this to keep making forward progress, go for it.

---

Is there a plan to swap out the current `pkg/kubecost.AllocationFilter` usage with the new `pkg/filter.Filter[*kubecost.Allocation]` version? I worry a little bit about diverging implementations in the long run.

Actually, does the existing `pkg/kubecost.AllocationFilter` already match the new generic interface? Does 
```go
// I _think_ this is the correct way to write this. This is trying to assert
// that one interface matches another.
var _ filter.Filter[*kubecost.Allocation] = (kubecost.AllocationFilter)(nil)
```
 compile? That would mean you wouldn't have to duplicate all the code (`pkg/filter/allocation` and `pkg/filter/allocation_test.go`) just yet.

pkg/filter/filter.go
Sean-Holcomb: In general, when adding new types that you want filtered you are only adding in a base filter. Equals and Flatten are trivial for a both cases, so while it does end up being boiler plate, I think what flatten and Equals could be with the logical operators makes the filters so much cooler.

Sean-Holcomb: One thing that comes to mind for dealing this is a `BaseFilter[T any]` which has the equals that casts to casts to its own type and then an Flatten which just returns itself. Other base filters could extend this to cut down on the boiler plate.

pkg/filter/window.go
Sean-Holcomb: I do think this is a distinct operation from the other filter op and don't think they should be conflated. I would almost rename both to be `StringContains` and `WindowContains`

michaelmdresser: I think that explicit distinction is a great idea. What about `WindowContains = "windowcontains"` and changing the old one to `ArrayContains = "arraycontains"`? (The original contains op isn't actually a string contains op, see the comment).

pkg/filter/window.go
mbolt35: I'm fine with polish notation on operators within the AST. That's not exclusive to Lisp either :)

Sean-Holcomb:
Moved to cloud cost branch

PR #1417, opened 2022-10-10
Make Base_URL Configurable

stephenatwell:
Signed-off-by: Stephen Atwell <78044713+stephenatwell@users.noreply.github.com>## What does this PR change?
* This PR allows users of the UI to override the path to use when connecting to the cost model. This allows, for example, for the cost model to be access through an ingress resource or a loadBalancer instead of being port forwarded to the local host. Here is a slack thread discussing this and addressed by this PR: https://cloud-native.slack.com/archives/C03D56FPD4G/p1665377461919949

* No

* It will allow users of the ui to optionally override the path used to connect to the cost model. 

* Unknown

## How was this PR tested?
* hand tested both with the base_URL environment variable set to change the URL used to connect to the allocation, and with it not set to ensure preexisting behavior still works for existing users.

* Ideally we would document that when deploying the UI you can set the environment variable 'BASE_URL' in order to change the URL of the cost model. However, deploying the UI is currently undocumented in its entirely.

* I do not see an easy way to do this, I think I lack the needed permission. Ideally this PR would be part of the next OpenCost release.


AjayTripathy (approved):

AjayTripathy:
Thank you for this PR @stephenatwell ! LGTM.

PR #1421, opened 2022-10-12
GP3 volumes support

AndrewChubatiuk:
## What does this PR change?
Added GP3 volumes support in AWS

## How was this PR tested?
Was tested in my setup, which uses GP3

No

Have no clue how to label


AjayTripathy:
Thanks for this PR @AndrewChubatiuk . Is the most minimal set of changes to just modify aws.provider.go? I think the PR scope should be just limited to that-- we can discuss what's going on in queries.go in a separate PR or discussion.

pkg/prom/queries.go
```diff
@@ -0,0 +1,483 @@
+package prom
```
AjayTripathy: Is this necessary within this PR? Happy to discuss changes here in a separate one!

AjayTripathy (changes requested):

pkg/prom/queries.go
AndrewChubatiuk: Sorry, unintended file here

AndrewChubatiuk: @AjayTripathy updated PR

AjayTripathy (approved):

pkg/prom/queries.go
AjayTripathy: Thank you!

PR #1422, opened 2022-10-12
build docker image with static UI

AndrewChubatiuk:
## What does this PR change?
Added simple docker image for UI
Added simple CI build for docker image

## How was this PR tested?
Was tested in my EKS cluster

### DIFF
--- a/ui/default.nginx.conf
+++ b/ui/default.nginx.conf
@@ -1,0 +1,70 @@
+gzip_static  on;
+gzip on;
+gzip_min_length 50000;
+gzip_proxied expired no-cache no-store private auth;
+gzip_types
+    application/atom+xml
+    application/geo+json
+    application/javascript
+    application/x-javascript
+    application/json
+    application/ld+json
+    application/manifest+json
+    application/rdf+xml
+    application/rss+xml
+    application/vnd.ms-fontobject
+    application/wasm
+    application/x-web-app-manifest+json
+    application/xhtml+xml
+    application/xml
+    font/eot
+    font/otf
+    font/ttf
+    image/bmp
+    image/svg+xml
+    text/cache-manifest
+    text/calendar
+    text/css
+    text/javascript
+    text/markdown
+    text/plain
+    text/xml
+    text/x-component
+    text/x-cross-domain-policy;
+server {
+    server_name _;
+    root /var/www;
+    index index.html;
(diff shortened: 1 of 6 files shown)


ui/src/Reports.js
```diff
@@ -154,12 +154,6 @@ const ReportsPage = () => {
       if (resp.data && resp.data.length > 0) {
         const allocationRange = resp.data
         for (const i in allocationRange) {
-          // update cluster aggregations to use clusterName/clusterId names
```
AjayTripathy, line 157: Curious what this removal is for?

AndrewChubatiuk: Tested it on setup with single victoria metrics cluster, where we aggregate metrics from multiple EKS clusters. We do not want to setup prometheus on each EKS as it's describe in doc. Reverted these changes back

AndrewChubatiuk: @AjayTripathy all builds have passed

AjayTripathy (approved):

AjayTripathy:
Thanks @AndrewChubatiuk !

ui/default.nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+
+    add_header Cache-Control "max-age=300";
+    location / {
+        try_files $uri $uri/ /index.html;
+    }
+
+    add_header ETag "1.96.0";
+    listen 9090;
+    listen [::]:9090;
+    resolver kube-dns.kube-system.svc.cluster.local valid=5s;
```
a7i, line 54: Is there a reason this way added? kube-dns is long gone

ui/default.nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    location /healthz {
+        return 200 'OK';
+    }
+    location /allocation {
+        proxy_connect_timeout       180;
+        proxy_send_timeout          180;
+        proxy_read_timeout          180;
+        set $server http://cost-analyzer.kubecost.svc.cluster.local:9003;
```
a7i, line 62: why is this not set to opencost? but rather kubecost?

ui/default.nginx.conf
AjayTripathy: whoops. Good catch, thank you. Let's get rid of this.

PR #1424, opened 2022-10-13
Add test to ensure AllocationSet.accumulate is equivalent to AllocationSetRange.Accumulate

saweber:
## What does this PR change?
* unit test for AllocationSet.accumulate

* no

* no impact

* no

## How was this PR tested?
* ran unit tests

* no

* no changes outside of testing


pkg/kubecost/allocation_test.go
```diff
@@ -2657,3 +2657,114 @@ func TestAllocationSetRange_Minutes(t *testing.T) {
...
+
+	expected, err := asr.Accumulate()
+	if err != nil {
+		t.Errorf("TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate: AllocationSetRange.Accumulate() returned an error\n")
+	}
+
+	var got *AllocationSet
+
+	for i := 0; i < len(allocationSets)-1; i++ {
+		got, err = allocationSets[i].Clone().accumulate(allocationSets[i+1])
```
nikovacevic: I think we actually don't want the `Clone()` here -- the idea being to test whether calling that function mutates the underlying data. With `Clone()` we're guaranteeing it won't because it's operating on a clone, not the original. That make sense, or am I missing context?

pkg/kubecost/allocation_test.go
```diff
@@ -2657,3 +2657,114 @@ func TestAllocationSetRange_Minutes(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate: allocationSets[0].accumulate(1) returned an error\n")
+	}
+
+	for i := 2; i < len(allocationSets); i++ {
+		got, err = got.accumulate(allocationSets[i].Clone())
+		if err != nil {
+			t.Errorf("TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate: got.accumulate(allocationSets[%d]) returned an error\n", i-1)
+		}
+	}
```
nikovacevic: What is this block for? I'm probably just missing something, but it looks like we've already done the full 0-through-n accumulation above?

pkg/kubecost/allocation_test.go
```diff
@@ -2657,3 +2657,114 @@ func TestAllocationSetRange_Minutes(t *testing.T) {
...
+		for i := 2; i < len(allocationSets); i++ {
+			sum, err = sum.Add(allocationSets[i].Allocations[key])
+			if err != nil {
+				t.Errorf("TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate: adding source allocation %d errored for key:%s\n", i, key)
+			}
+		}
+
+		if !allocation.Equal(sum) {
+			t.Fatalf("TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate: got: %v for allocation, expected: %v\n", allocation, sum)
+		}
```
nikovacevic: Looking for a little insight into why we're calling `Add` here, as opposed to just comparing with `Equals` the allocations in each set, back to the original.

nikovacevic (changes requested):
Just a few things to check on; happy to discuss!

nikovacevic (approved):
LGTM! Thank you!

PR #1427, opened 2022-10-14
Support MSI authentication for Azure by always returning config values if secret not found 

duncan485:
## What does this PR change?
* By always returning the configValues, we can use MSI (aadpodidentity) to authenticate for the ratecard query. This uses the already auth.NewAuthorizerFromEnvironment existing on https://github.com/opencost/opencost/blob/develop/pkg/cloud/azureprovider.go#L788. Using MSI, we'd only need the subscriptionId as config to create the NewRateCardClientWithBaseURI. However, before this PR, the subscriptionId would never be returned if the client-secret, client-id and tenant-id where not set. 

* no

* A user can now also use aadpodidentity to authenticate . Only config values to be set are:
  currencyCode
  azureBillingRegion
  azureOfferDurableID
  azureSubscriptionID

## How was this PR tested?
* Build the container and run on an AKS cluster with AAD podidentity

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -523,9 +523,23 @@ func (az *Azure) getAzureRateCardAuth(forceReload bool, cp *CustomPricing) (subs
 		tenantID = s.ServiceKey.Tenant
 		return
 	}
+	// 2. Check config values (set though endpoint)
+	if cp.AzureSubscriptionID != "" && cp.AzureClientID != "" && cp.AzureClientSecret != "" && cp.AzureTenantID != "" {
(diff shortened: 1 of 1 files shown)


AjayTripathy:
@Sean-Holcomb quick one if you could take a look?

Sean-Holcomb (changes requested):
Since `azureSubscriptionID` is still a required field can we keep a check for it. Additionally I would like to see a short comment about MSI. More than happy to merge with those changes

Sean-Holcomb (approved):

duncan485:
@AjayTripathy @Sean-Holcomb Thanks for the approval, can this one be merged ?

PR #1438, opened 2022-10-19
Make usage of Disk resilient to absence of metrics

michaelmdresser:
## What does this PR change?
Before this change, we were unable to distinguish between these scenarios:

A. Usage metric exists in Prometheus and has a value of 0 

B. Usage metric does not exist in Prometheus

Because we were using a float64, situation B would cause usage to be unset, making it the Go default value: 0. This is bad data because the usage is probably not 0.

With this commit, the Go default value becomes nil, making it easy to distinguish between situations A and B.

* Supercedes https://github.com/kubecost/kubecost-cost-model/pull/1008
* Required for https://github.com/kubecost/kubecost-cost-model/pull/1010

* Users of nightly images between the release of v1.97 and v1.98 may need to rebuild Asset data to correct byte usage data of Disks.

## How was this PR tested?
See KCM PR test notes.

### Deserializing schema v18 data built before this PR

We've never changed the type of a bingen-serialized field before. We considered it acceptable because this schema has never hit a stable or "proper" release of Kubecost. I wanted to verify that deserialized would fail as expected, thus forcing a standard rebuild of data created from the prerelase v18 schema.

I built two scenarios of data with the schema BEFORE this PR using unit tests and wrote them to disk:

```go
const marshalFile1 = "/tmp/marshaledAS_zero.bin"
const marshalFile2 = "/tmp/marshaledAS_nonzero.bin"

func TestWriteMarshaled1(t *testing.T) {
	windowStart := time.Now().UTC().Add(-4 * time.Hour).Truncate(time.Hour)
	windowEnd := windowStart.Add(1 * time.Hour)
	as := NewAssetSet(
		windowStart,
		windowEnd,
		&Disk{
			Properties: &AssetProperties{},
			Start:      windowStart,
			End:        windowEnd,

			ByteHoursUsed: 0,
		},
	)

	bytes, err := as.MarshalBinary()
	if err != nil {
		t.Fatalf("marshaling: %s", err)
	}

	err = ioutil.WriteFile(marshalFile1, bytes, 0600)
	if err != nil {
		t.Fatalf("writing: %s", err)
	}
}

func TestWriteMarshaled2(t *testing.T) {
	windowStart := time.Now().UTC().Add(-4 * time.Hour).Truncate(time.Hour)
	windowEnd := windowStart.Add(1 * time.Hour)
	as := NewAssetSet(
		windowStart,
		windowEnd,
		&Disk{
			Properties: &AssetProperties{},
			Start:      windowStart,
			End:        windowEnd,

			ByteHoursUsed: 300,
		},
	)

	bytes, err := as.MarshalBinary()
	if err != nil {
		t.Fatalf("marshaling: %s", err)
	}

	err = ioutil.WriteFile(marshalFile2, bytes, 0600)
	if err != nil {
		t.Fatalf("writing: %s", err)
	}
}
```

I then switched to this branch and attempted an unmarshal:
```go
func TestUnmarshalFile1(t *testing.T) {
	bytes, err := ioutil.ReadFile(marshalFile1)
	if err != nil {
		t.Fatalf("reading: %s", err)
	}

	var as AssetSet
	err = as.UnmarshalBinary(bytes)
	if err != nil {
		t.Fatalf("unmarshaling: %s", err)
	}
}

func TestUnmarshalFile2(t *testing.T) {
	bytes, err := ioutil.ReadFile(marshalFile2)
	if err != nil {
		t.Fatalf("reading: %s", err)
	}

	var as AssetSet
	err = as.UnmarshalBinary(bytes)
	if err != nil {
		t.Fatalf("unmarshaling: %s", err)
	}
}
```

Got expected (but funky!) unmarshal errors:
```
--- FAIL: TestUnmarshalFile1 (0.00s)
    migration_test.go:73: unmarshaling: runtime error: index out of range [16777216] with length 2
--- FAIL: TestUnmarshalFile2 (0.00s)
    migration_test.go:86: unmarshaling: runtime error: index out of range [-1073741824]

```

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -121,7 +121,7 @@ type Disk struct {
 	// a new set of metrics that may or may not be available. Thus, they must
 	// be nilable to represent the complete absence of the data.
 	//
-	// In other words, nilability here let's us distinguish between
+	// In other words, nilability here lets us distinguish between
 	// "metric is not available" and "metric is available but is 0".
 	//
 	// They end in "Ptr" to distinguish from an earlier version in order to
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -116,13 +116,24 @@ type Disk struct {
 	ClaimNamespace string
 	Cost           float64
 	Bytes          float64
-	BytesUsedAvg   float64
-	BytesUsedMax   float64
-	Local          bool
-	Start          time.Time
-	End            time.Time
-	Minutes        float64
-	Breakdown      *ClusterCostsBreakdown
+
+	// These two fields may not be available at all times because they rely on
+	// a new set of metrics that may or may not be available. Thus, they must
+	// be nilable to represent the complete absence of the data.
+	//
+	// In other words, nilability here let's us distinguish between
+	// "metric is not available" and "metric is available but is 0".
+	//
+	// They end in "Ptr" to distinguish from an earlier version in order to
+	// ensure that all usages are checked for nil.
+	BytesUsedAvgPtr *float64
+	BytesUsedMaxPtr *float64
+
+	Local     bool
+	Start     time.Time
+	End       time.Time
+	Minutes   float64
+	Breakdown *ClusterCostsBreakdown
 }
 
 type DiskIdentifier struct {
@@ -321,7 +332,7 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 				Local:     true,
 			}
 		}
-		diskMap[key].BytesUsedAvg = bytesAvg
+		diskMap[key].BytesUsedAvgPtr = &bytesAvg
 	}
 
 	for _, result := range resLocalStorageUsedMax {
@@ -346,7 +357,7 @@ func ClusterDisks(client prometheus.Client, provider cloud.Provider, start, end 
 				Local:     true,
 			}
 		}
-		diskMap[key].BytesUsedMax = bytesMax
+		diskMap[key].BytesUsedMaxPtr = &bytesMax
 	}
 
 	for _, result := range resLocalStorageBytes {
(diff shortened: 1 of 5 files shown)


pkg/costmodel/cluster.go
```diff
@@ -116,13 +116,24 @@ type Disk struct {
...
-	Start          time.Time
-	End            time.Time
-	Minutes        float64
-	Breakdown      *ClusterCostsBreakdown
+
+	// These two fields may not be available at all times because they rely on
+	// a new set of metrics that may or may not be available. Thus, they must
+	// be nilable to represent the complete absence of the data.
+	//
+	// In other words, nilability here let's us distinguish between
```
nikovacevic: let's -> lets

pkg/costmodel/cluster.go
```diff
@@ -116,13 +116,24 @@ type Disk struct {
...
+	// "metric is not available" and "metric is available but is 0".
+	//
+	// They end in "Ptr" to distinguish from an earlier version in order to
+	// ensure that all usages are checked for nil.
+	BytesUsedAvgPtr *float64
+	BytesUsedMaxPtr *float64
```
nikovacevic, line 130: Nice convention

pkg/kubecost/asset_json.go
```diff
@@ -259,7 +259,11 @@ func (d *Disk) MarshalJSON() ([]byte, error) {
 	jsonEncodeFloat64(buffer, "minutes", d.Minutes(), ",")
 	jsonEncodeFloat64(buffer, "byteHours", d.ByteHours, ",")
 	jsonEncodeFloat64(buffer, "bytes", d.Bytes(), ",")
-	jsonEncodeFloat64(buffer, "byteHoursUsed", d.ByteHoursUsed, ",")
+	if d.ByteHoursUsed == nil {
+		jsonEncode(buffer, "byteHoursUsed", nil, ",")
```
nikovacevic, line 263: Seems alright to me to just leave this out of the payload. What do you think of that? Like an `omitempty`?

nikovacevic (approved):

pkg/costmodel/cluster.go
mbolt35: Is it not clear that it's a pointer from the type?

michaelmdresser: I was using the compiler to help me find instances of the field being used so I could handle 'em safely. We can always change them back now if you want -- they aren't serialized.

PR #1443, opened 2022-10-21
Federated ETL opencost changes

kaelanspatel:
## What does this PR change?
Adds/changes a few opencost concepts in order to support Federated ETL/Federator:
* Adds `GetAssetKey`, which is just an exported version of the `key` function for assets; required to dedupe assets in the Federator.
* Adds `ListDirectories` for bucket storages, which only lists the (full) path for directories. This is required because the existing `List` truncates any directories and tries to return the base filename, meaning it's not possible to get a list of the directories in a location using it.
* Changes the config unmarshall in bucketstorage to a normal unmarshall instead of `UnmarshalStrict`. This was preventing creating multiple storages with the same config. Discussed with Bolt. 

* 

* 

* Closes ...

## How was this PR tested?
* See KCCM

* No.

*


pkg/storage/storage.go
```diff
@@ -51,6 +51,8 @@ type Storage interface {
 	// List uses the relative path of the storage combined with the provided path to return
 	// storage information for the files.
 	List(path string) ([]*StorageInfo, error)
+
+	ListDirectories(path string) ([]*StorageInfo, error)
```
mbolt35, line 58: Doc comment please. Public open source repository.

pkg/storage/filestorage.go
```diff
@@ -60,6 +61,10 @@ func (fs *FileStorage) List(path string) ([]*StorageInfo, error) {
 	return FilesToStorageInfo(files), nil
 }
 
+func (fs *FileStorage) ListDirectories(path string) ([]*StorageInfo, error) {
+	return []*StorageInfo{}, fmt.Errorf("ListDirectories not supported for filestorage")
```
mbolt35: Why not? This seems perfectly reasonable to implement?

mbolt35 (changes requested):
Add doc comment to public API ListDirectories, add implementation of ListDirectories for FileStorage

mbolt35 (approved):

PR #1445, opened 2022-10-24
Download pricing data in a goroutine to prevent deadlock when updating Azure configuration

nikovacevic:
## What does this PR change?
* Fixes an issue where updating configuration (e.g. currency) on Azure would deadlock. Now, it doesn't deadlock, and the configuration is saved.

* It will let them update their configuration!

* Closes https://github.com/kubecost/kubecost-cost-model/issues/992

## How was this PR tested?
* Manually, on `khand-dev-1` where repro steps were well documented and easy to follow! (Thank you @ivankube.) Before, the request never completed and the handler got deadlocked. After, everything works as intended.

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -19,6 +19,8 @@ import (
 	"github.com/opencost/opencost/pkg/util"
 	"github.com/opencost/opencost/pkg/util/fileutil"
 	"github.com/opencost/opencost/pkg/util/json"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 
 	"github.com/Azure/azure-sdk-for-go/services/preview/commerce/mgmt/2015-06-01-preview/commerce"
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2016-06-01/subscriptions"
@@ -39,6 +41,8 @@ const (
 	defaultSpotLabelValue            = "spot"
 	AzureStorageUpdateType           = "AzureStorage"
 )
+
+var toTitle = cases.Title(language.Und, cases.NoLower)
 
 var (
 	regionCodeMappings = map[string]string{
@@ -355,7 +359,7 @@ type AzureRetailPricing struct {
 	Count              int                            `json:"Count"`
 }
 
-//AzureRetailPricingAttributes struct for unmarshalling Azure Retail pricing api JSON response
+// AzureRetailPricingAttributes struct for unmarshalling Azure Retail pricing api JSON response
 type AzureRetailPricingAttributes struct {
 	CurrencyCode         string     `json:"currencyCode"`
 	TierMinimumUnits     float32    `json:"tierMinimumUnits"`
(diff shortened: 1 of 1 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -1208,35 +1210,46 @@ func (az *Azure) UpdateConfig(r io.Reader, updateType string) (*CustomPricing, e
...
 			c.AzureCloud = asc.AzureCloud
 		} else {
-			defer az.DownloadPricingData()
+			// This will block if not in a goroutine. It calls GetConfig(), which
+			// in turn calls GetCustomPricingData, which acquires the same lock
+			// that is acquired by az.Config.Update, which is the function to
+			// which this function gets passed, and subsequently called. Booo.
+			defer func() {
+				go az.DownloadPricingData()
+			}()
```
nikovacevic, line 1221: This is the fix. Everything else is error wrapping, gofmt, etc.

Sean-Holcomb (approved):
looks good!

AjayTripathy (approved):

AjayTripathy:
Can we make sure this gets cherrypicked to the release branch as well?

pkg/cloud/azureprovider.go
mbolt35: Not sure how to best protect ourselves against reentrant locking, or why golang doesn't support this. Provider APIs need a major face lift.

pkg/cloud/azureprovider.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/opencost/opencost/pkg/util"
 	"github.com/opencost/opencost/pkg/util/fileutil"
 	"github.com/opencost/opencost/pkg/util/json"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
```
mbolt35, line 23: What's up with these imports?

mbolt35 (changes requested):
Looks good to me outside of the imports. Maybe accidental import while debugging?

pkg/cloud/azureprovider.go
nikovacevic: Go is telling me that `strings.Title` is deprecated. This is the alternative. See:

```go
- kUpper := strings.Title(k)
```
```go
+ kUpper := cases.Title(language.Und, cases.NoLower).String(k)
```

nikovacevic: Honestly... could've just left it. But I can't resist a little clean up effort when it's there.

nikovacevic:
Cherry-picked here: https://github.com/opencost/opencost/pull/1450

pkg/cloud/azureprovider.go
mbolt35: Can we, at least do:
```
var toTitle cases.Caser = cases.Title(language.Und, cases.NoLower)

// ...

kUpper := toTitle.String(k)
```

I think this is the intentional way to use it? The title caser creation doesn't look like we should have to create it every time we want a title.

michaelmdresser:
Just a heads up here -- the cherry pick PR of this change is already merged. Let's be careful to not let things get out of sync if this PR changes.

nikovacevic:
I just need Bolt to hit the big green button

pkg/cloud/azureprovider.go
nikovacevic: Yep! Done, pushing now.

mbolt35 (approved):

PR #1447, opened 2022-10-25
WIP: v1.25 support: Update autoscaling API references from v2beta1 to v2, possibly more necessary

### LINKED ISSUE #1440: Support for k8s 1.25
Opencost is broken in k8s 1.25. Need to move from v2beta1 to v2

```
The Beta API version (autoscaling/v2beta1) of HorizontalPodAutoscaler is no longer served as of version 1.25. This API was deprecated in version 1.23. Migrate manifests and API clients to use the autoscaling/v2 HorizontalPodAutoscaler API version.
```

michaelmdresser:
## What does this PR change?
* Updates dependencies and K8s API references to support K8s v1.25

THIS IS WIP: Need to test further, especially on earlier K8s versions.

* Support added for K8s v1.25 clusters.

* Should close https://github.com/opencost/opencost/issues/1440

## How was this PR tested?
* WIP

* Unsure yet, depends on if we lose support for other K8s versions.

* Marked as v1.99, but may relabel to v1.98 if it becomes a priority.


AjayTripathy:
Is AutoscalingV2beta1() gone from v1.25 cli or just the API?

if its gone...I wonder if there's another version like 1.24 that has both and we can switch based on a flag. Or multi-import with replace directive.

michaelmdresser:
Closing in favor of https://github.com/opencost/opencost/pull/1452

PR #1450, opened 2022-10-25
Download pricing data in a goroutine to prevent deadlock when updating Azure configuration

nikovacevic:
## What does this PR change?
* Fixes an issue where updating configuration (e.g. currency) on Azure would deadlock. Now, it doesn't deadlock, and the configuration is saved.

* It will let them update their configuration!

* Closes https://github.com/kubecost/kubecost-cost-model/issues/992

## How was this PR tested?
* Manually, on `khand-dev-1` where repro steps were well documented and easy to follow! (Thank you @ivankube.) Before, the request never completed and the handler got deadlocked. After, everything works as intended.


nikovacevic:
Cherry pick of https://github.com/opencost/opencost/pull/1445

AjayTripathy (approved):

PR #1457, opened 2022-10-26
add aliased fields (labels, annotations) to filters

### LINKED ISSUE #1644: Annotations not showing up in aggregations but labels are
**Describe the bug**
Controller objects with annotations not showing up in aggregations in the UI but one that are labeled are showing up

**To Reproduce**
I am using chart version 1.96.0.
I have this relevant config in my kustomize `values.yaml`

```
kubecostMetrics:
  emitPodAnnotations: true
  emitNamespaceAnnotations: true
  emitKsmV1Metrics: true
  emitKsmV1MetricsOnly: false
  exporter:
    enabled: false

...

kubecostProductConfigs:
  currencyCode: "USD"
  labelMappingConfigs:
    enabled: true
    owner_label: "example_com_project"
    team_label: "example_com_service_owner"
    product_label: "app_kubernetes_io_name"
    owner_external_label: "example_com_project"
    team_external_label: "example_com_service_owner"
    product_external_label: "app_kubernetes_io_name"
```

I have some statefulsets with owner label `example.com/project` and some other statefulsets with annotations  `example.com/project`. The pods in the statefulsets are also labeled and annotated accordingly, because I am using `commonLabels` and `commonAnnotations` in kustomize.

In the UI when i select aggregation by "Owner", I was expecting to see all the statefulsets which are labeled as well as annotated. However, I am only seeing the ones with the label. The ones with annotation are not showing up in the aggregate. Same behavior when I use filter in the UI. Also, this seems to be the case for aggregations other than just owner, ie, same behavior for product etc.

**Expected behavior**
I was expecting to see all the statefulsets which are labeled as well as annotated with `example.com/project`

gz#2435<br><br><i>(related to <a href='https://kubecost.zendesk.com/agent/tickets/2435'>Zendesk ticket #2435</a>)</i>

nickcurie:

## What does this PR change?
* Adds ability to filter by annotations for owner, department, team, product, and environment

* 

* Users can filter by annotations for owner, department, team, product, and environment

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1644

## How was this PR tested?
* Manually, on FE and BE

* 

*

### DIFF
--- a/pkg/util/allocationfilterutil/queryfilters.go
+++ b/pkg/util/allocationfilterutil/queryfilters.go
@@ -194,19 +194,19 @@ func AllocationFilterFromParamsV1(
 	// Label-mapped queries require a label config to be present.
 	if labelConfig != nil {
 		if raw := qp.GetList("filterDepartments", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.DepartmentLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.DepartmentLabel))
 		}
 		if raw := qp.GetList("filterEnvironments", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.EnvironmentLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.EnvironmentLabel))
 		}
 		if raw := qp.GetList("filterOwners", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.OwnerLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.OwnerLabel))
 		}
 		if raw := qp.GetList("filterProducts", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.ProductLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.ProductLabel))
 		}
 		if raw := qp.GetList("filterTeams", ","); len(raw) > 0 {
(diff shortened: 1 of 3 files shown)


nikovacevic (approved):
Approving, contingent on unit testing ASAP!

pkg/util/allocationfilterutil/queryfilters.go
```diff
@@ -289,7 +289,7 @@ func filterV1LabelMappedFromList(rawFilterValues []string, labelName string) kub
 		filterValue, wildcard := parseWildcardEnd(filterValue)
 
 		subFilter := kubecost.AllocationFilterCondition{
-			Field: kubecost.FilterLabel,
+			Field: kubecost.FilterAlias,
```
michaelmdresser, line 292: I'd like to suggest a different approach here. If you look at the comment on line 26 of allocationfilter.go, you'll see this:

```
// Filtering based on label aliases (team, department, etc.) should be a
// responsibility of the query handler. By the time it reaches this
// structured representation, we shouldn't have to be aware of what is
// aliased to what.
```

I made this very intentionally so that filters wouldn't need to have any concept of aliasing. Instead, I think a slightly different approach could work here:

The `subFilter` could be `(or (= label x) (= annotation y))`. 

Does that make sense?

nikovacevic: See https://github.com/opencost/opencost/pull/1459#discussion_r1007176306

pkg/util/allocationfilterutil/queryfilters_test.go
```diff
@@ -628,6 +628,58 @@ func TestFiltersFromParamsV1(t *testing.T) {
...
+					},
+				}),
+				allocGenerator(kubecost.AllocationProperties{
+					Labels: kubecost.AllocationLabels{
+						"testowner": "notnick",
+					},
+					Annotations: kubecost.AllocationAnnotations{
+						"testowner": "nick",
+					},
+				}),
```
nikovacevic, line 680: @michaelmdresser specifically, this is the test case! Thank you @nickcurie for adding this.

nikovacevic: If it were just `label[testowner] == "nick" OR annotation[testowner] == "nick"` this would actually pass, but we want it to not pass, based on business requirements. (Whatever we may think about the requirements... is another story.)

PR #1459, opened 2022-10-26
V1.98 nick anno bugfix

### LINKED ISSUE #1644: Annotations not showing up in aggregations but labels are

nickcurie:
## What does this PR change?
* Adds ability to filter by annotations for owner, department, team, product, and environment

* 

* Users can filter by annotations for owner, department, team, product, and environment

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1644

## How was this PR tested?
* Built an image and tested the API and FE w/ drilldown

* 

*

### DIFF
--- a/pkg/util/allocationfilterutil/queryfilters.go
+++ b/pkg/util/allocationfilterutil/queryfilters.go
@@ -194,19 +194,19 @@ func AllocationFilterFromParamsV1(
 	// Label-mapped queries require a label config to be present.
 	if labelConfig != nil {
 		if raw := qp.GetList("filterDepartments", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.DepartmentLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.DepartmentLabel))
 		}
 		if raw := qp.GetList("filterEnvironments", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.EnvironmentLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.EnvironmentLabel))
 		}
 		if raw := qp.GetList("filterOwners", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.OwnerLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.OwnerLabel))
 		}
 		if raw := qp.GetList("filterProducts", ","); len(raw) > 0 {
-			filter.Filters = append(filter.Filters, filterV1LabelMappedFromList(raw, labelConfig.ProductLabel))
+			filter.Filters = append(filter.Filters, filterV1LabelAliasMappedFromList(raw, labelConfig.ProductLabel))
 		}
 		if raw := qp.GetList("filterTeams", ","); len(raw) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/util/allocationfilterutil/queryfilters.go
```diff
@@ -289,7 +289,7 @@ func filterV1LabelMappedFromList(rawFilterValues []string, labelName string) kub
 		filterValue, wildcard := parseWildcardEnd(filterValue)
 
 		subFilter := kubecost.AllocationFilterCondition{
-			Field: kubecost.FilterLabel,
+			Field: kubecost.FilterAlias,
```
michaelmdresser, line 292: I'd like to suggest a different approach here. If you look at the comment on line 26 of allocationfilter.go, you'll see this:

```
// Filtering based on label aliases (team, department, etc.) should be a
// responsibility of the query handler. By the time it reaches this
// structured representation, we shouldn't have to be aware of what is
// aliased to what.
```

I made this very intentionally so that filters wouldn't need to have any concept of aliasing. Instead, I think a slightly different approach could work here:

The `subFilter` could be `(or (= label x) (= annotation y))`. 

Does that make sense?

michaelmdresser: Sorry, just realized this is the cherry pick. Moved to main PR: https://github.com/opencost/opencost/pull/1457#discussion_r1006354483

nikovacevic: Thanks for chiming in, Michael. Unfortunately, that logic actually does not quite work. Happy to explain more, but I think this is actually the only approach that matches the precise business logic of:

```
if (label is set)
  match on label
else if (annotation is set)
  match on annotation
else
  no match
```

nikovacevic: At least, it's the only approach we could come up with on the spot. It's really a case of a regrettable product decision. Consider the case of:

* filtering on `team == "kubecost"`
* allocation with `label[team] = "monitoring"` and `annotation[team] = "kubecost"`

Nick's logic would return `false`, but your proposed change would return `true`.

nickcurie:
Closing this in favor of https://github.com/opencost/opencost/pull/1460

PR #1473, opened 2022-11-07
perf(query): Align start/end times with step

### LINKED ISSUE #1472: Downstream performance impacts on Metrics store
**Overview**

At Grafana, we've rolled out `opencost` to our fleet of clusters and are seeing a downstream impact on our Mimir databases read performance.
Our read performance SLO had ~10% drop once we rolled out `opencost` to all of our clusters(`2022-10-21`):

Image

We've been able to trace some of the slowest running queries to a number of `query_range` calls originating from `opencost`.
The main thing we noticed was the queries did not have their start/end times alinged with the step being called, which reduces the effectiveness of caching. 
We've been able to mitigate the impact of the queries coming from `opencost` by forcing Mimir to [align queries](https://grafana.com/docs/mimir/latest/operators-guide/architecture/components/query-frontend/#caching) to ensure our query frontends are using as much cached data as possible.
The SLO graph above doesn't provide much evidence of an improvement, however an adhoc query to track the time spent processing `opencost` queries is more drastic(we enforced step-aligned queries on `2022-10-28`):

Image

**Expected behavior**
I would expect range queries to have the start/end times aligned with the step to enable underlying metric systems to better utilize caching mechanisms.

`quay.io/kubecost1/kubecost-cost-model:prod-1.97.0`

**Additional Context**
While I can't go into too much details about our production systems, we've deployed `opencost` to all of our production clusters that are running in GKE, EKS, and AKS. 
The `opencost` instances query our Mimir cluster that's responsible for storing metrics for all of our clusters. 
We did not notice a major decrease in performance when running ~4 instances of `kubecost` and `opencost`.

After some personal investigation, I've been about to identify two areas where the start/end times could be aligned with the step:
- [costmodel.go](https://github.com/opencost/opencost/blob/c7706a4fb2920869500f46ead1a7c5304080187f/pkg/costmodel/costmodel.go#L1564-L1572)
- [query.go](https://github.com/opencost/opencost/blob/cb96d62cf394c41dea507c8e72e781ab84c900ab/pkg/prom/query.go#L319-L323)

Here's the code that Mimir uses to enforce the start/end times are aligned: [step_align.go](https://github.com/grafana/mimir/blob/83bca0c0aa9d4ddb5b282f97aa8f46e554fb5b90/pkg/frontend/querymiddleware/step_align.go).

My ask for the opencost folks is see if there are any potential problems with aligning the start/end times with the step for range queries. I'd be more then happy to submit a PR to review and see if we can collaborate on generating an image we can run across our fleet to test out the change.

Pokom:
Range queries will now have their start and end times aligned with the step. This will help metric stores take better advantage of caches for queries. See
(https://promlabs.com/blog/2020/06/18/the-anatomy-of-a-promql-query#range-queries) for details on range queries.

- closes #1472

## What does this PR change?
* Ensures the start and end times are aligned to the step for any range query.

* N\A

* In theory this should reduce the amount of time metric stores need to spend processing the queries. 

* closes #1472

## How was this PR tested?
* Added unit tests to validate the two added methods, will rely upon E2E tests from the kubecost team and am more then willing to run a generated container in production at Grafana

* N\A

*

### DIFF
--- a/pkg/prom/query.go
+++ b/pkg/prom/query.go
@@ -258,8 +258,17 @@ func (ctx *Context) query(query string, t time.Time) (interface{}, v1.Warnings, 
 	return toReturn, warnings, nil
 }
 
+// isRequestStepAligned will check if the start and end times are aligned with the step
+func (ctx *Context) isRequestStepAligned(start, end time.Time, step time.Duration) bool {
+	return start.Unix()%step.Milliseconds() == 0 && end.Unix()%step.Milliseconds() == 0
+}
+
 func (ctx *Context) QueryRange(query string, start, end time.Time, step time.Duration) QueryResultsChan {
 	resCh := make(QueryResultsChan)
+
+	if !ctx.isRequestStepAligned(start, end, step) {
+		start, end = ctx.alignWindow(start, end, step)
+	}
 
 	go runQueryRange(query, start, end, step, ctx, resCh, "")
 
@@ -357,6 +366,7 @@ func (ctx *Context) RawQueryRange(query string, start, end time.Time, step time.
 
 func (ctx *Context) queryRange(query string, start, end time.Time, step time.Duration) (interface{}, v1.Warnings, error) {
 	body, err := ctx.RawQueryRange(query, start, end, step)
+
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Awesome thank you! We will review this soon and are excited by the potential perf improvements!

Pokom:
@AjayTripathy I'm going to add some unit tests for aligning the start/end times, my main concern right now is around the correctness of the time conversions. Let me know if there's anything obviously off with it while I write up some tests.

Pokom:
I found an issue with the new method: `query.isRequestStepAligned` that I'm looking to resolve now. I was not sure which precision we needed on the time(started with milliseconds). I'm working on that now and some unit tests

Pokom:
@AjayTripathy I added some unit tests and fixed two bugs I found due to converting between seconds/milliseconds. Let me know when you get a chance to review the PR! I'm in the #opencost CNCF slack channel, feel free to reach out to me there as well :smile:

AjayTripathy (approved):

AjayTripathy:
This LGTM. @nikovacevic would be great if you could take a quick look but in general I think this is low risk since it only touches QueryRange and thus only /model/allocation/compute codepaths.

nikovacevic:
@AjayTripathy I don't think it even touches that. (All I see using `QueryRange` is `ComputeCostDataRange` -- i.e. `CostDataModelRange`, `AggregatedCostModel` -- `ClusterCostsOverTime`, and some health queries.)

Seems fine to me.

PR #1478, opened 2022-11-10
feat(config): Expose DisableCache as an environment variable

Pokom:
This adds the ability to set the value for `disableCache` via an environment variable `DISABLE_CACHE`. The default value is set to false. This will influence whether ComputeAggregateCostModel will use data if it's found in the cache.

- Relates to #1472

Signed-off-by: pokom <mark.poko@grafana.com>

## What does this PR change?
* Adds an environment variable `DISABLE_CACHE` that defaults to false

* #1473

* Users can now *opt in* to disabling the cache

* 

## How was this PR tested?
* Added unit tests to ensure the value defaults to false and can be updated

* Most likely need to add `DISABLE_CACHE` to environment variables that can be overriden

*

### DIFF
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -32,7 +32,7 @@ const (
 	CSVPathEnvVar                  = "CSV_PATH"
 	ConfigPathEnvVar               = "CONFIG_PATH"
 	CloudProviderAPIKeyEnvVar      = "CLOUD_PROVIDER_API_KEY"
-	DisableCache                   = "DISABLE_CACHE"
+	DisableAggregateCostModelCache = "DISABLE_AGGREGATE_COST_MODEL_CACHE"
 
 	EmitPodAnnotationsMetricEnvVar       = "EMIT_POD_ANNOTATIONS_METRIC"
 	EmitNamespaceAnnotationsMetricEnvVar = "EMIT_NAMESPACE_ANNOTATIONS_METRIC"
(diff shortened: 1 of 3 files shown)


AjayTripathy:
@nikovacevic for review,  think this makes sense to bring back as an option in light of increased usage of the direct open source APIs?

nikovacevic:
Yeah, I'm not opposed. But we'll need a more specific name for the cache.

pkg/env/costmodelenv.go
```diff
@@ -239,6 +240,12 @@ func GetInsecureSkipVerify() bool {
 	return GetBool(InsecureSkipVerify, false)
 }
 
+// IsCacheDisabled returns the environment variable value for DisableCache which
+// will inform the aggregator on whether to load cached data. Defaults to false
+func IsCacheDisabled() bool {
```
nikovacevic: @Pokom would you be open to renaming this function and the associated variable `IsAggregateCostModelCacheDisabled()` and `DISABLE_AGGREGATE_COST_MODEL_CACHE`, respectively?

Pokom: Certainly! I just pushed a commit to update the references to be explicit in what cache is being set. In addition to the method definitions, I updated the other `disableCache` variables to be `disableAggregateCostModelCache`. Let me know if that's a problem!

nikovacevic: Not at all -- very helpful, thank you.

nikovacevic (approved):

nikovacevic:
@AjayTripathy feel free to take a final pass if you like, or merge when ready

PR #1484, opened 2022-11-15
strictly define minute boundaries to avoid some bad data edge cases

AjayTripathy:
Signed-off-by: Ajay Tripathy <4tripathy@gmail.com>

## What does this PR change?
*  Very rarely, kube-state-metrics reports garbage data for pods that translates into bad data in opencost. This enforces logic around it by strictly forcing prometheus data points to be between the windows, or be disregarded. Here's a sample of the bad data that we intend to mend:

* No

* this bad data should disappear.
```
{
name: "",
properties: {
cluster: "prod--us-east-1--prod",
node: "ip-100-xx-xx-244.ec2.internal",
controller: "canary",
controllerKind: "deployment",
namespace: "wam",
pod: "canary-76445dfcdb-6rrjc",
providerID: "i-0892a2e4a0c7861b8"
},
window: {
start: "2022-10-30T00:00:00Z",
end: "2022-10-31T00:00:00Z"
},
start: "2027-01-17T10:31:11Z",
end: "2018-08-12T13:18:48Z",
minutes: -4435032.380797,
cpuCores: 0,
cpuCoreRequestAverage: 0,
cpuCoreUsageAverage: 0,
cpuCoreHours: -138894009504962.98,
cpuCost: -1.970269753034173e+25,
cpuCostAdjustment: 0,
cpuEfficiency: 0,
gpuCount: 0,
gpuHours: 0,
gpuCost: 0,
gpuCostAdjustment: 0,
networkTransferBytes: 0,
networkReceiveBytes: 324702793.142857,
networkCost: 0,
networkCostAdjustment: 0,
loadBalancerCost: 0,
loadBalancerCostAdjustment: 0,
pvBytes: 0,
pvByteHours: 0,
pvCost: 0,
pvs: null,
pvCostAdjustment: 0,
ramBytes: 0,
ramByteRequestAverage: 0,
ramByteUsageAverage: 0,
ramByteHours: -11238632322353.56,
ramCost: -3009.52252,
ramCostAdjustment: 0,
ramEfficiency: 0,
sharedCost: 0,
externalCost: 0,
totalCost: -1.970269753034173e+25,
totalEfficiency: 0,
rawAllocationOnly: null
},```

* Possibly closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1803

## How was this PR tested?
* Manually by simulating data. It's meant to stop bad data emission so

* 

*


michaelmdresser (commented):
If I'm not mistaken, we're getting the start and end of a given Pod from the _Prometheus_ time (`datum.Timestamp`): https://github.com/opencost/opencost/blob/1795bcddb1d91d3e60772030528274c4dff29185/pkg/costmodel/allocation_helpers.go#L2018-L2040

 Can you help me understand how KSM emitting "garbage data" can mess up the timestamps in Prometheus itself? My assumption is that Prometheus itself is responsible for setting timestamps on data. I could be wrong!

However, if that is the case, it seems to me like either one of two things could cause the example you've attached:
- We're executing a bad query. I suspect this isn't it, because the Window on that Allocation looks good.
- The Prometheus or Prometheus lookalike responding to the query isn't respecting the time range bounds.

Is there a way to check for the above?

AjayTripathy:
It's the latter; I've run queries with users over the years to verify this. Timestamps are determined by machine time which we have no real control over:

https://github.com/prometheus/prometheus/issues/6554#issuecomment-644178203

PR #1487, opened 2022-11-17
Add optional configuration of regions for AWS, Azure, GCP, and Scaleway

### LINKED ISSUE #1382: Make AWS regions configurable
In our environment, opencost is not allowed IAM permissions to access resources in all AWS regions. 
This causes a lot of IAM errors with PermissionDenied due to the following hardcoded list of regions (https://github.com/opencost/opencost/blob/cda202ddedb57d02bc307ea63052b29f1a504c97/pkg/cloud/awsprovider.go#L111-L138)

This is used in the GetDisks https://github.com/opencost/opencost/blob/cda202ddedb57d02bc307ea63052b29f1a504c97/pkg/cloud/awsprovider.go#L1528 and GetAddresses functions https://github.com/opencost/opencost/blob/cda202ddedb57d02bc307ea63052b29f1a504c97/pkg/cloud/awsprovider.go#L1445

Is there any way that this could be made configurable, so that we could specify only the regions we are interested in?

jcttrll:
## What does this PR change?
* Add optional configuration of regions for AWS, Azure, GCP, and Scaleway, instead of strictly hardcoded regions
* A change to how GCP is identified as a provider was made to facilitate testing - this needs review by someone more familiar with this project

* No

* If desired, users will be able to add `"regions":["name", ...]` to their cloud configuration files.

* Closes #1382

## How was this PR tested?
* Tests were added to `test/cloud_test.go` to confirm proper functionality under happy path and other cases.

* Possibly. Comments on #1382 indicate that issue was going to be ignored because cloud providers will, in the future, not be part of OpenCost. In that case, it may be preferred to leave this as an undocumented feature.

* No. This does not seem like my decision, but that of the project maintainers.

### DIFF
--- a/pkg/cloud/provider.go
+++ b/pkg/cloud/provider.go
@@ -138,67 +138,68 @@ type OutOfClusterAllocation struct {
 }
 
 type CustomPricing struct {
-	Provider                     string `json:"provider"`
-	Description                  string `json:"description"`
-	CPU                          string `json:"CPU"`
-	SpotCPU                      string `json:"spotCPU"`
-	RAM                          string `json:"RAM"`
-	SpotRAM                      string `json:"spotRAM"`
-	GPU                          string `json:"GPU"`
-	SpotGPU                      string `json:"spotGPU"`
-	Storage                      string `json:"storage"`
-	ZoneNetworkEgress            string `json:"zoneNetworkEgress"`
-	RegionNetworkEgress          string `json:"regionNetworkEgress"`
-	InternetNetworkEgress        string `json:"internetNetworkEgress"`
-	FirstFiveForwardingRulesCost string `json:"firstFiveForwardingRulesCost"`
-	AdditionalForwardingRuleCost string `json:"additionalForwardingRuleCost"`
-	LBIngressDataCost            string `json:"LBIngressDataCost"`
-	SpotLabel                    string `json:"spotLabel,omitempty"`
-	SpotLabelValue               string `json:"spotLabelValue,omitempty"`
-	GpuLabel                     string `json:"gpuLabel,omitempty"`
-	GpuLabelValue                string `json:"gpuLabelValue,omitempty"`
-	ServiceKeyName               string `json:"awsServiceKeyName,omitempty"`
-	ServiceKeySecret             string `json:"awsServiceKeySecret,omitempty"`
-	SpotDataRegion               string `json:"awsSpotDataRegion,omitempty"`
-	SpotDataBucket               string `json:"awsSpotDataBucket,omitempty"`
-	SpotDataPrefix               string `json:"awsSpotDataPrefix,omitempty"`
-	ProjectID                    string `json:"projectID,omitempty"`
-	AthenaProjectID              string `json:"athenaProjectID,omitempty"`
-	AthenaBucketName             string `json:"athenaBucketName"`
-	AthenaRegion                 string `json:"athenaRegion"`
-	AthenaDatabase               string `json:"athenaDatabase"`
-	AthenaTable                  string `json:"athenaTable"`
-	AthenaWorkgroup              string `json:"athenaWorkgroup"`
-	MasterPayerARN               string `json:"masterPayerARN"`
-	BillingDataDataset           string `json:"billingDataDataset,omitempty"`
-	CustomPricesEnabled          string `json:"customPricesEnabled"`
-	DefaultIdle                  string `json:"defaultIdle"`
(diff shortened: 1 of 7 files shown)


pkg/cloud/provider.go
```diff
@@ -517,7 +518,7 @@ func getClusterProperties(node *v1.Node) clusterProperties {
 		accountID:      "",
 		projectID:      "",
 	}
-	if metadata.OnGCE() {
+	if metadata.OnGCE() || strings.HasPrefix(providerID, "gce") {
```
jcttrll, line 526: This change, in particular, needs review to ensure it's appropriate. `metadata.OnGCE()` internally uses `sync.Once`, which means, in the context of `cloud_test.go`, either you're on GCP for all the tests or you're not on GCP for any of the tests.

I don't know the history of why this condition uses `metadata.OnGCE()`, instead of looking at the provider ID prefix, as the other conditions do. I know next to nothing about GCP or what its provider IDs would look like in KubeCost. Based on some other things--in particular, `providerGCERegex` in `provider.go`--I think the `gce` prefix is appropriate. If this is OK, is `metadata.OnGCE()` needed at all?

AjayTripathy:
Amazing! Thank you so much for the PR @jcttrll 
@kaelanspatel could you help me take this review?

pkg/cloud/provider.go
kaelanspatel: From my understanding, this merely an "official" GCP client library way to determine this, and is therefore more advisable to use than the prefix check.

Is it possible to work around the testing limitation? Perhaps change `MetadataHostEnv` in the tests to reflect the provider that should be used (see [here](https://github.com/googleapis/google-cloud-go/blob/601c77a69a27b5f13ebb4508f8222a98c8a904bc/compute/metadata/metadata.go#L114-L189))?

dwbrown2:
@kaelanspatel @AjayTripathy any updates? Planning for 1.99?

cc @Adam-Stack-PM

kaelanspatel:
Hey, didn't see this earlier @dwbrown2 sorry.

Changes look okay overall, tested default behavior (default config) _en vivo_ and found no issues. Was not able to get custom regions working, though. We may also need a helm chart PR, is my guess.

Also not sure about the issue with `metadata.OnGCE()`. I don't _think_ it's a big risk as-is, fwiw.

AjayTripathy:
@kaelanspatel guessing this won't make 1.99. It seems like there are conflicts to resolve etc. Maybe you can pick up the conflict resolution and get this across the finish line shortly after release if this mostly looks good?

Adam-Stack-PM:
@kaelanspatel Any updates on the above? If possible would love to get this into v1.100.

kaelanspatel:
Hey @jcttrll, sorry for the long interval without a check-in. 

Any updates on the above conversation? Even if not, I've looked this over and after testing _en vivo_, I'm comfortable merging with that change to the `OnGCE` line.

As a final change, could `cloud_test.go` be moved into `pkg/cloud`? I'll approve and merge after that.

jcttrll:
Hi, @kaelanspatel and @AjayTripathy. Here's a little background on this PR...

A team at the company where I work bought a KubeCost license. They installed it, and it started triggering security logs in CloudTrail by violating our policies and accessing forbidden regions. After several weeks of back-and-forth with KubeCost support, they finally understood the issue and added these comments to an existing issue:

* https://github.com/opencost/opencost/issues/1382#issuecomment-1318871625
* https://github.com/opencost/opencost/issues/1382#issuecomment-1319163096

I wanted to fix the security violations, which cause a problem for me. I looked at the source and found the regions had been hardcoded, rather than configurable. At the time, f797a15da92b92e7cf10f7edcc5be0aac481739a was the current state. I was puzzled by why the testing was done in such a non-idiomatic way for Go, with a `test/cloud_test.go` file lumping all the "cloud tests" together (rather than having `*_test.go` files alongside the other code). I figured this was just because the developers on this project were new to Go and didn't know all the idioms, yet, and followed the "When in Rome..." principle, updating `cloud_test.go` accordingly.

Unfortunately, in the months that have passed since November, I no longer have more free time to spend on this, so I won't be able to refactor your tests for you. I understand this issue has been a low priority; while I'm disappointed my company's licensing fee didn't increase that priority, I understand that your team, also, doesn't have the time to work on this. I think it's time to just let it go. I'll close the MR. Thanks for your time in looking at it, anyway!

AjayTripathy:
Hey @jcttrll thank you for the context and feedback. Let us take over this PR and make sure we get the logs resolved and #1382 fixed.

PR #1488, opened 2022-11-17
Support Download pricing + Node pricing on Aliyun aka Alibaba cloud provider 

avrodrigues5:
## What does this PR change?
* Download pricing and Node pricing minus the local disk price adjust as a starting PR for initial Alibaba cloud aka aliyun provider support in Opencost

* [KCM PR](https://github.com/kubecost/kubecost-cost-model/pull/1074)
* [Helm Chart PR](https://github.com/kubecost/cost-analyzer-helm-chart/pull/1811)

* Can be able to see node hourly price on pay-as-you-go node subscription on Alibaba cluster.

* Closes ...

## How was this PR tested?
* Get the information about the K8s cluster on aliyun provider with 3 nodes displayed in <IP>/metrics, prometheus server and <IP>/allNodePricing

* 
Screen Shot 2022-11-17 at 12 58 37 PM
Screen Shot 2022-11-17 at 12 58 53 PM
Screen Shot 2022-11-17 at 12 59 44 PM

* v.1.99


configs/aliyun.json
```diff
@@ -0,0 +1,12 @@
+{
+    "provider": "Aliyun",
+    "description": "Aliyun Json have to figure out what the contents need to look like. Defaulting it to AWS json for now",
```
michaelmdresser: I think stealing the default node prices from the AWS JSON is fine here.  Let's make this description a little more professional, though.

Also, for my own info, "Aliyun" vs. "Alibaba" here? Is "Aliyun" specific to the cloud provider offering of Alibaba?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+}
+
+// AliyunDiskAttributes represents metadata about the product used to map to a PV.
+// Basic Attributes needed atleast to get the keys, Some attributes from k8s Node response
+// be populated directly into *PV object.
+type AliyunPVAttributes struct {
+	DiskType         int32  `json:"diskType"`
+	DiskCategory     string `json:"diskCategory"`
+	PerformanceLevel string `json:"performanceLevel"`
+}
```
michaelmdresser: Is this type used only for PVs, or also for local disks?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+
+// Stage 1 support will be Pay-As-You-Go with HourlyPrice equal to TradePrice with PriceUnit as Hour
+// TO-DO: Subscription and Premptible support, need to find how to distinguish node into these categories]
+// TO-DO: Open question Subscription would be either Monthly or Yearly, Firstly Data retrieval/population
+// TO-DO:  need to be tested from describe price API, but how would you calculate hourly price, is it PRICE_YEARLY/HOURS_IN_THE_YEAR?
+type AliyunPricingDetails struct {
+	HourlyPrice  float32 `json:"hourlyPrice"`  // Represents hourly price for the given Aliyun Product
```
michaelmdresser: Rather than trailing these comments after the line, would you put them above? E.g.

```go
type Foo struct {
    // Bar is XYZ. Like other docstrings, this will show up
    // in official documentation on the field because it is on
    // the line before the field and starts with the same name
    // as the field.
    Bar float32
}
```

The same goes for other trailing comments on struct fields in this PR

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+		InstanceType:       instanceType,
+		RegionID:           regionID,
+		PriceUnit:          priceUnit,
+		MemorySizeInKiB:    memorySizeInKiB,
+		IsIoOptimized:      isIOOptimized,
+		OSType:             osType,
+		ProviderID:         providerID,
+		InstanceTypeFamily: instanceTypeFamily,
+	}
+}
```
michaelmdresser, line 127: I don't have really strong feelings here, but is there any reason to include a `New...()` function for a struct if you could just as easily initialize it directly (e.g. `&SlimK8sNode{ ... }`)? I see merits to both, just curious about your thoughts.

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	NodeAttributes *AliyunNodeAttributes
+	PVAttributes   *AliyunPVAttributes
+	PricingTerms   *AliyunPricingTerms
+	Node           *Node
+	PV             *PV
+}
+
+// Aliyun's Provider struct
+type Aliyun struct {
+	Pricing                 map[string]*AliyunPricing // Data to store Aliyun(Alibaba cloud) pricing struct
```
michaelmdresser: What does the string in this map represent? That would be very helpful in a comment.

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	DownloadPricingDataLock sync.RWMutex              // Lock Needed to provide thread safe
+	Clientset               clustercache.ClusterCache
+	Config                  *ProviderConfig
+	serviceAccountChecks    *ServiceAccountChecks
+	clusterAccountId        string
+	clusterRegion           string
+	loadedAccessKey         bool                             // Check if Aliyun is authenticated
+	accessKey               *credentials.AccessKeyCredential // Aliyun Access key used specifically in signer interface used to sign API calls
+	clients                 map[string]*sdk.Client           // Map of regionID to sdk.client to call API for that region
+	*CustomProvider
```
michaelmdresser: Is there a reason why some of these fields are exported (starting with a capital letter) and some aren't?

If there is, would you please separate them into exported and unexported blocks to read a little more easily? A comment explaining why would also be helpful! Like this:
```go
type Foo struct {
    Exported1
    Exported2

    // The following fields are unexported because XYZ.

    unexported1
    unexported2
}
```

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+// Aliyun's Provider struct
+type Aliyun struct {
+	Pricing                 map[string]*AliyunPricing // Data to store Aliyun(Alibaba cloud) pricing struct
+	DownloadPricingDataLock sync.RWMutex              // Lock Needed to provide thread safe
+	Clientset               clustercache.ClusterCache
+	Config                  *ProviderConfig
+	serviceAccountChecks    *ServiceAccountChecks
+	clusterAccountId        string
+	clusterRegion           string
+	loadedAccessKey         bool                             // Check if Aliyun is authenticated
```
michaelmdresser: I like to use "is" for variables like this, so you get something like `Aliyun.isAccessKeyLoaded` or `Aliyun.accessKeyIsLoaded`. That makes it really clear that its a boolean flag. No requirement here, just a thought!

michaelmdresser: And perhaps another question -- is this variable always equivalent to `Aliyun.accessKey != nil`? If so, maybe we don't need this variable at all and could replace it with a method like `func (a *Aliyun) accessKeyIsLoaded() { return a.accessKey != nil }`

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	accessKey               *credentials.AccessKeyCredential // Aliyun Access key used specifically in signer interface used to sign API calls
+	clients                 map[string]*sdk.Client           // Map of regionID to sdk.client to call API for that region
+	*CustomProvider
+}
+
+// GetAliyunAccessKey return the Access Key used to interact with the Alibaba cloud, if not set it
+// set it first by looking at env variables else load it from secret files.
+// <IMPORTANT>Ask in PR what is the exact purpose of so many functions to set the key in AWS providers, am i missing something here!!!!!
```
michaelmdresser, line 229: This might be a good question for Sean H, Bianca, or Niko

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	}
+
+	err = env.Set(env.AliyunAccessKeyIDEnvVar, ak.AccessKeyID)
+	if err != nil {
+		return err
+	}
+	err = env.Set(env.AliyunAccessKeySecretEnvVar, ak.SecretAccessKey)
+	if err != nil {
+		return err
+	}
```
michaelmdresser: Setting environment variables mid-run of a program and planning to read them back later is a really bizarre pattern. I'm not sure if I've ever seen it before -- it bothers me because the environment becomes a mutable thing and thus very difficult to reason about.

This may be mimicking other provider code, though...

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	defer aliyun.DownloadPricingDataLock.Unlock()
+
+	var aak *credentials.AccessKeyCredential
+	var err error
+
+	if !aliyun.loadedAccessKey {
+		aak, err = aliyun.GetAliyunAccessKey()
+		if err != nil {
+			log.Errorf("Unable to get the access key information")
+			return fmt.Errorf("unable to get the access key information")
```
michaelmdresser: Please wrap the error, rather than throwing it away

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+
+	c, err := aliyun.Config.GetCustomPricingData()
+	if err != nil {
+		log.Errorf("Error downloading default pricing data: %s", err.Error())
+		return fmt.Errorf("error downloading default pricing data: %s", err.Error())
+	}
+
+	// // Get the nodes in Aliyun provider, Once alibaba cloud is setup
+	// // and populate data from the node object to resemble the data in hardcodeNodes
+	nodeList := aliyun.Clientset.GetAllNodes()
```
michaelmdresser: This comment seems a bit out of date, would you please update it?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	var lookupKey string
+	aliyun.clients = make(map[string]*sdk.Client)
+	aliyun.Pricing = make(map[string]*AliyunPricing)
+
+	// TO-DO: Add disk price adjustment by parsing the local disk information and putting it as a param in describe Price function.
+	for _, node := range nodeList {
+		slimK8sNode := generateSlimK8sNodeFromV1Node(node)
+		lookupKey, err = determineKeyForPricing(slimK8sNode)
+		if _, ok := aliyun.Pricing[lookupKey]; ok {
+			log.Infof("Pricing information for node with same features %s already exists hence skipping", lookupKey)
```
michaelmdresser: This seems like a debug-level log

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+			continue
+		}
+		// TO-DO: Check if Node pricing already available , skip it if available
```
michaelmdresser: Is this TODO not what the above logic is doing?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+				return fmt.Errorf("access key provided does not have access to location %s", slimK8sNode.RegionID)
+			}
+			aliyun.clients[slimK8sNode.RegionID] = client
+		}
+		signer = signers.NewAccessKeySigner(aak)
+		pricingObj, err = processDescribePriceAndCreateAliyunPricing(client, slimK8sNode, signer, c)
+
+		if err != nil {
+			return err
+		}
```
michaelmdresser: Would you please wrap this error return with a helpful description of what was being attempted?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	// Get node features for the key
+	keyFeature := key.Features()
+
+	pricing, ok := aliyun.Pricing[keyFeature]
+	if !ok {
+		log.Infof("Node pricing information not found for node with feature: %s", keyFeature)
+		return &Node{}, nil
+	}
+
+	log.Infof("returing the node price for the node with feature: %s", keyFeature)
```
michaelmdresser: This seems like a debug- or trace-level log

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+
+// NodePricing gives a specific node for the key
+func (aliyun *Aliyun) NodePricing(key Key) (*Node, error) {
+	aliyun.DownloadPricingDataLock.RLock()
+	defer aliyun.DownloadPricingDataLock.RUnlock()
+
+	// Get node features for the key
+	keyFeature := key.Features()
+
+	pricing, ok := aliyun.Pricing[keyFeature]
```
michaelmdresser: What is a "feature"? Is there a better name?

pkg/cloud/aliyunprovider.go
michaelmdresser:
```suggestion
	log.Infof("returning the node price for the node with feature: %s", keyFeature)
```

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	keyFeature := pvk.Features()
+
+	pricing, ok := aliyun.Pricing[keyFeature]
+
+	if !ok {
+		log.Infof("Persistent Volume pricing not found for PV with feature: %s", keyFeature)
+		return &PV{}, nil
+	}
+
+	log.Infof("returing the PV price for the node with feature: %s", keyFeature)
```
michaelmdresser: Similar log leveling, also typo
```suggestion
	log.Infof("returning the PV price for the node with feature: %s", keyFeature)
```

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+// the input parameter.
+func (aliyun *Aliyun) loadAliyunAuthSecretAndSetEnv(force bool) error {
+	if !force && aliyun.loadedAccessKey {
+		return nil
+	}
+
+	exists, err := fileutil.FileExists(authSecretPath)
+	if !exists || err != nil {
+		return fmt.Errorf("Failed to locate service account file: %s", authSecretPath)
+	}
```
michaelmdresser: Including the underlying `err` might be helpful in this return

pkg/cloud/aliyunprovider.go
michaelmdresser: Please wrap these error returns with some context

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	return aliyunRegions
+}
+
+// ClusterInfo returns information about ALiyun cluster, as provided by metadata. TO-DO: Look at this function closely at next PR iteration
+func (aliyun *Aliyun) ClusterInfo() (map[string]string, error) {
+
+	c, err := aliyun.GetConfig()
+	if err != nil {
+		log.Errorf("Error opening config: %s", err.Error())
+	}
```
michaelmdresser: Should this be a `return nil, fmt.Errorf(...)` rather than just a log?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+
+	var clusterName string
+	if c.ClusterName != "" {
+		clusterName = c.ClusterName
+	}
+
+	// Use a default name if none has been set until this point
+	if clusterName == "" {
+		clusterName = "Aliyun Cluster #1"
+	}
```
michaelmdresser: Why do we have a default at all? Is it any worse to use empty string, or `env.GetClusterID()` for this? I know I've heard SEs talk about recommending cluster name and ID be set the same, but there may be some quirk of logic here that I'm not aware of.

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+func (aliyunNodeKey *AliyunNodeKey) GPUType() string {
+	return ""
+}
+
+func (aliyunNodeKey *AliyunNodeKey) GPUCount() int {
+	return 0
+}
+
+// Get's the key for the k8s node input
+func (aliyun *Aliyun) GetKey(mapValue map[string]string, node *v1.Node) Key {
```
michaelmdresser: Is the `mapValue` argument necessary because we're fulfulling an interface?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+// When supporting different new type of instances like Compute Optimized, Memory Optimized etc make sure you add the instance type
+// in unit test and check if it works or not to create the ack request and processDescribePriceAndCreateAliyunPricing function
+// else more paramters need to be pulled from kubernetes node response or gather infromation from elsewhere and function modified.
+// TO-DO: Add disk adjustments to the node , Test it out!
+func createDescribePriceACSRequest(i interface{}) (*requests.CommonRequest, error) {
+	request := requests.NewCommonRequest()
+	request.Method = requests.GET
+	request.Product = ALIYUN_ECS_PRODUCT_CODE
+	request.Domain = ALIYUN_ECS_DOMAIN
+	request.Version = ALIYUN_ECS_VERSION
```
michaelmdresser: Is it "ECS" or "ACS"? Or are they separate things?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+		disk := i.(*SlimK8sDisk)
+		request.QueryParams["RegionId"] = disk.RegionID
+		request.QueryParams["ResourceType"] = ALIYUN_DISK_RESOURCE_TYPE
+		request.QueryParams["DataDisk.1.Category"] = disk.DiskCategory
+		request.QueryParams["DataDisk.1.Size"] = fmt.Sprintf("%d", disk.SizeInGiB)
+		request.QueryParams["PriceUnit"] = disk.PriceUnit
+		request.TransToAcsRequest()
+		return request, nil
+	default:
+		return nil, fmt.Errorf("unsupported ECS type for DescribePrice at this time")
```
michaelmdresser: You can add the type to help the user:
```suggestion
	default:
		return nil, fmt.Errorf("unsupported ECS type (%T) for DescribePrice at this time", i)
```

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	}
+}
+
+// determineKeyForPricing generate a unique key from SlimK8sNode object that is construct from v1.Node object.
+// This function is
+func determineKeyForPricing(i interface{}) (string, error) {
+	switch i.(type) {
+	case *SlimK8sNode:
```
michaelmdresser: Looking for the same `nil` checks on `i` here!

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+func createDescribePriceACSRequest(i interface{}) (*requests.CommonRequest, error) {
+	request := requests.NewCommonRequest()
+	request.Method = requests.GET
+	request.Product = ALIYUN_ECS_PRODUCT_CODE
+	request.Domain = ALIYUN_ECS_DOMAIN
+	request.Version = ALIYUN_ECS_VERSION
+	request.Scheme = requests.HTTPS
+	request.ApiName = ALIYUN_DESCRIBE_PRICE_API_ACTION
+	switch i.(type) {
+	case *SlimK8sNode:
```
michaelmdresser, line 668: Please make sure `i != nil` before type switching on it and accessing it.

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	}
+}
+
+func (aliyunNodeKey *AliyunNodeKey) ID() string {
+	return aliyunNodeKey.ProviderID
+}
+
+func (aliyunNodeKey *AliyunNodeKey) Features() string {
+	keyLookup := stringutil.DeleteEmptyStringsFromArray([]string{aliyunNodeKey.RegionID, aliyunNodeKey.InstanceType, aliyunNodeKey.OSType, aliyunNodeKey.OptimizedKeyword})
+	return strings.Join(keyLookup, "::")
```
michaelmdresser: Does Ali support some of these parameters (e.g. InstanceType) being absent from this "lookup" key? Is that why we have the `DeleteEmptyStringsFromArray` here?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+type DescribePriceResponse struct {
+	RequestId string    `json:"RequestId"`
+	PriceInfo PriceInfo `json:"PriceInfo"`
+}
+
+// processDescribePriceAndCreateAliyunPricing processes the DescribePrice API and generates the pricing information for alibaba node resource.
+func processDescribePriceAndCreateAliyunPricing(client *sdk.Client, i interface{}, signer *signers.AccessKeySigner, custom *CustomPricing) (pricing *AliyunPricing, err error) {
+	pricing = &AliyunPricing{}
+	var response DescribePriceResponse
+	log.Infof("type is %v", i)
```
michaelmdresser: Is this log necessary? If so, let's put it at Debug or Trace level

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+	switch i.(type) {
+	case *SlimK8sNode:
```
michaelmdresser, line 744: Please make sure `i` is not nil

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+		node := i.(*SlimK8sNode)
+		req, err := createDescribePriceACSRequest(node)
+		log.Debugf("Request is : %v", req)
+		if err != nil {
+			return nil, err
+		}
+		resp, err := client.ProcessCommonRequestWithSigner(req, signer)
+		log.Infof("value is %s", resp.GetHttpContentString())
```
michaelmdresser: Similar log cleanup question

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+// processDescribePriceAndCreateAliyunPricing processes the DescribePrice API and generates the pricing information for alibaba node resource.
+func processDescribePriceAndCreateAliyunPricing(client *sdk.Client, i interface{}, signer *signers.AccessKeySigner, custom *CustomPricing) (pricing *AliyunPricing, err error) {
+	pricing = &AliyunPricing{}
+	var response DescribePriceResponse
+	log.Infof("type is %v", i)
+	switch i.(type) {
+	case *SlimK8sNode:
+		node := i.(*SlimK8sNode)
+		req, err := createDescribePriceACSRequest(node)
+		log.Debugf("Request is : %v", req)
```
michaelmdresser: If we're going to keep this log, let's prefix it with at least "Aliyun" or something like that to speed up understanding of the source of the log.

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+		if err != nil {
+			return nil, err
+		}
+		resp, err := client.ProcessCommonRequestWithSigner(req, signer)
+		log.Infof("value is %s", resp.GetHttpContentString())
+		pricing.NodeAttributes = NewAliyunNodeAttributes(node)
+		if err != nil || resp.GetHttpStatus() != 200 {
+			// Can be defaulted to some value here?
+			return nil, fmt.Errorf("unable to fetch information for node with InstanceType: %v", node.InstanceType)
```
michaelmdresser, line 754: Please include `err` in the error message

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+		} else {
+			// This is where population of Pricing happens
+			err = json.Unmarshal(resp.GetHttpContentBytes(), &response)
+			if err != nil {
+				return nil, fmt.Errorf("unable to unmarshall json response to custom struct, possible change in json response of DescribePrice")
```
michaelmdresser: Please include `err` in the error message.

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+			}
+			// TO-DO : Ask in PR How to get the defaults is it equal to AWS/GCP defaults? And what needs to be returned
+			pricing.Node = &Node{
+				Cost:         fmt.Sprintf("%f", response.PriceInfo.Price.TradePrice),
+				BaseCPUPrice: custom.CPU,
+				BaseRAMPrice: custom.RAM,
+				BaseGPUPrice: custom.GPU,
+			}
```
michaelmdresser, line 767: I suspect the defaults are the same, but I could absolutely be wrong. It's a perfectly reasonable place to start, at least!

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+			// TO-DO : Currently with Pay-As-You-go Offering TradePrice = HourlyPrice , When support happens to other type HourlyPrice Need to be determined.
+			pricing.PricingTerms = NewAliyunPricingTerms(ALIYUN_PAY_AS_YOU_GO_BILLING, NewAliyunPricingDetails(response.PriceInfo.Price.TradePrice, ALIYUN_HOUR_PRICE_UNIT, response.PriceInfo.Price.TradePrice, response.PriceInfo.Price.Currency))
```
michaelmdresser: Do we log some sort of helpful error or warning if the user has something other than "pay as you go" billing?

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+			return nil, err
+		}
+		resp, err := client.ProcessCommonRequestWithSigner(req, signer)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch information for disk with DiskType: %v", disk.DiskType)
+		} else {
+			// This is where population of Pricing happens
+			err = json.Unmarshal(resp.GetHttpContentBytes(), &response)
+			if err != nil {
+				return nil, fmt.Errorf("unable to unmarshall json response to custom struct, possible change in json response of DescribePrice")
```
michaelmdresser: Error wrapping

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+			}
+			pricing.PVAttributes = &AliyunPVAttributes{}
+			pricing.PV = &PV{
+				Cost: fmt.Sprintf("%f", response.PriceInfo.Price.TradePrice),
+			}
+
+		}
+	default:
+		return nil, fmt.Errorf("unsupported ECS pricing component at this time")
```
michaelmdresser:
```suggestion
	default:
		return nil, fmt.Errorf("unsupported ECS pricing component (%T) at this time", i)
```

pkg/cloud/aliyunprovider.go
```diff
@@ -0,0 +1,828 @@
...
+		// HIGHLY UNLIKELY THAT THIS LABEL WONT BE THERE.
+		log.Debugf("OS type undetected for the node: %s", node.Name)
+	}
+	if instanceType, ok = node.Labels["node.kubernetes.io/instance-type"]; !ok {
+		// HIGHLY UNLIKELY THAT THIS LABEL WONT BE THERE.
+		log.Debugf("Instance Type undetected for the node: %s", node.Name)
+	}
+
+	instanceFamily = getInstanceFamilyFromType(instanceType)
+	memorySizeInKiB = fmt.Sprintf("%v", node.Status.Capacity.Memory())
```
michaelmdresser: `%v` is a bit of an odd formatting choice here for a non-log string. Is there a reason why you chose it over something like `%f` or `%s`?

pkg/cloud/provider.go
```diff
@@ -531,6 +542,11 @@ func getClusterProperties(node *v1.Node) clusterProperties {
 	} else if strings.HasPrefix(providerID, "scaleway") { // the scaleway provider ID looks like scaleway://instance/<instance_id>
 		cp.provider = kubecost.ScalewayProvider
 		cp.configFileName = "scaleway.json"
+	} else if strings.Contains(node.Status.NodeInfo.KubeletVersion, "aliyun") { // provider ID is not prefix with any distinct keyword like other providers
+		// TO-DO: Check for Nil conditions
```
michaelmdresser: TODO?

pkg/cloud/provider.go
```diff
@@ -531,6 +542,11 @@ func getClusterProperties(node *v1.Node) clusterProperties {
...
+		log.Infof("I assgined the cp provider and json")
```
michaelmdresser: Is this log necessary?

configs/aliyun.json
avrodrigues5: Alibaba is the company and Alibaba Cloud or Aliyun is the cloud offering. We could use Alibaba Cloud but looks like Aliyun is most used name is sdks, api etc. FOr example : 
https://www.alibabacloud.com/help/en/alibaba-cloud-cli/latest/macos - "Aliyun CLI"
https://github.com/aliyun/alibaba-cloud-sdk-go - "Aliyun/alibaba-cloud".

I was actually torn between aliyun , alibabaCloud or aliCloud. Choose aliyun but someone can veto me here.

pkg/cloud/aliyunprovider.go
avrodrigues5: Deferring to when i implement PV, but i can say intend was to have localDisk alone but now want it to use for PV .

pkg/cloud/aliyunprovider.go
avrodrigues5: So for readability i created New<createStructfunction> and maintained it throughout. I dont prefer one over other let me know if you want me to change it.

pkg/cloud/aliyunprovider.go
avrodrigues5: Mostly copied from AWSProvder with similar scope private or Public but mentioned in update PR why i think secret is good to be unexported.

pkg/cloud/aliyunprovider.go
avrodrigues5: Key Interface has feature [method](https://github.com/opencost/opencost/blob/develop/pkg/cloud/provider.go#L120), ideally its something used to get the information for node/pv matching this feature.

pkg/cloud/aliyunprovider.go
avrodrigues5: Just to support interface

pkg/cloud/aliyunprovider.go
avrodrigues5: ECS is the Elastic compute service like EC2, and this PRODUCT_CODE which is "ecs" is used to get the price of the backing instance type of the node. I tried searching for "ACS" product on alibaba cloud and its not present.

pkg/cloud/aliyunprovider.go
avrodrigues5: removed all of these, I was using this during initial code and forgot to remove.

pkg/cloud/aliyunprovider.go
avrodrigues5: https://go.dev/play/p/_XcViAvD0SP, surprising the nil go to default case but i have added the il check with explicit error message.

pkg/cloud/aliyunprovider.go
avrodrigues5: Sadly at this time i need to read a bit how to distinguish between these different offered node pricing either via v1.Node labels(which when i looked doesn't seem like it has any information) or have to use a different k8s function to get this information. I'll get back to you on this.

pkg/cloud/aliyunprovider.go
avrodrigues5: It was just to be extra cautious, if node data is not populated with information but this function can be removed!

configs/aliyun.json
dwbrown2: My two cents: everybody that knows Aliyun will know of Alibaba, but not everybody that knows Alibaba will know of Aliyun. I'd go with Alibaba based on that, but really think that either should be fine.

michaelmdresser: I like Webb's track here -- let's call it `"provider": "Alibaba"` (and the struct `Alibaba`, etc.)

pkg/cloud/aliyunprovider.go
michaelmdresser: Understood! I'm fine with leaving this around until you have a PR for PVs.

pkg/cloud/aliyunprovider.go
michaelmdresser: I'm fine with New! Let's leave as-is.

pkg/cloud/aliyunprovider.go
michaelmdresser: Gotcha!

pkg/cloud/aliyunprovider.go
michaelmdresser: Okay, should we call this function `createDescribePriceECSRequest`, then? (rather than `...ACSRequest`)

pkg/cloud/aliyunprovider.go
michaelmdresser: Whatever you feel is safer works for me, just want to make sure we had thought about it.

pkg/cloud/aliyunprovider.go
michaelmdresser: Gotcha. I'm fine with approving as-is, I know we've intentionally scoped to pay as you go. I assume we'll just document that for now and come back to this.

michaelmdresser (approved):
As long as you change all instances of "Aliyun" to "Alibaba", I think this PR is ready to go, with the understanding that we'll continue to work on improvements to the Alibaba cloud support as time goes on and customer needs increase (in future PRs).

pkg/cloud/aliyunprovider.go
avrodrigues5: Ah now i see what you referred by ACS and ECS, https://github.com/aliyun/alibaba-cloud-sdk-go/blob/v1.62.26/sdk/requests/acs_request.go#L61, this is the interface for Alibaba Cloud Services (ACS) Request and Describe Price is one of such request Action name or API name on Product code ECS. So maybe the name seems fine but im okay to change it.

avrodrigues5:
@michaelmdresser @nikovacevic , could you please merge this PR as I have lack of permission (not a OpenCost maintainer)

pkg/cloud/aliyunprovider.go
michaelmdresser: Aha! All becomes clear, thank you

PR #1489, opened 2022-11-17
getdisks for azure

nickcurie:

## What does this PR change?
* Adds orphaned resources struct and integrates it for Azure

* https://github.com/kubecost/kubecost-cost-model/pull/1081

* Users can get orphaned resources data for clusters on Azure

* 

## How was this PR tested?
* Unit tests and on an Azure cluster

* 

* v1.99


.gitignore
```diff
@@ -6,3 +6,4 @@ ui/.cache
 ui/dist
 ui/node_modules/
 cmd/costmodel/costmodel
+pkg/cloud/azureorphan_test.go
```
nikovacevic, line 9: Might we want to remove this? Or is there a good reason to keep it?

pkg/cloud/azureprovider.go
```diff
@@ -521,7 +522,7 @@ func (ask *AzureServiceKey) IsValid() bool {
 }
 
 // Loads the azure authentication via configuration or a secret set at install time.
-func (az *Azure) getAzureRateCardAuth(forceReload bool, cp *CustomPricing) (subscriptionID, clientID, clientSecret, tenantID string) {
+func (az *Azure) getAzureAuth(forceReload bool, cp *CustomPricing) (subscriptionID, clientID, clientSecret, tenantID string) {
```
nikovacevic: What's the reasoning behind renaming this?

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1170,118 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+			return nil, fmt.Errorf("error getting next page: %v", err)
+		}
+	}
+
+	return disks, nil
+}
+
+func isDiskOrphaned(disk *compute.Disk) bool {
+	//TODO: needs better algorithm
+	return disk.DiskState == "Unattached" || disk.DiskState == "Reserved"
```
nikovacevic, line 1250: This seems like a fine start to me. Why is "Reserved" considered orphaned?

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1170,118 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+		return nil, err
+	}
+
+	return json.Marshal(disks)
+}
+
+func (az *Azure) authAndGetDisks() ([]*compute.Disk, error) {
+	//TODO: do I need this, probably not
+	az.DownloadPricingDataLock.Lock()
+	defer az.DownloadPricingDataLock.Unlock()
```
nikovacevic: Did you determine whether or not you need this?

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1170,118 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+func (az *Azure) GetDisks() ([]byte, error) {
+	disks, err := az.authAndGetDisks()
+	if err != nil {
+		return nil, err
+	}
+
+	return json.Marshal(disks)
+}
+
+func (az *Azure) authAndGetDisks() ([]*compute.Disk, error) {
```
nikovacevic: I would recommend just calling this "getDisks" -- the "auth" part is implied, or is an implementation detail. Any concerns with that?

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1170,118 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+				return nil, err
+			}
+			authorizer = a
+		}
+	}
+	client := compute.NewDisksClient(config.AzureSubscriptionID)
+	client.Authorizer = authorizer
+	if err != nil {
+		return nil, fmt.Errorf("failed to create client: %v", err)
+	}
```
nikovacevic: This block looks disconnected from the error it's intended to check. Where is that `err` and why should we check it here?

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1170,118 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+		if isDiskOrphaned(d) {
+			or := OrphanedResource{
+				Kind:   "disk",
+				Region: *d.Location,
+				Description: map[string]string{
+					"disk state":   string(d.DiskState),
+					"time created": d.TimeCreated.String(),
+				},
+				Size:        d.DiskSizeGB,
+				MonthlyCost: 5.00, //!not implemented
```
nikovacevic: Let's implement it! What's holding us back here?

pkg/cloud/gcpprovider.go
```diff
@@ -382,6 +382,10 @@ func (gcp *GCP) GetDisks() ([]byte, error) {
 
 }
 
+func (*GCP) GetOrphanedResources() ([]OrphanedResource, error) {
+	return nil, nil
```
nikovacevic: These should return `nil, errors.New("not implemented")` so that callers get an indication that this is not yet implemented and don't take `err == nil` to mean "there are results".

pkg/cloud/provider.go
```diff
@@ -104,6 +104,15 @@ type Network struct {
 	InternetNetworkEgressCost float64
 }
 
+type OrphanedResource struct {
+	Kind        string            `json:"resourceKind"`
+	Region      string            `json:"region"`
+	Description map[string]string `json:"description"`
+	Size        *int32            `json:"diskSizeInGB, omitempty"`
+	Address     string            `json:"ipAddress, omitempty"`
```
nikovacevic: No space after the comma with these `omitempty`s. That's likely why the omission is not happening.

nikovacevic (changes requested):
Looking pretty good so far! Let's try to get this one across the finish line today. Let me know how I can help.

pkg/cloud/azureprovider.go
nickcurie: From the docs: 
image
This implies to me that if a VM is unused then the disk that VM uses is also unused, but the user is still paying for it. Maybe this would be a good place to use the time fields provided by the API

nikovacevic: Seems reasonable!

.gitignore
nickcurie: I had this here so that I wasn't accidentally pushing code that has our azure secrets in it. It can be removed once I'm done working on the feature as well as that test file

pkg/cloud/azureprovider.go
nickcurie: As far as I could tell this function wasn't explicitly related to rate cards so I renamed it so that its a little clearer, especially because its used in multiple places now.

nikovacevic: That may be true, but it's also good to ask, "Is there a reason it has this name?" You can ask the person who named it, by checking the git blame.

In this case, that yields the interesting result that the function used to be named `getAzureAuth` and then Sean renamed it to `getAzureRateCardAuth`: https://github.com/opencost/opencost/commit/a1bd4042b072b232f5731b01ec42734d80e5274c

So, you could ask him what motivated that, if you're interested in changing it. But I would be somewhat cautious about just changing function names without looking into the original author's intent :wink:

nikovacevic: To be clear, he may not mind! But good to ask, especially given that this would be reverting to an old name.

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1172,142 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+
+func (az *Azure) GetOrphanedResources() ([]OrphanedResource, error) {
+	disks, err := az.getDisks()
+	if err != nil {
+		return nil, err
+	}
+
+	//! running this is very slow, there is probably a better way to get costs
+	err = az.DownloadPricingData()
+	if err != nil {
```
nickcurie: Wanted to point this out. This is making this function call very slow, so there's likely a better way to figure out the costs for disks

nikovacevic: Why are we downloading pricing data here at all? What happens if you simply remove that function?

.gitignore
nikovacevic: Let's remember to return here. It'd be cool to commit the tests, too, but with some disabling feature. Like... set the auth to empty strings, and only run the tests if the auth information is given? So that your work on that doesn't go to waste? Eventually, we should have nice ways to inject sensitive auth info into tests.

pkg/cloud/azureprovider.go
```diff
@@ -40,6 +41,8 @@ const (
 	defaultSpotLabel                 = "kubernetes.azure.com/scalesetpriority"
 	defaultSpotLabelValue            = "spot"
 	AzureStorageUpdateType           = "AzureStorage"
+
+	hoursPerMonth = 730
```
nikovacevic: We have a `timeutil` package for this!
```go
timeutil.HoursPerMonth
```

pkg/cloud/azureprovider.go
nikovacevic: Ping!

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1172,136 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+	storageClass := string(d.Sku.Name)
+	if strings.EqualFold(storageClass, "Premium_LRS") {
+		storageClass = AzureDiskPremiumSSDStorageClass
+	} else if strings.EqualFold(storageClass, "StandardSSD_LRS") {
+		storageClass = AzureDiskStandardSSDStorageClass
+	} else if strings.EqualFold(storageClass, "Standard_LRS") {
+		storageClass = AzureDiskStandardStorageClass
+	}
+
+	key := *d.Location + "," + storageClass
```
nikovacevic, line 1298: These derefs look a little unsafe. Let's check if `d` is `nil` at the top and early return if it's `nil`

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1172,136 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+				MonthlyCost: cost,
+			}
+			orphanedResources = append(orphanedResources, or)
+		}
+	}
+
+	return orphanedResources, nil
+}
+
+func (az *Azure) findCostForDisk(d *compute.Disk) (*float64, error) {
```
nikovacevic: Let's use the zero value for `float64`. As a general rule, pointers to basic types is a pattern to avoid, unless you really need a `nil` as opposed to a `0.0`. Are you alright with returning `(float64, error)` and `0.0, err` if `err != nil`?

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1172,136 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+		storageClass = AzureDiskPremiumSSDStorageClass
+	} else if strings.EqualFold(storageClass, "StandardSSD_LRS") {
+		storageClass = AzureDiskStandardSSDStorageClass
+	} else if strings.EqualFold(storageClass, "Standard_LRS") {
+		storageClass = AzureDiskStandardStorageClass
+	}
+
+	key := *d.Location + "," + storageClass
+
+	diskCost, err := strconv.ParseFloat(az.Pricing[key].PV.Cost, 64)
```
nikovacevic: So, the claim is that this `diskCost` is "cost for this particular PV per GB per hour? Can we name this variable accordingly, something like `diskPricePerGBHour`?

nikovacevic (changes requested):
Let me know if you have any questions about this final pass of review! Should be quick to address. Very close.

Also, checked on PV pricing in `DownloadPricingData` and your monthly cost looks right to me. The disks we have orphaned look suspiciously cheap, but I don't see any errors...

pkg/cloud/azureprovider.go
```diff
@@ -1169,8 +1171,139 @@ func (*Azure) GetAddresses() ([]byte, error) {
...
+			if err != nil {
+				return nil, err
+			}
+
+			or := OrphanedResource{
+				Kind:   "disk",
+				Region: *d.Location,
+				Description: map[string]string{
+					"disk state":   string(d.DiskState),
+					"time created": d.TimeCreated.String(),
```
nikovacevic: Can we camel-case these, too? Spaces in the JSON keys looks a bit odd.

nikovacevic: diskState, timeCreated

nickcurie: Yep, agreed

nikovacevic (approved):

PR #1507, opened 2022-11-23
PV pricing for cloud disk attached to the Alibaba K8s Cluster

avrodrigues5:
## What does this PR change?
* Gives insights into PV pricing for Physical volumes that are of type cloud disk.

* Continuation of work from [1488](https://github.com/opencost/opencost/pull/1488)

* Gives them visibility into PV pricing.

* Closes ...

## How was this PR tested?
* Unit test
* creating PV and PVC with various combination of different available cloud disk options and seeing if results match with the actual Alibaba price.

Images listing the allNodePricing information = hourly price for the corresponding Alibaba's disk prices and the prometheus having the values in pv_hourly_cost
Screen Shot 2022-11-23 at 9 10 43 AM
Screen Shot 2022-11-23 at 9 10 48 AM
Screen Shot 2022-11-23 at 9 07 50 AM

* 

* v1.99

### DIFF
--- a/pkg/cloud/aliyunprovider.go
+++ b/pkg/cloud/aliyunprovider.go
@@ -153,10 +153,14 @@ func NewSlimK8sNode(instanceType, regionID, priceUnit, memorySizeInKiB, osType, 
 // Basic Attributes needed atleast to get the key, Some attributes from k8s Node response
 // be populated directly into *Node object.
 type AlibabaNodeAttributes struct {
-	InstanceType    string `json:"instanceType"`
+	// InstanceType represents the type of instance.
+	InstanceType string `json:"instanceType"`
+	// MemorySizeInKiB represents the size of memory of instance.
 	MemorySizeInKiB string `json:"memorySizeInKiB"`
-	IsIoOptimized   bool   `json:"isIoOptimized"`
-	OSType          string `json:"osType"`
+	// IsIoOptimized represents the if instance is I/O optimized.
+	IsIoOptimized bool `json:"isIoOptimized"`
+	// OSType represents the OS installed in the Instance.
+	OSType string `json:"osType"`
 }
 
 func NewAlibabaNodeAttributes(node *SlimK8sNode) *AlibabaNodeAttributes {
@@ -410,14 +414,14 @@ func (alibaba *Alibaba) DownloadPricingData() error {
 	return nil
 }
 
-// AllNodePricing returns all the billing data for nodes and pvs
+// AllNodePricing returns all the pricing data for all nodes and pvs
 func (alibaba *Alibaba) AllNodePricing() (interface{}, error) {
 	alibaba.DownloadPricingDataLock.RLock()
 	defer alibaba.DownloadPricingDataLock.RUnlock()
 	return alibaba.Pricing, nil
 }
 
-// NodePricing gives a specific node pricing information given by the key
+// NodePricing gives pricing information of a specific node given by the key
 func (alibaba *Alibaba) NodePricing(key Key) (*Node, error) {
 	alibaba.DownloadPricingDataLock.RLock()
 	defer alibaba.DownloadPricingDataLock.RUnlock()
@@ -437,7 +441,7 @@ func (alibaba *Alibaba) NodePricing(key Key) (*Node, error) {
 	return returnNode, nil
 }
 
-// PVPricing gives a specific PV price for the PVkey
+// PVPricing gives a pricing information of a specific PV gien by PVkey
 func (alibaba *Alibaba) PVPricing(pvk PVKey) (*PV, error) {
 	alibaba.DownloadPricingDataLock.RLock()
 	defer alibaba.DownloadPricingDataLock.RUnlock()
@@ -957,19 +961,10 @@ func generateSlimK8sDiskFromV1PV(pv *v1.PersistentVolume, regionID string) *Slim
 		}
 	}
 
-	// Highly unlikely that label "csi.alibabacloud.com/disktype" doesn't exist but if occured default to cloud (most basic disk type)
+	// Highly unlikely that label pv.Spec.CSI.VolumeAttributes["type"] doesn't exist but if occured default to cloud (most basic disk type)
 	if diskCategory == "" {
 		diskCategory = ALIBABA_DISK_CLOUD_CATEGORY
 	}
 
-	return &SlimK8sDisk{
-		DiskType:         diskType,
-		RegionID:         regionID,
-		PriceUnit:        priceUnit,
-		SizeInGiB:        sizeInGiB,
-		DiskCategory:     diskCategory,
-		PerformanceLevel: performanceLevel,
-		ProviderID:       providerID,
-		StorageClass:     pv.Spec.StorageClassName,
-	}
+	return NewSlimK8sDisk(diskType, regionID, priceUnit, diskCategory, performanceLevel, providerID, pv.Spec.StorageClassName, sizeInGiB)
 }
--- a/pkg/cloud/aliyunprovider.go
+++ b/pkg/cloud/aliyunprovider.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/cloud/aliyunprovider.go
```diff
@@ -322,59 +370,58 @@ func (alibaba *Alibaba) DownloadPricingData() error {
...
+	// set the first occurance of region from the node
+	i := 0
+	for alibaba.clusterRegion == "" && i < len(nodeList) {
+		regionID, ok := nodeList[i].Labels["topology.kubernetes.io/region"]
+		if ok {
+			alibaba.clusterRegion = regionID
+		} else {
+			i += 1
+		}
+	}
```
michaelmdresser: I find this loop a little difficult to read. I'll give you another option, let me know how you feel about it. Hopefully I'm understanding the logic here right!

```go
if alibaba.clusterRegion == "" {
    for _, node := range nodeList {
        if regionID, ok := node.Labels["topology.kubernetes.io/region"]; ok {
            alibaba.clusterRegion = regionID
            break
        }
    }
}
```

pkg/cloud/aliyunprovider.go
```diff
@@ -322,59 +370,58 @@ func (alibaba *Alibaba) DownloadPricingData() error {
...
+
+	// PV pricing for only Cloud Disk for now.
+	// TO-DO: Support both NAS(Network Attached storage) and OSS(Object Storage Service) type PVs
+
+	pvList := alibaba.Clientset.GetAllPersistentVolumes()
+
+	for _, pv := range pvList {
+		pricingObj := &AlibabaPricing{}
+		// Note for PR: Region ID defaulted to first occurance of node region for now, need to know the implication here!
+		// Can Pvs be in seperate Region than the Cluster? Is there ever a Multi Region k8s cluster?
```
michaelmdresser: Thanks for calling this out in a comment! I think most cloud providers support multi-regional/multi-zonal clusters, i.e. nodes in different regions. If Alibaba supports that, then I suspect that there could be PVs in different regions but all in the same cluster.

Have you checked if the PVs/PVCs in the cluster have a `topology` annotation of any kind?

pkg/cloud/aliyunprovider.go
```diff
@@ -842,3 +919,52 @@ func generateSlimK8sNodeFromV1Node(node *v1.Node) *SlimK8sNode {
 
 	return NewSlimK8sNode(instanceType, regionID, priceUnit, memorySizeInKiB, osType, providerID, instanceFamily, IsIoOptimized)
 }
+
+// generateSlimK8sDiskFromV1PV function generates SlimK8sDisk from v1.PersistentVolume and DescribeDisk API(If required) of alibaba
+// to generate slim disk type that can be used to fetch pricing information.
+func generateSlimK8sDiskFromV1PV(pv *v1.PersistentVolume, regionID string) *SlimK8sDisk {
+	//Regular expression to get the GiB storage size for the API call
+	sizeRegEx := regexp.MustCompile("(.*?)Gi")
```
michaelmdresser: MustCompile is a little spicy -- would you please write a really simple unit test that calls this function so that we can know for certain that the regex compiles at compile/test time?

I'd also recommend pulling the compilation of the regex into a top-level var (outside of a function scope) so that we aren't recompiling this regex every time we call this function.

pkg/cloud/aliyunprovider.go
```diff
@@ -842,3 +919,52 @@ func generateSlimK8sNodeFromV1Node(node *v1.Node) *SlimK8sNode {
...
+
+	sizeQuantity := fmt.Sprintf("%s", pv.Spec.Capacity.Storage())
+
+	res := sizeRegEx.FindAllStringSubmatch(sizeQuantity, 1)
+
+	// This is the default value used for the DescribePrice DataDisk size, if any error occured defaulting it to 2000GiB's price
+	sizeInGiB := "2000"
+	if len(res) != 0 {
+		sizeInGiB = res[0][1]
+	}
```
michaelmdresser: This could panic, I think: https://go.dev/play/p/lHfBVsr2nSM

michaelmdresser (commented):
Fast work! I really appreciate the effort put into unit tests.

My main outstanding question is how we might get region information from a PV/PVC. Let's check if they are labeled/annotated in K8s similar to how nodes are. If not, we might be able to guess region based on the node they are attached to, if we have that information. If neither of those are feasible, I think we can move forward and come back to it.

pkg/cloud/aliyunprovider.go
nikovacevic: Definitely need to pull this out into a var outside the func

pkg/cloud/aliyunprovider.go
avrodrigues5: thanks Michael , it's cleaner agreed and I adopted it

pkg/cloud/aliyunprovider.go
avrodrigues5: Adopted it, Thanks!

pkg/cloud/aliyunprovider.go
avrodrigues5: Very unlikely scenario to occur since Ali cloud clearly states on UI atleast to create 20 Gi PVs and doesnt allow, but I went ahead and incorporated the ask and unit test cases around it!
Screen Shot 2022-11-23 at 3 14 11 PM

pkg/cloud/aliyunprovider.go
avrodrigues5: This is definitely most tricking PR comment and closest thing you are suggesting via PV describe is in node affinity part of it but it gives affinity zone, I would have to programmatically align it with a region. What do you think about that?
Like in the below scenario us-east-1a is the zone i'll have to make sure it's region is us-east-1.
Screen Shot 2022-11-23 at 3 16 27 PM

avrodrigues5: as an FYI 

```apiVersion: v1
kind: PersistentVolume
metadata:
  name: yaml-pv
  labels:
    alicloud-pvname: static-disk-yaml-pv
spec:
  capacity:
    storage: 25Gi
  accessModes:
    - ReadWriteOnce
  persistentVolumeReclaimPolicy: Retain
  csi:
    driver: diskplugin.csi.alibabacloud.com
    volumeHandle: "d-0xi2gxpy6za03i2cffff"
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: topology.diskplugin.csi.alibabacloud.com/zone
          operator: In
          values:
          - "us-east-1a"
```

I was only able to create pv with one node affinity. It doesnt allow me to affiliate a PV with multiple zone such as us-east-1a, us-east1b let alone multiple region. Can safely say first occurance of value and determining region based on that should be enough .

pkg/cloud/aliyunprovider.go
```diff
@@ -322,59 +378,59 @@ func (alibaba *Alibaba) DownloadPricingData() error {
...
+		for _, node := range nodeList {
+			if regionID, ok := node.Labels["topology.kubernetes.io/region"]; ok {
+				alibaba.clusterRegion = regionID
+				break
+			}
+		}
+	}
+
+	// PV pricing for only Cloud Disk for now.
+	// TO-DO: Support both NAS(Network Attached storage) and OSS(Object Storage Service) type PVs
```
nikovacevic, line 392: This is still outstanding, right? If so, we can open an issue for it after this gets merged.

pkg/cloud/aliyunprovider.go
```diff
@@ -322,59 +378,59 @@ func (alibaba *Alibaba) DownloadPricingData() error {
...
+	for _, pv := range pvList {
+		pvRegion := determinePVRegion(pv)
+		if pvRegion == "" {
+			pvRegion = alibaba.clusterRegion
+		}
+		pricingObj := &AlibabaPricing{}
+		slimK8sDisk := generateSlimK8sDiskFromV1PV(pv, pvRegion)
+		lookupKey, err = determineKeyForPricing(slimK8sDisk)
+		if _, ok := alibaba.Pricing[lookupKey]; ok {
+			log.Debugf("Pricing information for pv with same features %s already exists hence skipping", lookupKey)
```
nikovacevic, line 405: `%w`, nice! I need to start doing this. I keep using `%s`.

pkg/cloud/aliyunprovider.go
```diff
@@ -389,10 +445,12 @@ func (alibaba *Alibaba) NodePricing(key Key) (*Node, error) {
...
 
 	log.Debugf("returning the node price for the node with feature: %s", keyFeature)
-	return pricing.Node, nil
+	returnNode := pricing.Node
+
+	return returnNode, nil
 }
 
-// PVPricing gives a specific PV price for the PVkey
+// PVPricing gives a pricing information of a specific PV gien by PVkey
```
nikovacevic: `gien` -> `given`

nikovacevic (approved):
LGTM! Thanks, Alan!

pkg/cloud/aliyunprovider.go
avrodrigues5: https://github.com/opencost/opencost/issues/1510

PR #1508, opened 2022-11-23
Cloud Cost: core types; generic filtering; coverage; helper functions

nikovacevic:
Signed-off-by: Niko Kovacevic <nikovacevic@gmail.com>

## Release notes headline (Required or include Not Relevant) 
* Alpha version of the Cloud Costs pipeline and API, which replaces the Cloud Usage pipeline and API.

## What does this PR change?
* Alpha version of the Cloud Costs pipeline and API, which replaces the Cloud Usage pipeline and API.

* https://github.com/kubecost/kubecost-cost-model/pull/1083

* There will be a new pipeline that (hopefully) reduces memory, and people can turn off the old pipeline.

* Closes https://github.com/kubecost/kubecost-cost-model/issues/1056

## How was this PR tested?
* Manually, unit testing, benchmarks

* Yes, we'll follow up with that later

### DIFF
--- a/pkg/filter/and.go
+++ b/pkg/filter/and.go
@@ -1,0 +1,36 @@
+package filter
+
+import (
+	"fmt"
+)
+
+// And is a set of filters that should be evaluated as a logical
+// AND.
+type And[T any] struct {
+	Filters []Filter[T]
+}
+
+func (a And[T]) String() string {
+	s := "(and"
+	for _, f := range a.Filters {
+		s += fmt.Sprintf(" %s", f)
+	}
+
+	s += ")"
+	return s
+}
+
+func (a And[T]) Matches(that T) bool {
+	filters := a.Filters
+	if len(filters) == 0 {
+		return true
+	}
+
+	for _, filter := range filters {
+		if !filter.Matches(that) {
+			return false
+		}
+	}
+
+	return true
+}
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -1,0 +1,20 @@
+package filter
+
+// Filter represents anything that can be used to filter given generic type T.
+//
+// Implement this interface with caution. While it is generic, it
+// is intended to be introspectable so query handlers can perform various
+// optimizations. These optimizations include:
+// - Routing a query to the most optimal cache
+// - Querying backing data stores efficiently (e.g. translation to SQL)
+//
+// Custom implementations of this interface outside of this package should not
+// expect to receive these benefits. Passing a custom implementation to a
+// handler may in errors.
+type Filter[T any] interface {
+	String() string
+
+	// Matches is the canonical in-Go function for determining if T
+	// matches a filter.
+	Matches(T) bool
+}
--- a/pkg/filter/not.go
+++ b/pkg/filter/not.go
@@ -1,0 +1,17 @@
+package filter
+
+import "fmt"
+
+// Not negates any filter contained within it
+type Not[T any] struct {
+	Filter Filter[T]
+}
+
+func (n Not[T]) String() string {
+	return fmt.Sprintf("(not %s)", n.Filter.String())
+}
+
+// Matches inverts the result of the child filter
+func (n Not[T]) Matches(that T) bool {
+	return !n.Filter.Matches(that)
+}
--- a/pkg/filter/or.go
+++ b/pkg/filter/or.go
@@ -1,0 +1,36 @@
+package filter
+
+import (
+	"fmt"
+)
+
+// Or is a set of filters that should be evaluated as a logical
+// OR.
+type Or[T any] struct {
+	Filters []Filter[T]
+}
+
+func (o Or[T]) String() string {
+	s := "(or"
+	for _, f := range o.Filters {
+		s += fmt.Sprintf(" %s", f)
+	}
+
+	s += ")"
+	return s
+}
+
+func (o Or[T]) Matches(that T) bool {
+	filters := o.Filters
+	if len(filters) == 0 {
+		return true
+	}
+
+	for _, filter := range filters {
+		if filter.Matches(that) {
+			return true
+		}
+	}
+
+	return false
+}
--- a/pkg/filter/stringmapproperty.go
+++ b/pkg/filter/stringmapproperty.go
@@ -1,0 +1,83 @@
+package filter
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/opencost/opencost/pkg/log"
+)
+
+const unallocatedSuffix = "__unallocated__"
+
+type StringMapPropertied interface {
+	StringMapProperty(string) (map[string]string, error)
+}
+
+// StringMapOperation is an enum that represents operations that can be performed
+// when filtering (equality, inequality, etc.)
+type StringMapOperation string
+
+const (
+	// StringMapHasKey passes if the map has the provided key
+	StringMapHasKey StringMapOperation = "stringmapcontains"
+
+	StringMapStartsWith = "stringmapstartswith"
+
+	// StringMapEquals when the given key and value match
+	StringMapEquals = "stringmapequals"
+)
+
+// StringMapProperty is the lowest-level type of filter. It represents
+// a filter operation (equality, inequality, etc.) on a property that contains a string map
+type StringMapProperty[T StringMapPropertied] struct {
+	Field string
+	Op    StringMapOperation
+	Key   string
+	Value string
+}
+
+func (smp StringMapProperty[T]) String() string {
+	return fmt.Sprintf(`(%s %s[%s] "%s")`, smp.Op, smp.Field, smp.Key, smp.Value)
+}
+
+func (smp StringMapProperty[T]) Matches(that T) bool {
+
+	thatMap, err := that.StringMapProperty(smp.Field)
+	if err != nil {
+		log.Errorf("Filter: StringMapProperty: could not retrieve field %s: %s", smp.Field, err.Error())
+		return false
+	}
+
+	valueToCompare, keyIsPresent := thatMap[smp.Key]
+
+	switch smp.Op {
+	case StringMapHasKey:
+		return keyIsPresent
+	case StringMapEquals:
+		// namespace:"__unallocated__" should match a.Properties.Namespace = ""
+		// label[app]:"__unallocated__" should match _, ok := Labels[app]; !ok
+		if !keyIsPresent || valueToCompare == "" {
+			return smp.Value == unallocatedSuffix
+		}
+
+		if valueToCompare == smp.Value {
+			return true
+		}
+
+	case StringMapStartsWith:
+		if !keyIsPresent {
+			return false
+		}
+
+		// We don't need special __unallocated__ logic here because a query
+		// asking for "__unallocated__" won't have a wildcard and unallocated
+		// properties are the empty string.
+
+		return strings.HasPrefix(valueToCompare, smp.Value)
+	default:
+		log.Errorf("Filter: StringMapProperty: Unhandled filter op. This is a filter implementation error and requires immediate patching. Op: %s", smp.Op)
+		return false
+	}
+
+	return false
+}
--- a/pkg/filter/stringproperty.go
+++ b/pkg/filter/stringproperty.go
@@ -1,0 +1,83 @@
+package filter
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/opencost/opencost/pkg/log"
+)
+
+// StringPropertied is used to validate the name of a property field and return its value
+type StringPropertied interface {
+	// StringProperty acts as a validator and getter for a structs string properties
+	StringProperty(string) (string, error)
+}
+
+// StringOperation is an enum that represents operations that can be performed
+// when filtering (equality, inequality, etc.)
+type StringOperation string
+
+// If you add a FilterOp, MAKE SURE TO UPDATE ALL FILTER IMPLEMENTATIONS! Go
+// does not enforce exhaustive pattern matching on "enum" types.
+const (
+	// StringEquals is the equality operator
+	// "kube-system" FilterEquals "kube-system" = true
+	// "kube-syste" FilterEquals "kube-system" = false
+	StringEquals StringOperation = "stringequals"
+
+	// StringStartsWith matches strings with the given prefix.
+	// "kube-system" StartsWith "kube" = true
+	//
+	// When comparing with a field represented by an array/slice, this is like
+	// applying FilterContains to every element of the slice.
+	StringStartsWith = "stringstartswith"
+)
+
+// StringProperty is the lowest-level type of filter. It represents
+// a filter operation (equality, inequality, etc.) on a field with a string value (namespace,
+// node, pod, etc.).
+type StringProperty[T StringPropertied] struct {
+	Field string
+	Op    StringOperation
+
+	// Value is for _all_ filters. A filter of 'namespace:"kubecost"' has
+	// Value="kubecost"
+	Value string
+}
+
+func (sp StringProperty[T]) String() string {
+	return fmt.Sprintf(`(%s %s "%s")`, sp.Op, sp.Field, sp.Value)
+}
+
+func (sp StringProperty[T]) Matches(that T) bool {
+
+	thatString, err := that.StringProperty(sp.Field)
+	if err != nil {
+		log.Errorf("Filter: StringProperty: could not retrieve field %s: %s", sp.Field, err.Error())
+		return false
+	}
+
+	switch sp.Op {
+	case StringEquals:
+		// namespace:"__unallocated__" should match a.Properties.Namespace = ""
+		if thatString == "" {
+			return sp.Value == unallocatedSuffix
+		}
+
+		if thatString == sp.Value {
+			return true
+		}
+	case StringStartsWith:
+
+		// We don't need special __unallocated__ logic here because a query
+		// asking for "__unallocated__" won't have a wildcard and unallocated
+		// properties are the empty string.
+
+		return strings.HasPrefix(thatString, sp.Value)
+	default:
+		log.Errorf("Filter: StringProperty: Unhandled filter op. This is a filter implementation error and requires immediate patching. Op: %s", sp.Op)
+		return false
+	}
+
+	return false
+}
--- a/pkg/filter/util/cloudcostaggregate.go
+++ b/pkg/filter/util/cloudcostaggregate.go
@@ -1,0 +1,70 @@
+package util
+
+import (
+	"strings"
+
+	"github.com/opencost/opencost/pkg/filter"
+	"github.com/opencost/opencost/pkg/kubecost"
+	"github.com/opencost/opencost/pkg/util/mapper"
+)
+
+func parseWildcardEnd(rawFilterValue string) (string, bool) {
+	return strings.TrimSuffix(rawFilterValue, "*"), strings.HasSuffix(rawFilterValue, "*")
+}
+
+func CloudCostAggregateFilterFromParams(pmr mapper.PrimitiveMapReader) filter.Filter[*kubecost.CloudCostAggregate] {
+	filter := filter.And[*kubecost.CloudCostAggregate]{
+		Filters: []filter.Filter[*kubecost.CloudCostAggregate]{},
+	}
+
+	if raw := pmr.GetList("filterAccounts", ","); len(raw) > 0 {
+		filter.Filters = append(filter.Filters, filterV1SingleValueFromList(raw, kubecost.CloudCostAccountProp))
+	}
+
+	if raw := pmr.GetList("filterProjects", ","); len(raw) > 0 {
+		filter.Filters = append(filter.Filters, filterV1SingleValueFromList(raw, kubecost.CloudCostProjectProp))
+	}
+
+	if raw := pmr.GetList("filterProviders", ","); len(raw) > 0 {
+		filter.Filters = append(filter.Filters, filterV1SingleValueFromList(raw, kubecost.CloudCostProviderProp))
+	}
+
+	if raw := pmr.GetList("filterServices", ","); len(raw) > 0 {
+		filter.Filters = append(filter.Filters, filterV1SingleValueFromList(raw, kubecost.CloudCostServiceProp))
+	}
+
+	if raw := pmr.GetList("filterLabelValues", ","); len(raw) > 0 {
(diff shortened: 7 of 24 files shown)


pkg/filter/and.go
```diff
@@ -0,0 +1,111 @@
+package filter
+
+import (
+	"fmt"
+	"sort"
+)
+
+// AllocationFilterOr is a set of filters that should be evaluated as a logical
+// AND.
+type And[T any] struct {
```
michaelmdresser: This comment is a bit confused

pkg/filter/and.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	if len(flattenedFilters) == 1 {
+		return flattenedFilters[0]
+	}
+
+	// If more than half of the children are Not filters And should have DeMorgans rule applied to it
+	if percentNot(flattenedFilters) > 0.5 {
+		return Not[T]{Or[T]{negateFilters(flattenedFilters)}}
+	}
```
michaelmdresser: Interesting! Why?

pkg/filter/filter.go
```diff
@@ -0,0 +1,120 @@
+package filter
+
+// Filter represents anything that can be used to filter an
+// Allocation.
```
michaelmdresser: Maybe not be specific to Allocation?

pkg/filter/filter.go
```diff
@@ -0,0 +1,120 @@
...
+	// When exactly half of the filters are Not, And is preferred.
+	//
+	// It returns AllPass if the filter is filtering nothing.
+	//
+	//
+	// Example:
+	// (and (or (namespaceequals "kubecost")) (or)) ->
+	// (namespaceequals "kubecost")
+	//
+	// (and (or)) -> nil
```
michaelmdresser: This is `(and (or)) -> (AllPass)` now, right?

pkg/filter/filter.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	var dedupedFilters []Filter[T]
+	for _, filter := range filterSet {
+		dedupedFilters = append(dedupedFilters, filter)
+	}
+	return dedupedFilters
+}
+
+// checkForNegations assumes that []Filter[T] has already been flattend
+func checkForNegations[T any](flattenedFilters []Filter[T]) bool {
```
michaelmdresser: I'm having trouble understanding what cases this handles, would you show a couple of examples in the comment, just for my sake?

pkg/filter/not.go
```diff
@@ -0,0 +1,49 @@
...
+	return fmt.Sprintf("(not %s)", n.Filter.String())
+}
+
+// Matches inverts the result of the child filter
+func (n Not[T]) Matches(that T) bool {
+	return !n.Filter.Matches(that)
+}
+
+// Flattened flattens callers child filter. Cancel out child not and inverts AllPass and AllCut filters
+func (n Not[T]) Flattened() Filter[T] {
```
michaelmdresser: This function is awesome!

pkg/filter/stringmapproperty.go
```diff
@@ -0,0 +1,96 @@
...
+	StringMapHasKey StringMapOperation = "stringmapcontains"
+
+	StringMapStartsWith = "stringmapstartswith"
+
+	// StringMapEquals when the given key and value match
+	StringMapEquals = "stringmapequals"
+)
+
+// StringMapProperty is the lowest-level type of filter. It represents
+// a filter operation (equality, inequality, etc.) on a property that contains a string slice
```
michaelmdresser: "a string slice"?

pkg/filter/stringproperty.go
```diff
@@ -0,0 +1,95 @@
...
+	//
+	// When comparing with a field represented by an array/slice, this is like
+	// applying FilterContains to every element of the slice.
+	StringStartsWith = "stringstartswith"
+)
+
+// AllocationCondition is the lowest-level type of filter. It represents
+// the a filter operation (equality, inequality, etc.) on a field (namespace,
+// label, etc.).
+type StringProperty[T StringPropertied] struct {
```
michaelmdresser: Comment needs an update

pkg/filter/stringproperty.go
```diff
@@ -0,0 +1,95 @@
...
+
+// StringPropertied is used to validate the name of a property field and return its value
+type StringPropertied interface {
+	// StringProperty acts as a validator and getter for a structs string properties
+	StringProperty(string) (string, error)
+}
+
+// StringOperation is an enum that represents operations that can be performed
+// when filtering (equality, inequality, etc.)
+type StringOperation string
```
michaelmdresser, line 18: This new approach to filters is neat, like a slightly better type system in the filter representation!

pkg/filter/stringmapproperty.go
```diff
@@ -0,0 +1,96 @@
...
+	return false
+}
+
+func (smp StringMapProperty[T]) Matches(that T) bool {
+
+	thatMap, err := that.StringMapProperty(smp.Field)
+	if err != nil {
+		log.Errorf("Filter: StringMapProperty: could not retrieve field %s: %s", smp.Field, err.Error())
+		return false
+	}
```
michaelmdresser, line 49: Will this return an empty map e.g. if no labels are present? Just want to make sure that the `__unallocated__` logic works correctly

pkg/filter/util/cloudcostaggregate.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if raw := pmr.GetList("filterLabelValues", ","); len(raw) > 0 {
+		filter.Filters = append(filter.Filters, filterV1SingleValueFromList(raw, kubecost.CloudCostLabelProp))
+	}
+
+	if len(filter.Filters) == 0 {
+		return nil
+	}
+
+	return filter
```
michaelmdresser, line 44: Could you just return `filter.Flattened()`?

pkg/filter/window.go
```diff
@@ -0,0 +1,60 @@
+package filter
+
+// import (
```
michaelmdresser: Intentionally committing a commented-out file?

pkg/kubecost/cloudcostaggregate.go
```diff
@@ -0,0 +1,440 @@
...
+		return ccap.Project
+	case CloudCostServiceProp:
+		return ccap.Service
+	case CloudCostLabelProp:
+		// TODO:cloudcost will we need the LabelName here?... can move it down from the Set
+		return ccap.LabelValue
+	}
+
+	// TODO:cloudcost error?
+	return ""
```
michaelmdresser: Have these TODOs been resolved?

pkg/kubecost/cloudcostaggregate.go
```diff
@@ -0,0 +1,440 @@
...
+}
+
+// CloudCostAggregate represents an aggregation of Billing Integration data on the properties listed
+type CloudCostAggregate struct {
+	Properties        CloudCostAggregateProperties `json:"properties"`
+	KubernetesPercent float64                      `json:"kubernetesPercent"`
```
michaelmdresser, line 84: Would you add a docstring on `KubernetesPercent` to explain it?

pkg/kubecost/cloudcostaggregate.go
```diff
@@ -0,0 +1,440 @@
...
+	Cost              float64                      `json:"cost"`
+	Credit            float64                      `json:"credit"`
```
michaelmdresser, line 86: Same with `Credit` -- cost and properties are pretty intuitive

pkg/kubecost/cloudcostaggregate.go
```diff
@@ -0,0 +1,440 @@
...
+
+	cca.Cost = sumCost
+	cca.Credit += that.Credit
+	cca.KubernetesPercent = k8sPct
+}
+
+type CloudCostAggregateSet struct {
+	CloudCostAggregates map[string]*CloudCostAggregate `json:"items"`
+	AggregationProperty string                         `json:"-"`
+	Integration         string                         `json:"-"`
```
michaelmdresser: Any particular reason why these aren't returned in the JSON?

pkg/kubecost/cloudcostaggregate.go
```diff
@@ -0,0 +1,440 @@
...
+
+	return result, nil
+}
+
+func (ccas *CloudCostAggregateSet) Filter(filters filter.Filter[*CloudCostAggregate]) *CloudCostAggregateSet {
+	if ccas == nil {
+		return nil
+	}
+
+	result := ccas.Clone()
```
michaelmdresser, line 207: I'm guessing we might have to remove this Clone() later down the line, but I think its fine for a first pass!

pkg/kubecost/cloudcostaggregate.go
```diff
@@ -0,0 +1,440 @@
...
+		CloudCostAggregateSets = append(CloudCostAggregateSets, ccas)
+	}
+	return CloudCostAggregateSets, nil
+}
+
+// LoadCloudCostAggregateSets creates and loads CloudCostAggregates into provided CloudCostAggregateSets. This method makes it so
+// that the input windows do not have to match the one day frame of the Athena queries. CloudCostAggregates being generated from a
+// CUR which may be the identical except for the pricing model used (default, RI or savings plan)
+// are accumulated here so that the resulting CloudCostAggregate with the 1d window has the correct price for the entire day.
+func LoadCloudCostAggregateSets(itemStart time.Time, itemEnd time.Time, properties CloudCostAggregateProperties, K8sPercent, cost, credit float64, CloudCostAggregateSets []*CloudCostAggregateSet) {
```
michaelmdresser, line 375: Fancy! Should we have a unit test or three for this logic?

pkg/kubecost/cloudcostitem.go
```diff
@@ -0,0 +1,362 @@
...
+	if filters == nil {
+		return ccis.Clone()
+	}
+
+	result := NewCloudCostItemSet(*ccis.Window.start, *ccis.Window.end)
+
+	for _, cci := range ccis.CloudCostItems {
+		if filters.Matches(cci) {
+			// TODO:cloudcost ideally... this would be a Clone, but performance?
+			result.Insert(cci)
```
michaelmdresser: Making a new result but inserting an uncloned CCI feels a little off. Either new result + clone the CCI, or just filter the original `ccis` in place. What do you think?

pkg/kubecost/cloudcostitem.go
```diff
@@ -0,0 +1,362 @@
...
+		ccis.CloudCostItems = map[string]*CloudCostItem{}
+	}
+
+	// Add the given CloudCostItem to the existing entry, if there is one;
+	// otherwise just set directly into allocations
+	if _, ok := ccis.CloudCostItems[that.Key()]; !ok {
+		ccis.CloudCostItems[that.Key()] = that.Clone()
+	} else {
+		ccis.CloudCostItems[that.Key()].add(that)
+		//return fmt.Errorf("cannot re-insert %s", that.Key())
```
michaelmdresser: Comment?

pkg/kubecost/cloudcostitem.go
```diff
@@ -0,0 +1,362 @@
...
+		//    pct := 4.0 / 4.0 = 1.0 for window 1
+		//    pct := 0.0 / 4.0 = 0.0 for window 2
+		//    pct := 0.0 / 4.0 = 0.0 for window 3
+		//
+		// 2. item overlaps multiple CloudCostItemSet windows
+		//    |      ----|----------|--        |
+		//    totalMins = 16.0
+		//    pct :=  4.0 / 16.0 = 0.250 for window 1
+		//    pct := 10.0 / 16.0 = 0.625 for window 2
+		//    pct :=  2.0 / 16.0 = 0.125 for window 3
```
michaelmdresser: Is this the exact same logic as the other one? Is it possible to put in a single function and unit test just that, so we can be sure that there isn't implementation drift if some future dev doesn't realize that its duplicated?

pkg/kubecost/cloudcostitem.go
```diff
@@ -0,0 +1,362 @@
...
+			Credit:       credit * pct,
+		}
+		err := ccis.Insert(cci)
+		if err != nil {
+			log.Errorf("LoadCloudCostItemSets: failed to load CloudCostItem with key %s and window %s: %s", cci.Key(), window.String(), err.Error())
+		}
+	}
+}
+
+// TODO:cloudcost bingen for time.Duration
```
michaelmdresser: :eyes:

pkg/kubecost/coverage.go
```diff
@@ -0,0 +1,108 @@
+package kubecost
+
+import (
+	"time"
+
+	"github.com/opencost/opencost/pkg/filter"
+)
+
+// Coverage This is a placeholder struct which can be replaced by a more specific implementation later
+type Coverage struct {
```
michaelmdresser, line 10: A good docstring here would be great, but maybe you're waiting to discuss more in depth?

pkg/kubecost/coverage.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (c *Coverage) Clone() *Coverage {
+	return &Coverage{
+		Window:   c.Window.Clone(),
+		Type:     c.Type,
+		Count:    c.Count,
+		Updated:  c.Updated,
+		Errors:   append(c.Errors),
+		Warnings: append(c.Warnings),
```
michaelmdresser: I don't think this `append` usage is a true copy: https://go.dev/play/p/xoieGEkcNxP

pkg/kubecost/coverage.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+// Coverage This is a placeholder struct which can be replaced by a more specific implementation later
+type CoverageSet struct {
+	Window Window               `json:"window"`
```
michaelmdresser, line 54: Why do both `Coverage` and `CoverageSet` have windows? It feels to me like the Window of a `CoverageSet` should be derived (e.g. `.Window()`) from the stored Coverages' Windows. Is that wrong?

pkg/kubecost/window.go
```diff
@@ -689,6 +689,74 @@ func (w Window) DurationOffsetStrings() (string, string) {
 	return timeutil.DurationOffsetStrings(dur, off)
 }
 
+// GetWindows returns a slice of Window with equal size between the given start and end. If windowSize does not evenly
+// divide the period between start and end, the last window is not added
+func GetWindows(start time.Time, end time.Time, windowSize time.Duration) ([]Window, error) {
```
michaelmdresser, line 738: These new funcs are neat, should we add at least one unit test apiece?

michaelmdresser (approved):
Approving because I think the general picture here is good. The only thing I really want to see addressed before a merge is the usage of `append` in the `Clone`.

pkg/filter/window.go
nikovacevic: I'll come back through and remove. Unfortunately, this causes a circular dep :(

pkg/kubecost/cloudcostaggregate.go
nikovacevic: I have them resolved in a branch. I'll merge my work ASAP and those will go away.

pkg/kubecost/cloudcostaggregate.go
nikovacevic: They're just for internal use. Not relevant to the end user.

pkg/kubecost/cloudcostaggregate.go
nikovacevic: I'm going Clone() maximalist from the start, and then we can get the memory improvement GAINZ we need later on.

nikovacevic: We'll have a more-or-less functional API for a bit! Yay!

pkg/kubecost/cloudcostitem.go
nikovacevic: This has been addressed in my "Address TODOs" PR, and we do Clone there.

pkg/kubecost/cloudcostitem.go
nikovacevic: I'll clean that up, good catch.

pkg/kubecost/cloudcostitem.go
nikovacevic: Cleaned up

pkg/filter/filter.go
```diff
@@ -0,0 +1,120 @@
...
+	//
+	// Example:
+	// (and (or (namespaceequals "kubecost")) (or)) ->
+	// (namespaceequals "kubecost")
+	//
+	// (and (or)) -> nil
+	Flattened() Filter[T]
+
+	// equals checks that the input filter is exactly equal. non-flattend filters should not use this method
+	equals(Filter[T]) bool
```
mbolt35: Already talked with @Sean-Holcomb a bit about this. This should really just be `Equals(Filter[T])` - using lower case here would cause some very unintentional consequences.

For example, `Filter[T]` is a public API, therefore importable from any other code base/package. Let's say my project imports opencost and wants to write a `Filter[T]` implementation - this is prevented due to this interface. 

This problem isn't local to external projects either. Just trying to create a `Filter` implementation in another package will not work.

mbolt35 (dismissed):
`equals(Filter[T]) bool` should be `Equals(Filter[T]) bool`

pkg/filter/filter.go
mbolt35: Additionally, comments like the following: 
```
non-flattend filters should not use this method
```
indicate to me that this should not be part of the interface.

pkg/filter/and.go
mbolt35: @michaelmdresser I had to write out some basic examples to test this:

Assuming that the variables in the following are a "push" (just a bool), you can count how many resulting operations there are. This _also_ assumes all operations have equal weight:

### 2 Nots out of 3

```
!a && !b && c
```
Ops: `And`, `Not`, `Not`
Total: `3`

Applying the 50% Not rule:

```
!(a || b || !c)
```
Ops: `Not`, `Or`, `Not`
Total: `3`

These are equivalent in total ops: 🚫 

---- 

### 3 Nots out of 4

```
!a && !b && !c && d
```
Ops: `And`, `Not`, `Not`, `Not`
Total: `4`

Applying the 50% Not rule:

```
!(a || b || c || !d)
```
Ops: `Not`, `Or`, `Not`
Total: `3`

Applying DeMorgan's yields 1 less operation: ✔️ 

---- 

### 3 Nots out of 5

```
!a && !b && !c && d && e
```
Ops: `And`, `Not`, `Not`, `Not`
Total: `4`

Applying the 50% Not rule:

```
!(a || b || c || !d || !e)
```
Ops: `Not`, `Or`, `Not`, `Not`
Total: `4`

These are equivalent in total ops: 🚫 

----

It looks like the `test` to see if a combination of filters can be reduced should be: 
```go
totalOps := // ...
totalNotOps := // ... 
remainingOps := totalOps - totalNotOps  

// Because we are adding +1 NotOp in exchange for TotalNotOps, but that involves flipping RemainingOps, 
// TotalNotOps has to be > RemainingOps + 1
if totalNotOps > remainingOps + 1 { 
    return Not[T]{Or[T]{negateFilters(flattenedFilters)}}
}
```

mbolt35:
See the following comment - I think there is a bug where `50%` isn't enough to justify applying DeMorgan's rule: 

https://github.com/opencost/opencost/pull/1508#discussion_r1038569755

pkg/filter/or.go
```diff
@@ -0,0 +1,112 @@
...
+	if len(flattenedFilters) == 0 {
+		return AllPass[T]{}
+	}
+
+	if len(flattenedFilters) == 1 {
+		return flattenedFilters[0]
+	}
+
+	// If more than half of the children are Not filters, Or should have DeMorgans rule applied to it
+	if percentNot(flattenedFilters) >= 0.5 {
```
mbolt35: This appears to be different than the `And` filter condition to apply demorgan's rule. Shouldn't these be equivalent with respect to _when_ to apply the rule? See here: https://github.com/opencost/opencost/pull/1508#discussion_r1038569755

pkg/filter/and.go
mbolt35: Now that I'm following a little closer, I see that the point is to prove equivalency. The only concern I have is:

`!(a || b || !c || !d)` == `(!a && !b && c && d)` when flattening would become:

`!!(!a && !b && c && d)` == `(!a && !b && c && d)`

Do we recursively simplify here? Does the `!!` cancel?

michaelmdresser: I think the !! does cancel, but check `Not[T] Flatten()` to be sure

pkg/kubecost/coverage.go
mbolt35: It's possible you could use something like `append([]string{}, c.Errors...)` instead. However, the following is preferable:

```go
var errors []string 
if len(c.Errors) > 0 {
    errors = make([]string, len(c.Errors))
    copy(errors, c.Errors)
}
```

pkg/kubecost/coverage.go
mbolt35: @michaelmdresser I think that deriving that logic from the stored coverage windows would be annoying to calculate every time you wanted the range if your coverages are stored in an unsorted list. It's likely easier to modify the window of the set as new coverages are added (via the `Expand` method). 

However, you _could_ store coverages in order based on their windows, and be able to quickly access `coverages[0].Start -> coverages[len].End`

pkg/filter/and.go
mbolt35: I mean, I see that `Not[T].Flatten()` does remove double bang, but what we return from the DeMorgan's rule is:
```go
return Not[T]{Or[T]{negateFilters(flattenedFilters)}}
```

I think we would have to recursively Flatten() in order for it to work.

Sean-Holcomb: Flatten is recursive, and yes proving equality was the idea here (There needed to be a deterministic criteria when there were equal number of negations and non-negations). As fun of problem as this is to work through, Bolt put it best in our side chat that flatten and equality isn't really necessary. My next move here is going to be to remove these features from the api since there are still existing problems.

pkg/filter/or.go
Sean-Holcomb: I think you figured out already but this was about ensuring that conjunctions with even numbers of elements where half were negations, end up in the same format.

However I will be removing equality and flatten from filters to address this

PR #1516, opened 2022-11-30
Adjusting Node Pricing based on retrieving system disk information and passing it as param to Alibaba's pricing API

avrodrigues5:
## What does this PR change?
* Before this PR, Node pricing was calculated with default system disk information, now we retrieve the system disk information and pass it along to get the actual node hourly price based on both the node type as well as the system disk attached to the node instance. 

* Enhancement to initiate push of [1488](https://github.com/opencost/opencost/pull/1488)

* accurate node hourly price 

* Closes [#1055](https://github.com/kubecost/kubecost-cost-model/issues/1055)

## How was this PR tested?
*  Alibaba's pricing calculator for ecs.g5.xlarge type is 0.252$ in us-east-1 region with system disk ESSD 120 GiB
In test Alibaba cluster, before this fix our hourly price with default params  default system disk to lowest priority disk category cloud_ssd and 40 GiB resulting in price of only 0.194$ now with the fix it's exactly accurate at 0.252$

Alibaba's price for the instance
Screen Shot 2022-11-30 at 10 53 35 AM

Before this PR node metrics in Prometheus
Screen Shot 2022-11-30 at 10 58 05 AM

After this PR node metrics in prometheus
Screen Shot 2022-11-30 at 10 59 57 AM

* 

* v1.99


pkg/cloud/aliyunprovider.go
```diff
@@ -142,23 +142,25 @@ type SlimK8sNode struct {
...
 		IsIoOptimized:      isIOOptimized,
 		OSType:             osType,
+		SystemDisk:         systemDiskInfo,
 		ProviderID:         providerID,
 		InstanceTypeFamily: instanceTypeFamily,
 	}
 }
 
-// AlibabaNodeAttributes represents metadata about the product pricing information used to map to a node.
+// AlibabaNodeAttributes represents metadata about the Node in it's pricing information.
```
nikovacevic: "it's" => "its"

pkg/cloud/aliyunprovider.go
```diff
@@ -170,19 +172,28 @@ type AlibabaNodeAttributes struct {
...
-		MemorySizeInKiB: node.MemorySizeInKiB,
-		IsIoOptimized:   node.IsIoOptimized,
-		OSType:          node.OSType,
+		InstanceType:               node.InstanceType,
+		MemorySizeInKiB:            node.MemorySizeInKiB,
+		IsIoOptimized:              node.IsIoOptimized,
+		OSType:                     node.OSType,
+		SystemDiskCategory:         node.SystemDisk.DiskCategory,
+		SystemDiskSizeInGiB:        node.SystemDisk.SizeInGiB,
+		SystemDiskPerformanceLevel: node.SystemDisk.PerformanceLevel,
```
nikovacevic: Let's just be careful here... I believe `node.SystemDisk` is a pointer, so we might want a `nil` check before trying to access `PerformanceLevel` on it. (Even if you're certain your code doesn't panic, someone else might call this function without a valid `SystemDisk` later on.)

nikovacevic: For that matter, we might want to add an overall `nil` check to the top on `node`, which is a pointer, itself. In that case, we can just return `nil`.

pkg/cloud/aliyunprovider.go
```diff
@@ -170,19 +172,28 @@ type AlibabaNodeAttributes struct {
...
 	}
 }
 
-// AlibabaPVAttributes represents metadata about the product pricing information used to map to a PV.
-// Basic Attributes needed atleast to get the keys.Some attributes from k8s PV response
+// AlibabaPVAttributes represents metadata the PV in it's pricing information.
```
nikovacevic: "it's" => "its"

pkg/cloud/aliyunprovider.go
```diff
@@ -904,7 +1026,80 @@ func getInstanceFamilyFromType(instanceType string) string {
...
+			return
+		}
+		// Every instance should only have one system disk per Alibaba Cloud documentation https://www.alibabacloud.com/help/en/elastic-compute-service/latest/block-storage-overview-disks,
+		// if TotalCount is not 1 just return empty and let it not impact default pricing.
+		if response.TotalCount != 1 {
+			log.Warnf("Total count of system disk for node with InstanceID: %s is not 1, hence defaulting it to an empty system disk to pass through to defaults", instanceID)
+			return
+		}
+		// TO-DO: When supporting Subscription type disk, you can leverge the disk.DiskChargeType here to map it to subscription type.
+		systemDisk := response.Disks.Disk[0]
```
nikovacevic, line 1127: Looks like we check `response.TotalCount` above. Do you think it's also worth checking `response.Disks == nil` and `len(response.Disks.Disk)` here? Might be a nice add, even though it seems ridiculous.

nikovacevic (approved):
Left some improvements, but overall this seems good to me. Thank you @avrodrigues5!

pkg/cloud/aliyunprovider.go
avrodrigues5: Agreed , i think there was a possibility of panic in  current code too. Thanks for pointing it out.

PR #1523, opened 2022-12-05
Ignoring Idle Summary allocation in total efficiency computation of Summary allocation set 

avrodrigues5:
## What does this PR change?
* Parity between old UI and new UI of Kubecost calling the Opencost package 

## Does this PR relate to any other PRs?
* None

## How will this PR impact users?
* Consistency between efficiency between old UI and New UI

Old UI with Idle not impacting the total efficiency
Screen Shot 2022-12-05 at 3 30 57 PM
Screen Shot 2022-12-05 at 3 31 25 PM

New UI without the above fix where total efficiency is impacted with idle true
Screen Shot 2022-12-05 at 3 30 52 PM
Screen Shot 2022-12-05 at 3 31 20 PM

## Does this PR address any GitHub or Zendesk issues?
* Closes https://github.com/kubecost/kubecost-cost-model/issues/1057

## How was this PR tested?
* Unit tests

## Does this PR require changes to documentation?
* No

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next Opencost release? If not, why not?
* v1.99 (Can maintainer label it please)

PR #1537, opened 2022-12-07
Taking into account Summary Allocation not run for entire duration of SAS set window for RAM and CPU efficiency calculation 

avrodrigues5:
## What does this PR change?
* Fixing UI mismatch between old UI and new UI resulting from the impact of calculating RAM and CPU efficiency of allocations not active for entire duration of the requested window.

* Enhancement to #1523 after investigation

* Absolute concurrency between old UI and new UI to the decimal place

* Closes https://github.com/kubecost/kubecost-cost-model/issues/1057

## How was this PR tested?
* Uploading to my instance and testing old and new UI along with drilldowns

* None

* v.199


nikovacevic (approved):
Awesome find!!! This is great.

PR #1538, opened 2022-12-07
Nick/azure orphaned resources disknames

nickcurie:
## What does this PR change?
* Adds disk names to Azure's orphaned resources

* https://github.com/opencost/opencost/pull/1489

* Users will see disk names

* 

## How was this PR tested?
* On Azure cluster

* No

* v1.99

### DIFF
--- a/pkg/cloud/azureprovider.go
+++ b/pkg/cloud/azureprovider.go
@@ -3,7 +3,6 @@ package cloud
 import (
 	"context"
 	"fmt"
-	"github.com/opencost/opencost/pkg/kubecost"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


pkg/cloud/azureprovider.go
```diff
@@ -1274,6 +1274,7 @@ func (az *Azure) GetOrphanedResources() ([]OrphanedResource, error) {
 					"timeCreated": d.TimeCreated.String(),
 				},
 				Size:        d.DiskSizeGB,
+				DiskName:    *d.Name,
```
nikovacevic: Given that this is a pointer, I'm guessing there's a chance it could be `nil`. Let's check that nil, just to be safe?

```
diskName := ""
if d.Name != nil {
    diskName = *d.Name
}

...

DiskName: diskName,
```

nikovacevic (changes requested):
Just that one nil check to add! Otherwise, looks good.

pkg/cloud/azureprovider.go
nickcurie: Should I do the same with region?

michaelmdresser (approved):

nikovacevic (approved):

PR #1539, opened 2022-12-08
Noticed a Panic during system disk retrieval for a node when Alibaba service key is not given

avrodrigues5:
## What does this PR change?
* Fixes the panic

* Fix for Alibaba provider 

* No panic when servicekey is not given

* Closes ...

## How was this PR tested?
* Negative case when panic occurs, positive case with alibaba service key provided

* None

* v1.99

### DIFF
--- a/pkg/cloud/aliyunprovider.go
+++ b/pkg/cloud/aliyunprovider.go
@@ -481,8 +481,8 @@ func (alibaba *Alibaba) NodePricing(key Key) (*Node, error) {
 
 	pricing, ok := alibaba.Pricing[keyFeature]
 	if !ok {
-		log.Warnf("Node pricing information not found for node with feature: %s", keyFeature)
-		return &Node{}, nil
+		log.Errorf("Node pricing information not found for node with feature: %s", keyFeature)
+		return nil, fmt.Errorf("Node pricing information not found for node with feature: %s letting it use default values", keyFeature)
 	}
 
 	log.Debugf("returning the node price for the node with feature: %s", keyFeature)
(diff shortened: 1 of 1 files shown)


michaelmdresser:
That `Bump to v1.99.0-rc.0` commit shouldn't be in here if we're merging into `develop`, would you please remove it?

pkg/cloud/aliyunprovider.go
```diff
@@ -481,8 +481,8 @@ func (alibaba *Alibaba) NodePricing(key Key) (*Node, error) {
 
 	pricing, ok := alibaba.Pricing[keyFeature]
 	if !ok {
-		log.Warnf("Node pricing information not found for node with feature: %s", keyFeature)
-		return &Node{}, nil
+		log.Errorf("Node pricing information not found for node with feature: %s", keyFeature)
+		return nil, fmt.Errorf("Node pricing information not found for node with feature: %s letting it use default values", keyFeature)
```
nikovacevic, line 485: Are these returns (instead of `nil` with empty `&Node{}`) what fixes the panic? Could you add to the description of the PR where, precisely, the panic was occurring, and why?

pkg/cloud/aliyunprovider.go
```diff
@@ -782,24 +791,11 @@ func (alibaba *Alibaba) GetKey(mapValue map[string]string, node *v1.Node) Key {
 		client, err = sdk.NewClientWithAccessKey(slimK8sNode.RegionID, aak.AccessKeyId, aak.AccessKeySecret)
```
avrodrigues5, line 796: @nikovacevic , the panic was here, before the code fix i was setting `skipSystemDiskRetrieval` to true and then code would follow into create a sdk client with the keys `aak.AccessKeyId` `aak.AccessKeySecret` expected to be initialized but where nil...

pkg/cloud/aliyunprovider.go
avrodrigues5: Im sending an error here because i want Opencost to calculate cost using the BaseCPUprice when error encountered. https://github.com/opencost/opencost/blob/develop/pkg/costmodel/costmodel.go#L959-L971

pkg/cloud/aliyunprovider.go
nikovacevic: So `alibaba.accessKey` was `nil`? And you were trying to access `alibaba.accessKey.AccessKeyId` and `alibaba.accessKey.Secret`? If so, I would recommend a manual `nil` check there (`if aak == nil`), even if your change here appears to solve it. Best to handle such things explicitly.

avrodrigues5: Thanks @nikovacevic . I have added this needed condition.

nikovacevic (approved):

PR #1574, opened 2022-12-16
Allow ui to run from non privileged

torreytsui:
## What does this PR change?
**Fix this issue**

> [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
> [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)

```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2022/12/16 12:45:34 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2022/12/16 12:45:34 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
```

* No

* Fix the frontend ui so that it can be run on non privileged k8s cluster

* No

## How was this PR tested?
* Tested on a private cluster which run on non privileged mode

* No

* v1.99.1

### DIFF
--- a/ui/default.nginx.conf
+++ b/ui/default.nginx.conf
@@ -31,6 +31,13 @@ gzip_types
     text/xml
     text/x-component
     text/x-cross-domain-policy;
+
+upstream model {
+    # Update to the cost model endpoint
+    # Example: host.docker.internal:9003;
+    server 0.0.0.0:9003;
+}
+
 server {
     server_name _;
     root /var/www;
@@ -51,7 +58,6 @@ server {
     add_header ETag "1.96.0";
     listen 9090;
     listen [::]:9090;
-    resolver kube-dns.kube-system.svc.cluster.local valid=5s;
     location /healthz {
         return 200 'OK';
     }
@@ -59,8 +65,7 @@ server {
         proxy_connect_timeout       180;
         proxy_send_timeout          180;
         proxy_read_timeout          180;
-        set $server http://opencost.opencost.svc.cluster.local.:9003;
-        proxy_pass $server;
+        proxy_pass http://model/;
         proxy_redirect off;
         proxy_http_version 1.1;
         proxy_set_header Connection "";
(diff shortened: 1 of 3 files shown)


dwbrown2:
Thanks for this contribution! @AjayTripathy can you help triage for review?

ui/default.nginx.conf
```diff
@@ -31,6 +31,13 @@ gzip_types
     text/xml
     text/x-component
     text/x-cross-domain-policy;
+
+upstream model {
+    # Update to the cost model endpoint
+    # Example: host.docker.internal:9003;
+    server 0.0.0.0:9003;
```
torreytsui, line 38: Most likely, deployed into the same pod as the cost model - therefore it should resolve to 0.0.0.0:9003.

Once we have the helm chart, we could update this to point at the service in the same namespace.

AjayTripathy (approved):

PR #1577, opened 2022-12-16
feat(metrics): Add opencost_build_info metric

### LINKED ISSUE #1576: Add opencost_build_info prometheus metric
Currently there is no way of determining what version of `opencost` via PromQL. This is important for two reasons:
1. Internally we rely upon querying `${SERVICE}_build_info` to determine if a new version is available for us to upgrade to
2. When having _many_ `opencost` deployments, it's helpful to understand which cluster has which version of `opencost` running

We should export an addition metric `opencost_build_info` that has at least a `version` label that is inferred at build time of the container. 

We could manually release upgrades, but that's tedious and goes against our internal goal of staying relatively up to do. 
We could also write custom scripts to determine the versions per cluster, but again would prefer for this to align with our existing dashboards.

Pokom:
## What does this PR change?
* Adds a `opencost_build_info` metric with a `version` label

* No

* Limited impact, will add one more metric

* Closes #1576 

## How was this PR tested?
* Locally

* N/A

I'm looking for some input from other maintainers on:
- How to best get the version information(Git SHA, ldflags set a build time etc..)
- Is this the right place into the codebase to set the build information?


AjayTripathy:
https://github.com/opencost/opencost/blob/v1.99/pkg/env/costmodelenv.go#L170 Just a note that the env variable of published version seems relevant!

Pokom:
Thanks for the nudge @AjayTripathy! After looking through, I think the two version variables *should* be the same, but I am hesitant to rely upon one that can be overridden by users for the `opencost_build_info` metric. I feel it would be pretty easy to set that environment variable at runtime(IE, helm/jsonnet/manually) and then forget about it. 

Out of curiosity, what is the use case for being able to override `APP_VERSION`?

Pokom:
@AjayTripathy I think this PR is ready for review at this point. It should be good enough for a first pass to start exporting the version which will be enough for me to setup rolling out `opencost` in an automated fashion. Let me know if there's anything you'd like to see tested prior to a review.

AjayTripathy:
SGTM will take a look @Pokom ! Just curious how are you planning to automate the rollouts?

pkg/metrics/telemetry.go
```diff
@@ -42,7 +51,7 @@ func InitKubecostTelemetry(config *MetricsConfig) {
 			Help: "kubecost_http_response_size_bytes Response size in bytes",
 		}, []string{"handler", "method", "code"})
 
-		prometheus.MustRegister(requestsCount, responseTime, responseSize)
+		prometheus.MustRegister(requestsCount, responseTime, responseSize, buildInfo)
```
AjayTripathy, line 54: Did you mean to do this on the requestsCounts metric? I don't think that's necessary.

AjayTripathy (changes requested):

AjayTripathy:
@Pokom quick question? I think the metric got accidentaly added as a label on another request?

pkg/metrics/telemetry.go
Pokom: @AjayTripathy I didn't think this impacted the other metrics. I was piggy backing off of the existing `prometheus.MustRegister` calls . If needed, I could move out buildInfo to it's own separate `prometheus.MustRegister` call

Pokom:
> @AjayTripathy Just curious how are you planning to automate the rollouts?

We have internal tooling setup to periodically check if there's a new container image in the registry by querying prometheus for the specific `*_build_info` metric and then does some GitOps magic under the hood to stage a rollout. 

>  I think the metric got accidentaly added as a label on another request?

Provided my response inline. I can double check the exported metrics, but I thought that `prometheus.MustRegister` accepted a variable amount of collectors and registered them independently. I don't think it will impact the other metrics.

pkg/metrics/telemetry.go
AjayTripathy: Oh NVM I think it's fine.

AjayTripathy (approved):

PR #1592, opened 2023-01-03
Add endpoints to set and get log levels

saweber:
Signed-off-by: saweber <saweber@gmail.com>

## What does this PR change?
* addition of GET /logs/level and POST /logs/level endpoints

* no

* ability to get and set log levels without restarting application

* Closes issue in private repo

## How was this PR tested?
* manually

* yes - addition of GET /logs/level and POST /logs/level endpoints

* I can't.

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -1398,6 +1398,47 @@ func (a *Accesses) Status(w http.ResponseWriter, r *http.Request, _ httprouter.P
 	}
 }
 
+type LogLevelRequestResponse struct {
+	Level string `json:"level"`
+}
+
+func (a *Accesses) GetLogLevel(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+
+	level := log.GetLogLevel()
+	llrr := LogLevelRequestResponse{
+		Level: level,
+	}
+
+	body, err := json.Marshal(llrr)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("unable to retrive log level"), http.StatusInternalServerError)
+		return
+	}
+	_, err = w.Write(body)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("unable to write response: %s", body), http.StatusInternalServerError)
+		return
+	}
+}
+
+func (a *Accesses) SetLogLevel(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	params := LogLevelRequestResponse{}
+	err := json.NewDecoder(r.Body).Decode(&params)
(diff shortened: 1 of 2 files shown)


pkg/costmodel/router.go
```diff
@@ -1398,6 +1398,38 @@ func (a *Accesses) Status(w http.ResponseWriter, r *http.Request, _ httprouter.P
...
+	if err != nil {
+		http.Error(w, fmt.Sprintf("unable to retrive log level"), http.StatusInternalServerError)
+		return
+	}
+	w.Write(body)
+}
+
+func (a *Accesses) SetLogLevel(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	qp := httputil.NewQueryParams(r.URL.Query())
+	level := qp.Get("level", "")
```
michaelmdresser: If this is a POST request, perhaps we should get the data from the body? Frankly, I'm not sure if that's the right convention as I'm not an HTTP expert.

pkg/costmodel/router.go
```diff
@@ -1398,6 +1398,38 @@ func (a *Accesses) Status(w http.ResponseWriter, r *http.Request, _ httprouter.P
...
+	err := log.SetLogLevel(level)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("level must be a valid log level according to zerolog; level given: %s", level), http.StatusBadRequest)
+		return
+	}
```
michaelmdresser: Great error message. Would you also include `err` somewhere in the message, just in case its something unexpected?

michaelmdresser (approved):
This is awesome, thanks for picking this up and nice encapsulation of zerolog-specific calls in `pkg/log`!

pkg/costmodel/router.go
saweber: Hopefully we don't go too deep down the REST rabbit hole here... :smile: 

1) Since this endpoint is idempotent, the 'more correct' verb is PUT. I chose to stick with POST for consistency with other endpoints in OpenCost, as well as closed source. I'm open to making this `PUT`. I just did not want this to be the only `PUT` endpoint.
2) After investigating and refreshing on REST, using the body is not required, but is generally the better practice. Also after looking into the closed source endpoints, those consistently use the body for data instead of a query parameter.

TL;DR -> Yes, I'll make the change.

michaelmdresser: Love it, thanks for the detail. I'm down with whatever you think is best.

nikovacevic:
Closes https://github.com/kubecost/kubecost-cost-model/issues/932

nikovacevic (approved):

nikovacevic:
@saweber I'll be happy to merge this as soon as the merge conflicts are resolved. Perhaps you can rebase tomorrow, and then let me know so I can merge?

saweber:
@nikovacevic Done!

PR #1594, opened 2023-01-04
allocation struct to have  Network breakdown values of Zone, Region and Internet costs

avrodrigues5:
## What does this PR change?
* Gives visibility into underlying network cost subtypes of Zone cost, region cost and internet cost

* No

* kets the /model/allocation have data for the three new fields, have to rebuild allocation ETL to have parity between networkCost with NetworkZoneCost, RegionCost and InternetCost.

* Closes https://github.com/kubecost/kubecost-cost-model/issues/1141

## How was this PR tested?
* Test case1: Network costs enabled and dev image uploaded immediately
* Test case2: Network costs enabled, let it run for a bit and dev image uploaded after few hours
* Test case3: Network costs enabled dev image uploaded later swapped for prior v1.99 version of kcm and made it run for a but then later uploaded dev image.

* [Add fields here](https://github.com/kubecost/docs/blob/master/allocation.md), will sync up with this later

* Cant label can someone label v1.100 please


dwbrown2:
@avrodrigues5 @mbolt35 should we go ahead and split/create fields for egress vs ingress? I think we'll want the ability to track both going forward..

Also, should we track cost vs byes transfered?

Lastly, thoughts on these names... NetworkZoneCost --> CrossNetworkZoneCost, RegionCost --> CrossRegionCost

AjayTripathy:
Webb to your points:
1. I think these are all egress and ingress is separate since ingress generally does not have charges. We could track ingress somewhere else, but it may not belong on an allocation ETL since ingress happens at the cluster or loadbalancer level.
2. Preferring fewer fields over more fields seems better if fields can be recovered, to keep our footprint small. Pricing doesn't fluctuate frequently so in general I think the bytes is recoverable via division of the config pricing.
3. Renaming is really hard in ETL so I'd favor consistency given we've already thrown one name on here. We can always rename at the view layer later.

dwbrown2:
1. Ingress _can_ have charges and just trying to avoid us backing ourselves in a corner that makes it tricky to add later

AjayTripathy:
Fair enough, but I don't think we currently track eg "cross region ingress" in a meaningful way, so maybe just a single field for ingress?

dwbrown2:
> Fair enough, but I don't think we currently track eg "cross region ingress" in a meaningful way, so maybe just a single field for ingress?

My understanding is that we were tracking in Prom metrics, but I may be mistaken.

avrodrigues5:
https://docs.kubecost.com/general/user-metrics#kubecost-network-costs, I dont think we have any ingress cost metrics in prometheus at this time, only egress. I looked at pkg/costmodel/metrics.go and this doc seems to be up to date.

dwbrown2:
Doesn't this resource highlight `kubecost_pod_network_ingress_bytes_total`?

avrodrigues5:
To calculate egress cost for zone, region and internet we use `kubecost_pod_network_egress_bytes_total` along with `kubecost_network_zone_egress_cost`, `kubecost_network_region_egress_cost` and `kubecost_network_region_egress_cost` respectively. we dont have a metrics similar like that for ingress at this time in prometheus AFAIK and only `kubecost_pod_network_ingress_bytes_total`. 

As always I might be wrong and open to be corrected by others :)

AjayTripathy:
> My understanding is that we were tracking in Prom metrics, but I may be mistaken.

Just `kubecost_pod_network_ingress_bytes_total` as noted. I'm not opposed to a field here, but to avoid scope creep I think we can defer it for a later time since it's not used in cost calculations at all today.

mbolt35:
I'm just skimmed through a lot of this thread, and from what I'm gathering, it sounds as if everyone is on the same page in terms of what is currently available:
* `kubecost_pod_network_ingress_bytes_total` is being emitted by the network-costs pod and scraped. 
  * This _does_ include the category of traffic based on the fields:
    * `internet`
    * `sameRegion`
    * `sameZone`
* Nothing in the cost-model queries `kubecost_pod_network_ingress_bytes_total`

For pricing `egress` we use the different categories of pricing traditional to egress pricing models: 
* in-zone
* cross-zone (same-region)
* cross-region
* internet

In-Zone is assumed to be free, so I don't believe there exists a price metric for it. However, as noted, the other price metrics do exist. While there are a lot of providers who don't charge for ingress in similar categories, it does seem like we should have the same pricing consistent to egress, no? 

Additionally, if we're wanting to add more visibility into the "type" or "category" in ETL, my suggestion is _add_ a field that is a struct `NetworkCosts` that contains something like:

```go
type NetworkCost struct {
    CrossZoneBytes float64
    CrossZoneCost float64 
    CrossRegionBytes float64 
    CrossRegionCost float64 
    InternetBytes float64 
    InternetCost float64 
}

type NetworkCosts struct {
    Egress *NetworkCost
    Ingress *NetworkCost 
}
````

I would not suggest we try to add metrics incrementally - it's better to make these changes up front and stay consistent to avoid schema or metrics changes down the road.

File: pkg/costmodel/allocation_helpers.go [hunk 1]
```diff
@@ -749,10 +751,26 @@ func applyNetworkAllocation(podMap map[podKey]*pod, resNetworkGiB []*prom.QueryR
...
+				case networkZoneCost:
+					alloc.NetworkZoneCost = currentNetworkSubCost
+				case networkRegionCost:
+					alloc.NetworkRegionCost = currentNetworkSubCost
+				case networkInternetCost:
+					alloc.NetworkInternetCost = currentNetworkSubCost
+				default:
+					if !slices.Contains(unknownNetworkTypes, networkCostSubType) {
+						unknownNetworkTypes = append(unknownNetworkTypes, networkCostSubType)
+					}
```
nikovacevic: No need to collect these this way -- let's just log here directly and move forward. As written, we'll never end up here, anyways, as we control the only call to this private function. (Bigger deal if this were a public function.)

pkg/costmodel/allocation.go
```diff
@@ -58,6 +58,13 @@ const (
 	queryFmtLBActiveMins             = `count(kubecost_load_balancer_cost) by (namespace, service_name, %s)[%s:%s]`
 )
 
+// Constants for Network Cost Subtype
+const (
+	networkZoneCost     = "NetworkZoneCost"
+	networkRegionCost   = "NetworkRegionCost"
```
nikovacevic: Going to check if we should call these "CrossZone" and "CrossRegion" or "XZone" and "XRegion" instead of just "Zone" and "Region".

nikovacevic: Just saw Webb's comment. Let's go with "NetworkCrossRegionCost" and "NetworkCrossZoneCost"

nikovacevic:
Closes https://github.com/kubecost/kubecost-cost-model/issues/1141 (just opened the issue)

nikovacevic (approved):
This looks good to me for what was originally asked. I think we should merge this. Then, after we clear up what we want to do next, I'll open a new ticket with that scope and assign it out.

nikovacevic (changes requested):
Oops! Let's change those names first, really quick.

pkg/kubecost/allocation.go
```diff
@@ -63,6 +63,9 @@ type Allocation struct {
 	NetworkTransferBytes       float64               `json:"networkTransferBytes"`
 	NetworkReceiveBytes        float64               `json:"networkReceiveBytes"`
 	NetworkCost                float64               `json:"networkCost"`
+	NetworkZoneCost            float64               `json:"networkZoneCost"`     // @bingen:field[version=16]
+	NetworkRegionCost          float64               `json:"networkRegionCost"`   // @bingen:field[version=16]
+	NetworkInternetCost        float64               `json:"networkInternetCost"` // @bingen:field[version=16]
```
nikovacevic: Also -- are these versions correct? Aren't we at a higher version now?

nikovacevic:
My only other comment here relates to @mbolt35 's proposal, which I agree is very nice and clean... but I'm worried about adding that many fields. 12 new floats is a ~30% increase all in one shot. SummaryAllocation will still compress that, so maybe nothing to worry about. But this type is already bloated and big changes to field counts make me nervous.

pkg/costmodel/allocation_helpers.go
avrodrigues5: addressed!

pkg/costmodel/allocation.go
avrodrigues5: Addressed by updating the variable name and regenerating the codec file

pkg/kubecost/allocation.go
avrodrigues5: It is at 16 itself for allocation struct

mbolt35:
So are we just lumping in ingress?

nikovacevic:
My proposal is just to merge Alan's changes as-is, then scope future changes. But if there's strong push back to that idea, we can hold off.

And in the medium-to-long term I'm hoping that a full, robust solution for network costs will come in the form of Network ETL -- including egress by x-zone, x-region, and Internet, as well as by cloud service; ingress; whatever else. Cramming too much in here is a mistake IMO. We're just trying to get the current network.html page working on ETL data quickly, is my understanding.

dwbrown2:
No objection here, just wanted to ask the questions

nikovacevic (approved):

PR #1597, opened 2023-01-05
aws and gcp orphaned resources

nickcurie:
## What does this PR change?
* Adds GetOrphanedResources functionality to AWS and GCP
* Changed OrphanedResource struct to use int64 for disk size rather than int32
  * Changed Azure implementation to account for this

* No

* Users can use the new `/savings/orphanedResources` endpoint on AWS and GCP clusters

* Closes https://github.com/kubecost/kubecost-cost-model/issues/1144

## How was this PR tested?
* In progress

* No

*


pkg/cloud/awsprovider.go
```diff
@@ -1517,6 +1526,14 @@ func (aws *AWS) GetAddresses() ([]byte, error) {
 	})
 }
 
+func (aws *AWS) isAddressOrphaned(address *ec2Types.Address) bool {
+	if address.AssociationId != nil {
+		return false
+	}
+
+	return true
```
nickcurie, line 1534: I made a bunch of private methods in this PR, just want to call that out in case that has some weird implications

pkg/cloud/awsprovider.go
```diff
@@ -1517,6 +1526,14 @@ func (aws *AWS) GetAddresses() ([]byte, error) {
 	})
 }
 
+func (aws *AWS) isAddressOrphaned(address *ec2Types.Address) bool {
```
nickcurie, line 1534: I made a bunch of private methods in this PR, just want to call that out in case that has some weird implications

pkg/cloud/awsprovider.go
```diff
@@ -1509,6 +1508,16 @@ func (aws *AWS) GetAddresses() ([]byte, error) {
 		return nil, fmt.Errorf("%d error(s) retrieving addresses: %v", len(errs), errs)
 	}
 
+	return addresses, nil
+}
+
+// GetAddresses retrieves EC2 addresses
+func (aws *AWS) GetAddresses() ([]byte, error) {
+	addresses, err := aws.getAllAddresses()
```
nikovacevic, line 1521: Slick move.

pkg/cloud/awsprovider.go
nikovacevic: No problem!

pkg/cloud/awsprovider.go
```diff
@@ -1610,8 +1636,87 @@ func (aws *AWS) GetDisks() ([]byte, error) {
 	})
 }
 
-func (*AWS) GetOrphanedResources() ([]OrphanedResource, error) {
-	return nil, errors.New("not implemented")
+func (aws *AWS) isDiskOrphaned(vol *ec2Types.Volume) bool {
+	// Do not consider volume orphaned if in use
+	if vol.State == "in-use" {
```
nikovacevic: This (and `"attached"`) might make nice `const` values.

pkg/cloud/awsprovider.go
```diff
@@ -1610,8 +1636,87 @@ func (aws *AWS) GetDisks() ([]byte, error) {
...
+	}
+
+	var orphanedResources []OrphanedResource
+
+	for _, volume := range volumes {
+		if aws.isDiskOrphaned(volume) {
+			cost := aws.findCostForDisk(volume)
+
+			var volumeSize int64
+			volumeSize = -1
```
nikovacevic: Why this?

pkg/cloud/awsprovider.go
```diff
@@ -1610,8 +1636,87 @@ func (aws *AWS) GetDisks() ([]byte, error) {
...
+			}
+
+			orphanedResources = append(orphanedResources, or)
+		}
+	}
+	return orphanedResources, nil
+}
+
+func (aws *AWS) findCostForDisk(disk *ec2Types.Volume) float64 {
+	//todo: use AWS pricing
```
nikovacevic: If this ends up being difficult, let me know.

nikovacevic: For this first pass, keeping parity with this hardcoding is not a huge problem.

pkg/cloud/azureprovider.go
```diff
@@ -1277,14 +1277,20 @@ func (az *Azure) GetOrphanedResources() ([]OrphanedResource, error) {
 				diskRegion = *d.Location
 			}
 
+			var diskSize int64
+			diskSize = -1
```
nikovacevic: Ditto

pkg/cloud/gcpprovider.go
```diff
@@ -355,11 +354,28 @@ func (gcp *GCP) GetAddresses() ([]byte, error) {
...
 
-// GetDisks returns the GCP disks backing PVs. Useful because sometimes k8s will not clean up PVs correctly. Requires a json config in /var/configs with key region.
-func (gcp *GCP) GetDisks() ([]byte, error) {
+func (gcp *GCP) isAddressOrphaned(address *compute.Address) bool {
+	// Do not consider address orphaned if it has more than 0 users
+	if len(address.Users) > 0 {
+		return false
+	}
+
+	return true
```
nikovacevic: I believe you could just say
```go
// Consider address orphaned if it has 0 users
return len(address.Users) == 0
```

nikovacevic: (No need to accept this as a change; just pointing it out!)

pkg/cloud/gcpprovider.go
```diff
@@ -379,12 +395,109 @@ func (gcp *GCP) GetDisks() ([]byte, error) {
...
+	}
+
+	for _, addressList := range addresses.Items {
+		if len(addressList.Addresses) == 0 {
+			continue
+		}
+
+		for _, address := range addressList.Addresses {
+			if gcp.isAddressOrphaned(address) {
+				//todo: use GCP pricing
```
nikovacevic, line 489: Ditto

pkg/cloud/gcpprovider.go
```diff
@@ -379,12 +395,109 @@ func (gcp *GCP) GetDisks() ([]byte, error) {
...
+				orphanedResources = append(orphanedResources, or)
+			}
+		}
+	}
+
+	return orphanedResources, nil
+}
+
+func (gcp *GCP) findCostForDisk(disk *compute.Disk) float64 {
+	//todo: use GCP pricing
```
nikovacevic: Ditto

nikovacevic (commented):
Looking really good so far!

pkg/cloud/awsprovider.go
```diff
@@ -1610,8 +1639,108 @@ func (aws *AWS) GetDisks() ([]byte, error) {
...
+	// price := 0.04
+	// if strings.Contains(string(disk.VolumeType), "ssd") {
+	// 	price = 0.17
+	// }
+	// if strings.Contains(string(disk.VolumeType), "gp2") {
+	// 	price = 0.1
+	// }
+	// cost := price * float64(*disk.Size)
+	// return &cost, nil
```
nikovacevic: Since we're keeping the actual code, can we clean up the old hard-coded logic?

pkg/cloud/awsprovider.go
```diff
@@ -1610,8 +1639,108 @@ func (aws *AWS) GetDisks() ([]byte, error) {
...
+				MonthlyCost: cost,
+			}
+
+			orphanedResources = append(orphanedResources, or)
+		}
+	}
+
+	for _, address := range addresses {
+		if aws.isAddressOrphaned(address) {
+			cost := timeutil.HoursPerMonth * 0.005
```
nikovacevic: Can we name this magic number? A `const` perhaps? Is this the hourly AWS public IP cost, I assume? (Know you've probably just copied this over from JS.)

pkg/cloud/gcpprovider.go
```diff
@@ -370,21 +385,144 @@ func (gcp *GCP) GetDisks() ([]byte, error) {
...
+
+	for _, addressList := range addresses.Items {
+		if len(addressList.Addresses) == 0 {
+			continue
+		}
+
+		for _, address := range addressList.Addresses {
+			if gcp.isAddressOrphaned(address) {
+				//todo: use GCP pricing
+				cost := 0.01 * timeutil.HoursPerMonth
```
nikovacevic: Likewise, can we name this?

pkg/cloud/gcpprovider.go
```diff
@@ -370,21 +385,144 @@ func (gcp *GCP) GetDisks() ([]byte, error) {
...
+
+func (gcp *GCP) findCostForDisk(disk *compute.Disk) (*float64, error) {
+	//todo: use GCP pricing struct
+	price := 0.04
+	if strings.Contains(disk.Type, "ssd") {
+		price = 0.17
+	}
+	if strings.Contains(disk.Type, "gp2") {
+		price = 0.1
+	}
```
nikovacevic: And, last one -- can we name these?

nikovacevic (approved):
This looks good! Is ready to merge as-is, but a little clean up (deleting old comments, naming magic variables) would be much appreciated. Thank you @nickcurie!

Issue #1590, opened 2022-12-30
Resource allocations are not tracked when scheduled, but only during Pod Running state


djmcgreal-cc:
**Describe the bug**

My immediate problem is specific to GPUs, but I think this may apply to resources more generally.

OpenCost only considers a resource to be allocated whilst its Pod is in a `Running` state. However, the resources on a node are already allocated to the Pod as soon as its scheduled (no other Pods can use it, at least for GPUs). In our case we're unable to assign costs to namespaces for Pods with containers in an `ImagePullBackoff` state.

**To Reproduce**

1. Run a Pod with a non-existent image
2. See that there are no `container_gpu_allocated` metrics for this Pod
3. Edit the Pod to point to an existent image
4. See that there will now be metrics for the Pod

**Expected behavior**

That allocation metrics are recorded as soon as the Pod is allocated to a node, e.g. that `spec.nodeName != null`

**Screenshots**

**Which version of OpenCost are you using?**

v1.97.0

**Additional context**

EKS 1.21, Nvidia GPUs (g4dn.xlarge)

Adam-Stack-PM:
@djmcgreal-cc,  Thanks for opening this issue and the additional context.  I will get eyes on this tomorrow.

AjayTripathy:
Hey @djmcgreal-cc I think this is an interesting point. I don't think pending pods block the scheduler but scheduled failed pods with imagepullbackoff do, and we don't currently account for them in our cost model. 

@nikovacevic thoughts on modifying our cost-model here to support imagepullbackoff? I think there is some technical complexity in implementing this quickly. We should at least document this for now.

dwbrown2:
Should we diagram all [pod states](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) and add to spec?

AjayTripathy:
That makes sense. Maybe all states and note what charge model is used for them? @bstuder99 can we add this note to documentation and get this knocked out? Basically a table like this:

State                      |  Cost Allocation         | Status
Running                 | max(usage, request) | Implemented
ImagePullBackoff | request                       | TODO, currently no charge

brstuder:
@AjayTripathy Yeah this idea sounds pretty easy to knock out, just need some clarifying details. We'll be implementing this in the OpenCost spec[ here](https://github.com/opencost/opencost/blob/develop/spec/opencost-specv01.md), right? For the example table you list, do we want to include other pod states in the article Webb linked?

[...]

PR #1598, opened 2023-01-06
Update opencost-specv01.md

brstuder:
Signed-off-by: bstuder99 <81711774+bstuder99@users.noreply.github.com>

## What does this PR change?
* Adds a section on cost allocation of pod states. There are other pod lifecycle states that I'm still waiting to hear from @AjayTripathy if we should cover, for now it explains running and `imagepullbackoff`.

* Based on https://github.com/opencost/opencost/issues/1590

* This PR only requires changes to the spec doc.


mattray (approved):

dwbrown2:
I like the idea of adding more detail on pod states, but is this not a bug/limitation vs something we want to adopt as a recommended practice?

brstuder:
If the idea was to document a bug initially, I feel as if it should exist in a known bugs/troubleshooting doc, if we want to format that info for a table, it would make sense to create a more cohesive approach to displaying all known information. That being said, Ajay's initial suggestion should be fine for informing users about this one problem.

mattray (approved):

PR #1602, opened 2023-01-10
Fix EKS Spot Node Detection

### LINKED ISSUE #1600: EKS: No spot instances detected
**Describe the bug**

`kubecost_node_is_spot` is returning no results for any spot nodes in EKS clusters. `kubecost_node_is_spot` _does_ return the correct *total* number of nodes in an EKS cluster. My hunch is that the code to check if a node is a spot is flaky and is causing the value to always return as false. GKE and AKS clusters are reporting the correct amount of "premptible" nodes.

**Expected behavior**

I would expect `kubecost_node_is_spot` to return the correct number of nodes that are spot. 

**Screenshots**

In source code, the following value is hard coded to check if a machine is a spot node or not: https://github.com/opencost/opencost/blob/7320861234ce317c99fff8f17b0fb7b0987aad4b/pkg/cloud/awsprovider.go#L49

However, looking at two clusters nodes, I see the label with the following value: 
image
image

From CLI, I can also confirm that the value looks like so: 

```
aws --region us-east-2 --profile workloads-prod eks describe-nodegroup --cluster-name XXXX --nodegroup-name XXXX |egrep capaci
        "capacityType": "SPOT",
```

Latest

Pokom:
Signed-off-by: pokom <mark.poko@grafana.com>

## What does this PR change?
* Detects if an EKS node is a spot type 

* 

* Users will now to be able to accurately calculate costs of workloads in EKS that run on SpotVMS

* closes #1600 

## How was this PR tested?
* Ran locally against multiple EKS clusters and cross referenced with actual data

* N\A

*

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -636,6 +636,9 @@ func (k *awsKey) ID() string {
 	return ""
 }
 
+// Features will return a comma seperated list of features for the given node
+// If the node has a spot label, it will be included in the list
+// Otherwise, the list include instance type, operating system, and the region
 func (k *awsKey) Features() string {
 
 	instanceType, _ := util.GetInstanceType(k.Labels)
@@ -657,8 +660,12 @@ func (k *awsKey) Features() string {
 	return key
 }
 
+// getUsageType returns the usage type of the instance
+// If the instance is a spot instance, it will return PreemptibleType
+// Otherwise returns an empty string
 func (k *awsKey) getUsageType(labels map[string]string) string {
-	if _, ok := labels["eks.amazonaws.com/capacityType"]; ok && labels["eks.amazonaws.com/capacityType"] == "SPOT" {
+	if label, ok := labels["eks.amazonaws.com/capacityType"]; ok && label == "SPOT" {
+		// We currently write out spot instances as "preemptible" in the pricing data, so these need to match
 		return PreemptibleType
 	}
 	return ""
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -638,7 +638,7 @@ func (k *awsKey) Features() string {
 	region, _ := util.GetRegion(k.Labels)
 
 	key := region + "," + instanceType + "," + operatingSystem
-	usageType := PreemptibleType
+	usageType := k.getUsageType(k.Labels)
 	spotKey := key + "," + usageType
 	if l, ok := k.Labels["lifecycle"]; ok && l == "EC2Spot" {
 		return spotKey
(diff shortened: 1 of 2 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -643,17 +643,27 @@ func (k *awsKey) Features() string {
...
 	key := region + "," + instanceType + "," + operatingSystem
-	usageType := PreemptibleType
+	usageType := k.getUsageType(k.Labels)
 	spotKey := key + "," + usageType
 	if l, ok := k.Labels["lifecycle"]; ok && l == "EC2Spot" {
 		return spotKey
 	}
 	if l, ok := k.Labels[k.SpotLabelName]; ok && l == k.SpotLabelValue {
 		return spotKey
 	}
```
Pokom, line 658: @AjayTripathy what's the intention of `k.SpotLabelName` and `k.SpotLabelValue`? When I run locally targeting an EKS cluster, these values are nil and can never enter this. The lines below work, but feels a bit hacky to me.

Pokom:
Current state is that the latest change will nodes as being spot in `kubecost_node_is_spot` metrics, but the cost is not being calculated correctly. I'm tracing the log warning spit out here:

```
 WRN Node i-06882d44775b17397 marked preemptible but we have no data in spot feed
```

This comes from https://github.com/opencost/opencost/blob/00fcbd80d96608606bd64059de91c7fe02c25d93/pkg/cloud/awsprovider.go#L1211, so I need to trace down why https://github.com/opencost/opencost/blob/00fcbd80d96608606bd64059de91c7fe02c25d93/pkg/cloud/awsprovider.go#L1190 does not get evaluated.

Pokom:
Sorry to do this, I'm not sure who the right person is to ask for a review. @AjayTripathy @dwbrown2 when you get an opportunity, can you review this PR? We're currently blocked on expanding usage of opencost metrics until we can get eks nodes marked as spot nodes.

AjayTripathy:
Sorry for the late response here @Pokom ... on my list for the week.

pkg/cloud/awsprovider.go
AjayTripathy: I think this was to support overriding in non EKS implementations of k8s on AWS like kops. Can safely be ignored on EKS, I would think.

Pokom: Got it, thank you! I added an additional check to see if the `usageType` returned is the same as `PreemptibleType` and return the `spotKey` if it is. I'm open to a different way of checking this.

AjayTripathy (approved):

AjayTripathy:
Thanks @Pokom ! Lgtm

PR #1608, opened 2023-01-13
fix GPU counting bug when separate GPU cost is defined but not CPU/RAM

AjayTripathy:
## What does this PR change?
* Fix issue where overall cost is not totalled correctly if line item GPU costs are defined, but only overall RAM/CPU costs

* No

* More accurate GPU pricing

* Closes ...

## How was this PR tested?
* by visiting the /metrics endpoint and noting node_total_hourly_cost

* no

* Yes, let's release in 1.100

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -1044,7 +1044,7 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 
 		if newCnode.GPU != "" && newCnode.GPUCost == "" {
 			// We couldn't find a gpu cost, so fix cpu and ram, then accordingly
-			log.Debugf("GPU without cost found for %s, calculating...", cp.GetKey(nodeLabels, n).Features())
+			log.Infof("GPU without cost found for %s, calculating...", cp.GetKey(nodeLabels, n).Features())
 
 			defaultCPU, err := strconv.ParseFloat(cfg.CPU, 64)
 			if err != nil {
@@ -1180,6 +1180,14 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 				if err != nil {
 					log.Warnf("Could not parse total node price")
 					return nil, err
+				}
+				if newCnode.GPUCost != "" {
+					gpuPrice, err := strconv.ParseFloat(newCnode.GPUCost, 64)
+					if err != nil {
+						log.Warnf("Could not parse node gpu price")
+						return nil, err
+					}
+					nodePrice = nodePrice - gpuPrice // remove the gpuPrice from the total, we're just costing out RAM and CPU.
 				}
 			} else {
 				nodePrice, err = strconv.ParseFloat(newCnode.VCPUCost, 64) // all the price was allocated to the CPU


AjayTripathy:
@nikovacevic quick one for you to review?

nikovacevic:
Haven't we done this before? Let's just be careful. I'm going to look at the git blame, because I swear that something like this we've done, and rolled back, before because it causes issues.

cc @Sean-Holcomb @michaelmdresser @kaelanspatel anyone remember a similar changeset in the past, which didn't work?

nikovacevic:
e.g. https://github.com/opencost/opencost/commit/959772b97f1e58cfaad021360a2415ccea584f34

pkg/costmodel/costmodel.go
```diff
@@ -1181,6 +1181,14 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 					log.Warnf("Could not parse total node price")
 					return nil, err
 				}
+				if newCnode.GPUCost != "" {
+					gpuPrice, err := strconv.ParseFloat(newCnode.GPUCost, 64)
+					if err != nil {
+						log.Warnf("Could not parse node gpu price")
+						return nil, err
+					}
+					nodePrice = nodePrice - gpuPrice // remove the gpuPrice from the total, we're just costing out RAM and CPU.
```
nikovacevic, line 1190: Based on the title of the PR ("GPU cost is defined, but not CPU/RAM) I'm trying to understand the contextual conditional scope of this change.

On line 1045 we start with with:
```go
if newCnode.GPU != "" && newCnode.GPUCost == ""
```

Then on 1135 we have:
```go
else if newCnode.RAMCost == ""
```

We're in that block. So we're really in the conditional scope:
```go
if !(newCnode.GPU != "" && newCnode.GPUCost == "") && newCnode.RAMCost == ""
```

Then within _that_ scope we hit:
```go
if newCnode.Cost != ""
```

So in total we're making this change when:
```go
if !(newCnode.GPU != "" && newCnode.GPUCost == "") && newCnode.RAMCost == "" && newCnode.Cost != ""
```
or equivalently
```go
if (newCnode.GPU == "" || newCnode.GPUCost != "") && newCnode.RAMCost == "" && newCnode.Cost != ""
```

In plain English, the relevant change would seem to happen when "(GPU **isn't** set OR GPUCost **is** set) and RAMCost **isn't** set, but Cost **is** set". So there's a cost, but no RAMCost, and either there is a GPUCost or there's not a GPU. Something like that.

Do I have that right? If so, and that makes sense, then I don't see any mistakes in the code. But the idea is still mostly opaque to me.

Anyways, I can approve this, but I'd be totally leaning on your testing. This is difficult to review.

nikovacevic (approved):

pkg/costmodel/costmodel.go
AjayTripathy: Yeah. Essentially this is saying-- I need to pull GPU costs in from a separate line item GPU and fulfill the given node price with them.

AjayTripathy: The key here is that when total cost is set and RAM cost is not, you need to perform inference. WHen you have GPU you need to take that into account when doing your inference.

PR #1616, opened 2023-01-17
Removed IsThanosEnabled check

jcharcalla:
Removed 'IsThanosEnabled' check as it is not required and possibly causing issues for some users.

Signed-off-by: Jason Charcalla <jason@kubecost.com>

## What does this PR change?
* It removes code that checks if thanos is enabled, and if not sets ETLResolutionSeconds to 1 min.

* [ZD: 3246](https://kubecost.zendesk.com/agent/tickets/3246)

## How was this PR tested?
* It was not tested


jcharcalla:
@AjayTripathy Here is the PR we discussed on todays call.

AjayTripathy:
We still need GetETLResolution, it's just the if statement that needs deletion.

jcharcalla:
Makes sense, I've updated it.

AjayTripathy:
@nikovacevic seems like a quick one, wdyt? I don't know why we ever set this to 1m

nikovacevic (approved):

nikovacevic:
Seems fine to me! Thank you @jcharcalla

PR #1618, opened 2023-01-20
Apply node labels to allocation data

### LINKED ISSUE #1609: Support node labels in Allocation API
Relates to https://github.com/kubecost/cost-analyzer-helm-chart/issues/1656

nikovacevic:
## What does this PR change?
* Applies node labels to allocation data. Note: labels may conflict! The order of precedence is pod labels supersede namespace labels, which supersede node labels. 

* Users will have access to node labels from the Allocation APIs (as long as namespace and pod labels do not overwrite them)

* Closes https://github.com/opencost/opencost/issues/1609
* Relates to https://github.com/kubecost/cost-analyzer-helm-chart/issues/1656

## Does this PR relate to other PRs?
* Ideally requires https://github.com/kubecost/cost-analyzer-helm-chart/pull/1925

* Added "documentation" to the helm chart. Is more needed? Defer to product. I think that is fine?

## How was this PR tested?
Tested on an EKS cluster. Tested three states:
1. default, i.e. enabled with default includeList
2. enabled with custom includeList
3. disabled

### Default (enabled)

Tested without actively setting `enabled` to `true` and verified that the feature works, with the default labels.

```yaml
kubecostModel:
  allocation:
    # nodeLabels:
    #   enabled: true
    #   includeList: "node.kubernetes.io/instance-type,topology.kubernetes.io/region,topology.kubernetes.io/zone"
```

```
INF [node-labels] using defaults
INF [node-labels] successfully applying node_kubernetes_io_instance_type=m5.large for node aws-dev-1-niko/ip-192-168-49-107.us-east-2.compute.internal
INF [node-labels] successfully applying topology_kubernetes_io_region=us-east-2 for node aws-dev-1-niko/ip-192-168-49-107.us-east-2.compute.internal
INF [node-labels] successfully applying topology_kubernetes_io_zone=us-east-2b for node aws-dev-1-niko/ip-192-168-49-107.us-east-2.compute.internal
INF [node-labels] using defaults
INF [node-labels] successfully applying node_kubernetes_io_instance_type=m5.large for node aws-dev-1-niko/ip-192-168-7-143.us-east-2.compute.internal
INF [node-labels] successfully applying topology_kubernetes_io_region=us-east-2 for node aws-dev-1-niko/ip-192-168-7-143.us-east-2.compute.internal
INF [node-labels] successfully applying topology_kubernetes_io_zone=us-east-2a for node aws-dev-1-niko/ip-192-168-7-143.us-east-2.compute.internal
INF [node-labels] using defaults
INF [node-labels] successfully applying node_kubernetes_io_instance_type=m5.large for node aws-dev-1-niko/ip-192-168-92-251.us-east-2.compute.internal
INF [node-labels] successfully applying topology_kubernetes_io_region=us-east-2 for node aws-dev-1-niko/ip-192-168-92-251.us-east-2.compute.internal
INF [node-labels] successfully applying topology_kubernetes_io_zone=us-east-2c for node aws-dev-1-niko/ip-192-168-92-251.us-east-2.compute.internal
```

Screenshot from 2023-01-26 17-10-05

### Enabled, custom include list

```yaml
kubecostModel:
  allocation:
    nodeLabels:
      enabled: true
      includeList: "node.kubernetes.io/instance-type"
```

```
INF [node-labels] successfully applying node_kubernetes_io_instance_type=m5.large for node aws-dev-1-niko/ip-192-168-7-143.us-east-2.compute.internal
INF [node-labels] successfully applying node_kubernetes_io_instance_type=m5.large for node aws-dev-1-niko/ip-192-168-92-251.us-east-2.compute.internal
INF [node-labels] successfully applying node_kubernetes_io_instance_type=m5.large for node aws-dev-1-niko/ip-192-168-3-17.us-east-2.compute.internal
```

Screenshot from 2023-01-26 17-24-31

### Disabled

```yaml
kubecostModel:
  allocation:
    nodeLabels:
      enabled: false
```

```
INF [node-labels] allocation node labels disabled
INF [node-labels] allocation node labels disabled
INF [node-labels] allocation node labels disabled
```

Screenshot from 2023-01-26 17-15-10

### DIFF
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -92,6 +92,9 @@ const (
 	IngestPodUIDEnvVar = "INGEST_POD_UID"
 
 	ETLReadOnlyMode = "ETL_READ_ONLY"
+
+	AllocationNodeLabelsEnabled     = "ALLOCATION_NODE_LABELS_ENABLED"
+	AllocationNodeLabelsIncludeList = "ALLOCATION_NODE_LABELS_INCLUDE_LIST"
 )
 
 var offsetRegex = regexp.MustCompile(`^(\+|-)(\d\d):(\d\d)$`)
@@ -502,4 +505,33 @@ func GetPromClusterLabel() string {
 // contents of podKeys in Allocation
 func IsIngestingPodUID() bool {
 	return GetBool(IngestPodUIDEnvVar, false)
+}
+
+func GetAllocationNodeLabelsEnabled() bool {
+	return GetBool(AllocationNodeLabelsEnabled, true)
+}
+
+var defaultAllocationNodeLabelsIncludeList []string = []string{
+	"cloud.google.com/gke-nodepool",
+	"eks.amazonaws.com/nodegroup",
+	"kubernetes.azure.com/agentpool",
+	"node.kubernetes.io/instance-type",
+	"topology.kubernetes.io/region",
+	"topology.kubernetes.io/zone",
+}
+
+func GetAllocationNodeLabelsIncludeList() []string {
+	// If node labels are not enabled, return an empty list.
+	if !GetAllocationNodeLabelsEnabled() {
+		return []string{}
+	}
+
+	list := GetList(AllocationNodeLabelsIncludeList, ",")
+
+	// If node labels are enabled, but the white list is empty, use defaults.
+	if len(list) == 0 {
+		return defaultAllocationNodeLabelsIncludeList
+	}
+
+	return list
 }
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -43,6 +43,7 @@ const (
 	queryFmtNetInternetCostPerGiB    = `avg(avg_over_time(kubecost_network_internet_egress_cost{}[%s])) by (%s)`
 	queryFmtNetReceiveBytes          = `sum(increase(container_network_receive_bytes_total{pod!="", container="POD"}[%s])) by (pod_name, pod, namespace, %s)`
 	queryFmtNetTransferBytes         = `sum(increase(container_network_transmit_bytes_total{pod!="", container="POD"}[%s])) by (pod_name, pod, namespace, %s)`
+	queryFmtNodeLabels               = `avg_over_time(kube_node_labels[%s])`
 	queryFmtNamespaceLabels          = `avg_over_time(kube_namespace_labels[%s])`
 	queryFmtNamespaceAnnotations     = `avg_over_time(kube_namespace_annotations[%s])`
 	queryFmtPodLabels                = `avg_over_time(kube_pod_labels[%s])`
@@ -393,6 +394,12 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
 	queryNetInternetCostPerGiB := fmt.Sprintf(queryFmtNetInternetCostPerGiB, durStr, env.GetPromClusterLabel())
 	resChNetInternetCostPerGiB := ctx.QueryAtTime(queryNetInternetCostPerGiB, end)
 
+	var resChNodeLabels prom.QueryResultsChan
+	if env.GetAllocationNodeLabelsEnabled() {
+		queryNodeLabels := fmt.Sprintf(queryFmtNodeLabels, durStr)
+		resChNodeLabels = ctx.QueryAtTime(queryNodeLabels, end)
+	}
+
 	queryNamespaceLabels := fmt.Sprintf(queryFmtNamespaceLabels, durStr)
 	resChNamespaceLabels := ctx.QueryAtTime(queryNamespaceLabels, end)
 
@@ -465,6 +472,12 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
 	resNetInternetGiB, _ := resChNetInternetGiB.Await()
 	resNetInternetCostPerGiB, _ := resChNetInternetCostPerGiB.Await()
 
+	var resNodeLabels []*prom.QueryResult
+	if env.GetAllocationNodeLabelsEnabled() {
+		if env.GetAllocationNodeLabelsEnabled() {
+			resNodeLabels, _ = resChNodeLabels.Await()
+		}
+	}
 	resNamespaceLabels, _ := resChNamespaceLabels.Await()
 	resNamespaceAnnotations, _ := resChNamespaceAnnotations.Await()
 	resPodLabels, _ := resChPodLabels.Await()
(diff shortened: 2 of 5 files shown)


dwbrown2:
@nikovacevic amazing to see this! Thinking 1.100 is feasible?

Are there label collisions here that could be problematic? @AjayTripathy thought a fair amount about this. I was originally picturing `node_label` as being different than workload `labels`...

nikovacevic:
@dwbrown2 yes definitely feasible. It's working now -- just need to be scale tested! I don't expect any nasty surprises.

@AjayTripathy do you have any context you want to share?

My thoughts on this design is that it's nice because it integrates seamlessly into existing label functionality. If we make a new field, it's kind of a pain in the butt because that takes effort... and then you have to go around everywhere that labels exist and add code to support the new field. BE, FE, everywhere.

There is a risk of label collisions, yes, but it seems like node labels have very specific naming that means the important ones are unlikely to be overwritten by a namespace label or a pod label. Any problematic ones of which you're aware?

AjayTripathy:
@nikovacevic I'm less concerned about label collisions and I guess more about data scale/keeping Summary fast and light. Can we:
1. Allow users to turn this on/off? For users with 1m+ pods this is a significant amount of data, right?
2. Put this in another field because that means it's easy to disambiguate and turn this off in Summary data/have more control over it. WDYT?

nikovacevic:
@AjayTripathy I always try to avoid even more helm flags, but if we must, we must. In terms of making a new field, we can of course do that. But then we have to make changes across the entire application for logic like: aggregating/filtering/etc. by a label? Okay, do you mean a label label, or a node label? Or maybe we make the "fallthrough" happen at query time, but that also takes work and testing and possibly creates problems. (Fallthrough-style filtering is notoriously problematic.)

So, in short:  this is very clean and compatible. If we go the other route, then it'll be at least a week of work on product, FE, and BE to figure out what the new behavior should be and to thread everything through the application and test that things are working as expected.. No long v1.100, but can be v1.101 if we're willing to commit the resources.

cc @Adam-Stack-PM for tracking, and still looking for feedback from product @dwbrown2 @kwombach12

AjayTripathy:
I think that's a fair point. I'm supportive of not making a new field, I'm just hesitant about adding 1m pods x 50 labels without the ability to turn them off. That seem fair? I know it sucks to add yet another helm value to turn on. What do you think about _just_ pulling through known labels like region and instancetype, then prefixing them with something like "node" so we can disambiguate? I would just ask the question what is the product value of getting EVERY node label. I think I can get behind one or two new labels x 1million pods but not an unbounded set of all labels.

nikovacevic:
Discussed out of band with @AjayTripathy. Plan is to create a helm white list for node labels, with sensible defaults (e.g. `topology.kubernetes.io/region`, `topology.kubernetes.io/zone`, etc... see https://kubernetes.io/docs/reference/labels-annotations-taints/) and only ingest those labels here. (All node labels will still be available in the Assets API.)

@srpomeroy @jessegoodier are there any specific labels you want to see here?

kwombach12:
@nikovacevic I think this is a really impactful feature. I will add this as a proposal for v1.101

nikovacevic:
@kwombach12 great! The work is practically done. If I can get a list of metrics, it can absolutely go into v1.100.

kwombach12:
Outside of region & zone, I think `node.kubernetes.io/instance-type` would be useful. @dwbrown2 any others you think are worth adding

nikovacevic:
Can add to the list here: https://github.com/kubecost/cost-analyzer-helm-chart/pull/1925/files

dwbrown2:
Sweet, nothing other stand out to me... you @AjayTripathy?

pkg/env/costmodelenv.go
```diff
@@ -509,3 +511,7 @@ func GetPromClusterLabel() string {
 func IsIngestingPodUID() bool {
 	return GetBool(IngestPodUIDEnvVar, false)
 }
+
+func GetAllocationNodeLabelWhiteList() []string {
+	return GetList(AllocationNodeLabelWhiteList, ",")
```
AjayTripathy: Can you copy the helm defaults here? Not every user uses helm to deploy and this helps standardize defaults in the off chance users use a legacy helm chart to deploy a new KC version.

AjayTripathy (approved):

AjayTripathy:
LGTM otherwise!

jessegoodier:
> Discussed out of band with @AjayTripathy. Plan is to create a helm white list for node labels, with sensible defaults (e.g. `topology.kubernetes.io/region`, `topology.kubernetes.io/zone`, etc... see https://kubernetes.io/docs/reference/labels-annotations-taints/) and only ingest those labels here. (All node labels will still be available in the Assets API.)
> 
> @srpomeroy @jessegoodier are there any specific labels you want to see here?

Common labels in the providers that have been requested (which may be good allow-list defaults?):
```
kubernetes.azure.com/agentpool
cloud.google.com/gke-nodepool
eks.amazonaws.com/nodegroup
```

and these more unique ones (perhaps not default?):
```
nodeGroup
nodeType
```

pkg/env/costmodelenv.go
nikovacevic: We can do that, yeah.

I was thinking that if the list is empty, that would indicate that you wanted this feature to be completely turned off. Instead do we need an `enabled` flag? And if enabled is `true`, but the list is empty, we default?

AjayTripathy: I think it's safe to turn this feature on for everyone...do we need enabled?

AjayTripathy:
Love it @jessegoodier ! Thanks for contributing. I've heard the nodepool request before enough to the point that we may even want a default filter option there!

srpomeroy:
I love the idea of nodepool/group! Potential separate FR to alias nodepool/group to its own dedication aggregation like we do with team/owner/department/product/environment labels.

nikovacevic:
Great. Thanks, everyone. This is ready for final review.

@srpomeroy @jessegoodier @AjayTripathy on labels for the default include-list we landed on:
```
"cloud.google.com/gke-nodepool",
"eks.amazonaws.com/nodegroup",
"kubernetes.azure.com/agentpool",
"node.kubernetes.io/instance-type",
"topology.kubernetes.io/region",
"topology.kubernetes.io/zone",
```

Last call (for now, at least)... anything else to add?

pkg/env/costmodelenv.go
nikovacevic: If you want empty list to mean "use defaults" and not "don't ingest any node labels" I think you need `enabled`. It looks fine to me that way! (See https://github.com/kubecost/cost-analyzer-helm-chart/pull/1925)

PR #1620, opened 2023-01-21
add url and description for providers

nickcurie:
## What does this PR change?
* Adds URL to AWS, GCP orphaned resources API

* Fixes https://github.com/opencost/opencost/pull/1597

* Users can see the resource in their providers console easier

* No

## How was this PR tested?
* Tested on GCP and AWS dev clusters

* 

* v1.100

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -1741,9 +1741,10 @@ func (aws *AWS) GetOrphanedResources() ([]OrphanedResource, error) {
 					continue
 				}
 				if tag.Value == nil {
-					continue
+					desc[*tag.Key] = ""
+				} else {
+					desc[*tag.Key] = *tag.Value
 				}
-				desc[*tag.Key] = *tag.Value
 			}
 
 			or := OrphanedResource{
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -1706,7 +1706,10 @@ func (aws *AWS) GetOrphanedResources() ([]OrphanedResource, error) {
 			}
 
 			// This is turning us-east-1a into us-east-1
-			zone := *volume.AvailabilityZone
+			var zone string
+			if volume.AvailabilityZone != nil {
+				zone = *volume.AvailabilityZone
+			}
 			var region, url string
 			region = regionRx.FindString(zone)
 			if region != "" {
@@ -1734,6 +1737,12 @@ func (aws *AWS) GetOrphanedResources() ([]OrphanedResource, error) {
 
 			desc := map[string]string{}
 			for _, tag := range address.Tags {
+				if tag.Key == nil {
+					continue
+				}
+				if tag.Value == nil {
+					continue
+				}
 				desc[*tag.Key] = *tag.Value
 			}
 
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -65,6 +65,7 @@ var (
 	provIdRx      = regexp.MustCompile("aws:///([^/]+)/([^/]+)")
 	usageTypeRegx = regexp.MustCompile(".*(-|^)(EBS.+)")
 	versionRx     = regexp.MustCompile("^#Version: (\\d+)\\.\\d+$")
+	regionRx      = regexp.MustCompile("([a-z]+-[a-z]+-[0-9])")
 )
 
 func (aws *AWS) PricingSourceStatus() map[string]*PricingSource {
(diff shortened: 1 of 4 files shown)


pkg/cloud/awsprovider.go
```diff
@@ -1704,11 +1705,22 @@ func (aws *AWS) GetOrphanedResources() ([]OrphanedResource, error) {
 				volumeSize = int64(*volume.Size)
 			}
 
+			// This is turning us-east-1a into us-east-1
+			zone := *volume.AvailabilityZone
```
nikovacevic: Any chance of this being a `nil` dereference? Might be good to check.

pkg/cloud/awsprovider.go
```diff
@@ -1720,9 +1732,16 @@ func (aws *AWS) GetOrphanedResources() ([]OrphanedResource, error) {
 		if aws.isAddressOrphaned(address) {
 			cost := AWSHourlyPublicIPCost * timeutil.HoursPerMonth
 
+			desc := map[string]string{}
+			for _, tag := range address.Tags {
+				desc[*tag.Key] = *tag.Value
```
michaelmdresser: Please add `nil` checks.

pkg/cloud/azureprovider.go
```diff
@@ -1282,13 +1282,15 @@ func (az *Azure) GetOrphanedResources() ([]OrphanedResource, error) {
 				diskSize = int64(*d.DiskSizeGB)
 			}
 
+			desc := map[string]string{}
+			for k, v := range d.Tags {
+				desc[k] = *v
```
michaelmdresser: Please add a `nil` check.

pkg/cloud/gcpprovider.go
```diff
@@ -468,12 +468,20 @@ func (gcp *GCP) GetOrphanedResources() ([]OrphanedResource, error) {
 					return nil, err
 				}
 
+				// GCP gives us description as a string formatted as a map[string]string, so we need to
+				// deconstruct it back into a map[string]string to match the OR struct
+				desc := map[string]string{}
+				if err := json.Unmarshal([]byte(disk.Description), &desc); err != nil {
+					return nil, err
```
michaelmdresser: Please wrap this error

nikovacevic (changes requested):
Some extra nil checking and error wrapping, then we should be good to go!

nickcurie:
Thanks y'all. Should be resolved now

pkg/cloud/awsprovider.go
```diff
@@ -1720,9 +1735,22 @@ func (aws *AWS) GetOrphanedResources() ([]OrphanedResource, error) {
...
 			cost := AWSHourlyPublicIPCost * timeutil.HoursPerMonth
 
+			desc := map[string]string{}
+			for _, tag := range address.Tags {
+				if tag.Key == nil {
+					continue
+				}
+				if tag.Value == nil {
+					continue
+				}
```
michaelmdresser: Is there a valid case where `tag.Key != nil` but `tag.Value == nil`?

pkg/cloud/azureprovider.go
```diff
@@ -1282,13 +1282,18 @@ func (az *Azure) GetOrphanedResources() ([]OrphanedResource, error) {
 				diskSize = int64(*d.DiskSizeGB)
 			}
 
+			desc := map[string]string{}
+			for k, v := range d.Tags {
+				if v == nil {
+					continue
+				}
```
michaelmdresser: Similar question if `v == nil` is actually a valid case, just a value-less tag

pkg/cloud/awsprovider.go
nickcurie: I'm not sure. The tags are coming directly from AWS's API so its entirely possible. Sounds like it could be worth doing `desc[*tag.Key] = ""` if value is empty

michaelmdresser: I think we should roll with that idea!

michaelmdresser (approved):

pkg/cloud/awsprovider.go
```diff
@@ -1720,9 +1735,23 @@ func (aws *AWS) GetOrphanedResources() ([]OrphanedResource, error) {
...
+			desc := map[string]string{}
+			for _, tag := range address.Tags {
+				if tag.Key == nil {
+					continue
+				}
+				if tag.Value == nil {
+					desc[*tag.Key] = ""
+				} else {
+					desc[*tag.Key] = *tag.Value
+				}
```
nikovacevic, line 1747: This will work, but I think it's also fine and simpler to just do:
```go
if tag.Key != nil && tag.Value != nil {
    desc[*tag.Key] = *tag.Value
}
```

I.e. no need to record an empty tag, right?

pkg/cloud/azureprovider.go
```diff
@@ -1282,13 +1282,19 @@ func (az *Azure) GetOrphanedResources() ([]OrphanedResource, error) {
 				diskSize = int64(*d.DiskSizeGB)
 			}
 
+			desc := map[string]string{}
+			for k, v := range d.Tags {
+				if v == nil {
+					desc[k] = ""
+				} else {
+					desc[k] = *v
+				}
```
nikovacevic, line 1291: Ditto

nikovacevic (commented):
Just a small final change there. Looking much better, much safer! Thank you!

pkg/cloud/awsprovider.go
nickcurie: So we just ignore the case where `tag.Key != nil` but `tag.Value == nil`?

nikovacevic (approved):

pkg/cloud/awsprovider.go
nikovacevic: Yeah, disregard. This will work just fine. I don't expect there to be many (any?) instances where we one is nil but the other isn't.

PR #1625, opened 2023-01-23
Update field names for cloud cost

Sean-Holcomb:
## What does this PR change?
Update to CloudCostItems and CloudCostAggregate Properties replacing Account and ProjectID with WorkGroupID and BillingID. This change comes as a result of a mismatch in terminology between providers, and to attempts to better capture a useful abstraction that cuts across cloud providers. That is a larger parent account from which billing is paid (GCP: billing_account_id, AWS: bill_payer_account_id, Azure: billingAccountId) and then a sub-account from which resources are run (GCP: project.id, AWS: line_item_usage_account_id, AZ: SubscriptionId). The new names were chosen with the goal of overloading terms in a multi-cloud context.

This change is being put into place as an renaming of fields within a existing bingen schema rather than changing the actual fields with Account > WorkGroupID, ProjectID > BillingID and Credit > NetCost. The reason this will be possible is because only AWS is live in 1.99. And AWS did not use the ProjectID or Credit field.

* https://github.com/kubecost/kubecost-cost-model/pull/1167

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/filter/util/cloudcostaggregate.go
+++ b/pkg/filter/util/cloudcostaggregate.go
@@ -17,12 +17,12 @@ func CloudCostAggregateFilterFromParams(pmr mapper.PrimitiveMapReader) filter.Fi
 		Filters: []filter.Filter[*kubecost.CloudCostAggregate]{},
 	}
 
-	if raw := pmr.GetList("filterAccounts", ","); len(raw) > 0 {
-		filter.Filters = append(filter.Filters, filterV1SingleValueFromList(raw, kubecost.CloudCostAccountProp))
+	if raw := pmr.GetList("filterBillingIDs", ","); len(raw) > 0 {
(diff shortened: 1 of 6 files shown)


nikovacevic:
How is it possible to map `Credit` > `NetCost`? Will that cause trouble in historic GCP data? (Isn't Net Cost the total with the discounts/credits baked in?)

mbolt35:
With respect to bingen, these changes look OK to me. However, be very careful with the changes to the properties map. You will likely want to check the old properties as well as the new properties to ensure backwards compatibility.

You can put in a pre-processor that checks for the old properties and migrates them to the new (that's sort of what is was made for).

nikovacevic (approved):

Sean-Holcomb:
closing in favor of https://github.com/opencost/opencost/pull/1632

PR #1626, opened 2023-01-25
remove label= from label values for allocations

nickcurie:
Signed-off-by: nickcurie <nick.curie64@gmail.com>

## What does this PR change?
* Remove `label=` from label values in allocations

* No

* When joining cloud costs and external allocations by label, there will no longer be a mismatch between labels
* Before: 
adv-reports-before
* After:
adv-reports-after

* Closes https://github.com/kubecost/kubecost-cost-model/issues/1163

## How was this PR tested?
* Tested on AWS dev-1

* 

* v1.100

### DIFF
--- a/pkg/kubecost/allocationprops.go
+++ b/pkg/kubecost/allocationprops.go
@@ -298,7 +298,7 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 			} else {
 				labelName := labelConfig.Sanitize(strings.TrimPrefix(agg, "label:"))
 				if labelValue, ok := labels[labelName]; ok {
-					names = append(names, fmt.Sprintf("%s=%s", labelName, labelValue))
+					names = append(names, fmt.Sprintf("%s", labelValue))
 				} else {
 					names = append(names, UnallocatedSuffix)
 				}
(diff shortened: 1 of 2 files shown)


pkg/kubecost/allocationprops.go
```diff
@@ -298,7 +298,7 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 			} else {
 				labelName := labelConfig.Sanitize(strings.TrimPrefix(agg, "label:"))
 				if labelValue, ok := labels[labelName]; ok {
-					names = append(names, fmt.Sprintf("%s=%s", labelName, labelValue))
+					names = append(names, fmt.Sprintf("%s", labelValue))
```
nickcurie, line 301: Do we want to do this same thing for annotations?

michaelmdresser: We've been treating labels and annotations as the "same" thing for a little while now, so I think the answer here is yes. @nikovacevic?

nikovacevic: Yeah, let's do it for annotations, too.

nikovacevic:
Love it!! The one-liner! (Well, two after we fix annotations, too.)

nikovacevic:
Unfortunately... you're going to have to change a bunch of tests. Are you seeing the failing unit tests @nickcurie?

You can test these locally and fix up the naming until tests pass again. Not a fun task, I'm sorry to say, but we'll need to do that before this merges.

nikovacevic (approved):
Fantastic! Thank you, Nick.

PR #1630, opened 2023-01-27
Add support for prefix parameter in bucket config

### LINKED ISSUE #1758: Allow use of `config.prefix` in `kubecostModel.etlBucketConfigSecret` secret by bumping to Thanos 0.27

I am trying to have multiple clusters backup ETL data to the same S3 bucket using `kubecostModel.etlBucketConfigSecret`. However when using a secret like this:

```yaml
apiVersion: v1
kind: Secret
type: Opaque
stringData:
  object-store.yaml: |
    type: S3
    config:
      bucket: "my-bucket"
      prefix: "my-prefix"
      endpoint: "s3.amazonaws.com"
      region: "us-east-1"
      insecure: false
      signature_version2: false
      put_user_metadata:
        "X-Amz-Acl": "bucket-owner-full-control"
```

I get the following warning:

```log
0000-00-00T22:57:58.831642422Z WRN ETL: failed to create bucket storage: create S3 client: yaml: unmarshal errors:
  line 4: field prefix not found in type storage.S3Config
```

However if we look at similar documenation to the [doc referenced](https://guide.kubecost.com/hc/en-us/articles/4407601811095-ETL-Backup#:~:text=This%20configuration%20secret%20follows%20the%20same%20layout%20documented%20for%20Thanos%20here%3A%20https%3A//thanos.io/v0.21/thanos/storage.md), in these [similar docs](https://github.com/thanos-io/objstore#s3:~:text=The%20field%20prefix,which%20blocks%20came) we can see that a `config.prefix` key is mentioned. (NOTE: I think the documentation actually has an error and the `prefix` key should be under the `config` key.)

Extend the unmarshalling functionality for the `storage.S3Config` type to be able to handle the `config.prefix` key. This may be as simple as just updating a dependency to a newer version (not sure).

None.

Users would be able to include a `config.prefix` key in their `kubecostModel.etlBucketConfigSecret` secret to be able to select a custom prefix inside of a custom bucket.

kaelanspatel:
## What does this PR change?
Adds a wrapper for bucket storage which appends a prefix to each path, allowing for different clusters to back up into the same bucket on different paths.

This behavior already exists in Thanos, so this, like our S3 storage, partially derives from that code.

No.

Allows for ETL backup to utilize `prefix` in config, as described [here](https://thanos.io/tip/thanos/storage.md/#:~:text=The%20field%20prefix,which%20blocks%20came).

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1758

## How was this PR tested?
Tested with and without prefix for ETL backup
Tested using internal validation functionality e.g. test file write, read, delete, exists etc.
Tested using manual validation e.g. after storage creation test all functions for proper path behavior

Yes, probably need to mention this under https://docs.kubecost.com/install-and-configure/install/etl-backup.

Yes.

### DIFF
--- a/pkg/storage/bucketstorage.go
+++ b/pkg/storage/bucketstorage.go
@@ -22,6 +22,7 @@ const (
 type StorageConfig struct {
 	Type   StorageProvider `yaml:"type"`
 	Config interface{}     `yaml:"config"`
+	Prefix string          `yaml:"prefix"`
 }
 
 // NewBucketStorage initializes and returns new Storage implementation leveraging the storage provider
@@ -52,6 +53,10 @@ func NewBucketStorage(config []byte) (Storage, error) {
 	}
 	if err != nil {
 		return nil, errors.Wrap(err, fmt.Sprintf("create %s client", storageConfig.Type))
+	}
+
+	if storageConfig.Prefix != "" {
+		return NewPrefixedBucketStorage(storage, storageConfig.Prefix)
 	}
 
 	return storage, nil
--- a/pkg/storage/prefixedbucketstorage.go
+++ b/pkg/storage/prefixedbucketstorage.go
@@ -1,0 +1,88 @@
+package storage
+
+// Partially forked from Thanos prefixed bucket for prefix handling
+// Licensed under the Apache License 2.0
+// https://github.com/thanos-io/objstore/blob/main/prefixed_bucket.go
+
+import (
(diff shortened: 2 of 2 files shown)


AjayTripathy:
This LGTM. Can you make the necessary docs update?

AjayTripathy (approved):

PR #1631, opened 2023-01-27
Bolt/command customization

mbolt35:
## What does this PR change?
* Addition to opencost argument processing to allow for the creation of custom commands. 

* No change

### DIFF
--- a/pkg/costmodel/clusters/clustermap.go
+++ b/pkg/costmodel/clusters/clustermap.go
@@ -22,6 +22,25 @@ const (
 	LoadRetryDelay time.Duration = 10 * time.Second
 )
 
+// The following constants are used as keys into the cluster info map data structure
+const (
+	ClusterInfoIdKey               = "id"
+	ClusterInfoNameKey             = "name"
+	ClusterInfoProviderKey         = "provider"
+	ClusterInfoProjectKey          = "project"
+	ClusterInfoAccountKey          = "account"
+	ClusterInfoRegionKey           = "region"
+	ClusterInfoProvisionerKey      = "provisioner"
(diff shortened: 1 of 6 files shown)


pkg/costmodel/clusters/clustermap.go
```diff
@@ -22,6 +22,25 @@ const (
 	LoadRetryDelay time.Duration = 10 * time.Second
 )
 
+// The following constants are used as keys into the cluster info map data structure
+const (
```
michaelmdresser, line 26: :heart:

pkg/costmodel/clusters/clustermap.go
```diff
@@ -391,3 +355,63 @@ func (pcm *PrometheusClusterMap) StopRefresh() {
...
+	if i, ok := info[ClusterInfoIdKey]; ok {
+		id = i
+	} else {
+		return nil, fmt.Errorf("Cluster Info Missing ID")
+	}
+	if n, ok := info[ClusterInfoNameKey]; ok {
+		name = n
+	} else {
+		name = id
+	}
```
michaelmdresser, line 374: This case used to be an error, what's the reason for the change?

mbolt35: Poor logic. If we have the id, the name should always inherit the id as the name. Additionally, as we expand the definition of ClusterInfo beyond the scope of just prom, being more precise here is important. Name is not a required cluster info configuration, therefore, we shouldn't error when it's missing.

michaelmdresser: Makes perfect sense to me!

michaelmdresser (approved):

PR #1632, opened 2023-01-27
update cloud cost structs

Sean-Holcomb:
This PR replaces: https://github.com/opencost/opencost/pull/1625
## What does this PR change?
Update to CloudCostItems and CloudCostAggregate Properties replacing Account and ProjectID with WorkGroupID and BillingID. This change comes as a result of a mismatch in terminology between providers, and to attempts to better capture a useful abstraction that cuts across cloud providers. That is a larger parent account from which billing is paid (GCP: billing_account_id, AWS: bill_payer_account_id, Azure: billingAccountId) and then a sub-account from which resources are run (GCP: project.id, AWS: line_item_usage_account_id, AZ: SubscriptionId). The new names were chosen with the goal of overloading terms in a multi-cloud context.

This change is being put into place as an renaming of fields within a existing bingen schema rather than changing the actual fields with Account > WorkGroupID, ProjectID > BillingID and Credit > NetCost. The reason this will be possible is because only AWS is live in 1.99. And AWS did not use the ProjectID or Credit field.

Add CloudCost Range helper functions

* 

* 

* Closes ...

## How was this PR tested?
* Unit test and and manual testing

* 

*

### DIFF
--- a/pkg/kubecost/cloudcostaggregate.go
+++ b/pkg/kubecost/cloudcostaggregate.go
@@ -11,33 +11,33 @@ import (
 )
 
 const (
-	CloudCostAccountProp  string = "account"
-	CloudCostProjectProp  string = "project"
-	CloudCostProviderProp string = "provider"
-	CloudCostServiceProp  string = "service"
-	CloudCostLabelProp    string = "label"
+	CloudCostBillingIDProp   string = "billingID"
+	CloudCostWorkGroupIDProp string = "workGroupID"
+	CloudCostProviderProp    string = "provider"
+	CloudCostServiceProp     string = "service"
+	CloudCostLabelProp       string = "label"
 )
 
 // CloudCostAggregateProperties unique property set for CloudCostAggregate within a window
 type CloudCostAggregateProperties struct {
-	Provider   string `json:"provider"`
-	Account    string `json:"account"`
-	Project    string `json:"project"`
-	Service    string `json:"service"`
-	LabelValue string `json:"label"`
+	Provider    string `json:"provider"`
+	WorkGroupID string `json:"workGroupID"`
+	BillingID   string `json:"billingID"`
+	Service     string `json:"service"`
+	LabelValue  string `json:"label"`
 }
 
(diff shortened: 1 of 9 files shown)


pkg/kubecost/cloudcostaggregate.go
```diff
@@ -351,51 +351,92 @@ func (ccas *CloudCostAggregateSet) Merge(that *CloudCostAggregateSet) (*CloudCos
...
+func NewCloudCostAggregateSetRange(start, end time.Time, window time.Duration, integration string, labelName string) (*CloudCostAggregateSetRange, error) {
+	windows, err := GetWindows(start, end, window)
 	if err != nil {
 		return nil, err
 	}
 
 	// Build slice of CloudCostAggregateSet to cover the range
-	CloudCostAggregateSets := []*CloudCostAggregateSet{}
-	for _, w := range windows {
+	cloudCostAggregateSets := make([]*CloudCostAggregateSet, len(windows))
```
nikovacevic, line 368: Yes, nice.

pkg/kubecost/cloudcostaggregate.go
```diff
@@ -351,51 +351,92 @@ func (ccas *CloudCostAggregateSet) Merge(that *CloudCostAggregateSet) (*CloudCos
...
-			Cost:              cost * pct,
-			Credit:            credit * pct,
+	for _, ccas := range ccasr.CloudCostAggregateSets {
+		pct := ccas.GetWindow().GetPercentInWindow(window)
+		if pct == 0 {
+			continue
+		}
+		cca := cloudCostAggregate
+		// If the current set Window only contains a portion of the CloudCostItem Window, insert costs relative to that portion
+		cca = &CloudCostAggregate{
```
nikovacevic: What's going on here? Am I misreading, or are we assigning and then immediately reassigning?

nikovacevic (approved):
Just that one line looks sus to me. Otherwise LGTM.

pkg/kubecost/cloudcostaggregate.go
Sean-Holcomb: missing `if pct < 1.0` there

PR #1629, opened 2023-01-26
Disable kube_pod_owner when emitting only KSM v1 metrics

kaelanspatel:
## What does this PR change?
Minimal change to solve https://github.com/opencost/opencost/issues/1571. Stops Kubecost from emitting the `kube_pod_owner` duplicated KSM metric when `EMIT_KSM_V1_METRICS_ONLY` is true.

As we recommend users with KSM v2.X to enable `EMIT_KSM_V1_METRICS_ONLY` and disable `EMIT_KSM_V1_METRICS`, we expect the minimal set of metrics not present in KSM v2 that Kubecost requires to be emitted. As `kube_pod_owner` exists in KSM v2, this PR allows the recommended behavior without that metric being duplicated. 

## Does this PR relate to any other PRs?
No.

## How will this PR impact users?
* Kubecost will no longer emit `kube_pod_owner` when EMIT_KSM_V1_METRICS_ONLY` is true.

## Does this PR address any GitHub or Zendesk issues?
* https://github.com/opencost/opencost/issues/1571

## How was this PR tested?
Tested manually by observing proper emission behavior.

## Does this PR require changes to documentation?
No.

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next Opencost release? If not, why not?
Yes.

Issue #1571, opened 2022-12-14
kube_pod_owner appears to be duplicated if EMIT_KSM_V1_METRICS to false and EMIT_KSM_V1_METRICS_ONLY to true. 


AjayTripathy:
**Describe the bug** 
kube_pod_owner appears to be duplicated if EMIT_KSM_V1_METRICS to false and EMIT_KSM_V1_METRICS_ONLY to true. 

**To Reproduce**
Set the above flags and see if multiple metrics are coming back

**Expected behavior**
EMIT_KSM_V1_METRICS to false and EMIT_KSM_V1_METRICS_ONLY to true should be a minimal set of metrics

**Which version of OpenCost are you using?**
1.99+

**Additional context**
https://github.com/opencost/opencost/issues/1509

robfaraj:
@AjayTripathy @teevans curious if this has been prioritized. have a couple folks asking about it

teevans:
@nikovacevic  - Would someone on your team have some time to look into this?

nikovacevic:
@teevans I'm probably the only one left to look into such things. Putting it on the list to investigate.

AjayTripathy:
We should probably make a call about migration to KSMv2 or at least better interoperability with it, but it's a fairly large effort since we'd have to migrate a bunch of metrics and that can be hard for historical data (if done carelessly you will lose the ability to do a rebuild/merge over KSM v1 metrics.

kaelanspatel:
Hi all, the above PR disables emission of `kube_pod_owner` when `EMIT_KSM_V1_METRICS_ONLY` is true. This should solve the specific problem this issue describes. 

I note, however, that the config we recommend for users running KSM v2 (EMIT_KSM_V1_METRICS_ONLY = true + EMIT_KSM_V1_METRICS = false) still emits several metrics which would be duplicated, as they are emitted by any KSM v2:
```
kube_pod_owner (as addressed in the above PR)
kube_pod_label
kube_namespace_labels
kube_node_labels
kube_node_status_condition
```

This seems counter to the described behavior of the config in the helm chart:
```
emitKsmV1MetricsOnly: false # emit only the KSM metrics missing from KSM v2. Advanced users only.
```
Is there a reason I'm missing for this? If not, IMO we should stop these from being emitted under the `EMIT_KSM_V1_METRICS_ONLY` so that behavior is consistent with what we describe.

cc @AjayTripathy

AjayTripathy:
[...]

Issue #1509, opened 2022-11-28
Remove kube-state-metrics related metrics


dudicoco:
**Is your feature request related to a problem? Please describe.**
Opencost currently exports duplicated KSM metrics while also depending on metrics exported by KSM, see https://docs.kubecost.com/architecture/ksm-metrics

Per that page, the current long term plan is to remove the dependency on KSM, this has a few downsides:
1. Maintaining code for all of the duplicated metrics 
2. Exporting many more metrics in addition to the cost metrics
3. Requires additional RBAC permissions

**Describe the solution you'd like**
Rather than removing the dependency on KSM I believe the better solution would be to only depend on KSM metrics and remove all of the duplicated metrics, while aligning with the metrics from the latest version of KSM.

Opencost already has a hard dependency on Prometheus (and possibly node-exporter and cadvisor?), so it could also have a hard dependency on KSM.

AjayTripathy:
Hi @dudicoco thanks for the observation here. Our philosophy is to use KSM if it exists, but not take a hard dependency on it. To that end, if you're running KSM v1, you can set EMIT_KSM_V1_METRICS to false and EMIT_KSM_V1_METRICS_ONLY to false. If you're running KSM v2, you can set EMIT_KSM_V1_METRICS to false and EMIT_KSM_V1_METRICS_ONLY to true. 

I agree we should update to support interoperability with KSM V2 better so you can have a simple flag to where Opencost emits no KSM metrics at all when KSMv2 is enabled. To that end, I propose we rename this issue to be something like "create flag to support ksm v2 with no metrics duplication". What do you think of that?

dudicoco:
Thanks @AjayTripathy.
Yes i'm aware of the different flags, though some of the KSM V1 metrics generated by opencost do exist in KSM V2, they only differ by a label or two.
For example the metric `kube_pod_owner` emits a lot of sample and it's a waste to duplicate it.

Can you please elaborate why should opencost be responsible for emitting KSM metrics but not node-exporter/cadvisor metrics which it also needs to operate?

AjayTripathy:
@dudicoco if there are specific metrics that seem to be duplicated, I'm very open to adding flags to remove them from opencost.

We chose to not take a hard dependency on KSM for two main reasons:
[...]

PR #1635, opened 2023-01-30
Drop duplicate metrics emitted for KSM v1 only

biancaburtoiu:
Signed-off-by: Bianca Burtoiu <bianca@kubecost.com>

## What does this PR change?
 Stops Kubecost from emitting several duplicated KSM metrics when `EMIT_KSM_V1_METRICS_ONLY` is true:
- `kube_pod_labels`
- `kube_namespace_labels`
- `kube_node_labels`
- `kube_node_status_condition`
- `kube_pod_owner` (built on top of @kaelanspatel's PR https://github.com/opencost/opencost/pull/1629)

As a consequence of this change, the **only metrics emitted** when `EMIT_KSM_V1_METRICS_ONLY = true & EMIT_KSM_V1_METRICS = false` (ie. the metrics that are not emitted in KSM V2) are:
- `kube_node_status_capacity`
- `kube_node_status_capacity_memory_bytes`
- `kube_node_status_capacity_cpu_cores`
- `kube_node_status_allocatable`
- `kube_node_status_allocatable_cpu_cores`
- `kube_node_status_allocatable_memory_bytes`

https://github.com/opencost/opencost/pull/1629
https://github.com/kubecost/docs/pull/538

Kubecost will no longer emit the metrics listed above when `EMIT_KSM_V1_METRICS_ONLY = true`

Address https://github.com/opencost/opencost/issues/1571

## How was this PR tested?
Locally with `EMIT_KSM_V1_METRICS_ONLY = true` and `EMIT_KSM_V1_METRICS = false`; dropped metrics do not appear as part of `/model/metrics` call

Yes: https://github.com/kubecost/docs/pull/538

v1.100


pkg/metrics/kubemetrics.go
```diff
@@ -108,15 +108,7 @@ func InitKubeMetrics(clusterCache clustercache.ClusterCache, metricsConfig *Metr
 				metricsConfig:    *metricsConfig,
 			})
 		} else if opts.EmitKubeStateMetricsV1Only {
-			prometheus.MustRegister(KubeNodeCollector{
```
biancaburtoiu, line 111: - Replaced `KubeNodeCollector` with `KubeNodeCollectorV1Only`, a collector emitting identical metrics to the original except for `kube_node_labels` and `kube_node_status_condition`
- Dropped `KubeNamespaceCollector` as it was emitting the `kube_namespace_labels` metric
- Dropped `KubePodLabelsCollector` as it was emitting the `kube_pod_owner` and `kube_pod_labels` metrics

pkg/metrics/podlabelmetrics.go
```diff
@@ -1,85 +0,0 @@
...
-	"github.com/opencost/opencost/pkg/prom"
-	"github.com/prometheus/client_golang/prometheus"
-)
-
-//--------------------------------------------------------------------------
-//  KubecostPodCollector
-//--------------------------------------------------------------------------
-
-// KubecostPodCollector is a prometheus collector that emits pod metrics
-type KubecostPodLabelsCollector struct {
```
biancaburtoiu, line 14: No usages

pkg/metrics/podlabelmetrics.go
```diff
@@ -1,85 +0,0 @@
...
-		}
-	}
-}
-
-//--------------------------------------------------------------------------
-//  KubePodLabelsCollector
-//--------------------------------------------------------------------------
-
-// KubePodLabelsCollector is a prometheus collector that emits pod labels only
-type KubePodLabelsCollector struct {
```
biancaburtoiu, line 44: Dropped since both metrics must be deregistered & no usages otherwise

teevans:
@AjayTripathy - Could I bother you to review?

pkg/metrics/podlabelmetrics.go
AjayTripathy: I have a feeling we want to leave this in because in the future we want to consider removing dependencies on KSM entirely. At the risk of incredibly annoying product configuration, maybe a third flag to emit these (that is hidden from users)?

AjayTripathy:
Yeah...just one quick note that there _should_ be an option to run without KSM entirely so keeping in the label emissions feels right to me. Maybe a hidden flag to emit those. I could imagine support using it to quickly troubleshoot issues where ksmv2 is missing metrics as well.

AjayTripathy:
Can we keep this in, merge this and _not_ cherrypick? I think we need to announce to users that the plan is to stop emitting these metric labels by default.

biancaburtoiu:
> just one quick note that there should be an option to run without KSM entirely so keeping in the label emissions feels right to me.

The metrics emitted when running without KSM entirely (`V1_METRICS=true & V1_METRICS_ONLY=false`) are the ones within [this code block](https://github.com/opencost/opencost/blob/c7706a4fb2920869500f46ead1a7c5304080187f/pkg/metrics/kubemetrics.go#L81-L109). 

In this scenario, the five metrics removed in this PR - `kube_pod_labels` `kube_namespace_labels` `kube_node_labels` `kube_node_status_condition` `kube_pod_owner` - will still be **emitted** as they are referenced within the `KubePodCollector`, `KubeNamespaceCollector` and `KubeNodeCollector` instances.

mattray:
@biancaburtoiu @AjayTripathy was closing this issue intentional? I believe it was still under consideration based on the conversation in https://github.com/opencost/opencost/issues/1509

PR #1644, opened 2023-02-02
code clean for log-format variable.

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

/kind cleanup

## What does this PR change?
Through cobra flag variable hint the user to input 'json' or 'pretty'.
 We should reduce the programe cost for the string match . 

like 
```
strings.ToLower(
```
.
maybe these purpose is transfer 'JSON' to 'json'.:grinning::grinning:

### DIFF
--- a/pkg/cmd/commands.go
+++ b/pkg/cmd/commands.go
@@ -71,7 +71,7 @@ func newRootCommand(costModelCmd *cobra.Command, cmds ...*cobra.Command) *cobra.
 
 	// Add our persistent flags, these are global and available anywhere
 	cmd.PersistentFlags().String("log-level", "info", "Set the log level")
-	cmd.PersistentFlags().String("log-format", "pretty", "Set the log format - Can be either 'JSON' or 'pretty'")
+	cmd.PersistentFlags().String("log-format", "pretty", "Set the log format - Can be either 'json' or 'pretty'")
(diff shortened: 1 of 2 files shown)


thomasvn:
@yanggangtony Thanks for the PR! So you're trying to reduce the program's overhead by removing a `strings.toLower` and instead requesting the user pass a lowercase string `"json"`?

I'm unsure whether the performance gains here outweigh the reduction in usability when setting `LOG_FORMAT=json`.

@teevans Thoughts?

yanggangtony:
@thomasvn 
thanks for your review.
yes , in my opinion , i just want to clean the code for the same style.
like log-format= json and log-format=pretty.
the same style with log-level=info .

mattray (approved):
This seems fine. According to the Viper docs, keys are case-insensitive.
https://github.com/spf13/viper#does-viper-support-case-sensitive-keys

yanggangtony:
image

yanggangtony:
yes, you are right , i just test the code in my local,  keys are case-insensitive..
So we should close these pr?  because we must lower it to match the right logic thing?
@mattray

mattray:
@yanggangtony yes, the PR should be closed. Thanks for the discussion though.

PR #1649, opened 2023-02-02
allocation GenerateKey - remove sanitization for proper label matching

### LINKED ISSUE #1750: Allocation API fails when aggregating on annotations with a JSON string value
**Describe the bug**

From another pod in kubernetes, when calling the allocation endpoint with the JSON format and with an aggregation on an annotation, e.g. `http://localhost:9090/model/allocation?format=json&aggregate=annotation%3Aad.datadoghq.com%2Ftags`, where the value of the annotation is a JSON string, e.g. `'{"key": "value"}'`, I receive an error. The workaround is to call the API with the csv format, then the key of the aggregation is the JSON string.

**To Reproduce**

1. Create a deployment / pod with an annotation, e.g. `ad.datadoghq.com/tags: '{"key": "value"}'`
1. Deploy the kubecost-analyzer
1. Call allocation endpoint with an aggregation of that particular annotation: `curl http://localhost:9090/model/allocation?format=json&aggregate=annotation%3Aad.datadoghq.com%2Ftags`
1. Returns 500 server error

**Expected behavior**

Calling the allocation endpoint with json format and with an aggregation on an annotation will return a JSON string.

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-199) by [Unito](https://www.unito.io)

saweber:
## What does this PR change?
* AllocationProperties.GenerateKey

* no

* Annotation names and label names that contain a '/' or other sanitized character will have a key name that matches what is actually stored in the allocation set.

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1750

## How was this PR tested?
* unit tests

* yes, if there any special guidance around annotation names or label names containing un-sanitized characters

* no permissions

### DIFF
--- a/pkg/kubecost/allocationprops.go
+++ b/pkg/kubecost/allocationprops.go
@@ -296,7 +296,7 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 			if labels == nil {
 				names = append(names, UnallocatedSuffix)
 			} else {
-				labelName := labelConfig.Sanitize(strings.TrimPrefix(agg, "label:"))
+				labelName := strings.TrimPrefix(agg, "label:")
 				if labelValue, ok := labels[labelName]; ok {
 					names = append(names, fmt.Sprintf("%s", labelValue))
 				} else {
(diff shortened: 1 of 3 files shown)


pkg/kubecost/mock.go
```diff
@@ -568,30 +568,44 @@ func GenerateMockAssetSets(start, end time.Time) []*AssetSet {
 //
 // | Asset                        | Cost |  Adj |
 // +------------------------------+------+------+
-//   cluster1:
```
saweber, line 571: autoformat ftw

pkg/kubecost/allocationprops.go
```diff
@@ -308,7 +308,7 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 			if annotations == nil {
 				names = append(names, UnallocatedSuffix)
 			} else {
-				annotationName := labelConfig.Sanitize(strings.TrimPrefix(agg, "annotation:"))
```
saweber, line 311: example (same for labels):
agg = "a/b"
(before) annotationName = "a_b"
annotations map contains "a/b", not "a_b", therefore allocation is aggregated to __unallocated__

PR #1653, opened 2023-02-03
Ensure permissions on copied files during build are world-readable.

mattray:
My umask is set to 077, so files copied over to the container retain their restrictive permissions of 600, which causes OpenCost to fail to load custom pricing in the `default.json`.

Signed-off-by: Matt Ray <github@mattray.dev>


AjayTripathy (approved):

AjayTripathy:
Thanks @mattray !

PR #1657, opened 2023-02-06
Fix typo

imp:
The change is cosmetic - it changes the text spelling in the error message.
[Wiktionary](https://en.wiktionary.org/wiki/arguement) suggests that the original spelling is either obsolete or incorrect.

## What does this PR change?
* Fixing typo

* Nop

* Should not

* No

## How was this PR tested?
* Not tested. It is based on the assumption that text in error message is not a contract API, and hence should not be relied upon.

* I believe it does not

* No, it seems I have no permissions to assign labels


michaelmdresser (approved):
Thank you for the contribution

PR #1666, opened 2023-02-08
use opencost json marshaler for allocation responses

saweber:
Signed-off-by: saweber <saweber@gmail.com>

## What does this PR change?
* Use Opencost JSON marshaling for building JSON response of an allocation - instead of constructing ourselves.

* PR closed/to be closed: for context/discussion: https://github.com/kubecost/kubecost-cost-model/pull/1201

* Querying allocations by an annotation where the value is JSON will work
* Improved performance on marshaling annotations - this needs to be confirmed, but I'm seeing a 10-15% reduction in overall response time on `/allocation` on a dev cluster where the query has no aggregation and a 24h window (as much JSON as possible).

* [Closes 1750](https://github.com/kubecost/cost-analyzer-helm-chart/issues/1750) 

## How was this PR tested?
* Manually + unit tests

* No, unless we want to specifically call out performance improvement

* no permissions

### DIFF
--- a/pkg/kubecost/allocation_json.go
+++ b/pkg/kubecost/allocation_json.go
@@ -1,59 +1,130 @@
 package kubecost
 
 import (
-	"bytes"
-	"encoding/json"
+	"fmt"
+	"github.com/opencost/opencost/pkg/util/json"
+	"math"
 	"time"
 )
 
+// AllocationJSON  exists because there are expected JSON response fields
+// that are calculated values from methods on an annotation
+type AllocationJSON struct {
+	Name                       string                 `json:"name"`
+	Properties                 *AllocationProperties  `json:"properties"`
+	Window                     Window                 `json:"window"`
+	Start                      string                 `json:"start"`
+	End                        string                 `json:"end"`
+	Minutes                    *float64               `json:"minutes"`
+	CPUCores                   *float64               `json:"cpuCores"`
+	CPUCoreRequestAverage      *float64               `json:"cpuCoreRequestAverage"`
+	CPUCoreUsageAverage        *float64               `json:"cpuCoreUsageAverage"`
+	CPUCoreHours               *float64               `json:"cpuCoreHours"`
+	CPUCost                    *float64               `json:"cpuCost"`
+	CPUCostAdjustment          *float64               `json:"cpuCostAdjustment"`
+	CPUEfficiency              *float64               `json:"cpuEfficiency"`
+	GPUCount                   *float64               `json:"gpuCount"`
+	GPUHours                   *float64               `json:"gpuHours"`
+	GPUCost                    *float64               `json:"gpuCost"`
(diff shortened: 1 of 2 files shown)


saweber:
@AjayTripathy - once this is approved for correctness - what performance testing do you want to see done on this?

pkg/kubecost/allocation_json.go
```diff
@@ -1,59 +1,130 @@
...
+	TotalCost                  *float64               `json:"totalCost"`
+	TotalEfficiency            *float64               `json:"totalEfficiency"`
+	RawAllocationOnly          *RawAllocationOnlyData `json:"rawAllocationOnly"`
+}
+
+func (aj *AllocationJSON) BuildFromAllocation(a *Allocation) {
+	if aj == nil {
+		return
+	}
+	aj.Name = a.Name
```
nikovacevic, line 62: Is it fair to say that the original bug was entirely a function of our bad JSON func? So we don't actually have to take any extra action here?

pkg/kubecost/allocation_json.go
```diff
@@ -1,59 +1,130 @@
...
+// formatFloat64ForResponse - take an existing float64, round it to 6 decimal places and return is possible, or return nil if invalid
+func formatFloat64ForResponse(f float64) *float64 {
+	if math.IsNaN(f) || math.IsInf(f, 0) {
+		return nil
+	}
+
+	// 6 digits of precision is the maximum the API should return
+	result := math.Round(f*100000) / 100000.0
+	return &result
+}
```
nikovacevic, line 116: LGTM! Thank you for asking clarifying questions.

nikovacevic (approved):
LGTM. Let's answer the question about profiling, then ready to merge.

(Early indications were that this might even be faster, right?)

saweber:
@nikovacevic - anecdotal testing on my local machine - is showing a 30-40% reduction in response time on `/allocation?window=24h&cache=false` against cluster `dev-1` and namespace `steven`. That is very optimistic - it would drop to near 0 for accumulations, smaller windows, or fewer allocations.

pkg/kubecost/allocation_json.go
saweber: Correct. With this code I can no longer reproduce the original issue.

nikovacevic:
@saweber aside from the performance testing, are you supportive of me merging? Unless we hear otherwise from @AjayTripathy tomorrow, I'm leaning towards merging and scale testing on a nightly build later. Don't want this to rot on the vine.

saweber:
@nikovacevic Agreed. I'm highly confident that this is very unlikely to cause a performance regression. I would still like to quantify this change (with some help), even if it is after the merge to properly document in the next release.

nikovacevic:
@saweber can you update this branch and ping me? I'll merge then and we can test perf later.

saweber:
@nikovacevic done

Issue #1664, opened 2023-02-07
some bugs caused by logic and grammar


wujunwei:
**Is your feature request related to a problem? Please describe.**
I am very interested in this project, when I look into this project , I found some bugs caused by logic or grammar .show as belows:

---
https://github.com/opencost/opencost/blob/eabe8bbfe4cad99a08b2523a117955b3df69418a/pkg/kubecost/window.go#L587-L590
> receiver `w` is not a pointer ,any operetion on it make no effect.

---
https://github.com/opencost/opencost/blob/232f9f6abbc4771ac617f6ce898112b99b0630fe/pkg/clustercache/clusterexporter.go#L29-L33
json tag **omitEmpty** can not be recognize by json option, **omitempty** can

---
https://github.com/opencost/opencost/blob/c7706a4fb2920869500f46ead1a7c5304080187f/pkg/costmodel/sql.go#L97-L101
when error occured, `db` could be null, call `db.Close() `  will lead to panic! `defer db.Close() ` shound move under err nil judge.

---
and so on ...

**Describe the solution you'd like**
All bugs methioned above can be foreseen and avoid by  adding some frequently-used linters in`.golangci.yml` and run it before merge PR in CI.

**Additional context**
None

thomasvn:
@wujunwei Thanks for pointing this out!

@mattray @teevans Thoughts on enabling `.golangci.yml` linter in Github Actions?

mattray:
@thomasvn @teevans I think that's a great idea, the more pedantic the better

AjayTripathy:
PR is merged, we can leave this open until some linting is introduced.

PR #1668, opened 2023-02-09
fix bugs referred by issue

wujunwei:
Signed-off-by: wujunwei <wjw3323@live.com>

## What does this PR change?
fix bugs reference by issue #1664

No

No

#1664

## How was this PR tested?
YES

NO

### DIFF
--- a/pkg/cloud/provider.go
+++ b/pkg/cloud/provider.go
@@ -619,6 +619,9 @@ func GetClusterMeta(cluster_id string) (string, string, error) {
 	address := env.GetSQLAddress()
 	connStr := fmt.Sprintf("postgres://postgres:%s@%s:5432?sslmode=disable", pw, address)
 	db, err := sql.Open("postgres", connStr)
+	if err != nil {
+		return "", "", err
+	}
 	defer db.Close()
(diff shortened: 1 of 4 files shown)


thomasvn:
This looks good to me, @AjayTripathy ?

AjayTripathy:
Everything lgtm but not sure about the change to *Window. @nikovacevic would be the expert there wdyt?

nikovacevic:
Definitely a bug, yes, but I'd actually prefer to fix it by deleting the function and the two calls. (That function is only called in deprecated code, and already does nothing, so no harm in deleting it.)

Or an alternative would be to replace
```go
window.Set(&s, &e)
```
with
```go
window = kubecost.NewWindow(&s, &e)
```

The only two references to that function are here: https://github.com/opencost/opencost/blob/develop/pkg/costmodel/aggregation.go

AjayTripathy:
I'm going to go ahead and approve this since removal is probably beyond the scope of an open source contribution and we can evaluate removal later!

Can we file an issue to remove later?

AjayTripathy (approved):

nikovacevic:
I'll just make the changes later today.

PR #1674, opened 2023-02-09
Fix upstream nginx to access cost-model API. Update README in /ui.

nikovacevic:
Previous PR broke the nginx config. This fixes that. Also updated the README in the ui folder.

### Testing
Did a basic install and it now works, whereas it previously 404'd and errored.

Screenshot from 2023-02-09 12-46-35

### DIFF
--- a/ui/default.nginx.conf
+++ b/ui/default.nginx.conf
@@ -62,7 +62,7 @@ server {
     location /healthz {
         return 200 'OK';
     }
-    location /allocation {
+    location /model/ {
         proxy_connect_timeout       180;
         proxy_send_timeout          180;
         proxy_read_timeout          180;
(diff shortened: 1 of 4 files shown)


Sean-Holcomb (approved):
Changes look good, need to sign commit though

nikovacevic:
This is the PR that broke it: https://github.com/opencost/opencost/pull/1574

Let's please test PRs before merging them.

ui/default.nginx.conf
```diff
@@ -62,7 +62,7 @@ server {
     location /healthz {
         return 200 'OK';
     }
-    location /allocation {
+    location /model/ {
```
nikovacevic, line 65: This was the main thing -- missing a trailing slash (h/t @nealormsbee!)

Also, we renamed it `/model/` because otherwise the URL to access the API would be `/allocation/allocation/compute`.

PR #1682, opened 2023-02-10
Export v1 Allocation filter params to support validation logic in handlers

michaelmdresser:
## What does this PR change?
* Exports the string names of filterV1 parameters so they can be used in validation. See https://github.com/kubecost/kubecost-cost-model/pull/1218 for example usage.

* Pairs with https://github.com/opencost/opencost/pull/1682
* Required for https://github.com/kubecost/kubecost-cost-model/pull/1218

* N/A

## How was this PR tested?
* In the KCM PR


nickcurie (approved):
LGTM

nikovacevic (approved):

PR #1684, opened 2023-02-11
get labels/annotations intersection

yue9944882:
## What does this PR change?
* currently `AllocationProperties#Intersection()` function doesnt process the intersection of labels/annotations which may contain useful information either inherited from belonging namespace or the pod per se. for my case, i'm managing workloads under namespaces controlled by [hierarchical-namespace-controller](https://github.com/kubernetes-sigs/hierarchical-namespaces), those namespaces has some additional metadata in labels

* N/A

* No, totally backward compatible

* N/A

## How was this PR tested?
* Yes

* 

*

### DIFF
--- a/pkg/kubecost/allocationprops.go
+++ b/pkg/kubecost/allocationprops.go
@@ -148,7 +148,9 @@ func (p *AllocationProperties) Clone() *AllocationProperties {
 }
 
 func (p *AllocationProperties) Equal(that *AllocationProperties) bool {
-	if p == nil || that == nil {
+	if p == nil && that == nil {
+		return true
+	} else if p == nil || that == nil {
 		return false
 	}
 
@@ -422,7 +424,10 @@ func (p *AllocationProperties) Intersection(that *AllocationProperties) *Allocat
 	if p == nil || that == nil {
 		return nil
 	}
-	intersectionProps := &AllocationProperties{}
+	intersectionProps := &AllocationProperties{
+		Labels:      make(map[string]string),
+		Annotations: make(map[string]string),
+	}
 	if p.Cluster == that.Cluster {
 		intersectionProps.Cluster = p.Cluster
 	}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
@nikovacevic correct me if I'm wrong but there are some scaling implications about processing the intersections here. While we could support taking an intersection here, we don't want to be rolling up n^2 by default. Want to weigh in?

yue9944882:
hi @AjayTripathy generally the cardinality of the labels/annotations attached to an k8s object should be at maximum ~100, n^2 wont cause any visible performance degrading i presume

nikovacevic:
Yeah, we had to disable this for performance reasons. The real issue here is, with N allocations and M labels/annotations each, when you aggregate you're talking about O(N) with respect to allocating new maps, and O(N*M) for string comparisons. We deemed that to be too slow for the Allocation API. Now that we have Summary and View, we can consider bringing it back, but with the caveat that it'll slow down current Allocation API users.

AjayTripathy:
@yue9944882 the cardinality when constructing the response object is 100x100x(number of pods) which can be in the millions.

I think there are a couple of approaches that work at scale here:
1. an API parameter to support using an intersection
2. a whitelist for labels that you want to be intersected

Thoughts on either of these @nikovacevic ?

nikovacevic:
Prefer (1) to (2). Would not be too hard to flag it. Though, again, we may be able to use offer intersection now that pressure is off the Allocation API in a way it wasn't before.

michaelmdresser:
I agree (1) is better than (2). 

> Though, again, we may be able to use offer intersection now that pressure is off the Allocation API in a way it wasn't before.

There are a few things (e.g. request sizing) that use the standard Allocation API that would suffer from default-enabling the intersection. As much as a pain as it is to increase the config surface, I think `EnableLabelIntersection bool` (thus defaulting to false) in `AllocationQueryOpts` is probably the play.

PR #1685, opened 2023-02-13
Support opt-out from alibaba cloud provider

yue9944882:
## What does this PR change?
* adding an env var for opt-out from the default Alibaba cloud provider implementation. This is useful for offline Alibaba cloud clusters of which the cost computing model is different/conflicting with the current one. these offline clusters are also using kubelet images suffixed `-aliyun-` platform version, this makes opencost agents not working in offline environment (because the provider detection logic for Alibaba cloud is based on kubelet version). I'm raising this pull because I'm currently working for alibaba cloud-native kubernetes related product and it looks fine for me to keep the current provider as default alibaba provider but just leaving an option to custom provider would be great.

### DIFF
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -19,6 +19,7 @@ const (
 
 	AlibabaAccessKeyIDEnvVar     = "ALIBABA_ACCESS_KEY_ID"
 	AlibabaAccessKeySecretEnvVar = "ALIBABA_SECRET_ACCESS_KEY"
+	AlibabaDisableEnvVar         = "ALIBABA_PROVIDER_DISABLE"
 
 	KubecostNamespaceEnvVar        = "KUBECOST_NAMESPACE"
 	PodNameEnvVar                  = "POD_NAME"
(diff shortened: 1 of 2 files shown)


pkg/env/costmodelenv.go
```diff
@@ -511,7 +518,7 @@ func GetAllocationNodeLabelsEnabled() bool {
 	return GetBool(AllocationNodeLabelsEnabled, true)
 }
 
-var defaultAllocationNodeLabelsIncludeList []string = []string{
```
yue9944882, line 514: this is an unintentional change from `go fmt`, but it looks like a positive change so I'm keeping it in this pull.

thomasvn:
Thanks for the PR @yue9944882  ! Just to double check, it seems like there are two issues here you intend to solve:

- your internet-disconnected Alibaba cloud clusters are charged differently compared to the public pricing APIs OpenCost is referencing?
- your internet-disconnected Alibaba kubelets are named differently, causing OpenCost to throw errors?

Is that correct? If so, would cloud billing reconciliation help address this concern?

@avrodrigues5 @kwombach12 Thoughts here on reconciliation, and on the ability to disable the Alibaba Cloud Integration?

kwombach12:
@thomasvn reconciliation for Alibaba is something we are actively working on now and will deliver in v1.101. I don't really have an issue with giving users the ability to disable Alibaba

AjayTripathy:
There should probably be a more general flag than an alibaba-specific one that forces use of the CSV provider specifically. Other than that this notion seems fine to me.

PR #1694, opened 2023-02-16
ensure /allocation response start and end are always in UTC format

saweber:
## What does this PR change?
* `/allocation` is sometimes returning local time rather than UTC

* yes; https://github.com/opencost/opencost/pull/1666

* user gets consistent timestamps

* no

## How was this PR tested?
* manually

* no

* no permissions


nikovacevic (approved):
Nice! LGTM!

PR #1709, opened 2023-02-23
fix(panic): Handle case where key "default" doesn't exist

### LINKED ISSUE #1710: Panic on Startup with no config file set
**Describe the bug**
There is a scenario on bootup where the app will crash due to a panic if the config file is not set. An easy way to replicate the panic is to not set `CONFIG_PATH` and then run  `./cmd/costmodel/main.go`.

**To Reproduce**
1. Unset `CONFIG_PATH`
2. `go run cmd/costmodel/main.go`
3. Observe the application crashing

**Expected behavior**
 If `CONFIG_PATH` _must_ be set, we should handle this error earlier in the bootup process and provide and appropriate message. Otherwise, I think the critical path in the code should be more defensive to handle the scenario where the pricing isn't set. 

The scenario I fear is that by this point, the `/health` and `/ready` endpoints are available and kubernetes would mark the pod as deployed. If that happens, the pod would panic and get stuck in a crashloopbackoff phase. If we handled the error when reading the configuration file and closed, neither endpoint would be available and the deployment wouldn't progress. 

Latest off of `develop`

**Additional context**
This scenario likely wouldn't have in production, but is something I feel strongly we should handle better in case it does.

Pokom:
There is a scenario on bootup where the app can crash due to a panic. An easy way to replicate the panic is to not set `CONFIG_PATH`. This change will handle this scenario better and set the cpu, memory and gpu prices to "0.0".

## What does this PR change?
* Behavior in custom provider when `default` pricing isn't found. 

* 

* 

* Closes #1710 

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/cloud/customprovider.go
+++ b/pkg/cloud/customprovider.go
@@ -3,7 +3,6 @@ package cloud
 import (
 	"errors"
 	"fmt"
-	"github.com/opencost/opencost/pkg/kubecost"
 	"io"
 	"strconv"
 	"sync"
@@ -11,6 +10,8 @@ import (
 
 	"github.com/opencost/opencost/pkg/clustercache"
 	"github.com/opencost/opencost/pkg/env"
+	"github.com/opencost/opencost/pkg/kubecost"
+	"github.com/opencost/opencost/pkg/log"
 	"github.com/opencost/opencost/pkg/util/json"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


AjayTripathy:
This LGTM! Thanks @Pokom .

Pokom:
Thanks @AjayTripathy! Would you mind taking a look at the issue linked as well(#1710) This certainly will stop the bleed, but I wanted to see if it would also make sense to catch the error earlier in the process and stop the application before health checks are available.

Pokom:
@teevans when you have a chance, mind giving this a one over for a review? Should be relatively straight forward to test out. 

Probably biggest caveat is if we wanted to handle startup slightly different. I highlighted the alternatives in #1710

AjayTripathy:
Hey @Pokom , I don't think it's necessary to catch this earlier in the pipeline since the application should still function without these prices set. The warning looks fine to me.

AjayTripathy (approved):

PR #1715, opened 2023-02-28
Fix for Opencost’s cpu & mem allocation metrics

### LINKED ISSUE #1702: Opencost doesn't take into account pod CPU usage
**Describe the bug**
Opencost doesn't take into account cpu consumption of the pods. Only requested value is used.

**To Reproduce**
I have a pod with the whole purpose of consuming CPU without requesting anything:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: busypod
  name: busypod
  namespace: busy
spec:
  replicas: 1
  selector:
    matchLabels:
      app: busypod
  template:
    metadata:
      labels:
        app: busypod
    spec:
      containers:
      - args:
        - -c
        - yes > /dev/null
        command:
        - sh
        image: alpine:latest
        imagePullPolicy: Always
        name: p1
      - args:
        - -c
        - yes > /dev/null
        command:
        - sh
        image: alpine:latest
        imagePullPolicy: Always
        name: p2
      restartPolicy: Always
```

**Expected behavior**

The cost of the pod/deployment should be above zero. The cost of the deployment should be roughly equal to 2 * core * time frame. (two containers, ~100% core each)

**Screenshots**

image

I believe it's the latest version:

```
  containerStatuses:
  - image: quay.io/kubecost1/kubecost-cost-model:latest
    imageID: quay.io/kubecost1/kubecost-cost-model@sha256:ced5420c7d3b3d7d2ba8c4f9fc4d2e6bf80187dbfe2e75c2a904891c7de5cdb2
  - image: quay.io/kubecost1/opencost-ui:latest
    imageID: quay.io/kubecost1/opencost-ui@sha256:340fb23b76e10a080dc8579d000b520105ee3336c091874c800747451eddac20
```

The opencost is installed following installation documentation. Prometheus helm chart version:

image

**Additional context**

I believe the issue caused by this query: 
https://github.com/opencost/opencost/blob/68a1ee2b969b2af31774febb01549dcd023ff9aa/pkg/costmodel/costmodel.go#L174

For the pod `container_cpu_allocation` metric from opencost reports 0:

image

`container_cpu_usage_seconds_total` metrics reports CPU usage:

image

`container_cpu_usage_seconds_total`  metrics comes from default prometheus setup and has `container` label, but doesn't have `container_name` label. `container_name!=""` filter in query returns empty result. It leads to ignoring CPU usage in `container_cpu_allocation` metric which leads to 0$ allocation cost in the UI.

thomasvn:
## What does this PR change?

- Previously, OpenCost’s `container_cpu_allocation` and `container_memory_allocation` were only reporting on requests, and not usage. This was due to OpenCost querying cAdvisor container usage metrics, and getting empty responses.
- This PR fixes those queries by now filtering for `container` instead of `container_name`
- This change also required additional metric relabeling in the query for compatibility reasons. (`container` -> `container_name`, `pod` -> `pod_name`)

Todo (probably for a future PR):

- [ ] Audit other cAdvisor metrics used by OpenCost to see which are also affected

- No

- **Backwards Compatibility:** We are now filtering on the `container` metric label, instead of `container_name`. Kubernetes clusters before v1.14 will be impacted, because the kubelet cAdvisor would still be emitting `container_name`. https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md?plain=1#L985

- Closes https://github.com/opencost/opencost/issues/1702

## How was this PR tested?

- Ran OpenCost locally and now see that OpenCost’s metrics (http://localhost:9003/metrics) for `container_cpu_allocation` and `container_memory_allocation` correctly return non-zero values for containers which do not have requests set.

- No

-


michaelmdresser:
@nikovacevic is it correct to say that this does not affect KCM because KCM relies on `pkg/costmodel/allocation.go`, not `pkg/costmodel/costmodel.go`'s `costDataRange`? I'm trying to understand the impact level of this bug outside of OC.

thomasvn:
@michaelmdresser I think that's correct. I didn't observe any issues with metrics emitted by KCM.

dudicoco:
@michaelmdresser @thomasvn do opencost and kubecost actually work differently?

The images seem to be at least named similarly which is confusing to me:
https://github.com/opencost/opencost/blob/develop/kubernetes/opencost.yaml#L143
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L329

thomasvn:
@dudicoco Correct, those images are different ^^. Kubecost is built on top of OpenCost and provides an additional set of features.

dudicoco:
@thomasvn but opencost also has the same baseline code that kubecost uses (`pkg/costmodel/allocation.go`), so why doesn't opencost make use of this code as well?

thomasvn:
@dudicoco Not sure I fully follow ^^, but I believe the reason why this is an issue in OpenCost but not Kubecost, is because Kubecost's bundled Prometheus is configured to perform metric relabeling from `container` -> `container_name`. Config [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.101/cost-analyzer/charts/prometheus/values.yaml#L1213-L1220).

Does that help clarify the issue?

dudicoco:
@thomasvn my last question was more specifically regarding why opencost and kubecost use different code within the same codebase - from what I understand `pkg/costmodel/allocation.go` is there only to support kubecost?

I guess I also find the whole separation model of kubecost and opencost very unusual and confusing (especially since kubecost ALSO has a free version).
This is very different than other open source projects which also have a paid model.

thomasvn:
@dudicoco Sorry if I mispoke. I believe both OpenCost and Kubecost are leveraging [`pkg/costmodel/allocation.go`](https://github.com/opencost/opencost/blob/develop/pkg/costmodel/allocation.go#L86).

dwbrown2:
> I guess I also find the whole separation model of kubecost and opencost very unusual and confusing (especially since kubecost ALSO has a free version).
This is very different than other open source projects which also have a paid model.

@dudicoco would love to hear more about your study of other open source projects and what you find unusual! In this case, the Kubecost helm chart bundles a Prometheus, whereas Opencost does not. Do you find that specifically to be unusual?

AjayTripathy:
@thomasvn core looks good but looks like there are some other changes in here? Can you isolate the PR to just be the relevant change?

thomasvn:
@AjayTripathy Will do. I think the extra commits were because I `git rebase`.

thomasvn:
Closing this PR, and resubmitting it as a new PR https://github.com/opencost/opencost/pull/1770 to fix DCO Signed Commit.

PR #1716, opened 2023-03-01
allocation - change accumulate to work by month, week, day, hour, true, or false

saweber:
## What does this PR change?
* AccumulateBy AllocationQueryOption is now a string, which maps to either '', 'day', 'week', 'month', or 'all' to support new aggregate queries in closed source

* yes- https://github.com/kubecost/kubecost-cost-model/pull/1246

* no impact for opencost users

* https://kubecost.atlassian.net/browse/CORE-49

## How was this PR tested?
* manually + unit tests

* not for opencost

* no permissions

### DIFF
--- a/pkg/costmodel/aggregation.go
+++ b/pkg/costmodel/aggregation.go
@@ -1076,7 +1076,7 @@ func (a *Accesses) ComputeAggregateCostModel(promClient prometheusClient.Client,
 		if durMins%60 != 0 || durMins < 3*60 { // not divisible by 1h or less than 3h
 			resolution = time.Minute
 		}
-	} else { // greater than 1d
+	} else {                    // greater than 1d
 		if durMins >= 7*24*60 { // greater than (or equal to) 7 days
 			resolution = 24.0 * time.Hour
 		} else if durMins >= 2*24*60 { // greater than (or equal to) 2 days
@@ -2191,12 +2191,11 @@ func (a *Accesses) ComputeAllocationHandlerSummary(w http.ResponseWriter, r *htt
 
 	// Accumulate, if requested
 	if accumulate {
-		as, err := asr.Accumulate()
+		asr, err = asr.Accumulate(kubecost.AccumulateOptionAll)
 		if err != nil {
 			WriteError(w, InternalServerError(err.Error()))
 			return
 		}
-		asr = kubecost.NewAllocationSetRange(as)
 	}
 
 	sasl := []*kubecost.SummaryAllocationSet{}
@@ -2244,10 +2243,15 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	// sums each Set in the Range, producing one Set.
 	accumulate := qp.GetBool("accumulate", false)
 
-	// AccumulateBy is an optional parameter that accumulates an AllocationSetRange
+	// Accumulate is an optional parameter that accumulates an AllocationSetRange
 	// by the resolution of the given time duration.
 	// Defaults to 0. If a value is not passed then the parameter is not used.
-	accumulateBy := qp.GetDuration("accumulateBy", 0)
+	accumulateBy := kubecost.AccumulateOption(qp.Get("accumulateBy", ""))
+
+	// if accumulateBy is not explicitly set, and accumulate is true, ensure result is accumulated
+	if accumulateBy == kubecost.AccumulateOptionNone && accumulate {
+		accumulateBy = kubecost.AccumulateOptionAll
+	}
 
 	// Query for AllocationSets in increments of the given step duration,
 	// appending each to the AllocationSetRange.
@@ -2277,19 +2281,13 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	}
 
 	// Accumulate, if requested
-	if accumulateBy != 0 {
-		asr, err = asr.AccumulateBy(accumulateBy)
-		if err != nil {
-			WriteError(w, InternalServerError(err.Error()))
-			return
-		}
-	} else if accumulate {
-		as, err := asr.Accumulate()
-		if err != nil {
-			WriteError(w, InternalServerError(err.Error()))
-			return
-		}
-		asr = kubecost.NewAllocationSetRange(as)
+	if accumulateBy != kubecost.AccumulateOptionNone {
+		asr, err = asr.Accumulate(accumulateBy)
+	}
+
+	if err != nil {
+		WriteError(w, InternalServerError(err.Error()))
+		return
 	}
 
 	w.Write(WrapData(asr, nil))
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -174,10 +174,17 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	// Accumulate to yield the result AllocationSet. After this step, we will
 	// be nearly complete, but without the raw allocation data, which must be
 	// recomputed.
-	result, err := asr.Accumulate()
+	resultASR, err := asr.Accumulate(kubecost.AccumulateOptionAll)
 	if err != nil {
 		return kubecost.NewAllocationSet(start, end), fmt.Errorf("error accumulating data for %s: %s", kubecost.NewClosedWindow(s, e), err)
 	}
+	if resultASR != nil && len(resultASR.Allocations) == 0 {
+		return kubecost.NewAllocationSet(start, end), nil
+	}
+	if length := len(resultASR.Allocations); length != 1 {
+		return kubecost.NewAllocationSet(start, end), fmt.Errorf("expected 1 accumulated allocation set, found %d sets", length)
+	}
+	result := resultASR.Allocations[0]
 
 	// Apply the annotations, labels, and services to the post-accumulation
 	// results. (See above for why this is necessary.)
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -2079,7 +2079,7 @@ func (asr *AllocationSetRange) Get(i int) (*AllocationSet, error) {
 
 // Accumulate sums each AllocationSet in the given range, returning a single cumulative
 // AllocationSet for the entire range.
-func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
+func (asr *AllocationSetRange) accumulate() (*AllocationSet, error) {
 	var allocSet *AllocationSet
 	var err error
 
@@ -2093,12 +2093,12 @@ func (asr *AllocationSetRange) Accumulate() (*AllocationSet, error) {
 	return allocSet, nil
 }
 
-// NewAccumulation clones the first available AllocationSet to use as the data structure to
+// newAccumulation clones the first available AllocationSet to use as the data structure to
 // Accumulate the remaining data. This leaves the original AllocationSetRange intact.
-func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
+func (asr *AllocationSetRange) newAccumulation() (*AllocationSet, error) {
 	// NOTE: Adding this API for consistency across SummaryAllocation and Assets, but this
 	// NOTE: implementation is almost identical to regular Accumulate(). The Accumulate() method
-	// NOTE: for Allocation returns Clone() of the input, which is required for AccumulateBy
+	// NOTE: for Allocation returns Clone() of the input, which is required for Accumulate
 	// NOTE: support (unit tests are great for verifying this information).
 	var allocSet *AllocationSet
 	var err error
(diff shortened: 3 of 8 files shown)


pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,63 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
-// at which point AccumulateBy will start summing from 0 until the requested resolution is met again.
-// If the requested resolution is smaller than the window of an AllocationSet then the resolution will default to the duration of a set.
-// Resolutions larger than the duration of the entire AllocationSetRange will default to the duration of the range.
-func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*AllocationSetRange, error) {
-	allocSetRange := NewAllocationSetRange()
-	var allocSet *AllocationSet
-	var err error
+// AccumulateBy sums AllocationSets based on the AccumulateByType (calendar week or calendar month).
+// The accumulated set is determined by the start of the window of the allocation set.
+func (asr *AllocationSetRange) AccumulateBy(accumulateBy AccumulateByType) (*AllocationSetRange, error) {
```
nikovacevic: `AccumulateByType` works, but exploring other naming ideas... what do you think of `AccumulationOption`? Or did you have other names in mind?

My thinking here is that "accumulate by type" has an odd ring to it because it can be read (accumulateBy)(type), but also (accumulate)(by type).

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,63 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+	if accumulateBy == AccumulateByNone {
```
nikovacevic: This is nice. Makes me think that we can also support `AccumulateByAll` and pass through to `Accumulate()`! What do you think of that?

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,63 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+		for i, as := range asr.Allocations {
+			dayOfWeek := as.Window.Start().Weekday()
+
+			if toAccumulate == nil {
+				toAccumulate = NewAllocationSetRange()
+			}
+
+			toAccumulate.Append(as)
+
+			// current assumption is the week always ends on Saturday
```
nikovacevic: That works :+1:

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,63 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+			if dayOfWeek == time.Saturday || i == len(asr.Allocations)-1 {
```
nikovacevic: So this is, "If we just finished the last day of the week, OR we just finished the last set in the range." Makes sense to me.

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,63 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+			}
+
+			toAccumulate.Append(as)
 
-			// check if end of asr to sum the final set
-			// If total asr accumulated sum <= resolution return 1 accumulated set
-			if allocSet.Window.Duration() >= resolution || i == len(asr.Allocations)-1 {
-				allocSetRange.Allocations = append(allocSetRange.Allocations, allocSet)
-				allocSet = NewAllocationSet(time.Time{}, time.Time{})
+			if month != nextDayMonth || i == len(asr.Allocations)-1 {
```
nikovacevic: And here it's "If this is the last day in the current month, OR the last set in the range." Yes, looks good.

pkg/kubecost/query.go
```diff
@@ -56,6 +56,14 @@ type AllocationQueryOptions struct {
 	Step                    time.Duration
 }
 
+type AccumulateByType string
```
nikovacevic: Let's talk naming! (This works, and we may well go with it. Just want to explore a couple other options, or hear if you had other options in mind.)

pkg/kubecost/query.go
```diff
@@ -56,6 +56,14 @@ type AllocationQueryOptions struct {
...
+
+const (
+	AccumulateByNone  AccumulateByType = ""
+	AccumulateByWeek  AccumulateByType = "week"
+	AccumulateByMonth AccumulateByType = "month"
```
nikovacevic: Could add `"all"` or not. What do you think?

nikovacevic (commented):
We'll want to add tests (which you know, and agree) -- but the logic looks sound! And it's implemented in exactly the way I'd have expected. Let's talk tests and naming, then we should be able to approve and merge. Thank you!

pkg/kubecost/allocation.go
saweber: That is an easy change - but is it useful? The only win I have is simplifying a couple of if else blocks in handlers (which still probably makes this worth doing).

pkg/kubecost/allocation.go
ghost: I'm seconding the consideration of other names for `AccumulateByType`; Maybe it's because I'm coming from Python, but I associate naming things as "x Type" as shorthand for, well, typechecking.

pkg/kubecost/allocation.go
saweber: Done.

pkg/costmodel/aggregation.go
```diff
@@ -2247,7 +2247,12 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	// AccumulateBy is an optional parameter that accumulates an AllocationSetRange
 	// by the resolution of the given time duration.
 	// Defaults to 0. If a value is not passed then the parameter is not used.
-	accumulateBy := qp.GetDuration("accumulateBy", 0)
+	accumulateBy := kubecost.AccumulateOption(qp.Get("accumulateBy", ""))
```
nikovacevic, line 2249: What do you think of the idea of lumping this all under `accumulate`? Don't worry about backwards-compatibility. Just in terms of an API, something like:
```
accumulate={false,true,hourly,daily,weekly,monthly}
```
instead of two params?

pkg/costmodel/aggregation.go
```diff
@@ -2247,7 +2247,12 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
...
+
+	// if accumulateBy is not explicitly set, and accumulate is true, ensure result is accumulated
+	if accumulateBy == kubecost.AccumulateOptionNone && accumulate {
+		accumulateBy = kubecost.AccumulateOptionAll
```
nikovacevic, line 2253: Love this.

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,133 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+		accumulated := NewAllocationSetRange(as)
+		return accumulated, nil
+	case AccumulateOptionHour:
+		return asr.accumulateByHour()
+	case AccumulateOptionDay:
+		return asr.accumulateByDay()
+	case AccumulateOptionWeek:
+		return asr.accumulateByWeek()
+	case AccumulateOptionMonth:
+		return asr.accumulateByMonth()
```
nikovacevic, line 2149: Very clean

pkg/kubecost/allocation.go
```diff
@@ -2435,3 +2534,16 @@ func (asr *AllocationSetRange) TotalCost() float64 {
 	}
 	return tc
 }
+
+// Clone returns a new AllocationSetRange cloned from the existing ASR
+func (asr *AllocationSetRange) Clone() *AllocationSetRange {
```
nikovacevic, line 2565: Are these new `Clone()` functions just used in testing?

nikovacevic (approved):
LGTM!

pkg/costmodel/aggregation.go
nikovacevic: Just thinking that we already have `aggregate` as a param, and having a single `accumulate` might mirror that nicely. Not a big deal either way, though, to be clear.

nikovacevic: In that case, we'd want to make sure that we had a little ParseAccumulateBy func that helped to map "true" to "all", "false" to "", etc.

saweber: That is a cleaner solution.

saweber:
~I still need to do manual testing on Monday; but this should be really close.~

Testing looks good.

pkg/costmodel/allocation.go
```diff
@@ -174,11 +174,16 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
...
-	result, err := asr.Accumulate()
+	resultASR, err := asr.Accumulate(kubecost.AccumulateOptionAll, false)
 	if err != nil {
 		return kubecost.NewAllocationSet(start, end), fmt.Errorf("error accumulating data for %s: %s", kubecost.NewClosedWindow(s, e), err)
 	}
 
+	if length := len(resultASR.Allocations); length != 1 {
+		return kubecost.NewAllocationSet(start, end), fmt.Errorf("expected 1 accumulated allocation set, found %d sets", length)
+	}
+	result := resultASR.Allocations[0]
```
michaelmdresser, line 187: Perhaps an outlier, but should this error if `len(resultASR.Allocations) == 0`? E.g. calling `Accumulate` on an ASR which is full of empty AllocationSets.

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,179 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
-// Resolutions not evenly divisible by the AllocationSetRange window durations Accumulate sets until a sum greater than or equal to the resolution is met,
-// at which point AccumulateBy will start summing from 0 until the requested resolution is met again.
-// If the requested resolution is smaller than the window of an AllocationSet then the resolution will default to the duration of a set.
-// Resolutions larger than the duration of the entire AllocationSetRange will default to the duration of the range.
-func (asr *AllocationSetRange) AccumulateBy(resolution time.Duration) (*AllocationSetRange, error) {
-	allocSetRange := NewAllocationSetRange()
-	var allocSet *AllocationSet
+// Accumulate sums AllocationSets based on the AccumulateOption (calendar week or calendar month).
+// The accumulated set is determined by the start of the window of the allocation set.
+func (asr *AllocationSetRange) Accumulate(accumulateBy AccumulateOption, clone bool) (*AllocationSetRange, error) {
```
nikovacevic: Is this `clone` flag the only option here? I'm afraid that we don't want to publicly expose the option to users. Would it be worth a quick conversation tomorrow morning to go over this?

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,179 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+		}
 
-			// check if end of asr to sum the final set
-			// If total asr accumulated sum <= resolution return 1 accumulated set
-			if allocSet.Window.Duration() >= resolution || i == len(asr.Allocations)-1 {
-				allocSetRange.Allocations = append(allocSetRange.Allocations, allocSet)
-				allocSet = NewAllocationSet(time.Time{}, time.Time{})
+		if clone {
+			as = as.Clone()
+		}
```
nikovacevic: This looks like the right idea! Just a little worried about exposing this internal "implementation detail" as a flag to the caller.

nikovacevic (commented):
Let's discuss the new Accumulate API ASAP -- if this is the only way, we can move forward with it. But I wonder if we can try to get back down to a single parameter.

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,179 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+}
 
-		if allocSet != nil {
+func (asr *AllocationSetRange) accumulateByMonth(clone bool) (*AllocationSetRange, error) {
+	var toAccumulate *AllocationSetRange
+	result := NewAllocationSetRange()
+	for i, as := range asr.Allocations {
+		if as.Window.Duration() != time.Hour*24 {
+			return nil, fmt.Errorf("window duration must equal 24 hours; got:%s", as.Window.Duration())
+		}
```
michaelmdresser, line 2226: In order to turn an hourly ASR into a monthly ASR (rare though it may be), this means I have to call `asr.Accumulate(day).Accumulate(month)`, right?

The case I'm thinking about is if we query Allocation data without specifying `step` and receive hourly data due to a <= 48h `window`, but then combine that with a `accumulateBy=month`. `/allocation?window=24h&accumulateBy=month` Will this logic cause such a query to error?

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,179 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
-			// If total asr accumulated sum <= resolution return 1 accumulated set
-			if allocSet.Window.Duration() >= resolution || i == len(asr.Allocations)-1 {
-				allocSetRange.Allocations = append(allocSetRange.Allocations, allocSet)
-				allocSet = NewAllocationSet(time.Time{}, time.Time{})
+		if clone {
+			as = as.Clone()
+		}
+
+		_, month, _ := as.Window.Start().Date()
+		_, nextDayMonth, _ := as.Window.Start().Add(time.Hour * 24).Date()
```
michaelmdresser, line 2229: Slick!

pkg/kubecost/allocation.go
saweber: Lets do that. I was hoping to not force the user to pay the penalties of cloning - but if we are fine with that, we can just make it always a clone. With the main penalty being the time to copy data (memory space should be minimal).

pkg/kubecost/allocation.go
```diff
@@ -2131,34 +2131,179 @@ func (asr *AllocationSetRange) NewAccumulation() (*AllocationSet, error) {
...
+
+		toAccumulate.Append(as)
+		asAccumulated, err := toAccumulate.accumulate()
+		if err != nil {
+			return nil, fmt.Errorf("error accumulating result: %s", err)
+		}
+		toAccumulate = NewAllocationSetRange(asAccumulated)
+
+		// current assumption is the week always ends on Saturday, or there are no more allocation sets
+		if dayOfWeek == time.Saturday || i == len(asr.Allocations)-1 {
```
michaelmdresser, line 2278: What if I have a (crazy) ASR that looks like this:
```
ASR{
    AS{ window: thursday }
    AS{ window: friday }
    AS{ window: sunday }
}
```

I think we might accumulate all of these ASes into one, rather than two like I would expect.

michaelmdresser: I suppose the same question goes for monthly accumulation.

michaelmdresser (commented):
Most of the way through, but want to get you what I have before I sign out.

pkg/costmodel/allocation.go
saweber: That is a very good callout! Will fix.

pkg/kubecost/allocation.go
saweber: Yes, your understanding is correct.

My thought was it only really makes sense to roll hourly windows up to a day, as long as 48h is our daily limit. If customers were configuring this to 168+ hours, then turning on weekly would start to make sense.

If this is a scenario we want to handle - its easy enough to do daily accumulation first.

pkg/kubecost/allocation.go
saweber: That is an accurate understanding. Does this scenario happen?

My main goal was to avoid difficult to understand date parsing code (especially when we have to come back and add a configurable start of the week day), but if we need it, we need it.

pkg/kubecost/allocation.go
michaelmdresser: As long as we're confident that in a first-install scenario (< 48h of data total) this won't cause an unnecessary query failure, I think its alright.

pkg/kubecost/allocation.go
michaelmdresser: Frankly, I'm not sure. @nikovacevic if Kubecost is down for >= 1 day, is it possible to query and get an ASR with entire AllocatioSet(s) missing?

nikovacevic: No, there shouldn't be any missing days. In that case, there would be an _empty_ day, or days, but that AllocationSet should still exist with a window.

michaelmdresser: I've seen something like my theoretical situation _serialized_ (JSON) as `ASR{ {}, {}, {}, {filled...} ...}`. It seems that `nil` ASes get turned into `{}` by our custom marshaler: https://github.com/opencost/opencost/blob/675a6af3f8ccb2e289fdc61d342821effd7c06cb/pkg/kubecost/allocation_json.go#L169-L175

That indicates that we could have a `nil` AS somewhere in an ASR, taking my earlier example:
```
ASR{
    AS{ window: thursday }
    AS{ window: friday }
    nil
    AS{ window: sunday }
}
```

I'll caveat this with the fact that I've never seen the `{}` show up in the _middle_ of an ASR, but maybe its possible?

pkg/kubecost/allocation.go
nikovacevic: In the medium-to-long term we should consider rolling up across multiple step sizes at once (e.g. hourly-to-monthly) but for now, we shouldn't ever encounter that.

E.g. in the fresh install scenario, we'll always have daily data no matter what -- it's perpetually cascading into the daily store. So we would never have to pull from hourly to fulfill a monthly request. (And, in fact, by the time 102 is released we plan to have weekly and monthly stores, too! So we should only ever accumulate-by at cascade-time.)

nikovacevic: TL;DR let's not let scope creep. This is fine for now!

pkg/kubecost/allocation.go
michaelmdresser: I'm still a little concerned about this, but I think it is sufficiently edge that we can proceed for now. If we do run into the hypothetical bug, it should at least show up in the resulting Window, meaning it will be debuggable.

michaelmdresser (approved):

pkg/kubecost/allocation.go
nikovacevic: I don't think concern is altogether unwarranted! Haha. You're right about the JSON side of things, but remember that empty AllocationSets that are _not_ null also encode that way. And because we never save `nil` sets, but we do save empty ones, I'd be very surprised if we ever had nils anywhere in SetRanges. I would actually consider that a high priority bug to iron out. They should always be empty sets.

(Feel free to keep this receipt in the case that I am wrong :wink: )

PR #1723, opened 2023-03-03
License documentation under the CC BY 4.0.

mattray:
Following guidance from the CNCF's IP policy https://github.com/cncf/foundation/blob/main/charter.md#11-ip-policy

The OpenCost code already has the Apache v2 License, this is for the documentation included in the repository.


Pokom:
This looks fine to me, with the caveat being I don't think I understand the legal implications enough. The way I'm viewing this is we already license OpenCost codebase as `Apache v2`, now we're explicitly marking the documentation as being licensed as `Apache V2`. Is that correct?

Pokom (approved):
See my comment on the PR. If my assumption is correct, feel free to merge in.

AjayTripathy (approved):

PR #1733, opened 2023-03-06
poc(s3): Initial blob storage cost exporter

Pokom:
This is a _very rough_ and broken implementation for cost exporter. I am attempting to use `aws-go-sdk-v2` to get all of the costs for S3 buckets in a specific region.

At this point I've been able to get all of the prices for a region and am working on parsing this out. This is a bit trickier then expected and will likely need to rely upon existing code to parse this.

This is primarily meant to be a discussion point for our weekly Cloud Costs conversation that happens on Tuesday's.


AjayTripathy:
So cool @Pokom ! Here's my general feedback:

1. I'm thinking that we want some sort of conditional compilation for these modules, since they are going to grow pretty rapidly. https://medium.com/@m_nakamura145/conditional-compilation-in-go-a4fde90c6dd5
2. What do you think of the directory structure? I am thinking we want these to live one directory deeper and perhaps refactor so we have something like pkg/cloud/aws/external/... and pkg/cloud/aws/kubernetes... wdyt?

Pokom:
> I'm thinking that we want some sort of conditional compilation for these modules, since they are going to grow pretty rapidly. https://medium.com/@m_nakamura145/conditional-compilation-in-go-a4fde90c6dd5

TIL! Let me look into this deeper and see what we can leverage. I love the idea that we could release this behind a custom build that most folks won't have access too.

I like going one level deeper for the directory structure. I have a few other refactors that will need to land before I can really kick off this work, so I will submit that as a separate PR with _just_ moving of files. How do you feel about starting that process for all the clouds? IE, go from the flat `pkg/cloud/*.go` to `pkg/cloud/[aws|google|azure|alibaba|...]/kubernetes`?

AjayTripathy:
Not the only stakeholder on a full directory structure change... @nikovacevic @Sean-Holcomb @mbolt35 wdyt?

Pokom:
I'm all in favor of creating an issue to call out the exact changes I'm
proposing so it's easier to follow.

AjayTripathy:
sg!

PR #1767, opened 2023-03-13
Create Response types to protect SummaryAllocation types against errors marshaling NaN

nikovacevic:
## What does this PR change?
* Create Response types to protect SummaryAllocation types against errors marshaling NaN

* Required for https://github.com/kubecost/kubecost-cost-model/pull/1297

* Prevents Summary Allocation API from returning an error if NaN values are encountered.

* Partially addresses https://kubecost.atlassian.net/browse/EV-35 (does not address the View API)

## How was this PR tested?
* Unit tests
* Manually

### DIFF
--- a/pkg/kubecost/summaryallocation_json.go
+++ b/pkg/kubecost/summaryallocation_json.go
@@ -1,0 +1,116 @@
+package kubecost
+
+import (
+	"math"
+	"time"
+)
+
+// SummaryAllocationResponse is a sanitized version of SummaryAllocation, which
+// formats fields and protects against issues like mashaling NaNs.
+type SummaryAllocationResponse struct {
+	Name                   string    `json:"name"`
+	Start                  time.Time `json:"start"`
+	End                    time.Time `json:"end"`
+	CPUCoreRequestAverage  *float64  `json:"cpuCoreRequestAverage"`
+	CPUCoreUsageAverage    *float64  `json:"cpuCoreUsageAverage"`
+	CPUCost                *float64  `json:"cpuCost"`
+	GPUCost                *float64  `json:"gpuCost"`
+	NetworkCost            *float64  `json:"networkCost"`
+	LoadBalancerCost       *float64  `json:"loadBalancerCost"`
+	PVCost                 *float64  `json:"pvCost"`
+	RAMBytesRequestAverage *float64  `json:"ramByteRequestAverage"`
+	RAMBytesUsageAverage   *float64  `json:"ramByteUsageAverage"`
+	RAMCost                *float64  `json:"ramCost"`
+	SharedCost             *float64  `json:"sharedCost"`
+	ExternalCost           *float64  `json:"externalCost"`
+}
+
+// ToResponse converts a SummaryAllocation to a SummaryAllocationResponse,
+// protecting against NaN and null values.
+func (sa *SummaryAllocation) ToResponse() *SummaryAllocationResponse {
+	if sa == nil {
+		return nil
+	}
+
+	return &SummaryAllocationResponse{
+		Name:                   sa.Name,
+		Start:                  sa.Start,
+		End:                    sa.End,
+		CPUCoreRequestAverage:  float64ToResponse(sa.CPUCoreRequestAverage),
+		CPUCoreUsageAverage:    float64ToResponse(sa.CPUCoreUsageAverage),
+		CPUCost:                float64ToResponse(sa.CPUCost),
+		GPUCost:                float64ToResponse(sa.GPUCost),
+		NetworkCost:            float64ToResponse(sa.NetworkCost),
+		LoadBalancerCost:       float64ToResponse(sa.LoadBalancerCost),
+		PVCost:                 float64ToResponse(sa.PVCost),
+		RAMBytesRequestAverage: float64ToResponse(sa.RAMBytesRequestAverage),
+		RAMBytesUsageAverage:   float64ToResponse(sa.RAMBytesUsageAverage),
+		RAMCost:                float64ToResponse(sa.RAMCost),
+		SharedCost:             float64ToResponse(sa.SharedCost),
+		ExternalCost:           float64ToResponse(sa.ExternalCost),
+	}
+}
+
+func float64ToResponse(f float64) *float64 {
+	if math.IsNaN(f) || math.IsInf(f, 0) {
+		return nil
+	}
+
+	return &f
+}
+
+// SummaryAllocationSetResponse is a sanitized version of SummaryAllocationSet,
+// which formats fields and protects against issues like marshaling NaNs.
+type SummaryAllocationSetResponse struct {
+	SummaryAllocations map[string]*SummaryAllocationResponse `json:"allocations"`
+	Window             Window                                `json:"window"`
+}
+
+// ToResponse converts a SummaryAllocationSet to a SummaryAllocationSetResponse,
+// protecting against NaN and null values.
+func (sas *SummaryAllocationSet) ToResponse() *SummaryAllocationSetResponse {
+	if sas == nil {
+		return nil
+	}
+
+	sars := make(map[string]*SummaryAllocationResponse, len(sas.SummaryAllocations))
+	for k, v := range sas.SummaryAllocations {
+		sars[k] = v.ToResponse()
+	}
+
+	return &SummaryAllocationSetResponse{
+		SummaryAllocations: sars,
+		Window:             sas.Window.Clone(),
+	}
+}
+
+// SummaryAllocationSetRangeResponse is a sanitized version of SummaryAllocationSetRange,
+// which formats fields and protects against issues like marshaling NaNs.
+type SummaryAllocationSetRangeResponse struct {
+	Step                  time.Duration                   `json:"step"`
+	SummaryAllocationSets []*SummaryAllocationSetResponse `json:"sets"`
+	Window                Window                          `json:"window"`
+}
+
+// ToResponse converts a SummaryAllocationSet to a SummaryAllocationSetResponse,
+// protecting against NaN and null values.
+func (sasr *SummaryAllocationSetRange) ToResponse() *SummaryAllocationSetRangeResponse {
+	if sasr == nil {
+		return nil
+	}
+
+	sasrr := make([]*SummaryAllocationSetResponse, len(sasr.SummaryAllocationSets))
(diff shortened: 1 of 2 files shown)


pkg/kubecost/summaryallocation_json.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"math"
+	"time"
+)
+
+// SummaryAllocationResponse is a sanitized version of SummaryAllocation, which
+// formats fields and protects against issues like mashaling NaNs.
+type SummaryAllocationResponse struct {
+	Name                   string    `json:"name"`
```
nik-kc, line 11: Taking this as a learning opportunity! This comment and those that follow will primarily be questions to help me learn the product, and a bit of Go in the process.

I'm assuming that `json:"name"` is how you specify how a given value should be encoded/decoded to/from JSON?

pkg/kubecost/summaryallocation_json.go
```diff
@@ -0,0 +1,116 @@
...
+		ExternalCost:           float64ToResponse(sa.ExternalCost),
+	}
+}
+
+func float64ToResponse(f float64) *float64 {
+	if math.IsNaN(f) || math.IsInf(f, 0) {
+		return nil
+	}
+
+	return &f
```
nik-kc, line 59: Do `*` and `&` in Go behave the same way as in C++? i.e. pointers and references?

pkg/kubecost/summaryallocation_json.go
nik-kc: In addition, it seems like Go has native JSON encoding/decoding capabilities. If so, that's super neat!

pkg/kubecost/summaryallocation_json_test.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"encoding/json"
+	"math"
+	"testing"
+	"time"
+
+	"github.com/opencost/opencost/pkg/util/timeutil"
+)
+
+func TestSummaryAllocationSetRangeResponse_MarshalJSON(t *testing.T) {
```
nik-kc, line 12: Pretty straightforward test, looks good to me!

pkg/kubecost/summaryallocation_json_test.go
```diff
@@ -0,0 +1,129 @@
...
+	// Set a 1-day (start, end)
+	s := time.Date(2023, time.March, 13, 0, 0, 0, 0, time.UTC)
+	e := s.Add(timeutil.Day)
+
+	// Set some basic numbers that can be used to asset accuracy later
+	adjustment := -0.14
+	bytes := 2183471523842.00
+	cores := 2.78
+	cost := 12.50
+	gpus := 1.00
```
nik-kc, line 22: Is it possible to have non-integer values for number of GPUs, cores, etc.?

nik-kc (approved):
No changes requested, just a few comments to help me get acquainted with the product and Go.

pkg/kubecost/summaryallocation_json.go
michaelmdresser: Yep, you've got the idea exactly. Go does indeed support JSON natively, even if you don't specify `json:"name"` (this is called a "struct tag"). It's worth noting that `encoding/json` is only able to marshal/unmarshal into exported (capitalized) fields by default. Otherwise you need to write a custom marshal/unmarshal function.

Proper docs here are pretty good: https://pkg.go.dev/encoding/json#Marshal

pkg/kubecost/summaryallocation_json.go
michaelmdresser: Yes, `*` is a pointer deref operator and `&` is a reference (get a pointer to the thing) operator. Additionally, pointer types are `*type` e.g. `*int` is the type of a pointer to an int.

pkg/kubecost/summaryallocation_json_test.go
michaelmdresser: Certainly! I might only use half a CPU core over the span of an hour. Measurement is also imprecise, so things like averages may always report fractional values even if the idea of consuming half a RAM byte doesn't quite sound right. GPUs are particularly weird here because K8s supports "fractional" GPUs.

pkg/kubecost/summaryallocation_json.go
```diff
@@ -0,0 +1,116 @@
...
+		PVCost:                 float64ToResponse(sa.PVCost),
+		RAMBytesRequestAverage: float64ToResponse(sa.RAMBytesRequestAverage),
+		RAMBytesUsageAverage:   float64ToResponse(sa.RAMBytesUsageAverage),
+		RAMCost:                float64ToResponse(sa.RAMCost),
+		SharedCost:             float64ToResponse(sa.SharedCost),
+		ExternalCost:           float64ToResponse(sa.ExternalCost),
+	}
+}
+
+func float64ToResponse(f float64) *float64 {
```
michaelmdresser, line 54: Is there a reason to define a new version of this? I already see `formatFloat64ForResponse` in `pkg/kubecost/allocation_json.go`.

pkg/kubecost/summaryallocation_json.go
```diff
@@ -0,0 +1,116 @@
+package kubecost
+
+import (
+	"math"
+	"time"
+)
+
+// SummaryAllocationResponse is a sanitized version of SummaryAllocation, which
+// formats fields and protects against issues like mashaling NaNs.
+type SummaryAllocationResponse struct {
```
michaelmdresser, line 10: Thoughts on calling this ...Response vs. ...JSON (like the existing `AllocationJSON`)? I don't have a strong opinion.

pkg/kubecost/summaryallocation_json.go
```diff
@@ -0,0 +1,116 @@
...
+	Name                   string    `json:"name"`
+	Start                  time.Time `json:"start"`
+	End                    time.Time `json:"end"`
```
michaelmdresser, line 13: `type AllocationJSON` types these as `string` rather than `time.Time`. I'm not sure why this is, but maybe there was a good reason for it and we should follow the convention here?

michaelmdresser (approved):
Meant to approve, none of my feedback hardblocks. Also you need to sign off

pkg/kubecost/summaryallocation_json.go
nik-kc: Very cool! I'll definitely take a look through those docs.

pkg/kubecost/summaryallocation_json.go
nik-kc: Intuitive! I like it.

pkg/kubecost/summaryallocation_json_test.go
nik-kc: Well when you explain it that way, it makes perfect sense! Thank you for taking the time to explain everything to me!

nik-kc (approved):

pkg/kubecost/summaryallocation_json.go
nikovacevic: The other func spends resources enforcing rounding after six decimals, which I did not want to do here. But I also didn't want to pull out rounding from there, in case it was important. (I'd prefer to do the latter... what do you think?)

pkg/kubecost/summaryallocation_json.go
nikovacevic: My feeling was just that the type could be used with other protocols, perhaps... I dunno. Today it's only for JSON. I don't feel strongly about it either.

pkg/kubecost/summaryallocation_json.go
nikovacevic: I will reach out to @saweber to ask about that. But `AllocationJSON` also uses the `Window` type, which encodes `time.Time`s, so even internal to `AllocationJSON` there are multiple conventions going on.

Trying to let the standard JSON marshaler do its thing, unless strictly necessary to protect against errors.

pkg/kubecost/summaryallocation_json.go
michaelmdresser: Gotcha. Fine as-is, IMO.

pkg/kubecost/summaryallocation_json.go
michaelmdresser: `Window` actual has a custom marshaler which ensures the `time.RFC3339` format as well as our weird string `"null"` return: https://github.com/opencost/opencost/blob/0ada30f9653a6fdcfa112c4f142c96aa7bb1fd17/pkg/kubecost/window.go#L482-L496

pkg/kubecost/summaryallocation_json.go
michaelmdresser: Not enforcing precision is a good-enough reason to be honest. I agree that that functionality from the other helper is the ideal case here. Not blocking, of course, so take as you will. Maybe if you talk with Steven about time marshaling you can cover the other float formatter and its precision enforcement.

pkg/kubecost/summaryallocation_json.go
saweber: I did not do anything special with `AllocationJSON` other than rely on the default JSON encoding for `type Window`. I naively relied on Window marshaling itself correctly.

@michaelmdresser interesting find!

pkg/kubecost/summaryallocation_json.go
saweber: I think response is a better name - it communicates the intent better and leaves us open for alternative encodings/structures for the future.

Not a big deal either way.

pkg/kubecost/summaryallocation_json.go
saweber: https://github.com/opencost/opencost/blob/675a6af3f8ccb2e289fdc61d342821effd7c06cb/pkg/kubecost/allocation_json.go#L108

saweber: The reason for enforcing precision was for making the response look as identical as possible (and intentionally naively assuming that that is correct behavior). (on formatFloatFor64Response)

saweber (approved):

pkg/kubecost/summaryallocation_json.go
nikovacevic: Wowow very interesting. (Makes sense, of course... the Window fields are private. SMDH.)

PR #1769, opened 2023-03-15
Make configs more generic

ausmith:
## What does this PR change?
* Makes the templated configs more generic
* Removes company identifying data that is long since dead internally

* No

* It won't

* No

## How was this PR tested?
* N/A

* No

* Not a release impacting change, only modifying the configuration templates

### DIFF
--- a/configs/gcp.json
+++ b/configs/gcp.json
@@ -5,7 +5,7 @@
     "spotCPU": "0.006655",
     "RAM": "0.004237",
     "spotRAM": "0.000892",
-    "projectID": "guestbook-227502",
+    "projectID": "my-awesome-project",
     "storage": "0.00005479452",
     "zoneNetworkEgress": "0.01",
     "regionNetworkEgress": "0.01",
--- a/configs/gcp.json
+++ b/configs/gcp.json
@@ -5,7 +5,7 @@
     "spotCPU": "0.006655",
     "RAM": "0.004237",
     "spotRAM": "0.000892",
-    "projectID": "my-awesome-project",
+    "projectID": "todo",
     "storage": "0.00005479452",
     "zoneNetworkEgress": "0.01",
     "regionNetworkEgress": "0.01",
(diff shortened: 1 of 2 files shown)


configs/gcp.json
```diff
@@ -5,7 +5,7 @@
     "spotCPU": "0.006655",
     "RAM": "0.004237",
     "spotRAM": "0.000892",
-    "projectID": "guestbook-227502",
+    "projectID": "my-awesome-project",
```
michaelmdresser: Is there a valid string we can provide that has invalid GCP project ID characters to guarantee no overlap with a valid ID? The same question goes for all fields; having something that is too short to be valid also counts, I think!

michaelmdresser (commented):
You'll need to set up signoff for the DCO check to pass (requirement from CNCF to have it as a blocker). You can use the provided instructions in the DCO workflow link on this PR. In the longer term, I recommend setting up some sort of automation in your Git-interaction tool of choice to auto-signoff all commits or just commits to your fork of `opencost`.

configs/gcp.json
ausmith: So GCP project IDs must be 6-30 characters in length [source](https://cloud.google.com/resource-manager/docs/creating-managing-projects). AWS S3 bucket names must be between 3 and 63 characters in length [source](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html). And unfortunately AWS Athena table names can be anything less than 255 characters [source](https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html). Values chosen for each accordingly.

nikovacevic:
cc @AjayTripathy can you confirm this won't have any unintended consequences?

michaelmdresser (approved):
Definitely should have Ajay take a look, but this LGTM. Thanks for addressing feedback as well as possible!

AjayTripathy:
its fine, thanks!

AjayTripathy (approved):

PR #1779, opened 2023-03-17
Make "/var/configs" a const and create new env helper called GetCostAnalyzerVolumeMountPath

nikovacevic:
## What does this PR change?
* Makes `"/var/configs"` a const that is exported by the `env` package to unify implementations and prevent typos
* Adds a more clear helper function for those seeking to get the mount path for the cost analyzer volume: `GetCostAnalyzerVolumeMountPath`

* Required for https://github.com/kubecost/kubecost-cost-model/pull/1303

## How was this PR tested?
* Builds work
* Unit tests pass
* Manually, everything looks right in cluster logs

* No

### DIFF
--- a/pkg/cmd/agent/agent.go
+++ b/pkg/cmd/agent/agent.go
@@ -181,7 +181,7 @@ func Execute(opts *AgentOpts) error {
 
 	clusterCache.SetConfigMapUpdateFunc(watchConfigFunc)
 
-	configPrefix := env.GetConfigPathWithDefault("/var/configs/")
+	configPrefix := env.GetConfigPathWithDefault(env.DefaultConfigMountPath)
 
 	// Initialize cluster exporting if it's enabled
 	if env.IsExportClusterCacheEnabled() {
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -99,6 +99,8 @@ const (
 	regionOverrideList = "REGION_OVERRIDE_LIST"
 )
 
+const DefaultConfigMountPath = "/var/configs"
+
 var offsetRegex = regexp.MustCompile(`^(\+|-)(\d\d):(\d\d)$`)
 
 func IsETLReadOnlyMode() bool {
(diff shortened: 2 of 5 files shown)


saweber (approved):

pkg/cmd/agent/agent.go
```diff
@@ -181,7 +181,7 @@ func Execute(opts *AgentOpts) error {
 
 	clusterCache.SetConfigMapUpdateFunc(watchConfigFunc)
 
-	configPrefix := env.GetConfigPathWithDefault("/var/configs/")
+	configPrefix := env.GetConfigPathWithDefault(env.DefaultConfigMountPath)
```
michaelmdresser, line 184: Now that we have `GetCostAnalyzerVolumeMountPath`, why keep using the `WithDefault` version in places like this where we know we want the mount path?

pkg/env/costmodelenv.go
```diff
@@ -99,6 +99,8 @@ const (
 	regionOverrideList = "REGION_OVERRIDE_LIST"
 )
 
+const DefaultConfigMountPath = "/var/configs"
```
michaelmdresser, line 102: I'm not sure that this should be exported, because theoretically it should only be accessed via one of the `env.Get` function. Were you thinking along different lines?

pkg/cmd/agent/agent.go
nikovacevic: Just trying to keep the change set minimally disruptive tbh. 1:1 string-for-const seemed perfectly unlikely to break anything. Felt that any more was scope creep -- we need to return to this question in the future. I just don't have time to revamp it all thoughtfully today.

nikovacevic: Really just trying to get https://github.com/kubecost/kubecost-cost-model/pull/1303 out the door!

nikovacevic: And make a modest improvement here.

pkg/env/costmodelenv.go
nikovacevic: In the long term, I'd like that. But for now, we need to just export that precise string, otherwise we risk breaking behavior. And I think this is just as good, if not better than:
```
func GetDefaultConfigMountPath() string {
    return "/var/configs"
}
```

My thought is:  today, a LOT of places are hardcoded to the string `"/var/configs"` and if we make that `GetConfigPathWithDefault("/var/configs")` then there's a chance we change behavior on some clusters.

Is that a problem? Definitely. Am I in a position to solve it today? I think... no, unfortunately.

nikovacevic: Propose we revisit later?

michaelmdresser (approved):
Minimal changeset does make sense, and we do need to keep progress rolling elsewhere.

PR #1791, opened 2023-03-21
Remove AppVersion for Version

mbolt35:
## What does this PR change?
* Removes the APP_VERSION env var usage, replaces with `version.Version` for up to date error reporting.

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -1482,7 +1482,7 @@ func Initialize(additionalConfigWatchers ...*watcher.ConfigMapWatcher) *Accesses
 
 	var err error
 	if errorReportingEnabled {
-		err = sentry.Init(sentry.ClientOptions{Release: env.GetAppVersion()})
+		err = sentry.Init(sentry.ClientOptions{Release: version.FriendlyVersion()})
 		if err != nil {
 			log.Infof("Failed to initialize sentry for error reporting")
 		} else {
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -10,8 +10,6 @@ import (
 )
 
 const (
-	AppVersionEnvVar = "APP_VERSION"
-
 	AWSAccessKeyIDEnvVar     = "AWS_ACCESS_KEY_ID"
 	AWSAccessKeySecretEnvVar = "AWS_SECRET_ACCESS_KEY"
 	AWSClusterIDEnvVar       = "AWS_CLUSTER_ID"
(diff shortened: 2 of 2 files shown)


pkg/costmodel/router.go
```diff
@@ -1482,7 +1482,7 @@ func Initialize(additionalConfigWatchers ...*watcher.ConfigMapWatcher) *Accesses
 
 	var err error
 	if errorReportingEnabled {
-		err = sentry.Init(sentry.ClientOptions{Release: env.GetAppVersion()})
+		err = sentry.Init(sentry.ClientOptions{Release: version.FriendlyVersion()})
```
michaelmdresser, line 1485: @nealormsbee does our Sentry setup have any particular requirements that this might break?

pkg/env/costmodelenv.go
```diff
@@ -171,12 +169,6 @@ func GetMetricsConfigmapName() string {
 	return Get(MetricsConfigmapName, "metrics-config")
 }
 
-// GetAWSAccessKeyID returns the environment variable value for AWSAccessKeyIDEnvVar which represents
-// the AWS access key for authentication
-func GetAppVersion() string {
-	return Get(AppVersionEnvVar, "1.91.0-rc.0")
-}
-
```
michaelmdresser, line 179: Since we're nuking this (which I think is the right call due to build args existing), would you please open a matching PR in the Helm chart that removes the env var from there?

pkg/costmodel/router.go
michaelmdresser: E.g. I'm not sure if `FriendlyVersion` is or is not prefixed with `v`.

pkg/env/costmodelenv.go
mbolt35: I would, but it appears that was already done at some point... 😬

michaelmdresser: Good. @nealormsbee may want to check that sentry setup anyway then eh? We may have been dropping stuff but shouldn't in the future.

michaelmdresser (approved):
LGTM, but we should make sure Neal gets a look just in case.

pkg/costmodel/router.go
mbolt35: There's two separate apps for Sentry, one for frontend and backend. We've been reporting backend errors/panics for the last year apparently using the same version `v1.91.0rc-0`. This will just allow us to actually tie a specific version to the error/panic.

nealormsbee: Definitely won't break the FE project. Should not cause any issue for the BE project either. Sentry will accept with or without the `v` prefix. I learned this the hard way, because the FE was using one format and our script to build sourcemaps was using another, creating two separate releases. One received errors and the other had the sourcemaps :facepalm:

Looks like they don't even have to be semver: https://docs.sentry.io/platforms/javascript/configuration/releases/#bind-the-version

>The value can be arbitrary, but we recommend either of these naming strategies:

nealormsbee (approved):
All LGTM. Shouldn't impact build scripts or the UI, and looks like it'll just make OpenCost report crashes to the correct release.

PR #1792, opened 2023-03-22
Add functionality to export opencost data to a CSV file in cloud storage.

r2k1:
## What does this PR change?

This pull request adds a background worker that periodically updates a CSV file in a cloud storage of choice (such as Azure Blob, Amazon S3, Google Cloud Storage) to allow for the storage of pre-aggregated data beyond the time limit set by Prometheus. The CSV format is user-friendly and easily ingested into SQL DB, Excel Spreadsheet or any ETL pipeline, allowing for easy aggregation of data from multiple clusters. 

The exported data is aggregated to a day granularity and the worker is set to run at 00:10 UTC every day, as well as at launch time to handle any previous failures. The first export includes all available data in Prometheus, while subsequent launches only export data for dates not already present in the existing export file.

An example of CSV file exported:

| Date     | Namespace   | ControllerKind | ControllerName                       | Pod                                                  | Container      | CPUCoreUsageAverage | CPUCoreRequestAverage | RAMBytesUsageAverage | RAMBytesRequestAverage | NetworkReceiveBytes | NetworkTransferBytes | GPUs | PVBytes | CPUCost     | RAMCost     | NetworkCost | PVCost | GPUCost | TotalCost   |
|----------|-------------|----------------|--------------------------------------|------------------------------------------------------|----------------|---------------------|-----------------------|----------------------|------------------------|---------------------|----------------------|------|---------|-------------|-------------|-------------|--------|---------|-------------|
| 20/03/23 | kube-system | deployment     | metrics-server                       | metrics-server-85bb67db6-nnmcc                       | metrics-server | 0.003727224         | 0.044                 | 35668923.73          | 57671680               | 0                   | 0                    | 0    | 0       | 0.036885024 | 0.00221332  | 0           | 0      | 0       | 0.039098344 |
| 20/03/23 | kube-system | daemonset      | csi-azuredisk-node                   | csi-azuredisk-node-fvfbm                             | azuredisk      | 0.000109207         | 0.01                  | 18809545.96          | 20971520               | 0                   | 0                    | 0    | 0       | 0.00838296  | 0.000804844 | 0           | 0      | 0       | 0.009187804 |
| 20/03/23 | prometheus  | deployment     | my-prometheus-prometheus-pushgateway | my-prometheus-prometheus-pushgateway-8c9c8cfdc-5jk9j | pushgateway    | 0.000212164         | 0                     | 23251003.73          | 0                      | 0                   | 0                    | 0    | 0       | 0           | 0           | 0           | 0      | 0       | 0           |
| 20/03/23 | kube-system | daemonset      | ama-logs                             | ama-logs-p89dr                                       | ama-logs       | 0.012029655         | 0.075                 | 295596916.6          | 340787200              | 0                   | 0                    | 0    | 0       | 0.0628722   | 0.013078711 | 0           | 0      | 0       | 0.075950911 |

[full-example.csv](https://github.com/opencost/opencost/files/11035209/test-opencost-export.4.csv)

### Configuration

To enable export, set the COST_EXPORT_FILE environment variable to the path of the file. The file can be a local file or a storage object in one of the clouds. Here are some usage examples:

| Provider |  Example |
|------|----------|
| Azure Blob Storage | `COST_EXPORT_FILE=https://azblobaccount.blob.core.windows.net/containername/path/to/file.csv` |
| AWS S3 | `COST_EXPORT_FILE=s3://bucketname/path/to/file.csv ` |
| Google Cloud Storage | `COST_EXPORT_FILE=gs://bucket-name/path/to/file.csv` |
| Local File |  `COST_EXPORT_FILE=export.csv ` |

Cloud Access is configured through the default mechanism provided by cloud SDKs. For example, Azure blob storage can be configured through AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET environment variables, or workload identity. S3 can be configured through AWS_ACCESS_KEY_ID, AWS_DEFAULT_REGION, AWS_SECRET_ACCESS_KEY, and so on.

## How was this PR tested?

Some tests has been added.

For manual testing using Azure Blob Storage I've been using prometheus + opencost setup running in a AKS cluster.

```sh
kubectl port-forward --namespace prometheus service/my-prometheus-server 9093:80
COST_EXPORT_FILE=export.csv go run ./cmd/costmodel/. 
COST_EXPORT_FILE=s3://bucket-name/path/to/file.csv go run ./cmd/costmodel/. 
COST_EXPORT_FILE=gs://bucket-name/path/to/file.csv go run ./cmd/costmodel/. 
COST_EXPORT_FILE=https://azblobaccount.blob.core.windows.net/containerName/path/to/file.csv go run ./cmd/costmodel/. 
```

The export can be run locally using credentials set by CLI by relevant cloud provider.

Once the configuration is established the documentation should be updated.

I have no permissions.

## TODO. Discussion points.
- Configuration. Cloud storage needs to be configured. I used existing code, but it's not used anywhere and there is no example of initialisation or documentation how it should be configured.
- Launching. My idea is to spawn an addition goroutine on start up. (If export path is provided) using `go UpdateCSVWorker(...)` method.
- CSV Format. I added some columns that I think is useful. But the format is flexible and code designed that it should be easy to change in future without breaking.
- Data aggregation. Currently the data is exported at the container level. But it's likely to be too noisy for most users. I think it may be grouped by controller, but it may posses some challenges if users want to aggregate data by certain labels.  Such aggregation will require some refactoring as existing logic is located at HTTP Handler level.


AjayTripathy:
Incredible work @r2k1 ! Going to review this tonight!

AjayTripathy:
> Configuration. Cloud storage needs to be configured. I used existing code, but it's not used anywhere and there is no example of initialisation or documentation how it should be configured.

@Sean-Holcomb you mentioned you were going to open source some more configuration code for cloud right?

pkg/costmodel/allocation.go
```diff
@@ -252,6 +257,26 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	return result, nil
 }
 
+// DateRange checks the data (up to 90 days in the past), and returns the oldest and newest sample timestamp from opencost scraping metric
+// it supposed to be a good indicator of available allocation data
+func (cm *CostModel) DateRange() (time.Time, time.Time, error) {
```
AjayTripathy, line 286: I'm wondering if a new metric here makes sense vs. just using an opencost specific metric, for example node_total_hourly_cost or similar to check the oldest/newest samples?

@nikovacevic I'm not the expert here... any thoughts?

r2k1: I don't think it's matter which metric is used in this case. Can be anything.

`up` is a prometheus built-in metric. It's sampled on every scrape. No scrape = no data.

https://prometheus.io/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series


It can be further generalized:
```
max_over_time(timestamp(group(up > 0))[1d:1h])
```

Sean-Holcomb:
Those storage configs are collectively called BucketStorage and you can see how they are initialized in pkg/storage/bucketstorage.go. 
https://github.com/opencost/opencost/blob/52fa38faaae046d3e546830c3ae9973bbce83181/pkg/storage/bucketstorage.go#L28-L30
Essentially you just need a yaml file in the formates at that [link](https://thanos.io/tip/thanos/storage.md/).
From there you just need to load up the file and plug it in
https://github.com/opencost/opencost/blob/52fa38faaae046d3e546830c3ae9973bbce83181/pkg/config/configmanager.go#L62-L66
We usually mount the file via secret

pkg/costmodel/allocation.go
AjayTripathy: Totally understand; I think some managed promQL systems do not have an "up" metric necessarily-- so maybe one that is emitted by the cost-model makes the most sense?

r2k1: Is there an opencost metric that is guaranteed to be produced (or required to generate a meaningful allocation data)?

r2k1:
I'm trying to think how users are going to configure it. 

What a user guide would look like?

It looks like next steps to me:
- create credentials
- create k8s secret with the credentials and storage configuration
- attach secret to the pod volume
- add ENV variable to the deployment with a path to the storage config file
- add ENV variable to the deployment with a path to the export file.

Am I missing something? Is there an easier way to achieve it?
For example, it seems easier if everything can be set through ENV variables.

mattray:
> I'm trying to think how users are going to configure it.
> 

What if it were simply a Kubernetes CronJob? It could get added to the Helm chart with the configuration exposed accordingly. It's a bit of a refactor, but adding it to the `cost-model` CLI to run it would make it pretty stand-alone. Just a thought.

AjayTripathy:
I don't know if maintaining cron jobs is within the scope of the project-- would prefer users find a way to configure the repetition themselves.

mattray:
@AjayTripathy makes sense. I just figured since Kubernetes has CronJobs built in it's one less moving part for the cost-model agent.

r2k1:
I believe that CronJob is an excellent option as it handles retries, provides visibility on failed jobs, and separates execution from the opencost agent, allowing the exporter to fail independently. However, configuring it may be challenging since opencost is currently distributed as a YAML file rather than a helm chart. Users may need to manually create a cronjob and duplicate the configuration in both places. 

Therefore, to make it more user-friendly, I suggest implementing it as a background worker for simplicity.

mattray:
The [OpenCost Helm chart](https://github.com/opencost/opencost-helm-chart) could easily be updated to support this, and it might be the sort of advanced feature more likely to be used by Helm chart users. I'll defer to @r2k1 since he's the implementer.

AjayTripathy:
We've just had a huge pain maintaining cronjobs in the past. I would strongly advise against it, I don't think we want to take it on in the opencost helm project. It feels outside of scope there too.

AjayTripathy:
(proposal would just be to send a POST request to an endpoint when they want to write a file and leave the scheduling outside the scope of opencost)

r2k1:
I think I achieved desired result. I updated the PR description with the way how I see it can be configured.

A few more options to execute it can be added later:
- execute export through a command-line argument, can be used in cronjobs
- an API endpoint to export data for a given time range.

With this PR I tried to focus only on option I believe will be most beneficial for the users as it easiest one to use.

In order to achieve it I had to implement my own simple wrapper for the cloud providers.

There is existing pkg/storage, but it has a few flaws:
- It's hard to configure. I want to configure the export through a single ENV variable.
- It's out of date
- It requires loading files in memory. A year worth of data can produce a large file to load it in memory.
- I encountered a few bugs when trying to use it.

It seems like it can be replaced by https://github.com/thanos-io/objstore . But it currently only used for loading a config file and such feature isn't documented so I wonder whether it's worth touching. And I didn't figure out how to make it configurable in the desired way.

r2k1:
@AjayTripathy do you want me to change anything else?

pkg/costmodel/allocation.go
AjayTripathy: yes-- sorry for the late response. it's container_memory_allocation_bytes https://github.com/opencost/opencost/blob/13f6525c09fd9427d18afea5efb26b3e27c7866d/pkg/costmodel/metrics.go#L205

AjayTripathy:
> @AjayTripathy do you want me to change anything else?

Nothing from my end... would like one more owner (@Sean-Holcomb who owns cloud integrations to take a pass), but I support in general.

AjayTripathy (approved):

r2k1:
Hi @Sean-Holcomb did you have any chance to take a look at it?

Sean-Holcomb (approved):
Looks good, let's get it in

PR #1804, opened 2023-03-24
Weekly allocation and asset support

saweber:
## What does this PR change?
* support for window and step to accept 'w' for 'week'
* prevent accumulateByWeek from producing an error is the data is already accumulated into a week

* yes; 
https://github.com/opencost/opencost/pull/1794
https://github.com/kubecost/kubecost-cost-model/pull/1315/files

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/opencost/opencost/pkg/log"
 	"github.com/opencost/opencost/pkg/util"
+	"github.com/opencost/opencost/pkg/util/timeutil"
 )
 
 // TODO Clean-up use of IsEmpty; nil checks should be separated for safety.
@@ -2253,6 +2254,10 @@ func (asr *AllocationSetRange) accumulateByMonth() (*AllocationSetRange, error) 
 }
 
 func (asr *AllocationSetRange) accumulateByWeek() (*AllocationSetRange, error) {
+	if len(asr.Allocations) > 0 && asr.Allocations[0].Window.Duration() == timeutil.Week {
+		return asr, nil
+	}
+
 	var toAccumulate *AllocationSetRange
 	result := NewAllocationSetRange()
 	for i, as := range asr.Allocations {
--- a/pkg/kubecost/window.go
+++ b/pkg/kubecost/window.go
@@ -22,8 +22,8 @@ const (
 )
 
 var (
-	durationRegex       = regexp.MustCompile(`^(\d+)(m|h|d)$`)
-	durationOffsetRegex = regexp.MustCompile(`^(\d+)(m|h|d) offset (\d+)(m|h|d)$`)
+	durationRegex       = regexp.MustCompile(`^(\d+)(m|h|d|w)$`)
+	durationOffsetRegex = regexp.MustCompile(`^(\d+)(m|h|d|w) offset (\d+)(m|h|d|w)$`)
 	offesetRegex        = regexp.MustCompile(`^(\+|-)(\d\d):(\d\d)$`)
 	rfc3339             = `\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\dZ`
 	rfcRegex            = regexp.MustCompile(fmt.Sprintf(`(%s),(%s)`, rfc3339, rfc3339))
@@ -46,11 +46,6 @@ func RoundBack(t time.Time, resolution time.Duration) time.Time {
 // in the given time's timezone.
 // e.g. 2020-01-01T12:37:48-0700, 24h = 2020-01-02T00:00:00-0700
 func RoundForward(t time.Time, resolution time.Duration) time.Time {
-	// if the duration is a week - roll forward to the following Sunday
-	if resolution == timeutil.Week {
-		return timeutil.RoundToStartOfFollowingWeek(t)
-	}
-
 	back := RoundBack(t, resolution)
 	if back.Equal(t) {
 		// The given time is exactly a multiple of the given resolution
--- a/pkg/util/timeutil/timeutil.go
+++ b/pkg/util/timeutil/timeutil.go
@@ -112,7 +112,8 @@ var unitMap = map[string]int64{
 	"s":  int64(time.Second),
 	"m":  int64(time.Minute),
 	"h":  int64(time.Hour),
-	"d":  int64(time.Hour * 24),
+	"d":  int64(Day),
+	"w":  int64(Week),
 }
 
 // goParseDuration is time.ParseDuration lifted from the go std library and enhanced with the ability to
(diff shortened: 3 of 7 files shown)


pkg/kubecost/allocation.go
```diff
@@ -2,6 +2,7 @@ package kubecost
 
 import (
 	"fmt"
+	"github.com/opencost/opencost/pkg/util/timeutil"
```
ameijer: consider grouping this with the other opencost imports right below?

pkg/util/timeutil/timeutil.go
```diff
@@ -267,13 +268,19 @@ func FormatDurationStringDaysToHours(param string) (string, error) {
...
+	}
 	daysFromMonday := int(date.Weekday())
 	return date.Add(-1 * time.Duration(daysFromMonday) * Day)
 }
 
 // RoundToStartOfFollowingWeek creates a new time.Time for the following Monday 00:00 UTC
 func RoundToStartOfFollowingWeek(t time.Time) time.Time {
 	date := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
+	if date.Weekday() == time.Sunday {
+		return date
```
ameijer: @saweber if I am trying to use this func on a sunday afternoon date, do we want it to return sunday 7 days from now?

pkg/util/timeutil/timeutil.go
```diff
@@ -267,13 +268,19 @@ func FormatDurationStringDaysToHours(param string) (string, error) {
...
 func RoundToStartOfWeek(t time.Time) time.Time {
 	date := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
+	if date.Weekday() == time.Sunday {
+		return date
+	}
 	daysFromMonday := int(date.Weekday())
 	return date.Add(-1 * time.Duration(daysFromMonday) * Day)
 }
 
 // RoundToStartOfFollowingWeek creates a new time.Time for the following Monday 00:00 UTC
```
ameijer: should we update these comments/var names? Weekday() has the following underlying implementation: 
```
const (
	Sunday Weekday = iota
	Monday
	Tuesday
	Wednesday
	Thursday
	Friday
	Saturday
)
```
... I think this is returning sunday dates, not mondays?

ameijer (commented):
hi @saweber took a quick pass - what do you think?

pkg/kubecost/allocation.go
saweber: Good catch! There was no reason for it to be there - other than is where IntelliJ autoimported it. :)

pkg/util/timeutil/timeutil.go
saweber: Another good catch! Bad naming on my part, though I believe the functionality is correct (will validate + add tests).

pkg/util/timeutil/timeutil.go
saweber: Good question. The reason for this is to prevent a window=1w/step=1w query for a week into the future.

Some additional unit tests to ensure correct edge case functionality. The intended functionality is that '1w' means the current week, with additional weeks calculating backwards from there.

pkg/kubecost/allocation.go
nikovacevic: We gotta get the IntelliJ crew on `gofmt`

nikovacevic: I have no idea why that IDE uses a different formatter, but `gofmt` is the standard, so the two camps war over imports.

pkg/kubecost/window.go
```diff
@@ -22,8 +22,8 @@ const (
 )
 
 var (
-	durationRegex       = regexp.MustCompile(`^(\d+)(m|h|d)$`)
-	durationOffsetRegex = regexp.MustCompile(`^(\d+)(m|h|d) offset (\d+)(m|h|d)$`)
+	durationRegex       = regexp.MustCompile(`^(\d+)(m|h|d|w)$`)
+	durationOffsetRegex = regexp.MustCompile(`^(\d+)(m|h|d|w) offset (\d+)(m|h|d)$`)
```
nikovacevic, line 26: If we're accepting `w` for durations, we should probably accept it for offsets, too.

nikovacevic: If we're accepting w for durations, we should probably accept it for offsets, too.

pkg/util/timeutil/timeutil.go
nikovacevic: Will echo Alex's concern here -- I would expect this function to return "the following Sunday" if I gave it a Sunday at noon. We should remember that this is a very general package, so others may use these functions for other reasons than the ones we have in mind. Will suggest we change this back, unless there's more context I'm missing.

nikovacevic (changes requested):

pkg/kubecost/allocation.go
saweber: IntelliJ gets thrown off by the whitespace line - and wants to alphabetize. I’ll make a note and figure out how to configure IntelliJ (and get it in here and KCM)

pkg/kubecost/window.go
saweber: @Sean-Holcomb has some specific concerns related to allowing an offset measured in weeks.

ameijer:
@saweber I think the goimports tool can handle the grouping/ordering for you. Perhaps you can configure it to work with intellij?

-------- Original Message --------

pkg/kubecost/allocation.go
nikovacevic: I think Intellij has a `gofmt` option. To format code on save. That should do it.

saweber: IntelliJ was already set to `go fmt`. For some reason, JetBrains' defaults are different than VSCode.

These settings seem to match y'alls preferences. I do not care how imports are sorted - let me know if this is acceptable and if we want to open separate PRs for OC/KCM to store this as project files so we don't have to repeat this for every IntelliJ/GoLand user.
Screenshot 2023-03-28 at 08 51 51

ameijer: that looks great! and yes I think committing these to a file is a great idea.

pkg/kubecost/window.go
Sean-Holcomb: Not so much a concern as it seems like something we want to enable in the prometheus context, my point on line 26 was that `w` should not be an option on for both because this regex is only used for prometheus formatting. At this point I think accepting it for offset is fine for the sake of consistency.

saweber: Discussed with Sean, we will go ahead and add it.

nikovacevic (approved):

PR #1813, opened 2023-03-27
Fix CPU max query by using already-existing irate recording rule

michaelmdresser:
## What does this PR change?
The query before this commit is broken; it looks like a max but is actually an average (or very close to) due to the usage of "rate" before maxing. The rate reduces the resolution to the first and last datapoint.

The recording rule handles this for us. At the time of this commit, the recording rule is set in the Helm values:

```
- expr: avg(irate(container_cpu_usage_seconds_total{container_name!="POD", container_name!=""}[5m])) by (container_name,pod_name,namespace)
  record: kubecost_container_cpu_usage_irate
```

https://github.com/kubecost/cost-analyzer-helm-chart/blob/b04da589a5a8b469fe13f8b25cd21023601ecca9/cost-analyzer/values.yaml#L609-L610

Design doc discussing this problem in more depth and proposing alternative solutions: https://docs.google.com/document/d/1j36iYFWXsG7CsU3H-jODBKL76RkRA5g4Rnu7BRogmSs/edit

https://kubecost.atlassian.net/browse/CORE-126

## How was this PR tested?
Create a Pod like described in Jesse's original bug report:
```
read -r -d '' WL << EOM
apiVersion: v1
kind: Pod
metadata:
  name: stress-ng2
  labels:
    app: stress-ng2
spec:
  containers:
  - image: litmuschaos/stress-ng
    args:
      - -c 1
    imagePullPolicy: IfNotPresent
    name: stress-ng2
    resources:
      requests:
        cpu: 50m
        memory: 40Mi
      limits:
        cpu: 50m
        memory: 40Mi
EOM

echo "${WL}" | kubectl apply -f -
date
```

```
: pod/stress-ng2 created
: Tue Mar 28 01:24:57 PM PDT 2023
```

Wait 10 min

Query APIs, check maxes in request sizing and allocation

```
date
curl -G 'localhost:9090/model/savings/requestSizingV2' \
    -L -s \
    --data-urlencode 'filter=namespace:"default"+container:"stress-ng2"' \
    -d 'window=today' \
    -d 'targetCPUUtilization=1' \
    -d 'targetRAMUtilization=1' \
    | jq '.Recommendations[]' \
    | jq 'del(.monthlySavings, .currentEfficiency)'

Tue Mar 28 02:41:16 PM PDT 2023
{
  "clusterID": "cluster-one",
  "namespace": "default",
  "controllerKind": "pod",
  "controllerName": "stress-ng2",
  "containerName": "stress-ng2",
  "recommendedRequest": {
    "cpu": "51m",
    "memory": "20Mi"
  },
  "latestKnownRequest": {
    "cpu": "50m",
    "memory": "40Mi"
  }
}
```

```
date
curl -G 'localhost:9090/model/allocation' \
    -L -s \
    --data-urlencode 'filter=namespace:"default"+container:"stress-ng2"' \
    -d 'window=today' \
    -d 'accumulate=false' \
    -d 'step=1h' \
    -d 'idle=false' \
    | jq '.data[] | select(. != null) | to_entries[].value' \
    | jq 'del(.properties.labels, .properties.annotations, .properties.providerID, .properties.node)' \
    | jq 'with_entries(select([.key] | inside(["properties", "window", "start", "end", "cpuCoreUsageAverage", "ramByteUsageAverage", "rawAllocationOnly"])))'

Tue Mar 28 02:41:22 PM PDT 2023
{
  "properties": {
    "cluster": "cluster-one",
    "container": "stress-ng2",
    "namespace": "default",
    "pod": "stress-ng2"
  },
  "window": {
    "start": "2023-03-28T20:00:00Z",
    "end": "2023-03-28T21:00:00Z"
  },
  "start": "2023-03-28T20:25:00Z",
  "end": "2023-03-28T21:00:00Z",
  "cpuCoreUsageAverage": 0.02918,
  "ramByteUsageAverage": 8896512,
  "rawAllocationOnly": {
    "cpuCoreUsageMax": 0.05006192877884747,
    "ramByteUsageMax": 8896512
  }
}
{
  "properties": {
    "cluster": "cluster-one",
    "container": "stress-ng2",
    "namespace": "default",
    "pod": "stress-ng2"
  },
  "window": {
    "start": "2023-03-28T21:00:00Z",
    "end": "2023-03-28T22:00:00Z"
  },
  "start": "2023-03-28T21:00:00Z",
  "end": "2023-03-28T21:35:00Z",
  "cpuCoreUsageAverage": 0.02906,
  "ramByteUsageAverage": 8896512,
  "rawAllocationOnly": {
    "cpuCoreUsageMax": 0.05011191291833002,
    "ramByteUsageMax": 8896512
  }
}
```

Prior to this change, the first hourly Allocation block would have a <0.05 max.

* https://app.gitbook.com/o/MQuX6uFwV0j7vIHtR15E/s/cjJbIkEBCZifkHo05tVh/~/changes/159/install-and-configure/install/custom-prom/~/comments/JyuJPE3SHR6Cfg59OiJI

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -74,7 +74,14 @@ const (
 	// This is the subquery equivalent of the above recording rule query. It is
 	// more expensive, but does not require the recording rule. It should be
 	// used as a fallback query if the recording rule data does not exist.
-	queryFmtCPUUsageMaxSubquery = `max(max_over_time(irate(container_cpu_usage_seconds_total{container_name!="POD", container_name!=""}[5m])[%s:1m])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
+	//
+	// The parameter after the colon [:<thisone>] in the subquery affects the
+	// resolution of the subquery.
+	// The parameter after the metric ...{}[<thisone>] should be set to 2x
+	// the resolution, to make sure the irate always has two points to query
+	// in case the Prom scrape duration has been reduced to be equal to the
+	// ETL resolution.
+	queryFmtCPUUsageMaxSubquery = `max(max_over_time(irate(container_cpu_usage_seconds_total{container_name!="POD", container_name!=""}[%s])[%s:%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
 )
 
 // Constants for Network Cost Subtype
@@ -361,7 +368,12 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
 	resCPUUsageMax, _ := resChCPUUsageMax.Await()
 	// If the recording rule has no data, try to fall back to the subquery.
 	if len(resCPUUsageMax) == 0 {
-		queryCPUUsageMax = fmt.Sprintf(queryFmtCPUUsageMaxSubquery, durStr, env.GetPromClusterLabel())
+		// The parameter after the metric ...{}[<thisone>] should be set to 2x
+		// the resolution, to make sure the irate always has two points to query
+		// in case the Prom scrape duration has been reduced to be equal to the
+		// resolution.
+		doubleResStr := timeutil.DurationString(2 * resolution)
+		queryCPUUsageMax = fmt.Sprintf(queryFmtCPUUsageMaxSubquery, doubleResStr, durStr, resStr, env.GetPromClusterLabel())
 		resChCPUUsageMax = ctx.QueryAtTime(queryCPUUsageMax, end)
 		resCPUUsageMax, _ = resChCPUUsageMax.Await()
 
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -22,7 +22,7 @@ const (
 	queryFmtCPUCoresAllocated        = `avg(avg_over_time(container_cpu_allocation{container!="", container!="POD", node!=""}[%s])) by (container, pod, namespace, node, %s)`
 	queryFmtCPURequests              = `avg(avg_over_time(kube_pod_container_resource_requests{resource="cpu", unit="core", container!="", container!="POD", node!=""}[%s])) by (container, pod, namespace, node, %s)`
 	queryFmtCPUUsageAvg              = `avg(rate(container_cpu_usage_seconds_total{container!="", container_name!="POD", container!="POD"}[%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
-	queryFmtCPUUsageMax              = `max(rate(container_cpu_usage_seconds_total{container!="", container_name!="POD", container!="POD"}[%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
+	queryFmtCPUUsageMax              = `max(max_over_time(kubecost_container_cpu_usage_irate{}[%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Thanks for this PR. Would love a fallback to a query if the recording rule does not exist but not 100% sure how to make that work.

michaelmdresser:
> Would love a fallback to a query if the recording rule does not exist but not 100% sure how to make that work.

Prometheus has set operators [1] (`OR` is what we could use) but I believe it would require calculating the expensive subquery every time, giving us no win.

We could try checking for the existence of the recording rule at startup or something like that, but I worry about fragility of the implementation. The other concern is unstable behavior based on environment (if it has or doesn't have the recording rule) which could make support challenging.

Of course, we have to weigh this against the cost of getting all users to adopt the recording rule, particular BYO Prometheus users.

[1] https://prometheus.io/docs/prometheus/latest/querying/operators/#logical-set-binary-operators

michaelmdresser:
@AjayTripathy based on a Slack thread, it seems like BYO Prometheus users don't have our recording rules. What do you think of the support/install overhead of the recording rule for BYO Prom users? I'm wary of the alternative, which is the subquery, because it is stated to have bad performance implications and BYO Prom users tend to be high-scale.

AjayTripathy:
> We could try checking for the existence of the recording rule at startup or something like that, but I worry about fragility of the implementation. 

This seems like the right way to do it. There are a couple of ways we could potentially do this.
1. Check the prometheus config YAML. We already grab this. Drawback is that not all promQL systems have the config available
2. Simply run a single check for data existing via recording rule in the query loop before running the full load of queries and switch based upon its existence. It's an additional query but all we need to do is run 1 and validate. This should always work. We may also want to run this on the Diagnostics page.
3. Some other idea I haven't considered.

I'd lean towards option 2, its one additional query but should be super lightweight. There are some particulars about the first window this data becomes available, but I think it's worth the long-term stability here. wdyt?

michaelmdresser:
@AjayTripathy I've implemented 2 in this PR now. I'm still working on testing, but if you have time for a review pass I would appreciate it.

pkg/costmodel/allocation.go
```diff
@@ -57,6 +56,25 @@ const (
...
+	//
+	// [1] https://prometheus.io/blog/2019/01/28/subquery-support/
+	//
+	// If changing the name of the recording rule, make sure to update the
+	// corresponding diagnostic query to avoid confusion.
+	queryFmtCPUUsageMaxRecordingRule = `max(max_over_time(kubecost_container_cpu_usage_irate{}[%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
+	// This is the subquery equivalent of the above recording rule query. It is
+	// more expensive, but does not require the recording rule. It should be
+	// used as a fallback query if the recording rule data does not exist.
+	queryFmtCPUUsageMaxSubquery = `max(max_over_time(irate(container_cpu_usage_seconds_total{container_name!="POD", container_name!=""}[5m])[%s:1m])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
```
AjayTripathy: Is there a good reason to hardcode window and resolution here? All other queries inherit the :1m resolution from whatever you specify as your resolution in ETL_RESOLUTION_SECONDS . I understand that this leads to a diff with the recording rule, but even if this leads to inaccuracy I would lean towards that + a description of the inaccuracy since resolution is a big component in scaling here.

michaelmdresser: Great point. The latest commit addresses this, would you let me know what you think of the implementation and comment update?

Sean-Holcomb (approved):
Looks robust with the backup

PR #1814, opened 2023-03-28
[OPENCOST-1753] Derive Default Cluster Name from CLUSTER_ID var

### LINKED ISSUE #1753: logs display "INF Unable to sniff out cluster ID, perhaps set $AWS_CLUSTER_ID to force one"
**Describe the bug**
Log error: 
```INF Unable to sniff out cluster-ID, perhaps set $AWS_CLUSTER_ID to force one```
It keeps appearing on the logs 

there is no info on the default installation about this environment variable:
https://raw.githubusercontent.com/opencost/opencost/develop/kubernetes/opencost.yaml

I guess if it's not set by default should be like AWS_CLUSTER_ID = CLUSTER_ID then we need just 1 env variable as both will be the same:
- here you can see that AWS_CLUSTER_ID will be equal to ClusterIdEnvVar 
https://github.com/opencost/opencost/blob/3507d95d08803ef541ce3f147b40db500d10c1b6/pkg/cloud/awsprovider.go#L300

- ClusterIdEnvVar is equal to CLUSTER_ID 
https://github.com/opencost/opencost/blob/3507d95d08803ef541ce3f147b40db500d10c1b6/pkg/env/costmodelenv.go#L25

I guess this is the code part affected:
https://github.com/opencost/opencost/blob/3507d95d08803ef541ce3f147b40db500d10c1b6/pkg/cloud/awsprovider.go#L1344

**To Reproduce**
Install Open cost check into the log error 

**Expected behavior**
N/A

**Screenshots**
N/A

Opencost v1.100

**Additional context**
N/A

ameijer:
## What does this PR change?
* Utilizes the CLUSTER_ID env var to set an AWS cluster name if the AWS_CLUSTER_ID var is not set.

* None 

* Users will now be able to use the CLUSTER_ID var to end their cluster name. This opens the door for the eventual deprecation of the AWS_CLUSTER_ID variable 

* Closes #1753 

## How was this PR tested?
* A local build was run, and cluster name was checked when the pricing-config configmap did not have any cluster name set. 

* AWS_CLUSTER_ID was not documented, so it is not likely

* Release 102

### DIFF
--- a/pkg/cloud/awsprovider.go
+++ b/pkg/cloud/awsprovider.go
@@ -1355,9 +1355,10 @@ func (awsProvider *AWS) ClusterInfo() (map[string]string, error) {
 		if awsClusterID != "" {
 			log.Infof("Returning \"%s\" as ClusterName", awsClusterID)
 			clusterName = awsClusterID
+			log.Warnf("Warning - %s will be deprecated in a future release. Use %f instead", env.AWSClusterIDEnvVar, env.ClusterIDEnvVar)
 		} else {
-			log.Infof("Unable to sniff out cluster ID, perhaps set $%s to force one", env.AWSClusterIDEnvVar)
-			clusterName = "AWS Cluster #1"
+			clusterName = env.GetClusterID()
+			log.Infof("Setting cluster name to %s from %s ", clusterName, env.ClusterIDEnvVar)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Sean-Holcomb (approved):
This looks good, love to depreciate these duplicate values

PR #1816, opened 2023-03-29
Simplify configuration logic

r2k1:
## What does this PR change?

Simplifies the configuration logic by removing the option to load configuration from cloud storage. This option is not documented, and its complexity makes it difficult to modify configuration options.

In addition to simplifying the logic, this change also reduces the number of dependencies required.

This change should not affect users, unless they are currently using the undocumented option to load configuration from cloud storage.

## How was this PR tested?
According to the docs there is not much can be configured. So it's not clear how and what needs to be tested.

No


AjayTripathy:
@r2k1 I think there may be a reason we did it this way for importers of the library... @Sean-Holcomb any thoughts?

Sean-Holcomb (changes requested):
As is this PR cannot be approved. The benefits of reduced dependency is irrelevant as we move more cloud integration functionality into OpenCost is not really relevant as many of those dependencies would be returning in the near future.  I am interested in understanding the intention behind these changes. "its complexity makes it difficult to modify configuration options" what kind of changes do you want to make here and how does cloud backup impede them. If this discussion is happening elsewhere please share the link.

AjayTripathy:
Hey @r2k1 -- sync'd with @Sean-Holcomb on this a little. We understand that this configuration option could use more documentation. It's difficult for us to change this on short notice since it's a public API and opencost is also used as a library in certain cases-- we'd need to give heads up that config is changing at a minimum. Happy to discuss more during our sync.

r2k1:
I had some issues trying to understand how it works when exploring configuration options for my feature. I made a low effort change to see if it would spark some discussion or someone would share what it shouldn't be done.

Here is my reasons:
- It's using older major versions of cloud libraries. When adding new functionality you may want to use different libraries or their newer major version.
- The configuration setup is quite complex, with around 3-4 layers of abstractions and 3k LOC just to load a configuration file.  Most of this functionality is already implemented (and documented) in https://github.com/spf13/viper, which is also used in this project. With a non-obvious way of how this functionality is utilised it it seems excessive.
- Storage abstractions seems to be copied from thanos. I attempted to use them for my feature, but they are quite limiting and more work needs to be done in this case. Thanos released a separate library with this functionality that can be used https://github.com/thanos-io/objstore

r2k1:
This change probably doesn't worth investing a lot of effort in it.
Feel free to close it if you have concerns.

PR #1796, opened 2023-03-22
CORE-145 Draft: add ProportionalAssetResourceCosts to Allocation API

nikovacevic:

PR #1819, opened 2023-03-29
[CORE-129] updates to get idle coeffs surfaced, also surface provider ID

ameijer:
## What does this PR change?
* Surfaces idle coefficients and provider IDs

* yes, updates https://github.com/opencost/opencost/pull/1796 so that can be closed

* introduces a new includeProportionalAssetResourceCosts query param that can be used to return proportional costs for instances or clusters. other API behavior is unaffected 

* Closes https://github.com/opencost/opencost/pull/1796

## How was this PR tested?
* confirmed no regressions using local build. confirmed numbers returned looked appropriate based on manual test

* there is an additional query param that can be documented 

* is a specific customer ask


pkg/costmodel/assets.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/opencost/opencost/pkg/kubecost"
+	"github.com/opencost/opencost/pkg/log"
+)
+
+func (cm *CostModel) ComputeAssets(start, end time.Time) (*kubecost.AssetSet, error) {
```
michaelmdresser, line 11: @nikovacevic is this going to become the canonical `ComputeAssets`, also used by KCM?

pkg/costmodel/costmodel.go
```diff
@@ -2293,3 +2293,125 @@ func measureTimeAsync(start time.Time, threshold time.Duration, name string, ch
...
+func (cm *CostModel) QueryAllocation(window kubecost.Window, resolution, step time.Duration, aggregate []string, includeIdle, idleByNode, includeProportionalAssetResourceCosts bool) (*kubecost.AllocationSetRange, error) {
+	// Validate window is legal
+	if window.IsOpen() || window.IsNegative() {
+		return nil, fmt.Errorf("illegal window: %s", window)
+	}
+
+	// Idle is required for proportional asset costs
+	if includeProportionalAssetResourceCosts {
+		includeIdle = true
+	}
```
michaelmdresser: I'm not a fan of silently changing parameters that are incorrectly combined. I'd rather a helpful error be thrown. I can be convinced otherwise, though!

pkg/costmodel/costmodel.go
```diff
@@ -2293,3 +2293,125 @@ func measureTimeAsync(start time.Time, threshold time.Duration, name string, ch
...
+	var allocTotals map[string]*kubecost.AllocationTotals
+	var assetTotals map[string]*kubecost.AssetTotals
+
+	if idleByNode {
+		allocTotals = kubecost.ComputeAllocationTotals(allocSet, kubecost.AllocationNodeProp)
+		assetTotals = kubecost.ComputeAssetTotals(assetSet, kubecost.AssetNodeProp)
+	} else {
+		allocTotals = kubecost.ComputeAllocationTotals(allocSet, kubecost.AllocationNodeProp)
+		assetTotals = kubecost.ComputeAssetTotals(assetSet, kubecost.AssetNodeProp)
+	}
```
michaelmdresser: These cases look identical! Am I missing something?

ameijer: ah good catch - we had been intending to fix that - changed those to be cluster properties

pkg/costmodel/costmodel.go
ameijer: I like the idea. A helpful error message to serve as a guardrail. @nikovacevic thoughts?

michaelmdresser (approved):
Approving to unblock. Looks like you're missing a DCO signoff on the latest commit.

I'd really like to see an improvement on the defaulting thread, but I certainly won't block the PR on it :slightly_smiling_face:

michaelmdresser (approved):

PR #1823, opened 2023-03-29
Open Source CloudConfig and Cloud Service Integrations

Sean-Holcomb:
## What does this PR change?
This PR makes available the configurations and service connections used by CloudCost. Once this PR is merged the work of releasing our CloudCost integrations can begin.

### Configurations
This package contains definitions for configuring the integrations with Cloud Service Providers used to populate CloudCost along with a paradigm for integrating with any other Service offered by the CSPs . Each config has a CSP specific `authorizer` that handles auth for the respective SDKs. Each CSP's `authorizer` should be able to support any of the auth methods allowed by the CSP's SDK. The `authorizer` is then able to provide the required authorization to create and use a Client for any of the Services in the SDK. 

### Service Connections
This package is the layer above config and for each service config defined there is a service integration. These integrations are meant to provide the base functionality of a given service that would be applicable for multiple features. As it stands many of these are only used in the context of CloudCosts and the distinction between where this layer should end has not always been clear. Moving forward I would like to see this layer become better defined as Contributors find more use cases.

*  Provide access to Cloud Configs and Service Integrations to OpenCost contributors, to help contribute to standard integration practices.

* Closes https://kubecost.atlassian.net/browse/CORE-240

## How was this PR tested?
* This PR was tested in Multi Cloud environments to ensure continued functionality of the features it supports

* This PR requires developer docs to help contributors understand what exists and contribute to it.

*


michaelmdresser (approved):

Pokom (approved):
Obviously lots of code here so I can't review all of it. The only suggestion I'd have is to add a README somewhere calling out what exactly these files are and how to use them. To me it's not clear how I would wire up using the implementations and what I would need a head of time, especially in terms of credentials and permissions.

Sean-Holcomb:
@Pokom doc for that coming today

PR #1840, opened 2023-04-05
Added critical Prometheus diagnostic metrics.

nik-kc:
## What does this PR change?
* Exposes a few cAdvisor and KSM metrics for the /diagnostics page

* Nope

* Will give more visibility into configuration issues

* [cost-analyzer-helm-chart #2066](https://github.com/kubecost/cost-analyzer-helm-chart/issues/2066)

## How was this PR tested?
* Tested locally, confirmed metrics appeared on /diagnostics page

* May require additional documentation [here](https://github.com/kubecost/docs/blob/master/diagnostics.md)

* 

metrics

### DIFF
--- a/pkg/prom/diagnostics.go
+++ b/pkg/prom/diagnostics.go
@@ -40,6 +40,18 @@ const (
 	// KubecostRecordingRuleCPUUsageID is the identifier for the query used to
 	// determine of the CPU usage recording rule is set up correctly.
 	KubecostRecordingRuleCPUUsageID = "kubecostRecordingRuleCPUUsage"
+
+	// CAdvisorWorkingSetBytesMetricID is the identifier for the query used to determine
+	// if cAdvisor working set bytes data is being scraped
+	CAdvisorWorkingSetBytesMetricID = "cadvisorWorkingSetBytesMetric"
+
+	// KSMCPUCapacityMetricID is the identifier for the query used to determine if
+	// KSM CPU capacity data is being scraped
+	KSMCPUCapacityMetricID = "ksmCpuCapacityMetric"
+
+	// KSMAllocatableCPUCoresMetricID is the identifier for the query used to determine
+	// if KSM allocatable CPU core data is being scraped
+	KSMAllocatableCPUCoresMetricID = "ksmAllocatableCpuCoresMetric"
 )
 
 const DocumentationBaseURL = "https://github.com/kubecost/docs/blob/master/diagnostics.md"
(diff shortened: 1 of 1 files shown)


pkg/prom/diagnostics.go
```diff
@@ -107,6 +119,24 @@ var diagnosticDefinitions map[string]*diagnosticDefinition = map[string]*diagnos
...
 	},
+	CAdvisorWorkingSetBytesMetricID: {
+		ID:          CAdvisorWorkingSetBytesMetricID,
+		QueryFmt:    `absent_over_time(container_memory_working_set_bytes{container="cost-model", container!="POD", instance!=""}[5m] %s)`,
+		Label:       "cAdvsior working set bytes metrics available",
+		Description: "Determine if cAdvisor working set bytes metrics are available during last 5 minutes.",
+	},
+	KSMCPUCapacityMetricID: {
+		ID:          KSMCPUCapacityMetricID,
+		QueryFmt:    `avg_over_time(kube_node_status_capacity_cpu_cores[5m] %s) == 0`,
```
ameijer: @nik-kc a lot of ideas being thrown around on this parent issue! I am wondering whether you want to check absent_over_time here rather than an average of 0? My interpretation is that if KSM is not emitting the metric, something is broken. 

In the same vein, couldn't we have a fully allocated node where everything is functioning? That would have a 0 allocatable core result, but would not necessarily be an error. 

Would be interested to hear your thoughts in favor of checking for a 0 value though!

nik-kc: Good catch! My thoughts were, I'm still new to Prometheus and its functions and their interactions with scalar vs. vector values, so this is what I got working at the time. But your ideas make much more sense! Keeps them consistent with the other diagnostic queries.

Feedback like this is invaluable, so thank you for taking a look!

ameijer (approved):
LGTM - reviewed the original thread @ https://github.com/kubecost/cost-analyzer-helm-chart/issues/2066 and this hits the three diagnostic metrics called out by @AjayTripathy

ameijer (approved):

teevans (approved):

dwbrown2:
QQ: why did we add new metrics here and not update the existing ksm/cadvisor metrics?

nik-kc:
> QQ: why did we add new metrics here and not update the existing ksm/cadvisor metrics?

My understanding, based on [this comment](https://github.com/kubecost/cost-analyzer-helm-chart/issues/2066#issuecomment-1489009793), was that queries would be added, not changed. If it'd be better to update rather than add, I can make that change.

michaelmdresser:
@dwbrown2 perhaps I'm misreading your comment, but there are no new metrics in this PR, just new diagnostic queries.

dwbrown2:
Yep, was just wondering if we want to modify the existing diagnostic queries vs adding new ones. Either way seems fine to me. 

@nealormsbee anything we need to do to now get these reflected on FE?

nealormsbee:
@dwbrown2 I wouldn't think so -- we just query `/diagnostics/prometheusMetrics` and show what's in the payload, so we should get it in the UI for free.

I'm a little confused. I see the new metrics rendered in the screenshot on this PR, but I don't see them in nightly (in the UI or the payload).

nik-kc:
@nealormsbee It seems nightly doesn't have the latest from develop. When I run develop locally, I see the new metrics in the UI.

nik-kc:
@nealormsbee These new metrics are confirmed present in nightly now.

Issue #1757, opened 2023-03-10
Document Azure support for OpenCost


mattray:
OpenCost needs to document how to properly configure Azure, based on 
https://docs.kubecost.com/install-and-configure/install/cloud-integration/azure-out-of-cluster/azure-config

CsBigDataHub:
@mattray Thanks for working in this. Looking forward to this feature.
Rate card seems to be from 2019 - https://github.com/opencost/opencost/blob/develop/configs/azure.json

AjayTripathy:
Would love to bump priority on this one, we will have more cool Azure integrations to support as well!

PR #1842, opened 2023-04-05
Support for Azure consumption pricesheet download API

babbageclunk:
## What does this PR change?
* It extends the Azure provider to get customer-specific prices using the consumption pricesheet download API, if the user provides a billing account and the configured credentials have access to request it.

* Not that I'm aware of.

* If their cluster is running on Azure and they have special prices negotiated, OpenCost will give more accurate pricing.

* No

## How was this PR tested?
* By running the custom image in a cluster with a billing account added to the `azure-service-key` secret, checking the logs and the `/pricingSourceStatus`, `/allNodePricing` API endpoints.
* Also by running the custom image without a billing account to ensure that the current ratecard handling still works.

* Yes, I'll need to add information on specifying the billing account in the `azure-service-key` - at the moment the only documentation for the `azure-service-key` is on the Kubecost side. It also needs details about how to grant the service principal access to the billing account.

* I'm not sure whether this should be in the next release - it's intended for the kubecon candidate.

### DIFF
--- a/pkg/cloud/azurepricesheet/client.go
+++ b/pkg/cloud/azurepricesheet/client.go
@@ -1,0 +1,126 @@
+package azurepricesheet
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	armruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+)
+
+const (
+	moduleName    = "armconsumption"
+	moduleVersion = "v1.0.0"
+)
+
+// At the moment the consumption pricesheet download API is not a)
+// documented or b) supported by the SDK. This is an implementation of
+// a client in the style of the Azure go SDK - once the API is
+// supported this will be removed.
+
+// PriceSheetClient contains the methods for the PriceSheet group.
+// Don't use this type directly, use NewPriceSheetClient() instead.
+type PriceSheetClient struct {
+	host             string
+	billingAccountID string
+	pl               runtime.Pipeline
+}
+
+// NewClient creates a new instance of PriceSheetClient with the specified values.
+// subscriptionID - Azure Subscription ID.
+// credential - used to authorize requests. Usually a credential from azidentity.
+// options - pass nil to accept the default values.
+func NewClient(billingAccountID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*PriceSheetClient, error) {
+	if options == nil {
+		options = &arm.ClientOptions{}
+	}
+	ep := cloud.AzurePublic.Services[cloud.ResourceManager].Endpoint
+	if c, ok := options.Cloud.Services[cloud.ResourceManager]; ok {
+		ep = c.Endpoint
+	}
+	pl, err := armruntime.NewPipeline(moduleName, moduleVersion, credential, runtime.PipelineOptions{}, options)
+	if err != nil {
+		return nil, err
+	}
+	client := &PriceSheetClient{
+		billingAccountID: billingAccountID,
+		host:             ep,
+		pl:               pl,
+	}
+	return client, nil
+}
+
+func (client *PriceSheetClient) BeginDownloadByBillingPeriod(ctx context.Context, billingPeriodName string) (*runtime.Poller[PriceSheetClientDownloadResponse], error) {
+	resp, err := client.downloadByBillingPeriodOperation(ctx, billingPeriodName)
+	if err != nil {
+		return nil, err
+	}
+	return runtime.NewPoller[PriceSheetClientDownloadResponse](resp, client.pl, nil)
+}
+
+type PriceSheetClientDownloadResponse struct {
+	ID         string                             `json:"id"`
+	Name       string                             `json:"name"`
+	StartTime  time.Time                          `json:"startTime"`
+	EndTime    time.Time                          `json:"endTime"`
+	Status     string                             `json:"status"`
+	Properties PriceSheetClientDownloadProperties `json:"properties"`
+}
+
+type PriceSheetClientDownloadProperties struct {
+	DownloadURL string `json:"downloadUrl"`
+	ValidTill   string `json:"validTill"`
+}
+
+// GetByBillingPeriod - Get the price sheet for a scope by subscriptionId and billing period. Price sheet is available via
+// this API only for May 1, 2014 or later.
+// If the operation fails it returns an *azcore.ResponseError type.
+// Generated from API version 2021-10-01
+// billingPeriodName - Billing Period Name.
+// options - PriceSheetClientGetByBillingPeriodOptions contains the optional parameters for the PriceSheetClient.GetByBillingPeriod
+// method.
+func (client *PriceSheetClient) downloadByBillingPeriodOperation(ctx context.Context, billingPeriodName string) (*http.Response, error) {
+	req, err := client.downloadByBillingPeriodCreateRequest(ctx, billingPeriodName)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.pl.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusAccepted) {
+		return nil, runtime.NewResponseError(resp)
+	}
+	return resp, nil
+}
+
+// getByBillingPeriodCreateRequest creates the GetByBillingPeriod request.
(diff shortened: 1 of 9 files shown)


CsBigDataHub:
Is the related to https://github.com/opencost/opencost/issues/1757 ?

pkg/cloud/azurepricesheet/client.go
```diff
@@ -0,0 +1,126 @@
...
+	if billingPeriodName == "" {
+		return nil, errors.New("parameter billingPeriodName cannot be empty")
+	}
+	urlPath = strings.ReplaceAll(urlPath, "{billingPeriodName}", url.PathEscape(billingPeriodName))
+	req, err := runtime.NewRequest(ctx, http.MethodGet, runtime.JoinPaths(client.host, urlPath))
+	if err != nil {
+		return nil, err
+	}
+	reqQP := req.Raw().URL.Query()
+	reqQP.Set("api-version", "2022-06-01")
```
ameijer, line 119: does this potentially conflict with the comment on line 86 re: a 2021-10-01 API?

pkg/cloud/azurepricesheet/client.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusAccepted) {
+		return nil, runtime.NewResponseError(resp)
+	}
+	return resp, nil
+}
+
+// getByBillingPeriodCreateRequest creates the GetByBillingPeriod request.
+func (client *PriceSheetClient) downloadByBillingPeriodCreateRequest(ctx context.Context, billingPeriodName string) (*policy.Request, error) {
+	urlPath := "/providers/Microsoft.Billing/billingAccounts/{billingAccountID}/billingPeriods/{billingPeriodName}/providers/Microsoft.Consumption/pricesheets/download"
```
ameijer: thoughts on making this a const then sprintf-ing both vars at the same time to render the final URL?

pkg/cloud/azureprovider.go
```diff
@@ -847,127 +858,378 @@ func (az *Azure) DownloadPricingData() error {
...
-		meterName := *v.MeterName
-		meterRegion := *v.MeterRegion
-		meterCategory := *v.MeterCategory
-		meterSubCategory := *v.MeterSubCategory
-
-		region, err := toRegionID(meterRegion, regions)
+		pricings, err := convertMeterToPricings(v, regions, baseCPUPrice)
 		if err != nil {
+			log.Warnf("converting meter to pricings: %s", err.Error())
 			continue
```
ameijer, line 866: do you think an error in this method is serious enough to return it immediately?

pkg/cloud/azureprovider.go
```diff
@@ -847,127 +858,378 @@ func (az *Azure) DownloadPricingData() error {
...
 		}
+		results[key] = pricing
 	}
+	return results, nil
 
-	az.Pricing = allPrices
-	az.RateCardPricingError = nil
+}
+
+type pricesheetDownloader struct {
```
ameijer: Perhaps I am misunderstanding, but what are your thoughts on placing this in the price sheet client file you developed? From my reading, the current purpose of the price sheet client is to obtain a download URL for the pricesheet, then this pricesheet downloader downloads the actual pricesheet itself. Would we do one operation without the other? If not, maybe that can just all be abstracted into one call to the client.

pkg/cloud/azureprovider.go
```diff
@@ -847,127 +858,378 @@ func (az *Azure) DownloadPricingData() error {
...
+
+	_, err = out.Seek(0, io.SeekStart)
+	if err != nil {
+		return nil, fmt.Errorf("seeking to start of file: %w", err)
+	}
+
+	return out, nil
+}
+
+func (d *pricesheetDownloader) readPricesheet(ctx context.Context, data io.Reader) (map[string]*AzurePricing, error) {
```
ameijer: @babbageclunk do you think this would be a good place for a unit test? What we have been trying to do is feed in a few lines of cloud provider pricing data into these response body readers (e.g., https://github.com/opencost/opencost/blob/develop/pkg/cloud/gcpprovider_test.go#L124) just to test the happy path.

ameijer (commented):
hello @babbageclunk! This looks awesome - a big step forward for this component. Took a quick pass. Would be interested to hear your perspective on getting some testing on this file in

pkg/cloud/azurepricesheet/client.go
babbageclunk: Good call - I'll do that. This was adapted pretty quickly from existing azure-sdk-for-go code, I'll do a pass over it to make sure there aren't any inconsistencies.

pkg/cloud/azurepricesheet/client.go
babbageclunk: Oops, like this!

pkg/cloud/azureprovider.go
babbageclunk: I was keeping this behaviour the same as it is for the ratecard handling code for consistency, but I'm alright making this an error that stops the pricesheet processing if you'd prefer.

pkg/cloud/azureprovider.go
babbageclunk: I did think about this but ended up not doing it because of the dependency it has on `convertMeterToPricings`, which is also used from the ratecard code. 

But now that I think about it more I could pass that function in as config. I'll do that, I think it makes sense to keep the cloud package tidier since all the providers are in it together.

pkg/cloud/azureprovider.go
babbageclunk: Yeah you're right, this is totally amenable to some unit tests, I'll add them here.

pkg/cloud/azureprovider.go
ameijer: your call, sounds like you considered the tradeoffs already. Just wanted to have a discussion!

ameijer:
@babbageclunk happy to give additional reviews as you would like, just re-request one from me when ready!

pkg/cloud/azurepricesheet/downloader.go
```diff
@@ -0,0 +1,252 @@
...
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/profiles/2020-09-01/commerce/mgmt/commerce"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+
+	"github.com/opencost/opencost/pkg/log"
+)
+
+type Downloader[T any] struct {
```
babbageclunk, line 23: I made this generic over the specific pricing type to avoid a circular dependency on `pkg/cloud.AzurePricing` in this code, but that might be too clever. Is there a sensible location I could move `AzurePricing` to so it can be imported from `pkg/cloud` and `pkg/cloud/azurepricesheet`?

babbageclunk: I guess I could just move `AzurePricing` into `pkg/cloud/azurepricesheet`

AjayTripathy:
Hey @babbageclunk did you want any help producing docs? and @ameijer are we ready to take another pass?

ameijer:
@AjayTripathy I was waiting for @babbageclunk 's ping that he had addressed the items from the last look before I take another pass. Am monitoring this closely though 👍

babbageclunk:
Just working on the unit test for the downloader now. 

My plan for docs is to pull [this page](https://docs.kubecost.com/install-and-configure/install/cloud-integration/azure-out-of-cluster/azure-config) across from the Kubecost documentation, adapt it to fit OpenCost, and then add information about the pricesheet configuration to it:
1. assigning the billing access role to the SP
2. passing the offer ID and billing account in environment variables

For 1, do you think it would be better to include the bash script to assign the role, or to spell out the steps? The steps are a bit messy since there isn't Azure CLI support for assigning a billing role to an SP as far as I can tell - you need to do a PUT to an ARM URL directly.

babbageclunk:
Actually, it would be helpful if someone there could migrate the Kubecost rate card documentation to the OpenCost docs?

pkg/cloud/azurepricesheet/downloader.go
ameijer: @babbageclunk what if you moved all of this + the existing azureProvider into a new package pkg/cloud/azure? You have made the decision (that I concur with) to introduce a new azure specific subpackage (azurepricesheet) - it could be refactored to just azure and include the azureprovider.go to become a one stop shop for all azure specific logic. 

This would almost certainly break our enterprise builds downstream but I am happy to prepare a companion PR and take care of things here on our end. 

That seems more maintainable than a pseudo-generic - what do you think?

ameijer (approved):
@babbageclunk this lgtm - tests look good as well. Perhaps consider merging this, then doing that refactor of azurepricing in a follow up PR since it is not a functionality blocker?

AjayTripathy (approved):

AjayTripathy:
Thanks so much @babbageclunk ! Anywhere I can get started with helping to write documentation for how to set up the key on the Azure side?

pkg/cloud/azurepricesheet/downloader.go
babbageclunk: That sounds like a great idea! @r2k1 had mentioned that there was a reorganisation happening in parallel to break the providers into separate packages. I'll make a quick PR to do that and remove the type parameter.

PR #1850, opened 2023-04-07
Nodepool percentage

ameijer:
## What does this PR change?
* creates an assets API on opencost

* https://github.com/kubecost/kubecost-cost-model/pull/1377

* provides a new API for the assets on opencost

* none

## How was this PR tested?
* confirmed expected behavior when running locally

* introduces a new API which might need to be documented

* 

## How this API works

This API returns raw asset data - not allocation data. Here is an example request/response: 

```
curl http://localhost:9003/assets?window=7d
{
  "code": 200,
  "status": "success",
  "data": {
    "AWS/__undefined__/__undefined__/Compute/cluster-localrun-default/Node/Kubernetes/i-00305755a4910b84d/ip-192-168-36-40.us-east-2.compute.internal": {
      "type": "Node",
      "properties": {
        "category": "Compute",
        "provider": "AWS",
        "service": "Kubernetes",
        "cluster": "cluster-localrun-default",
        "name": "ip-192-168-36-40.us-east-2.compute.internal",
        "providerID": "i-00305755a4910b84d"
      },
      "labels": {
        "instance": "10.100.182.2:9003",
        "job": "kubecost",
        "label_beta_kubernetes_io_arch": "amd64",
        "label_beta_kubernetes_io_instance_type": "m5.large",
        "label_beta_kubernetes_io_os": "linux",
        "label_eks_amazonaws_com_capacityType": "ON_DEMAND",
        "label_eks_amazonaws_com_nodegroup": "dev-1-ng-us-east2b-pub",
        "label_eks_amazonaws_com_nodegroup_image": "ami-0a3ff7ea25e73f7de",
        "label_failure_domain_beta_kubernetes_io_region": "us-east-2",
        "label_failure_domain_beta_kubernetes_io_zone": "us-east-2b",
        "label_kubernetes_io_arch": "amd64",
        "label_kubernetes_io_hostname": "ip-192-168-36-40.us-east-2.compute.internal",
        "label_kubernetes_io_os": "linux",
        "label_node_kubernetes_io_instance_type": "m5.large",
        "label_topology_kubernetes_io_region": "us-east-2",
        "label_topology_kubernetes_io_zone": "us-east-2b",
        "node": "ip-192-168-36-40.us-east-2.compute.internal"
      },
      "window": {
        "start": "2023-03-31T14:19:32Z",
        "end": "2023-04-07T14:19:32Z"
      },
      "start": "2023-04-05T17:45:00-04:00",
      "end": "2023-04-05T18:20:00-04:00",
      "minutes": 35,
      "nodeType": "m5.large",
      "pool": "dev-1-ng-us-east2b-pub",
      "cpuCores": 2,
      "ramBytes": 8121929728.000001,
      "cpuCoreHours": 1.166667,
      "ramByteHours": 4737792341.333334,
      "GPUHours": 0,
      "cpuBreakdown": {
        "idle": 1,
        "other": 0,
        "system": 0,
        "user": 0
      },
      "ramBreakdown": {
        "idle": 1,
        "other": 0,
        "system": 0,
        "user": 0
      },
      "preemptible": 0,
      "discount": 0,
      "cpuCost": 0.037162,
      "gpuCost": 0,
      "gpuCount": 0,
      "ramCost": 0.018837,
      "adjustment": 0,
      "totalCost": 0.055998
    },
```

The call to `/assets` returns all the available assets - the above response is just a segment. In the response object representing an asset, there are two fields of interest: `pool` which represents the node group/pool that the node is in; and `minutes`, which represents the number of minutes that the node was active during the window specified in the request. 

__To obtain the total minutes of all nodes in the nodepool, we can loop through the list of node assets in the response and keep a sum  of all the node `minutes` for each unique value of `pool`.__

__To find out what fraction of total nodepool time a given node consumed, we can just divide that nodes `minutes` by the sum computed for all nodes in the pool__


pkg/cloud/awsprovider.go
```diff
@@ -2314,3 +2315,15 @@ func (aws *AWS) PricingSourceSummary() interface{} {
 	// encode the pricing source summary as a JSON string
 	return aws.Pricing
 }
+
+func (aws *AWS) GetNodePoolName(labels map[string]string) string {
```
nik-kc: Could this be function be moved into provider.go so it's not repeated across providers?

pkg/cloud/awsprovider.go
```diff
@@ -2314,3 +2315,15 @@ func (aws *AWS) PricingSourceSummary() interface{} {
...
+	sanitizedLabel := regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(EKSNodepoolLabel, "_")
+
+	if poolName, found := labels[fmt.Sprintf("label_%s", sanitizedLabel)]; found {
+		return poolName
+	} else {
+		log.Warnf("unable to derive node pool name from node labels")
+		return ""
```
nik-kc: This is more of a curiosity, but why return an empty string instead of nil?

pkg/cloud/awsprovider.go
ameijer: An interesting proposal! I guess the reason I went with this implementation is that the label we use to determine the name of the node pool is different for each provider. For example, the label in azure is kubernetes.azure.com/agentpool, whereas in AWS it is  eks.amazonaws.com/nodegroup. Let me look at the approach again and see if I can optimize

nik-kc: That makes sense! I'm interested to see what you come up with.

pkg/cloud/awsprovider.go
ameijer: In go, you actually cannot have nil strings, unless you have a string pointer. So in stuff like structs, when you instantiate a new one, all of your string members are set to ""

nik-kc: Gotcha! I figured it was something like that. I'm still learning Go, so I appreciate you sharing that info.

pkg/costmodel/assets_handler.go
```diff
@@ -0,0 +1,34 @@
...
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/opencost/opencost/pkg/env"
+	"github.com/opencost/opencost/pkg/kubecost"
+	"github.com/opencost/opencost/pkg/util/httputil"
+)
+
+// ComputeAllocationHandler returns the assets from the CostModel.
+func (a *Accesses) ComputeAssetsHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
avrodrigues5: Should we just name it `handlers.go `? Are we going to have more handlers off costmodel.

Go lang file names are not delimited by underscore unless test files so going to be assetsHandlers.go if its retained.

pkg/kubecost/asset_json.go
```diff
@@ -477,6 +477,9 @@ func (n *Node) MarshalJSON() ([]byte, error) {
 	jsonEncodeString(buffer, "end", n.End.Format(time.RFC3339), ",")
 	jsonEncodeFloat64(buffer, "minutes", n.Minutes(), ",")
 	jsonEncodeString(buffer, "nodeType", n.NodeType, ",")
+	if n.Pool != "" {
+		jsonEncodeString(buffer, "pool", n.Pool, ",")
```
avrodrigues5: Question: What happens if n.Pool is empty? 

Shouldnt json encoding handle the case of an empty n.Pool or is your idea not to present n.Pool if it's not there?

avrodrigues5 (commented):
Minor review comments Alex. Generally LGTM

pkg/kubecost/asset_json.go
ameijer: Idea would be to ignore the pool if it is not there - an empty string might make someone think the node is not in a pool, when in reality all it meant is that we cannot infer the pool name

pkg/costmodel/assets_handler.go
ameijer: I think `handlers` is good!

pkg/cloud/awsprovider.go
ameijer: @nik-kc I took a stab at a new implementation. Interface func now just returns the bare minimum provider-specific label, all common logic refactored to the generic provider - see what you think!

nik-kc: Chef's kiss, love it. Great work!

avrodrigues5 (approved):

nik-kc (approved):

pkg/kubecost/asset.go
```diff
@@ -1772,6 +1772,7 @@ type Node struct {
 	RAMCost      float64
 	Discount     float64
 	Preemptible  float64
+	Pool         string
```
mbolt35: This needs a version annotation, and you need to update the assets version for bingen - anytime a primary pipeline data structure is changed, you have to do this. Read docs here: 

https://github.com/opencost/opencost/blob/develop/pkg/kubecost/bingen.go

mbolt35: Also, better docs here: https://github.com/kubecost/bingen

Specifically: https://github.com/kubecost/bingen#backwards-compatibility-and-field-versioning

pkg/kubecost/kubecost_codecs.go
```diff
@@ -1426,6 +1427,9 @@ func (target *AllocationProperties) UnmarshalBinaryWithContext(ctx *DecodingCont
 	target.Annotations = AllocationAnnotations(tt)
 	// --- [end][read][alias](AllocationAnnotations) ---
 
+	eee := buff.ReadBool() // read bool
+	target.AggregatedMetadata = eee
```
mbolt35: Is aggregated metadata supposed to be part of this PR?

mbolt35 (changes requested):
I think the aggregated metadata change from the allocation work snuck in this PR?

pkg/kubecost/kubecost_codecs.go
ameijer: fixed, this is ignored

mbolt35 (approved):
Changes look great! Nice work.

PR #1854, opened 2023-04-10
Update GKE Metadata Google Cloud Client Defaults

mbolt35:
## What does this PR change?
* The metadata client was hanging indefinitely in some cases running locally (where the endpoint was unreachable). Updating the HTTPClient used to match the default provided in the google cloud API: https://github.com/googleapis/google-cloud-go/blob/4c6bb3cb062b3e208a617462ee4c4c64660ea17f/compute/metadata/metadata.go#L66-L77
  * This also stopped the indefinite hang.


pkg/util/httputil/roundtrip.go
```diff
@@ -1,17 +1,20 @@
...
-	base      http.RoundTripper
 }
 
 // NewUserAgentTransport creates a RoundTripper that attaches the configured user agent.
 func NewUserAgentTransport(userAgent string, base http.RoundTripper) http.RoundTripper {
 	return &userAgentTransport{
-		userAgent: userAgent,
-		base:      base,
+		RoundTripper: base,
+		userAgent:    userAgent,
```
michaelmdresser: What's going on with this `base: base` to `RoundTripper: base` change?

michaelmdresser (approved):

AjayTripathy (approved):

pkg/util/httputil/roundtrip.go
mbolt35: It makes the base the actual base. When I made this change originally, I was moving most of the construction inside this type, but given that this just overrides the roundtripper, it made more sense just to move the HTTPClient defaults back out to the provider. This is what happened when I put it back together, and the way that it existed when I tested.

mbolt35: Semantically identical, I'd be glad to change it back if you want to roll the dice :)

michaelmdresser: Huh. You're definitely the expert here, and I trust that expertise.

mbolt35: No, you're right - I just sloppily threw it back together. The tests do pass, but I won't chance it. Will need to test live again, but I'll do that before merging.

PR #1855, opened 2023-04-11
Add labels column to CSV export 

r2k1:
## What does this PR change?
Add labels column to CSV export. Allow users to aggregate data by pod/namespace labels.

* This should update a feature not yet realised.

## How was this PR tested?
* Updated tests
* Run export job locally

* I will update a sample file in the documentation.


AjayTripathy:
As a heads up, the labels column could potentially be gigantic. We may want some configuration / flag to enable or disable.

r2k1:
Do you have a number in mind how gigantic they can be?

The data is stored in a file and labels are already in memory.
Do you have a concrete issue in mind? I expect the data to be processed with tools like SQL so visual clutter shouldn't matter.

If large amount of labels is an issue shouldn't be there an option at the different level that preventing loading them in memory/allocation model?

AjayTripathy:
You can potentially drop them at the prometheus ingestion level, but we have seen people with hundreds of labels for ~10k pods which could result in millions of items. By default these get auto-dropped when rolling up levels in aggregation.

I'm not opposed to just passing this through the pipeline to get started, or working on a more general "projection" idea to only pull requested data through.

r2k1:
It would be ideal to have configurable allowlist of labels for the exporter. I'm not sure how to configure it cohesively yet, but I'm also unsure if it's worth introducing another ENV variable for the CSV export yet. Is there a central definition of all available configuration parameters?

Considering the CSV exporter keeps little information in memory, I don't see any technical issues writing millions of strings in a file. Especially since the opencost already keeps them in memory without issues. However, I believe it is more dangerous to keep millions of strings in memory.

AjayTripathy (approved):

AjayTripathy:
Ok, agreed...let's accept for now but hold until after initial launch since we'd want to just verify by scale testing here.

r2k1:
@AjayTripathy After some addition thoughts I added few extra configuration parameters. Please take another look.

The main reason is extracting data from `json` column isn't trivial in some tools. For example both Microsoft Excel and Google Spreadsheet doesn't have built-in function to work with json.

So in some cases it can be easier for the user to have the label in a separate column.

By default none of labels is exported.

Here is an example of export with the next options:

```
EXPORT_CSV_FILE=export.csv
EXPORT_CSV_LABELS_ALL=true
EXPORT_CSV_LABELS_LIST=app,pod_template_generation
```
[export.csv](https://github.com/opencost/opencost/files/11391943/export.csv)

Sean-Holcomb (approved):

PR #1857, opened 2023-04-11
Core 140 attempt iii

### LINKED ISSUE #1223: Make namespace annotations/labels available when aggregating by namespace
When using the allocations API you should be able to aggregate by namespace and specify explicitly the annotations/labels you want to be included in the properties field. Today, you can pull in annotations only if you specifically ask to aggregate by the annotation. The reason for this is the possibility for these values to conflict with more granular level objects like pods.  The feature request would be to enable the ability to specify specific annotations set at the namespace level to be authoritatively pulled in to the dataset when aggregating by namespace, regardless of conflicts with the lower level attributes. 

Use Case: As a K8S operator the ability is needed to tag namespaces with specific business related metadata that will be used in other admin processes. We need the ability to pull in that metadata through kubecost and not have it affected by other objects set within the namespace. 

gz#1868<br><br><i>(related to <a href='https://kubecost.zendesk.com/agent/tickets/1868'>Zendesk ticket #1868</a>)</i>

ameijer:
## What does this PR change?
* Updates label logic to be governed by a query param, and returns aggregated annotations/labels 

* a non-rebased iteration of https://github.com/opencost/opencost/pull/1829

* This adds a new query param, includeAggregatedMeta, that governs the behavior

* Closes CORE-140, closes #1223 

## How was this PR tested?
* tested locally, confirmed expected labels and annotations were returned 

* yes, an additional query param 

*

### DIFF
--- a/pkg/kubecost/allocationprops.go
+++ b/pkg/kubecost/allocationprops.go
@@ -92,18 +92,20 @@ func ParseProperty(text string) (string, error) {
 
 // AllocationProperties describes a set of Kubernetes objects.
 type AllocationProperties struct {
-	Cluster            string                `json:"cluster,omitempty"`
-	Node               string                `json:"node,omitempty"`
-	Container          string                `json:"container,omitempty"`
-	Controller         string                `json:"controller,omitempty"`
-	ControllerKind     string                `json:"controllerKind,omitempty"`
-	Namespace          string                `json:"namespace,omitempty"`
-	Pod                string                `json:"pod,omitempty"`
-	Services           []string              `json:"services,omitempty"`
-	ProviderID         string                `json:"providerID,omitempty"`
-	Labels             AllocationLabels      `json:"labels,omitempty"`
-	Annotations        AllocationAnnotations `json:"annotations,omitempty"`
-	AggregatedMetadata bool                  `json:"-"`
+	Cluster        string                `json:"cluster,omitempty"`
+	Node           string                `json:"node,omitempty"`
+	Container      string                `json:"container,omitempty"`
+	Controller     string                `json:"controller,omitempty"`
+	ControllerKind string                `json:"controllerKind,omitempty"`
+	Namespace      string                `json:"namespace,omitempty"`
+	Pod            string                `json:"pod,omitempty"`
+	Services       []string              `json:"services,omitempty"`
+	ProviderID     string                `json:"providerID,omitempty"`
+	Labels         AllocationLabels      `json:"labels,omitempty"`
+	Annotations    AllocationAnnotations `json:"annotations,omitempty"`
+	// When set to true, maintain the intersection of all labels + annotations
+	// in the aggregated AllocationProperties object
+	AggregatedMetadata bool `json:"-"`
 }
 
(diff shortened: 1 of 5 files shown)


pkg/kubecost/allocationprops.go
```diff
@@ -92,17 +92,18 @@ func ParseProperty(text string) (string, error) {
...
+	Container          string                `json:"container,omitempty"`
+	Controller         string                `json:"controller,omitempty"`
+	ControllerKind     string                `json:"controllerKind,omitempty"`
+	Namespace          string                `json:"namespace,omitempty"`
+	Pod                string                `json:"pod,omitempty"`
+	Services           []string              `json:"services,omitempty"`
+	ProviderID         string                `json:"providerID,omitempty"`
+	Labels             AllocationLabels      `json:"labels,omitempty"`
+	Annotations        AllocationAnnotations `json:"annotations,omitempty"`
+	AggregatedMetadata bool                  `json:"-"`
```
michaelmdresser: Please add a docstring to this field so that future users understand what behavior it implies.

pkg/kubecost/allocationprops.go
```diff
@@ -439,7 +440,22 @@ func (p *AllocationProperties) Intersection(that *AllocationProperties) *Allocat
...
 		intersectionProps.Namespace = p.Namespace
+		// ignore the incoming labels from unallocated or unmounted special case pods
+		if p.AggregatedMetadata || that.AggregatedMetadata {
+			intersectionProps.AggregatedMetadata = true
+			if p.Container == UnmountedSuffix || p.Container == UnallocatedSuffix {
+				intersectionProps.Annotations = that.Annotations
+				intersectionProps.Labels = that.Labels
+			} else if that.Container == UnmountedSuffix || that.Container == UnallocatedSuffix {
+				intersectionProps.Annotations = p.Annotations
+				intersectionProps.Labels = p.Labels
```
michaelmdresser: My first thought would be that annotations and labels would be dropped entirely if we're dealing with an unmounted or unallocated Allocation. This looks like we're overriding the annotations/labels instead.

I'm certainly missing context here, so you may have already considered this.

michaelmdresser (commented):
One question, and I also think it would be beneficial to add at least one unit test for this logic, perhaps around here https://github.com/opencost/opencost/blob/107349113d025622b3035a50f499f39570496761/pkg/kubecost/allocation_test.go#L14

pkg/kubecost/allocationprops.go
ameijer: Yes, this logic looks weird and probably deserves a comment with what I am about to share: 

The core change is that we now maintain the intersection of the labels/annotations of the two AllocationProperties objects that we add. Normally, this works fine, however, our special case unallocated/unmounted Allocations are built by us and never have any labels or annotations. Because of this, as these get aggregated up (by namespace or otherwise) they have the effect of always clearing out the intersection, regardless if all the other actual pods/etc have them. 

This logic is designed to effectively ignore the unmounted/unallocated objects and just copy over the labels from the other object - we only take the intersection of 'legitimate' allocations.

michaelmdresser (approved):

PR #1864, opened 2023-04-17
Add CPU/RAM idle to exported metrics.

macb:
## What does this PR change?
* Adds `container_cpu_idle` and `container_ram_idle` to the metrics emitter. They represent the idle % for a container ie. `(request - usage)/request`

I'm not 100% sure this is the correct metric to export but it seemed most convenient with the available data. The intention is to allow users to determine the historical efficiency of workloads at a glance. It seems reasonable to limit this to containers with explicit requests, if instead we used max(usage, request) similar to allocations those without a request would consistently have 0% idle as the it would be equal to the usage.

* Users that scrape the metrics endpoint will be able to view historical idle values similar to allocations.

## How was this PR tested?
* Built locally using minikube and podman.

```
# HELP container_cpu_idle container_cpu_idle Ratio of CPU of containers
# TYPE container_cpu_idle gauge
container_cpu_idle{container="coredns",instance="minipod",namespace="kube-system",node="minipod",pod="coredns-787d4945fb-vznrv"} 0.09678320090805903
container_cpu_idle{container="etcd",instance="minipod",namespace="kube-system",node="minipod",pod="etcd-minipod"} 0.082127348502672
container_cpu_idle{container="kindnet-cni",instance="minipod",namespace="kube-system",node="minipod",pod="kindnet-n7gkh"} 0.09901240555918292
container_cpu_idle{container="kube-apiserver",instance="minipod",namespace="kube-system",node="minipod",pod="kube-apiserver-minipod"} 0.21464884710724458
container_cpu_idle{container="kube-controller-manager",instance="minipod",namespace="kube-system",node="minipod",pod="kube-controller-manager-minipod"} 0.18833326821991758
container_cpu_idle{container="kube-scheduler",instance="minipod",namespace="kube-system",node="minipod",pod="kube-scheduler-minipod"} 0.09697293807641634
container_cpu_idle{container="opencost",instance="minipod",namespace="opencost",node="minipod",pod="opencost-646b557747-66lff"} 0.008518372009752938
container_cpu_idle{container="opencost-ui",instance="minipod",namespace="opencost",node="minipod",pod="opencost-646b557747-66lff"} 0.01


....

# HELP container_ram_idle container_ram_idle Ratio of idle RAM of containers
# TYPE container_ram_idle gauge
container_ram_idle{container="coredns",instance="minipod",namespace="kube-system",node="minipod",pod="coredns-787d4945fb-vznrv"} 0.48306361607142856
container_ram_idle{container="etcd",instance="minipod",namespace="kube-system",node="minipod",pod="etcd-minipod"} 0.5233984375
container_ram_idle{container="kindnet-cni",instance="minipod",namespace="kube-system",node="minipod",pod="kindnet-n7gkh"} 0.5557421875
container_ram_idle{container="opencost",instance="minipod",namespace="opencost",node="minipod",pod="opencost-587c654bf-w2fqg"} 0.51872
container_ram_idle{container="opencost-ui",instance="minipod",namespace="opencost",node="minipod",pod="opencost-587c654bf-w2fqg"} 0.9263464727272728
```

* Potentially as this metric could be excluded via `disabledMetrics`.
* It also assumes a container has a cpu or ram request to record an idle metric.

* Not sure the process for this, so didn't. :)

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -128,6 +128,8 @@ var (
 	cpuAllocGv                 *prometheus.GaugeVec
 	gpuAllocGv                 *prometheus.GaugeVec
 	pvAllocGv                  *prometheus.GaugeVec
+	ramIdleGv                  *prometheus.GaugeVec
+	cpuIdleGv                  *prometheus.GaugeVec
 	networkZoneEgressCostG     prometheus.Gauge
 	networkRegionEgressCostG   prometheus.Gauge
 	networkInternetEgressCostG prometheus.Gauge
@@ -232,6 +234,22 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider model
 			toRegisterGV = append(toRegisterGV, pvAllocGv)
 		}
 
+		ramIdleGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "container_ram_idle",
+			Help: "container_ram_idle Ratio of idle RAM of containers",
+		}, []string{"namespace", "pod", "container", "instance", "node"})
+		if _, disabled := disabledMetrics["container_ram_idle"]; !disabled {
+			toRegisterGV = append(toRegisterGV, ramIdleGv)
+		}
+
+		cpuIdleGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "container_cpu_idle",
+			Help: "container_cpu_idle Ratio of CPU of containers",
+		}, []string{"namespace", "pod", "container", "instance", "node"})
+		if _, disabled := disabledMetrics["container_cpu_idle"]; !disabled {
+			toRegisterGV = append(toRegisterGV, cpuIdleGv)
+		}
+
 		networkZoneEgressCostG = prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: "kubecost_network_zone_egress_cost",
 			Help: "kubecost_network_zone_egress_cost Total cost per GB egress across zones",
@@ -312,6 +330,8 @@ type CostModelMetricsEmitter struct {
 	RAMAllocationRecorder         *prometheus.GaugeVec
 	CPUAllocationRecorder         *prometheus.GaugeVec
 	GPUAllocationRecorder         *prometheus.GaugeVec
+	CPUIdleRecorder               *prometheus.GaugeVec
+	RAMIdleRecorder               *prometheus.GaugeVec
 	ClusterManagementCostRecorder *prometheus.GaugeVec
 	LBCostRecorder                *prometheus.GaugeVec
 	NetworkZoneEgressRecorder     prometheus.Gauge
@@ -364,6 +384,8 @@ func NewCostModelMetricsEmitter(promClient promclient.Client, clusterCache clust
 		CPUAllocationRecorder:         cpuAllocGv,
 		GPUAllocationRecorder:         gpuAllocGv,
 		PVAllocationRecorder:          pvAllocGv,
+		RAMIdleRecorder:               ramIdleGv,
+		CPUIdleRecorder:               cpuIdleGv,
 		NetworkZoneEgressRecorder:     networkZoneEgressCostG,
 		NetworkRegionEgressRecorder:   networkRegionEgressCostG,
 		NetworkInternetEgressRecorder: networkInternetEgressCostG,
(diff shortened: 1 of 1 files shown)


mattray:
We're finalizing the 103 release right now, I'll get this added to the queue for review as soon as it's released.

pkg/costmodel/metrics.go
```diff
@@ -609,9 +631,29 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 				if len(costs.RAMAllocation) > 0 {
 					cmme.RAMAllocationRecorder.WithLabelValues(namespace, podName, containerName, nodeName, nodeName).Set(costs.RAMAllocation[0].Value)
 				}
+				if len(costs.RAMReq) > 0 && costs.RAMReq[0].Value > 0 {
+					req := costs.RAMReq[0].Value
+					usage := 0.0
+					if len(costs.RAMUsed) > 0 {
+						usage = costs.RAMUsed[0].Value
+					}
+					idle := (req - usage) / req
```
thomasvn: Thanks for the PR @macb! In cases where `usage > req`, this would result in negative idle? Is this expected behavior for the metric?

macb: Ya, I wasn't quite sure what the desired behavior would be. Setting negative idles to 0 means we'd lose the context that a pod is consistently overstepping its requests (but presumably not meeting its limits).

In the simple case of milliCPU:
Request: 1000m
Usage: 1500m
`(1000 - 1500) / 1000` would be and idle of `-0.5`

Converting that to an efficiency %:
`1 - (-0.5)` would be 1.5 or 150% efficiency

Keeping that context seemed better than constraining the value between 0 and 1.

thomasvn:
@macb Also just for your awareness, OpenCost's [`/allocation/compute`](https://www.opencost.io/api) does report `cpuEfficiency`, `ramEfficiency`, and `totalEfficiency` for every workload. And this accounts for requests/usage ([code ref](https://github.com/opencost/opencost/blob/v1.103/pkg/kubecost/allocation.go#L573-L625)). Would this fulfill your use case?

macb:
> @macb Also just for your awareness, OpenCost's [`/allocation/compute`](https://www.opencost.io/api) does report `cpuEfficiency`, `ramEfficiency`, and `totalEfficiency` for every workload. And this accounts for requests/usage ([code ref](https://github.com/opencost/opencost/blob/v1.103/pkg/kubecost/allocation.go#L573-L625)). Would this fulfill your use case?

Unfortunately, it doesn't quite fit. Prometheus makes for a really nice interface for accessing and aggregating the data for multiple clusters. 

Id be happy to change the bounds to match the API responses for consistency if that'd be more desirable.

thomasvn:
Yes, I think it would be good to match bounds & metric name to the API response!

Adding one new metric for every individual container has cpu/mem implications for users running at really large scale. Curious for opinions from @AjayTripathy @nikovacevic ?

pkg/costmodel/metrics.go
nikovacevic: I would strongly recommend using `allocated` (i.e. `max(usage, request)`) here instead of `request` for three reasons:
1. If you are summing this value over time, and you allow negative idle, then you make it possible to have a big positive idle followed by a big negative idle---which would sum to zero idle. I.e. the value zero no longer means "no idle" it just means "net zero idle" which is very different.
2. It's not uncommon for request to be 0.0, in which case you either have to divide by zero and yield `idle=NaN`, or not export a metric for a given container. Neither is particularly nice.
3. It makes sense that a workload using all of its requested resources _and more_ would have no idle. Zero idle seems more appropriate for such a case than a negative value that requires effort to interpret.

pkg/costmodel/metrics.go
```diff
@@ -232,6 +234,22 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider model
 			toRegisterGV = append(toRegisterGV, pvAllocGv)
 		}
 
+		ramIdleGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "container_ram_idle",
```
nikovacevic, line 238: This is a fine metric to measure. Unfortunately, there may be confusion caused by the use of the word "idle" in other contexts---most frequently to describe node resources that are not allocated to a workload. (Though sometimes we do just qualify the type of idle, as in "workload idle," and accept the fact that there can be multiple kinds of idle, and multiple kinds of efficiency.)

Do you have an opinion here @kwombach? or @AjayTripathy and @dwbrown2 ?

nikovacevic:
> Adding one new metric for every individual container has cpu/mem implications for users running at really large scale. Curious for opinions from @AjayTripathy @nikovacevic ?

Yes, we would expect this to increase Prometheus memory significantly for big users. A condition for approval here should probably involve a "default-off, opt-in" configuration.

pkg/costmodel/metrics.go
macb: I was specifically thinking of the "workload idle" referenced in the [opencost spec](https://github.com/opencost/opencost/blob/develop/spec/opencost-specv01.md#idle-costs) and thought that since it was referencing resources (cpu/mem) that'd fit to that case.

macb:
It follows the `disabledMetrics` pattern the rest of the collectors have so users would be able to opt-out similar to other metrics. Is there an example of another opt-in metric?

macb:
I'm not able to resolve the comment in https://github.com/opencost/opencost/pull/1864#discussion_r1210819644, but I moved from requests to allocations in https://github.com/opencost/opencost/pull/1864/commits/1a24cb188c5282a75f23bd773d024c1de0b5d886. I also added a bounds check to set any negative idles as 0 since I wasn't sure if the usage is guaranteed to be equivalent to the allocation's usage when usage exceeds request.

nikovacevic:
Thanks @macb. I've pinged the team to find out who needs to review this next, and we'll hopefully get back to you soon.

macb:
Hey @nikovacevic, just checking in for next steps. Would love to get this in and just want to do whatever it takes to move that along.

nikovacevic:
Thanks @macb — not sure what next steps are here. Maybe @AjayTripathy can weigh in?

macb:
Hey @AjayTripathy, any chance this could get a review and a timeline for inclusion in a release? I'm happy to address any feedback to make that happen.

macb:
Hey @mattray, not really sure what the next steps are but this PR has been open for ~3mo's. Some idea of what needs to happen to get this in would be great!

mattray:
I've raised this for review by Kubecost Engineering's backlog.

mattray:
Assigning to Kubecost Engineering

nikovacevic:
@mattray this is blocked on product, not engineering. The question is whether or not we want to write (and support) an extra high-cardinality metric for this particular use case, and if so, whether it should be named this or named something else.

I do not have any answers to these questions. Maybe the following people can weigh in: @cliffcolvin @teevans @AjayTripathy

r2k1:
This is fairly easy to achieve through a prometheus query. 

If it's a common request, such query can be added somewhere to the documentation instead.

PR #1865, opened 2023-04-18
Single Cloud Cost type

Sean-Holcomb:
## What does this PR change?
Merge Cloud Cost into one type, update fields to match focus spec, Remove monitoring key logic from Cloud Cost and Assets

### New Cloud Cost Type
Unified Cloud Cost types into a single type that supports this the needs of both. This is something that needed to be done now before Cloud Cost was switched to default because changing this later would require migrations for all users. This needed to be done because the aggregate type does not serve a purpose that cannot be fulled by the item type if it can be aggregated and accumulated (This is the primary feature ask that this PR is supporting). 

The new Cloud Cost type has support for filtering aggregation and accumulation. Additionally its fields have been updated to match the [focus spec](https://github.com/finopsfoundation/finops-open-cost-usage-spec/blob/main/specification_sheet_import.md) more closely to better support multi cloud environments. These changes can be seen in property changes
- workGroupID -> accountID
- billingAccountID -> invoiceEntityID
For Cost there are now 4 costMetrics that are supported
- ListCost
- NetCost
- NetAmortizedCost
- InvoicedCost
The details of each are outlined in the focus spec. Each `costMetric` also has a `kubernetesPercent` value. Un-aggregated this should be 1 or 0. However during aggregation, `kubernetesPercent` is determined by multiplying the `costMetric` cost by its `kubernetesPercent` and aggregating that value as `kubernetesCost` for that `costMetric`. That `kubernetesCost` is then divided by the aggregated Total costs to determine the new `kubernetesPercent`. Since this process results in unique values for each `costMetric` this value is included as part of the cost metric

### Remove monitoring key logic from Cloud Cost and Asset
This logic has been removed from Open Cost as its sole purpose is to support KCM functionality. It has been added to the appropriate package in KCM.

* https://github.com/kubecost/kubecost-cost-model/pull/1396

* 

* Closes ...

## How was this PR tested?
* See https://github.com/kubecost/kubecost-cost-model/pull/1396

* 

*


pkg/kubecost/kubecost_codecs.go
```diff
@@ -33,6 +33,9 @@ const (
 )
 
 const (
+	// DefaultCodecVersion is used for any resources listed in the Default version set
+	DefaultCodecVersion uint8 = 17
+
```
ameijer, line 38: I have a TODO to sort these lexicographically before writing them back out to the file. This is making your changeset look larger than it actually is :(

pkg/kubecost/bingen.go
```diff
@@ -73,20 +73,13 @@ package kubecost
...
-// @bingen:generate:CloudCostAggregateSetRange
-// @bingen:generate:CloudCostAggregateProperties
-// @bingen:generate:CloudCostAggregateLabels
-// @bingen:end
-
-// @bingen:set[name=CloudCostItem,version=2]
-// @bingen:generate:CloudCostItem
-// @bingen:generate[stringtable]:CloudCostItemSet
-// @bingen:generate:CloudCostItemSetRange
-// @bingen:generate:CloudCostItemProperties
```
ameijer, line 88: @Sean-Holcomb will removing these break backwards compat?

pkg/kubecost/cloudcostprops.go
```diff
@@ -0,0 +1,213 @@
...
+func (ccp *CloudCostProperties) Intersection(that *CloudCostProperties) *CloudCostProperties {
+	if ccp == nil || that == nil {
+		return nil
+	}
+
+	if ccp.Equal(that) {
+		return ccp
+	}
+	intersectionCCP := &CloudCostProperties{}
+	if ccp == intersectionCCP || that == intersectionCCP {
```
ameijer: do you need to use the Equal method here? is this doing pointer comparison?

pkg/kubecost/cloudcostprops_test.go
```diff
@@ -0,0 +1,126 @@
...
+				},
+			},
+			expectedCCP: &CloudCostProperties{
+				Provider:        "CustomProvider",
+				ProviderID:      "ProviderID1",
+				AccountID:       "",
+				InvoiceEntityID: "InvoiceEntityID1",
+				Service:         "",
+				Category:        "Category1",
+				Labels: map[string]string{
```
ameijer, line 111: consider adding a test case where the labels differ to exercise the label intersection logic

pkg/kubecost/cloudcost.go
```diff
@@ -0,0 +1,547 @@
...
+func (cc *CloudCost) add(that *CloudCost) {
+	if cc == nil {
+		log.Warnf("cannot add to nil CloudCost")
+		return
+	}
+
+	// Preserve properties of cloud cost  that are matching between the two CloudCost
+	cc.Properties = cc.Properties.Intersection(that.Properties)
+
+	// Compute KubernetesPercent for sum
```
ameijer: dup of comment in add func for CostMetric

pkg/kubecost/cloudcost.go
```diff
@@ -0,0 +1,547 @@
...
+}
+
+type CloudCostSet struct {
+	CloudCosts            map[string]*CloudCost `json:"cloudCosts"`
+	Window                Window                `json:"window"`
+	Integration           string                `json:"-"`
+	AggregationProperties []string              `json:"aggregationProperties"`
+}
+
+// NewAssetSet instantiates a new AssetSet and, optionally, inserts
```
ameijer:
```suggestion
// NewCloudCostSet instantiates a new CloudCostSet and, optionally, inserts
```

pkg/kubecost/cloudcost.go
```diff
@@ -0,0 +1,547 @@
...
+// the given list of Assets
```
ameijer:
```suggestion
// the given list of CloudCosts
```

ameijer (commented):
Also, an aside question, is this clouse cost stuff in the Kubecost cloud? Will this change break that downstream?

pkg/kubecost/allocation.go
```diff
@@ -86,7 +86,7 @@ type Allocation struct {
 	// allocation as a percentage of the per-resource total cost of the
 	// asset on which the allocation was run. It is optionally computed
 	// and appended to an Allocation, and so by default is is nil.
-	ProportionalAssetResourceCosts ProportionalAssetResourceCosts `json:"proportionalAssetResourceCosts"`
+	ProportionalAssetResourceCosts ProportionalAssetResourceCosts `json:"proportionalAssetResourceCosts"` //@bingen:field[ignore]
```
michaelmdresser, line 89: Why are we ignoring this field in serialization?

pkg/kubecost/cloudcost.go
```diff
@@ -0,0 +1,547 @@
...
+		return cc.InvoicedCost, nil
+	}
+	return CostMetric{}, fmt.Errorf("invalid Cost Metric: %s", costMetricName)
+}
+
+type CloudCostSet struct {
+	CloudCosts            map[string]*CloudCost `json:"cloudCosts"`
+	Window                Window                `json:"window"`
+	Integration           string                `json:"-"`
+	AggregationProperties []string              `json:"aggregationProperties"`
```
michaelmdresser, line 147: I think these two fields would particularly benefit from docstrings

pkg/kubecost/cloudcost.go
```diff
@@ -0,0 +1,547 @@
...
+// CloudCostSet for the entire range.
+func (ccsr *CloudCostSetRange) Accumulate() (*CloudCostSet, error) {
+	var cloudCostSet *CloudCostSet
+	var err error
+
+	for _, ccs := range ccsr.CloudCostSets {
+		if cloudCostSet == nil {
+			cloudCostSet = ccs.cloneSet()
+			continue
+		}
```
michaelmdresser: I think this means that the first `CloudCostSet` in `ccsr` won't be accumulated correctly because `cloneSet()` doesn't copy over the CloudCosts. I think this can be fixed either by removing the `continue` or turning this into a `.Clone()`.

pkg/kubecost/cloudcost.go
```diff
@@ -0,0 +1,547 @@
...
+			log.Errorf("CloudCostSetRange: LoadCloudCost: failed to load CloudCost with window %s: %s", setWindow.String(), err.Error())
+		}
+
+		// If all cost has been inserted then finish
+		totalPct += pct
+		if totalPct >= 1.0 {
+			return
+		}
+	}
+}
```
michaelmdresser: Is it worth having a check after the loop? If we don't exit in the loop, we've probably messed up the percentage distribution and cost will be lost; might be worth an error log!

pkg/kubecost/cloudcost.go
```diff
@@ -0,0 +1,547 @@
...
+		k8sPct = (thisK8sCost + thatK8sCost) / sumCost
+	}
+
+	return CostMetric{
+		Cost:              sumCost,
+		KubernetesPercent: k8sPct,
+	}
+}
+
+// percent returns the product of the give percent and the cost, kubecost percent remians the same
```
michaelmdresser:
```suggestion
// percent returns the product of the give percent and the cost, Kubernetes percent remains the same
```

pkg/kubecost/bingen.go
Sean-Holcomb: It will, this will mean that on start up they will rebuild, cloud cost, given that this is still a beta feature this should not be an issue

pkg/kubecost/allocation.go
Sean-Holcomb: It is calculated query time, it was added without running bingen again and breaks bingen runs if not ignored.

pkg/kubecost/cloudcost.go
Sean-Holcomb: This is an early return condition.
Say we have a range with 3 days but the asset is all on the first day. totalPct will be 1.0 but it will continue to check the window for the days later in the range.

pkg/kubecost/bingen.go
ameijer: OK makes sense if it is still in beta

pkg/kubecost/cloudcost.go
Sean-Holcomb: I updated the comment, so that is more clear.

ameijer (approved):

michaelmdresser (approved):

PR #1867, opened 2023-04-19
Parcs accum support

ameijer:
## What does this PR change?
* Adds support for accumulation of PARCS

* None

* Allows users to accumulate with PARCS

## How was this PR tested?
* ensured both accumulated and unaccumlated results were plausible when run locally

* No

*

### DIFF
--- a/pkg/costmodel/aggregation.go
+++ b/pkg/costmodel/aggregation.go
@@ -2241,6 +2241,19 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 
 	// IncludeIdle, if true, uses Asset data to incorporate Idle Allocation
 	includeIdle := qp.GetBool("includeIdle", false)
+	// Accumulate is an optional parameter, defaulting to false, which if true
+	// sums each Set in the Range, producing one Set.
+	accumulate := qp.GetBool("accumulate", false)
+
+	// Accumulate is an optional parameter that accumulates an AllocationSetRange
+	// by the resolution of the given time duration.
+	// Defaults to 0. If a value is not passed then the parameter is not used.
+	accumulateBy := kubecost.AccumulateOption(qp.Get("accumulateBy", ""))
+
+	// if accumulateBy is not explicitly set, and accumulate is true, ensure result is accumulated
+	if accumulateBy == kubecost.AccumulateOptionNone && accumulate {
+		accumulateBy = kubecost.AccumulateOptionAll
+	}
 
 	// IdleByNode, if true, computes idle allocations at the node level.
 	// Otherwise it is computed at the cluster level. (Not relevant if idle
@@ -2262,6 +2275,15 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 		}
 
 		return
+	}
+
+	// Accumulate, if requested
+	if accumulateBy != kubecost.AccumulateOptionNone {
+		asr, err = asr.Accumulate(accumulateBy)
+		if err != nil {
+			WriteError(w, InternalServerError(err.Error()))
+			return
+		}
 	}
 
 	w.Write(WrapData(asr, nil))
--- a/pkg/costmodel/aggregation.go
+++ b/pkg/costmodel/aggregation.go
@@ -2282,7 +2282,8 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	if accumulateBy != kubecost.AccumulateOptionNone {
 		asr, err = asr.Accumulate(accumulateBy)
 		if err != nil {
-			WriteError(w, InternalServerError(err.Error()))
+			log.Errorf("error accumulating by %v: %s", accumulateBy, err)
+			WriteError(w, InternalServerError(fmt.Errorf("error accumulating by %v: %s", accumulateBy, err).Error()))
 			return
 		}
 	}
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -246,16 +246,62 @@ func (pva *PVAllocation) Equal(that *PVAllocation) bool {
 		util.IsApproximately(pva.Cost, that.Cost)
 }
 
-type ProportionalAssetResourceCost struct {
-	Cluster                    string  `json:"cluster"`
-	Node                       string  `json:"node,omitempty"`
-	ProviderID                 string  `json:"providerID,omitempty"`
-	CPUPercentage              float64 `json:"cpuPercentage"`
-	GPUPercentage              float64 `json:"gpuPercentage"`
-	RAMPercentage              float64 `json:"ramPercentage"`
-	NodeResourceCostPercentage float64 `json:"nodeResourceCostPercentage"`
+// used to compute the average of 2 PARCs. A PARC is a dimensonless
+// percentage that loses information needed when finding an average
+// to that end, we track the percentage and resource usage of the
+// component used to make the parc, to allow us to average two PARCs
+type ParcsComponent struct {
+	TotalCost       float64
+	UsageProportion float64
 }
 
+func (p *ParcsComponent) Clone() ParcsComponent {
+	return ParcsComponent{
+		TotalCost:       p.TotalCost,
+		UsageProportion: p.UsageProportion,
+	}
+}
+
+type ProportionalAssetResourceCost struct {
+	Cluster                    string           `json:"cluster"`
+	Node                       string           `json:"node,omitempty"`
+	ProviderID                 string           `json:"providerID,omitempty"`
+	CPUPercentage              float64          `json:"cpuPercentage"`
+	GPUPercentage              float64          `json:"gpuPercentage"`
+	RAMPercentage              float64          `json:"ramPercentage"`
+	NodeResourceCostPercentage float64          `json:"nodeResourceCostPercentage"`
+	GPUComponents              []ParcsComponent `json:"-"`
+	CPUComponents              []ParcsComponent `json:"-"`
+	RAMComponents              []ParcsComponent `json:"-"`
+}
+
+func (parc ProportionalAssetResourceCost) Clone() ProportionalAssetResourceCost {
+	gpuComps := []ParcsComponent{}
+	cpuComps := []ParcsComponent{}
+	ramComps := []ParcsComponent{}
+
+	for _, gpuComp := range parc.GPUComponents {
+		gpuComps = append(gpuComps, gpuComp.Clone())
+	}
+	for _, cpuComp := range parc.CPUComponents {
+		cpuComps = append(cpuComps, cpuComp.Clone())
+	}
+	for _, ramComp := range parc.RAMComponents {
+		ramComps = append(ramComps, ramComp.Clone())
+	}
+	return ProportionalAssetResourceCost{
+		Cluster:                    parc.Cluster,
+		Node:                       parc.Node,
+		ProviderID:                 parc.ProviderID,
+		CPUPercentage:              parc.CPUPercentage,
+		GPUPercentage:              parc.GPUPercentage,
+		RAMPercentage:              parc.RAMPercentage,
+		NodeResourceCostPercentage: parc.NodeResourceCostPercentage,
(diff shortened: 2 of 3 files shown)


pkg/kubecost/allocation.go
```diff
@@ -266,34 +276,57 @@ func (parc ProportionalAssetResourceCost) Key(insertByNode bool) string {
...
+}
+
+func (parcs ProportionalAssetResourceCosts) Insert(parc ProportionalAssetResourceCost, insertByNode, isAccumulation bool) {
 	if !insertByNode {
 		parc.Node = ""
 		parc.ProviderID = ""
 	}
 	if curr, ok := parcs[parc.Key(insertByNode)]; ok {
+
+		CPUPercentage := curr.CPUPercentage + parc.CPUPercentage
```
nik-kc, line 332: Can you explain the logic/math behind adding/averaging the percentages?

ameijer: @nik-kc does my comment @ https://github.com/opencost/opencost/pull/1867/files#diff-7c339daaa6b130f0e74be88b76e4c35d3941edf78f886fae793772b8f566c58cR298 make sense? happy to add further detail if not

nik-kc: I think what's confusing me is why adding percentages works to begin with, and by extension, why averaging them would also work, given that the windows could be different. I'm having a hard time seeing how that math works out.

ameijer: I see. So basically, there are two different dimensions we are summing across: resource and time. If pod A and pod B at the same instant are using 1% each of a node's CPU, then when we combine their allocations, it makes sense to sum them for a total of 2%. That's the first dimension and what this highlighted line handles. 

The second dimension is time. If pod A+B is using 2% each min, when we accumulate those (which is just a time based aggregation under the hood), over 10 minutes then if we just sum, it looks like Pod A+B is using 20% of the node... when in reality it was just using 2% each minute over the 10 minute period. In the time case, we want the _average_ percentage that it used over the period. 

Because of this, we have no choice but to pass down the context to the addition function, and tell it whether we want the average (when we are combining these over time) or the sum (when we are combining these across resources)

nik-kc: That makes sense! Since I'm not super familiar with what the data is expected to look like, would it be valid in the non-accumulation case to have a percentage over 100%? e.g. given pod A and B in your first example, if they had percentages of 80% and 60%, respectively.

ameijer: my understanding is that would be an acceptable outcome due to exactly that reason - at different times, pods can consume a majority of resources, so if those pods get aggregated together, the usage would be > 100%

nik-kc: Awesome, makes sense. Thanks for explaining!

nik-kc (approved):

pkg/kubecost/allocation.go
michaelmdresser: I think I haven't followed this fully. Let me pose an example and see how it works out. I'll stick with just CPU to avoid overloading with identical math.

Say I have a **one-node cluster** with **three Pods** (each with one container). The **node has 4 CPU**.

Over the span of hour 1:
- Pod 1 runs for 30 minutes, consuming 1 CPU while alive. PARC: 12.5% (0.5 core-hours / 4 available core-hours)
- Pod 2 runs for 1 hour, consuming 2 CPU while alive. PARC: 50% (2 core-hours)
- Pod 3 runs for 1 hour, consuming 1 CPU while alive. PARC: 25% (1 core-hour)

Over the span of hour 2:
- Pod 1 does not run. PARC: 0% (0 core-hours / 4 available core-hours)
- Pod 2 runs for 30 minutes, consuming 2 CPU while active. PARC: 25% (1 core-hour)
- Pod 3 runs for 1 hour, consuming 1 CPU while active. PARC: 25% (1 core-hour)

Over the span of hour 3:
- Pod 1 does not run. PARC: 0% (0 core-hours / 4 available)
- Pod 2 runs for 30 minutes, consuming 3 CPU while active. PARC: 37.5% (1.5 core-hours)
- Pod 3 runs for 1 hour, consuming 1 CPU while active. PARC: 25% (1 core-hour)

I'd expect the following accumulated PARC:
- Pod 1: (0.5 + 0 + 0) core-hours used / (4 + 4 + 4) core-hours available = 0.5/12 = **4.16%**
- Pod 2: (2 + 1 + 1.5) / (4 + 4 + 4) = 4.5/12 = **37.5%**
- Pod 3: (1 + 1 + 1) / (4 + 4 + 4) = 3/12 = **25%**

The logic I'm seeing in this PR will do this:
- Pod 1: `12.5%`-> `(12.5% + 0%) / 2` -> `((12.5% + 0%) / 2) + 0%) / 2` = **3.125%**
- Pod 2: `50%` -> `(50% + 37.5%) / 2` -> `((50% + 37.5%) / 2) + 37.5%) / 2` = **40.625%**
- Pod 3: `25%` -> `(25% + 25%) / 2` -> `((25% + 25%) / 2) + 25%) / 2` = **25%**

This leads me to think that the logic in this PR is incorrect. However, I may be misunderstanding what PARC is supposed to represent.

michaelmdresser: Before I forget: what if the node (and workloads) go away halfway through hour 3? How does that affect PARC by node? And by cluster?

Or, what if halfway through hour 3 the existing node was replaced by one with double the capacity? All workloads get rescheduled onto the new node.

ameijer: This has been remediated by tracking the individual components comprising the PARC during the computation process. We have a unit test covering this exact scenario and the numbers match now

pkg/costmodel/aggregation.go
```diff
@@ -2265,6 +2278,15 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
+	// Accumulate, if requested
+	if accumulateBy != kubecost.AccumulateOptionNone {
+		asr, err = asr.Accumulate(accumulateBy)
+		if err != nil {
+			WriteError(w, InternalServerError(err.Error()))
```
michaelmdresser: Would you please wrap this error? e.g. "error accumulating: %s"

pkg/kubecost/allocation.go
```diff
@@ -246,16 +246,62 @@ func (pva *PVAllocation) Equal(that *PVAllocation) bool {
...
+	Cluster                    string           `json:"cluster"`
+	Node                       string           `json:"node,omitempty"`
+	ProviderID                 string           `json:"providerID,omitempty"`
+	CPUPercentage              float64          `json:"cpuPercentage"`
+	GPUPercentage              float64          `json:"gpuPercentage"`
+	RAMPercentage              float64          `json:"ramPercentage"`
+	NodeResourceCostPercentage float64          `json:"nodeResourceCostPercentage"`
+	GPUComponents              []ParcsComponent `json:"-"`
+	CPUComponents              []ParcsComponent `json:"-"`
+	RAMComponents              []ParcsComponent `json:"-"`
```
michaelmdresser, line 275: Scale testing is topical; what do you expect the cardinality of these slices to be? Will there be substantial memory overhead?

pkg/kubecost/allocation.go
```diff
@@ -1774,15 +1909,35 @@ func deriveProportionalAssetResourceCostsFromIdleCoefficients(idleCoeffs map[str
...
+			TotalCost:       totals[idleId]["gpu"],
+			UsageProportion: idleCoeffs[idleId][allocation.Name]["gpu"],
+		},
+	}
+	parc.RAMComponents = []ParcsComponent{
+		{
+			TotalCost:       totals[idleId]["ram"],
+			UsageProportion: idleCoeffs[idleId][allocation.Name]["ram"],
+		},
+	}
```
michaelmdresser: Any particular reason why you chose to set these outside of the main struct initialization?

michaelmdresser (commented):
Logic looks solid to me. The only reason I'm not immediately approving is possible mem overhead; I just don't have a good grasp of this bit of data so I don't know what the size looks like.

r2k1:
I'm not familiar with this part of the codebase. Just a random thought.
Couldn't you achieve the same outcome just by moving PARC calculation down the pipeline?

In this function:
https://github.com/opencost/opencost/blob/develop/pkg/kubecost/allocation.go#L933-L977

Is it possible to move step 2b to happen after accumulation?https://github.com/opencost/opencost/blob/develop/pkg/costmodel/aggregation.go#L2194-L2200

ameijer:
@r2k1 I looked into this - Unfortunately, I don't think it can work. We need to compute the PARC before any kind of aggregation/accumulation, since that is the only time we know the specific node that a given pod/etc is on. Once we start aggregating or accumulating (which if you think about it, is really just an aggregation over time), that information is lost, it is only kept in the PARC itself which was computed before accumulation/aggregation.

pkg/kubecost/allocation.go
ameijer: this was done elsewhere but moved here. fixed this

pkg/kubecost/allocation.go
```diff
@@ -246,16 +246,62 @@ func (pva *PVAllocation) Equal(that *PVAllocation) bool {
...
+// used to compute the average of 2 PARCs. A PARC is a dimensonless
+// percentage that loses information needed when finding an average
+// to that end, we track the percentage and resource usage of the
+// component used to make the parc, to allow us to average two PARCs
+type ParcsComponent struct {
+	TotalCost       float64
+	UsageProportion float64
+}
+
+func (p *ParcsComponent) Clone() ParcsComponent {
```
nikovacevic, line 258: Just a note--it's a bit unusual to `Clone` and receive a different type (i.e. pointer-to-value).

pkg/kubecost/allocation.go
```diff
@@ -246,16 +246,62 @@ func (pva *PVAllocation) Equal(that *PVAllocation) bool {
...
+	NodeResourceCostPercentage float64          `json:"nodeResourceCostPercentage"`
+	GPUComponents              []ParcsComponent `json:"-"`
+	CPUComponents              []ParcsComponent `json:"-"`
+	RAMComponents              []ParcsComponent `json:"-"`
 }
 
+func (parc ProportionalAssetResourceCost) Clone() ProportionalAssetResourceCost {
+	gpuComps := []ParcsComponent{}
+	cpuComps := []ParcsComponent{}
+	ramComps := []ParcsComponent{}
```
nikovacevic, line 281: Might want to set capacities on those slices.

pkg/kubecost/allocation.go
```diff
@@ -267,35 +313,111 @@ func (parc ProportionalAssetResourceCost) Key(insertByNode bool) string {
...
+			CPUPercentage:              CPUPercentage,
+			GPUPercentage:              GPUPercentage,
+			RAMPercentage:              RAMPercentage,
 			NodeResourceCostPercentage: curr.NodeResourceCostPercentage + parc.NodeResourceCostPercentage,
+			CPUComponents:              append(curr.CPUComponents, parc.CPUComponents...),
+			GPUComponents:              append(curr.GPUComponents, parc.GPUComponents...),
+			RAMComponents:              append(curr.RAMComponents, parc.RAMComponents...),
 		}
+
+		if isAccumulation {
```
nikovacevic, line 349: This pattern definitely makes thing spicier. That said, this can of course work. But it may be more straightforward to have an implementation that uses purely cumulative units so that accumulation is the same as aggregation. (As we discussed on Slack, that would mean recording cumulative cost per-resource, along with total cost per-asset-resource, and summing those. Then dividing to yield a percentage only at the very end.

That may or may not be too much trouble for the short term. But conditionals like `isAccumulation` within these kinds of functions have historically caused headaches.

nikovacevic (approved):
Having different logic for aggregation and accumulation is certainly a risk, and one we've tried to avoid historically. That said, while I'm trusting the test cases to some extent, I think that this likely works. So the real question here is whether or not we want to try to convert to cumulative terms (i.e. $) and divide at the end so that the math is more straightforward. No feelings on that judgment call, but happy to talk through it.

ameijer:
closing in favor of https://github.com/opencost/opencost/pull/1920

Issue #1788, opened 2023-03-20
Refactor `pkg/cloud` directory 


Pokom:
Looking to kick off a conversation about starting to split out
the CSP's into dedicated modules. The primary motivation is to support
calculating out of cluster costs. My proposal is that our cloud directoy
starts to look like the following: `pkg/cloud/${CSP}/${csp.go}`. Then we
can start placing external costs code in a dedicated file
(`${csp}_external.go` or specific folders.

One thing I discovered from this process is that there is a _lot_ of
coupling between csp's that make changing code hard. I discovered a number of
import cycles that existed. In order to decouple the CSP's and resolve
the import cycles, I started to split out shared code into three
directories:

- `models`
- `utils`
- `provider`

Models has shared definitions for structs and interfaces that don't
change csp to csp.
Utils has some shared utilities across all the cps's.
Provider has the `providerconfig` struct that is used across all of the
CSP's.

I recognize this is a _massive_ change based upon LoC, and so I suggest
we first evaluate whether or not this structure makes sense. If it does,
my proposal is to get a build available that I can roll out across the
three primary CSP's in Grafana's dev environment. This can help sus out
major issues early on.

Pokom:
@Sean-Holcomb @AjayTripathy @mbolt35 @nikovacevic I wanted to get this in front of everyone for review(Ajay said everyone here is a stakeholder). I'm proposing we create a deeper directory level in the `cloud` folder so we can start adding external asset implementations per cloud provider. I've detailed some of the challenges I ran into simply moving providers to dedicated modules and wanted to get some thoughts on whether or not this is the right direction.

Sean-Holcomb:
Hey Pokom, thanks for writing this up and the draft PR. We have been doing some refactoring on our end that is somewhat akin to this. There a few things to talk about with what you have said here. The first one is that while CSP division makes sense here because the way functionality is grouped in the provider interface, in our closed source we have some more modular interfaces which we have packaged separately (also under a `cloud` pkg). Having seen some of the work that you have done I am hoping that in the near term we can open-source some of our cloud configuration and integration which you could leverage or emulate the patterns of to continue our ongoing effort to make csp integration configuration more uniform across the application. In this world our file structure would look something like:
```
cloud
|--config
|--serviceintegrations
|--provider
```

[...]

PR #1787, opened 2023-03-20
Refactor cloud directory

Pokom:
## What does this PR change?
* Refactor `cloud` directory to create dedicated modules per CSP

## Does this PR relate to any other PRs?
* N\A

## How will this PR impact users?
* Ideally it won't

## Does this PR address any GitHub or Zendesk issues?
* #1788 

## How was this PR tested?
* Smoke Tests, Canary in Grafana Labs environment, E2E tests from Kubecost

## Does this PR require changes to documentation?
* No

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next Opencost release? If not, why not?
*

PR #1869, opened 2023-04-20
Refactor Azure provider into pkg/cloud/azure

babbageclunk:
## What does this PR change?
It combines the Azure cloud provider and `pkg/cloud/azurepricesheet` into one more-coherent `pkg/cloud/azure` package. To avoid circular imports between `pkg/cloud` and `pkg/cloud/azure` the core types shared between the providers have been pulled out into `pkg/cloud/models`.

```mermaid
graph LR;
subgraph before
cloud --> azp[cloud/azurepricesheet];
end
subgraph oops[circular imports]
c2[cloud] --> az[cloud/azure]
az -- uh oh! --> c2
end

subgraph after
c3[cloud]-->models[cloud/models]
c3 --> az2[cloud/azure]
az2 --> models
end

before --> oops --> after
```

The type parameter on `pkg/cloud/azurepricesheet.Downloader` was a workaround to avoid an import cycle as well, so it's been removed from `pkg/cloud/azure.PriceSheetDownloader`.

* #1842 - based on discussion from @ameijer during review.

* No behaviour change, this is purely code reorganisation.

* No.

## How was this PR tested?
* Running unit tests and running costmodel locally with rate card and price sheet downloading turned on, checking that both of those worked.

* No.

* I think it should have "next release", but it doesn't seem I can change labels.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	cloud.google.com/go/storage v1.28.1
 	github.com/Azure/azure-pipeline-go v0.2.3
 	github.com/Azure/azure-sdk-for-go v65.0.0+incompatible
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.4.1-0.20230323231529-14c481f239ec
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.5.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.2.2
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.0.0
 	github.com/Azure/azure-storage-blob-go v0.15.0
@@ -66,7 +66,7 @@ require (
 	cloud.google.com/go v0.110.0 // indirect
 	cloud.google.com/go/compute v1.18.0 // indirect
 	cloud.google.com/go/iam v0.12.0 // indirect
-	github.com/Azure/azure-sdk-for-go/sdk/internal v1.2.0 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/internal v1.3.0 // indirect
 	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
 	github.com/Azure/go-autorest/autorest/azure/cli v0.4.5 // indirect
 	github.com/Azure/go-autorest/autorest/date v0.3.0 // indirect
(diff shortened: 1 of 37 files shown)


mattray:
@Sean-Holcomb @Pokom does this relate to https://github.com/opencost/opencost/pull/1787 and https://github.com/opencost/opencost/issues/1788?

ameijer:
@mattray this is a follow up from our discussion here: https://github.com/opencost/opencost/pull/1842#pullrequestreview-1383469369

babbageclunk:
Oh darn, I hadn't seen that PR - I'm going to look through it now and see if I can harmonise this more.

babbageclunk:
Ok - it looks like @Pokom and I took very similar approaches, so the PRs are pretty compatible. The differences as I see them:

1. Where this PR has `pkg/cloud/types`, 1787 has `pkg/cloud/models` - I like that name better, I'll update this to match.
2. 1787 moves the `*ClusterMeta` functions to `pkg/cloud/utils` where I just dumped them into `pkg/cloud/types` with [some reservations](https://github.com/opencost/opencost/pull/1869/files#diff-e539669992808c43bd36d6cbe919d16f26b2c58da51baa5ffaf8632e3ab6ed00R18). I'll change this, although I'm always a bit uncomfortable calling a package `utils` - would `pkg/cloud/meta` be a better name?
3. I didn't move all of the providers since I was primarily focused on fixing the `azurepricesheet` wart I'd created, but that would have been a natural next step for the other ones.
4. Handling `ProviderConfig` - I introduced a `ProviderConfig` interface in `types` and updated the Azure provider to take the interface rather than a concrete type, and left the implementation where it was in `pkg /cloud`, while 1787 moved it into `pkg/cloud/provider`. I slightly prefer my one here, but not for any super firm reason - it just seems cleaner to leave it in the `cloud` package.

What do people think?

babbageclunk:
> I'll change this, although I'm always a bit uncomfortable calling a package `utils` - would `pkg/cloud/meta` be a better name?

I ended up keeping the name utils since `ToTitle` is also in there.

Pokom:
@babbageclunk sorry for the delay here, was on vacation last week! My apologies for not taking better notes and dropping them into #1787 after meeting with @Sean-Holcomb, but the short of it was the refactor was welcomed but likely too large and would require a lot of testing on the kubecost team to see through. In general, we agreed upon the overall structure that both of our PR's happened to land upon. I'd _love_ to see this PR work its way through and be the foundation for our to refactor the remaining cloud packages.

I can take a deeper look at your PR over the week, however your summary is pretty spot on. Here's my attempt to respond to each of your points:
1. `pkg/cloud/types` or `pkg/cloud/models` works for me, don't have a hard preference
2. I strongly favor `pkg/cloud/meta` over `utils`, at the time I just couldn't think of a better name
3. Starting small here has a better chance of seeing the refactor land
4. Leaving the provider in `pkg/cloud` makes sense as well, I don't see a need to have it in a dedicated ~model~ module

Love the contribution and I think it addresses a common challenge folks have. Really cool to see us both land on similar implementations :wink:

Sean-Holcomb (approved):
These Changes look good, lets get you guys unblocked

PR #1871, opened 2023-04-21
[Incubation] Inline Idle with Allocation

mbolt35:
## What does this PR change?
* Setup incubating conditional compilation for features which can be inflight but aren't ready for full release. This allows other dependencies to leverage these features without requiring core changes. 
* Incubation Feature: Appends node-related queries to allocation compute such that we can calculate idle inline.

Requires building with `-tags incubating`.

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -115,7 +115,8 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 
 	// If the duration is short enough, compute the AllocationSet directly
 	if end.Sub(start) <= cm.MaxPrometheusQueryDuration {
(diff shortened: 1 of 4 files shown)


AjayTripathy:
@Sean-Holcomb was also thinking about making this default. I think that makes a lot of sense and can save a bunch of time computing allocations.

mbolt35:
@AjayTripathy This was actually Sean's idea. Although, there maybe some edge cases with on-prem that prevent this from being an easy roll out. With the current behavior of idle being calculated each query for on prem, there's no chance for desync with reconciled, discounts, etc...

Speaking of which, I think we need to include discount in this PR as well.

Sean-Holcomb (approved):
Glad to see this moving forward, incubating flag makes this very low risk

PR #1876, opened 2023-04-25
unit tests for unmounted allocations bug

nickcurie:
## What does this PR change?
* Prevent allocations with services, but other unmounted properties from being listed separately than __unmounted__ allocation

* No

* 

* Closes https://kubecost.atlassian.net/browse/BURNDOWN-80

## How was this PR tested?
* Unit tests

* No

* v1.103

### DIFF
--- a/pkg/kubecost/allocation_test.go
+++ b/pkg/kubecost/allocation_test.go
@@ -2770,12 +2770,12 @@ func TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate(t *testin
 	}
 }
 
-func Test_AggregateBy(t *testing.T) {
+func Test_AggregateByService_UnmountedLBs(t *testing.T) {
 	end := time.Now().UTC().Truncate(day)
 	start := end.Add(-day)
 
 	normalProps := &AllocationProperties{
-		Cluster:        "kc-demo-stage",
+		Cluster:        "cluster-one",
 		Container:      "nginx-plus-nginx-ingress",
 		Controller:     "nginx-plus-nginx-ingress",
 		ControllerKind: "deployment",
@@ -2788,8 +2788,8 @@ func Test_AggregateBy(t *testing.T) {
 		},
 	}
 
-	weirdProps := &AllocationProperties{
-		Cluster:    "kc-demo-stage",
+	problematicProps := &AllocationProperties{
+		Cluster:    "cluster-one",
 		Container:  UnmountedSuffix,
 		Namespace:  UnmountedSuffix,
 		Pod:        UnmountedSuffix,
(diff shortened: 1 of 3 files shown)


pkg/kubecost/allocationprops.go
```diff
@@ -288,10 +288,16 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 				// Indicate that allocation has no services
 				names = append(names, UnallocatedSuffix)
 			} else {
-				// This just uses the first service
-				for _, service := range services {
-					names = append(names, service)
-					break
+				// sometimes load balancers will retain their services when unmounted
+				// so if we have an unmounted pod, make sure the key shows unmounted
```
nickcurie: not sure if this comment is accurate

pkg/kubecost/allocationprops.go
```diff
@@ -288,10 +288,16 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
...
+				if p.Pod == UnmountedSuffix {
```
nickcurie: Should we be checking more than just the pod? Maybe namespace and container as well?

pkg/kubecost/allocation_test.go
```diff
@@ -2757,3 +2759,80 @@ func TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate(t *testin
...
+				t.Error("pod unmounted")
+			}
+			if props.Controller == UnmountedSuffix {
+				t.Error("controller unmounted")
+			}
+		}
+	}
+
+	spew.Config.DisableMethods = true
+	fmt.Printf("%s", spew.Sdump(set.Allocations))
```
michaelmdresser: Please change this to a `t.Logf` so that we don't unnecessarily fill up stdout if the test passes.

pkg/kubecost/allocation_test.go
```diff
@@ -2757,3 +2759,80 @@ func TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate(t *testin
 		}
 	}
 }
+
+func Test_AggregateBy(t *testing.T) {
```
michaelmdresser:
```suggestion
func Test_AggregateByService_UnmountedLBs(t *testing.T) {
```

pkg/kubecost/allocation_test.go
```diff
@@ -2757,3 +2759,80 @@ func TestAllocationSet_Accumulate_Equals_AllocationSetRange_Accumulate(t *testin
...
+			"nginx-plus-nginx-ingress",
+			"ingress-nginx-controller",
+			"pacman",
+		},
+	}
+
+	idle := NewMockUnitAllocation(fmt.Sprintf("kc-demo-stage/%s", IdleSuffix), start, day, &AllocationProperties{
+		Cluster: "kc-demo-stage",
+	})
+	one := NewMockUnitAllocation("kc-demo-stage//__unmounted__/__unmounted__/__unmounted__", start, day, weirdProps)
```
michaelmdresser: If you can find some way to explain in a comment that this Allocation is the main point of the test, that would be very helpful.

pkg/kubecost/allocationprops.go
michaelmdresser: I might phrase it as

> Unmounted load balancers lead to \_\_unmounted\_\_ Allocations whose services field is populated. If we don't have a special case, the \_\_unmounted\_\_ Allocation will be transformed into a regular Allocation, causing havoc with AggregateBy and drilldown.

pkg/kubecost/allocationprops.go
michaelmdresser: I think that's a good idea! Perhaps `p.Pod == Unmounted || p.Namespace ... || p.Container ...`?

michaelmdresser (commented):
Nice job, I know this was thorny. Please make sure to update the Jira ticket/parent issue/etc. with the status and what we've learned.

michaelmdresser (approved):

PR #1878, opened 2023-04-26
Fix panic in CSV Export

### LINKED ISSUE #1877: CSV Export Panics When No Metrics Available
**Describe the bug**
In the case where the CSV uploader initializes without node_cpu_hourly_cost available, there is a panic trying to determine the start time of the CSV
```
goroutine 376 [running]:
[github.com/opencost/opencost/pkg/costmodel.(*CostModel).DateRange(0x40011ccd20?)](http://github.com/opencost/opencost/pkg/costmodel.(*CostModel).DateRange(0x40011ccd20?))
    /app/pkg/costmodel/allocation.go:293 +0x2c0
[github.com/opencost/opencost/pkg/costmodel.(*csvExporter).availableAllocationDates(0x4000100400?)](http://github.com/opencost/opencost/pkg/costmodel.(*csvExporter).availableAllocationDates(0x4000100400?))
    /app/pkg/costmodel/csv_export.go:132 +0x2c
[github.com/opencost/opencost/pkg/costmodel.(*csvExporter).Update](http://github.com/opencost/opencost/pkg/costmodel.(*csvExporter).Update)(0x4000e55f88, {0x27d8970, 0x4000058018})
    /app/pkg/costmodel/csv_export.go:41 +0x34
[github.com/opencost/opencost/pkg/costmodel.UpdateCSV(...)](http://github.com/opencost/opencost/pkg/costmodel.UpdateCSV(...))
    /app/pkg/costmodel/csv_export.go:31
[github.com/opencost/opencost/pkg/cmd/costmodel.StartExportWorker.func1()](http://github.com/opencost/opencost/pkg/cmd/costmodel.StartExportWorker.func1())
    /app/pkg/cmd/costmodel/costmodel.go:74 +0x1a4
created by [github.com/opencost/opencost/pkg/cmd/costmodel.StartExportWorker](http://github.com/opencost/opencost/pkg/cmd/costmodel.StartExportWorker)
    /app/pkg/cmd/costmodel/costmodel.go:64 +0x130
```

**To Reproduce**
Run OpenCost with the CSV Exporter enabled 

**Expected behavior**
No panic-- an error log explaining this would be great, though.

kc-eu-2023

r2k1:
Closes #1877


pkg/cmd/costmodel/costmodel.go
```diff
@@ -62,6 +61,13 @@ func StartExportWorker(ctx context.Context, model costmodel.AllocationModel) {
 		return
 	}
 	go func() {
+		// panic in a goroutine will crash the application. Log it and carry on.
+		defer func() {
+			if r := recover(); r != nil {
+				log.Errorf("panic in CSV Export Worker, stopping: %v", r)
+			}
+		}()
```
r2k1: I wonder if I should do another panic handler at the lower level (in addition to this one). To make worker keep trying to export according to the schedule (once a day).

But panic is something extraordinary that shouldn't happen.

Any preferences?

AjayTripathy: I don't think a panic handler feels super necessary, agree panics should be rare. I am ok to to just skip the panic altogether actually, since it feels like the issue is addressed.

AjayTripathy (changes requested):
The panic handler at the top doesn't feel super necessary to me, since this should be really rare and we don't use it often in the rest of the codebase. But if you feel it is helpful feel free to push back.

r2k1:
I think it depends on what is the desired outcome in a case of panic due to a bug in a CSV Export:
1) Crash the application (including the API). It will make the issue more visible, but increases the blast radius
2) Disable the CSV export and carry on. The API handler will survive, but it will make harder to track issues with the export.

I don't feel like a have a strong preference for either option. So I just removed the handler.

r2k1:
I think the rest of the code performs panic handling through an HTTP middleware.

And panic in the rest of the goroutines is unlikely to be possible to recover without causing more issues.

AjayTripathy (approved):

PR #1881, opened 2023-04-28
Default network costs and load balancer cost, Has a provision to have configs in opencost be second line of values in Closed source and fixes a bug

avrodrigues5:
## What does this PR change?
* Add default network costs and load balancer costs in Alibaba provider
* Noticed that in all provider even thou we pass /models in closed source the value used is defaultPricing values so added the config value as second line of information if user doesnt give their own values using /updateConfigByKey end point. This will only be in effect if user deleted the existing config else it continue to be default pricing itself. I dont want to be destructive.
* Fixes a bug wherein alibaba service key and secret were set to empty if there were passed in default config.

* None

* Give network costs and loadbalancer cost in their alibaba environment

* Closes ... [1251](https://github.com/kubecost/cost-analyzer-helm-chart/issues/2151)

## How was this PR tested?
* Tested on alibaba cluster to show the Network costs and LB costs
Screen Shot 2023-04-27 at 5 29 15 PM

* Prior to this fix if secret and key were passed in end point the overwrite of it would be empty and the REST end point would fail now it shows the node pricing and PV pricing... Might solve [1250](https://github.com/kubecost/features-bugs/issues/74) if user was doing the same.

* Verified that the default pricing gets created with alibaba.json in config if current written with default pricing is removed.

Screen Shot 2023-04-27 at 5 33 18 PM
Screen Shot 2023-04-27 at 5 33 32 PM
 

* Maybe requesting user to edit the network pricing for their region and zone give by alibaba documentation would be great 
https://www.alibabacloud.com/help/en/cloud-data-transmission/latest/cross-region-data-transfers
* Alibaba's load balancer is all divided into 4 types and each have different prices 
https://www.alibabacloud.com/product/server-load-balancer
Defaulted to Classic load balancer https://www.alibabacloud.com/help/en/server-load-balancer/latest/pay-as-you-go

I have documented this in code but would be great to have documentation around it!

* v1.103 but not release blocking can go to v1.104

### DIFF
--- a/pkg/cloud/providerconfig.go
+++ b/pkg/cloud/providerconfig.go
@@ -2,6 +2,8 @@ package cloud
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 	gopath "path"
 	"reflect"
 	"strconv"
@@ -13,6 +15,8 @@ import (
 	"github.com/opencost/opencost/pkg/log"
 	"github.com/opencost/opencost/pkg/util/json"
 )
+
+const closedSourceConfigMount = "models/"
 
 var sanitizePolicy = bluemonday.UGCPolicy()
 
@@ -91,6 +95,15 @@ func (pc *ProviderConfig) loadConfig(writeIfNotExists bool) (*CustomPricing, err
 	if !exists {
 		log.Infof("Could not find Custom Pricing file at path '%s'", pc.configFile.Path())
 		pc.customPricing = DefaultPricing()
+		// If config file is not present use the contents from mount models/ as pricing data
+		// in closed source rather than from from  DefaultPricing as first source of truth.
+		// since most images will already have a mount, to avail this facility user needs to delete the
+		// config file manually from configpath else default pricing still holds good.
+		fileName := filenameInConfigPath(pc.configFile.Path())
+		defaultPricing, err := ReturnPricingFromConfigs(fileName)
+		if err == nil {
+			pc.customPricing = defaultPricing
+		}
 
 		// Only write the file if flag enabled
 		if writeIfNotExists {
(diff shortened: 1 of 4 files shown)


avrodrigues5:
@mbolt35 When you have time please review my PR addressing all the key points related to network costs and loadbalancer cost as discussed
!

pkg/cloud/providerconfig.go
```diff
@@ -91,6 +95,15 @@ func (pc *ProviderConfig) loadConfig(writeIfNotExists bool) (*CustomPricing, err
...
 		pc.customPricing = DefaultPricing()
+		// If config file is not present use the contents from mount models/ as pricing data
+		// in closed source rather than from from  DefaultPricing as first source of truth.
+		// since most images will already have a mount, to avail this facility user needs to delete the
+		// config file manually from configpath else default pricing still holds good.
+		fileName := filenameInConfigPath(pc.configFile.Path())
+		defaultPricing, err := ReturnPricingFromConfigs(fileName)
+		if err == nil {
+			pc.customPricing = defaultPricing
+		}
```
mbolt35, line 106: Looks awesome, were you able to verify that this worked well with the other providers?

mbolt35 (approved):
This looks good to me. I would just make sure that the ProviderConfig updates you made work on other providers as well.

pkg/cloud/providerconfig.go
avrodrigues5: Verified in GCP and Azure:

Test strategy was: 
1. use rc-6 upload a default price config and then set the new image does it show up with the same sset default - yes it did.
2. when image was set to fixed version the configs mounted on /model showed up in the /var/config as expected.

Spoke to Bolt this looks good and merging the code!

PR #1884, opened 2023-04-28
costmodel: select not all metrics

### LINKED ISSUE #1883: fmtQueryCPUModePct is costly, consider adding label matchers
**Describe the bug**

This query is being consistently executed: https://github.com/opencost/opencost/blob/55d77f1de56d47ecadd6f3bd38c76e8fb9a4136a/pkg/costmodel/cluster.go#L896. Because it selects `node_cpu_seconds_total` from _all_ servers, it means that it leads to huge load on the metrics stack. Consider adding selectors such as `{cluster="..."}` or something else to make the load smaller.

**To Reproduce**

1. Run OpenCost

**Expected behavior**

OpenCost not pulling in a bunch of data from unrelated servers

**Screenshots**

N/A

N/A

**Additional context**

N/A

GiedriusS:
## What does this PR change?

* Add a non-empty label matcher to not select CPU data of all nodes.

* No

* Less unneeded resource usage

Fixes https://github.com/opencost/opencost/issues/1883.

## How was this PR tested?

Ran `go test`

* No.

* It most likely should be part of the next release but I'll defer this to maintainers


KipsasJaujoj:
Tested this on our opencost deployment. It reduced thanos load considerably.

AjayTripathy (approved):

AjayTripathy:
In general this LGTM but would it be possible to get a dump of data of before/after attached to the PR? We don't have unit test coverage here so go test is not going to be sufficient.

GiedriusS:
What kind of information would you be looking for? We're still new to OpenCost so we're not sure what you'd be looking for.

mattray:
@GiedriusS perhaps just printing out the values of `queryCPUModePct` before and after the patch?

KipsasJaujoj:
@mattray Our `PROM_CLUSTER_ID_LABEL` is set to 'cluster'.
Before patch:
`2023-05-08T14:49:35.765808819Z INF [Custom log] queryCPUModePct:
		sum(rate(node_cpu_seconds_total[3h]offset 1m)) by (cluster, mode) / ignoring(mode)
		group_left sum(rate(node_cpu_seconds_total[3h]offset 1m)) by (cluster)`
After patch:
`2023-05-08T14:52:21.555548261Z INF [Custom log] queryCPUModePct:
		sum(rate(node_cpu_seconds_total{cluster=~".+"}[3h]offset 1m)) by (cluster, mode) / ignoring(mode)
		group_left sum(rate(node_cpu_seconds_total{cluster=~".+"}[3h]offset 1m)) by (cluster)
`

mattray:
@AjayTripathy good to merge?

KipsasJaujoj:
@AjayTripathy @mattray Anything else you need for this PR? :)

Sean-Holcomb (changes requested):
The main issue with this PR is that the filter is only applied to one of several queries

Another contributor has added in filters on these queries to ensure that the results are all from the querying cluster.
https://github.com/opencost/opencost/blob/9ae9cb21f9b4ecfc05ad37f709405d3ff32cf21c/pkg/costmodel/cluster.go#L965-L967

Your non empty filter could be added to all relevant queries using this function
https://github.com/opencost/opencost/blob/9ae9cb21f9b4ecfc05ad37f709405d3ff32cf21c/pkg/env/costmodelenv.go#L297-L302

However it cannot be the default because this will break non-federated deployments that do not have a cluster label. To that end you will have to add an env variable to enable the filter. The filter function mentioned above is also used near universally, and you will have to consider if you don't want include clusterless metrics in those other places

PR #1896, opened 2023-05-03
Imp ns labels annots

ameijer:
## What does this PR change?
* This PR introduces new fields, namespaceLabels and namespaceAnnotations, that store namespace labels and annotations, respectively

* none

* Provides additional API fields

* Partially addresses https://github.com/kubecost/cost-analyzer-helm-chart/issues/2027

## How was this PR tested?
* Ensured that namespace labels/annotations were included in returned value

* potential additional API field that needs to be documented

*


mbolt35:
Everything that I have looked through so far looks great. There is also an `allocation_json.go` file I think that will need to add these fields manually (if sending the label data to the client via JSON is necessary).

ameijer:
Thanks @mbolt35, going to keep after this. I checked out the allocationJSON object, and it looks like it just contains a regular AllocationProperties object: 

```
type AllocationJSON struct {
	Name                           string                          `json:"name"`
	Properties                     *AllocationProperties     
```
So I think we are covered there. I also did do a test via API, and do get these namespace labels/annotations back in their new fields so I think the json is serializing correctly.

pkg/kubecost/allocationprops.go
```diff
@@ -476,15 +530,26 @@ func (p *AllocationProperties) Intersection(that *AllocationProperties) *Allocat
...
 	}
 	if p.ProviderID == that.ProviderID {
 		intersectionProps.ProviderID = p.ProviderID
 	}
+
 	return intersectionProps
 }
 
+func copyStringMap(original map[string]string) map[string]string {
+	copy := make(map[string]string)
```
nikovacevic, line 545: Small enhancement---we could set the capacity of `copy` here to `len(original)`

nikovacevic (approved):

PR #1902, opened 2023-05-05
refactor(cloud): Move aws and gcp to dedicated modules

Pokom:
Following the move by Azure folks to refactor the azure module in #1869, I am refactoring AWS and GCP into their own dedicated folders. This was for the most part straight forward. The one caveat is that the customprovider had hard dependencies on AWS resources. This didn't feel right to me, so instead of the two sharing certain structs, I opted to copy over the pvKey implementation to customprovider and a regex.

## What does this PR change?
* refactor `cloud` aws and gcp packages

* #1869 

* There should be no impact

* 

## How was this PR tested?
* Ran locally against AWS and GKE clusters and compared output to previous change

* 

*

### DIFF
--- a/pkg/cloud/customprovider.go
+++ b/pkg/cloud/customprovider.go
@@ -14,6 +14,7 @@ import (
 	"github.com/opencost/opencost/pkg/env"
 	"github.com/opencost/opencost/pkg/kubecost"
 	"github.com/opencost/opencost/pkg/log"
+	"github.com/opencost/opencost/pkg/util"
 	"github.com/opencost/opencost/pkg/util/json"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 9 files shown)


Pokom:
@babbageclunk not a hard ask, but if you have some time mind reviewing this PR?

pkg/cloud/customprovider.go
```diff
@@ -38,6 +39,31 @@ type CustomProvider struct {
...
+
+type customPVKey struct {
+	Labels                 map[string]string
+	StorageClassParameters map[string]string
+	StorageClassName       string
+	Name                   string
+	DefaultRegion          string
+	ProviderID             string
+}
+
```
Pokom, line 66: Just to call out, these were previously available to _every_ file within the cloud module and owned by the `awsprovider.go`. It didn't feel right for these to be shared, so my initial pass was to copy them from aws to customprovider and then we can iterate on decoupling the implementations if need be.

Pokom:
@Sean-Holcomb rebased and ready for a review :rocket: I had called out in comments the changes I had to make that weren't simply moving files around. `customprovider.go` had a number of hard dependencies on  `awsprovider.go` that IMO don't make sense to share. I didn't change any logic, just copied over the `awsprovider.go` over to `customprovider.go` and renamed the structs to be specific to customprovider.

Sean-Holcomb (approved):
Looks good, I will merge as soon as I have made the changes on our side.

Sean-Holcomb (approved):
Looks good, I will merge as soon as I have made the changes on our side.

PR #1912, opened 2023-05-12
Proposal: Measure node overhead as percentage of capacity

ameijer:
## What does this PR change?
* Introduces the concept of node overhead 

* none

* provides a new API field giving the node overhead as a percentage 

* Closes ...

## How was this PR tested?
* checked percentages on dev 1 against values returned by kubectl describe 

* 

*


pkg/costmodel/assets.go
```diff
@@ -133,6 +133,14 @@ func (cm *CostModel) ComputeAssets(start, end time.Time) (*kubecost.AssetSet, er
...
 		node.GPUCount = n.GPUCount
 		node.RAMCost = n.RAMCost
+		combinedRamCPUCost := n.CPUCost + n.RAMCost
+
+		node.Overhead = &kubecost.NodeOverhead{
+			RamOverheadFraction: n.Overhead.RamOverheadFraction,
+			CpuOverheadFraction: n.Overhead.CpuOverheadFraction,
+			OverheadCostFraction: (n.Overhead.CpuOverheadFraction * (n.CPUCost / combinedRamCPUCost)) +
+				(n.Overhead.RamOverheadFraction * (n.RAMCost / combinedRamCPUCost)),
+		}
```
r2k1, line 142: I think this can be inaccurate for node with GPUs (or any extra associated cost). Something like this should take it into account:

```go
OverheadCostFraction: (n.Overhead.CpuOverheadFraction * n.CPUCost  +
				n.Overhead.RamOverheadFraction * n.RAMCost) / n.TotalCost,
```

pkg/kubecost/asset.go
```diff
@@ -1993,6 +2002,14 @@ func (n *Node) add(that *Node) {
 		n.RAMBreakdown.User = (n.RAMBreakdown.User*n.RAMCost + that.RAMBreakdown.User*that.RAMCost) / totalRAMCost
 	}
 
+	if n.Overhead != nil && that.Overhead != nil {
+
+		n.Overhead.RamOverheadFraction = (n.Overhead.RamOverheadFraction*n.RAMCost + that.Overhead.RamOverheadFraction*that.RAMCost) / totalRAMCost
+		n.Overhead.CpuOverheadFraction = (n.Overhead.CpuOverheadFraction*n.CPUCost + that.Overhead.CpuOverheadFraction*that.CPUCost) / totalCPUCost
+		combinedRamCPUCost := n.RAMCost + n.CPUCost
+		n.Overhead.OverheadCostFraction = (n.Overhead.CpuOverheadFraction * (n.CPUCost / combinedRamCPUCost)) + (n.Overhead.RamOverheadFraction * (n.RAMCost / combinedRamCPUCost))
```
r2k1: Same can be applied here

```go
n.Overhead.OverheadCostFraction = (n.Overhead.CpuOverheadFraction * n.CPUCost + n.Overhead.RamOverheadFraction * n.RAMCost) / n.TotalCost
```

pkg/costmodel/assets.go
ameijer: fixed!

pkg/costmodel/cluster_helpers.go
```diff
@@ -426,6 +426,44 @@ func buildCPUBreakdownMap(resNodeCPUModeTotal []*prom.QueryResult) map[nodeIdent
 	return cpuBreakdownMap
 }
 
+func buildOverheadMap(capRam, allocRam, capCPU, allocCPU map[nodeIdentifierNoProviderID]float64) map[nodeIdentifierNoProviderID]*NodeOverhead {
+
+	m := make(map[nodeIdentifierNoProviderID]*NodeOverhead)
```
nikovacevic, line 431: Nit: could add capacity here for perf improvement.

nikovacevic: Nit: could add capacity here for perf improvement.

pkg/costmodel/cluster_helpers.go
```diff
@@ -426,6 +426,44 @@ func buildCPUBreakdownMap(resNodeCPUModeTotal []*prom.QueryResult) map[nodeIdent
...
+
+	for identifier, ramCapacity := range capRam {
+		allocatableRam, ok := allocRam[identifier]
+		if !ok {
+			log.Warnf("Could not find allocatable ram for node %s", identifier.Name)
```
nikovacevic, line 435: So, if we are missing one or more of the (capacity, allocatable) figures, we just don't enter an overhead value under that node's provider ID. Does that mean we'd default to "no overhead" in a failure case? Seems reasonable. Just checking failure case.

nikovacevic: I suppose we could try to guess, too, based on documented figures for certain node types. But that seems over-the-top.

pkg/kubecost/asset_json.go
```diff
@@ -494,7 +494,11 @@ func (n *Node) MarshalJSON() ([]byte, error) {
 	jsonEncodeFloat64(buffer, "gpuCount", n.GPUs(), ",")
 	jsonEncodeFloat64(buffer, "ramCost", n.RAMCost, ",")
 	jsonEncodeFloat64(buffer, "adjustment", n.Adjustment, ",")
+	if n.Overhead != nil {
+		jsonEncode(buffer, "overhead", n.Overhead, ",")
```
nikovacevic, line 498: Nothing to do with your code, but goodness I detest our JSON encoding for Assets. We did it better in Allocation. I might just fix all this crap this week... haha.

nikovacevic (approved):
LGTM!

nikovacevic:
@ameijer would love to hear about the product vision, or your vision, for this!

nikovacevic:
I can immediately see implications for Cluster Sizing, for instance. cc @biancaburtoiu @michaelmdresser

ameijer:
I agree in that a straightforward (but maybe not easy, am not familiar with the implementation) product usage of this would be to incorporate it into node/cluster sizing decisions. 

In my experience, people just don't talk about this or think about it. For that reason, I think also surfacing it in the observability side of things would be neat - it could help people understand why you can't schedule 20 pods requesting 100mcpu on a 2 CPU node. The overhead can be substantial - for example, in our GKE dev-1, the e2-medium nodes have 53% CPU overhead - that is more than 1 VCPU that we cannot actually use. 

If I can indulge my academic side, an exciting thing to me in the immediate term is to conduct a study of node types on the major cloud providers and compare overhead of nodes. could be a good blog post or conference paper? The questions we could ask could be: "does overhead decrease with increasing node size?", "does the operating system one uses for their managed nodes affect overhead?", and "which cloud providers have the least overhead?" Even if just an empirical study, I think would be very interesting.

pkg/costmodel/cluster_helpers.go
ameijer: yes - in the case where we can calculate either the CPU or RAM overhead, then the other would be 0. in the case where we cannot calculate either - then the overhead object in the resulting asset would be nil

pkg/kubecost/asset_json.go
ameijer: I did spend some time searching to see if we did allocations in the same way when I was making some other change... if nothing else, the consistency will help the maintainability

michaelmdresser:
I'll take a look at this in detail soon, merging.

r2k1:
@ameijer here is information about overhead for AKS clusters https://learn.microsoft.com/en-us/azure/aks/concepts-clusters-workloads#resource-reservations

Bigger nodes have less overhead.

ameijer:
@r2k1 very interesting reading, thank you! Lower overhead on bigger nodes is an interesting savings vector 🤔

pkg/kubecost/kubecost_codecs.go
```diff
@@ -40,10 +40,10 @@ const (
 	CloudCostCodecVersion uint8 = 1
 
 	// DefaultCodecVersion is used for any resources listed in the Default version set
-	DefaultCodecVersion uint8 = 17
+	DefaultCodecVersion uint8 = 19
```
Sean-Holcomb, line 43: It looks like you ran bingen with a 19 flag but did not update the command in the bingen file. The flag corresponds to this block
https://github.com/opencost/opencost/blob/87b4c630de3d99bfc791c8409b565a14be685feb/pkg/kubecost/bingen.go#L23-L26
and I don't think that you changed any of those types here. Increasing the version number is fine but you did not also update this line.
https://github.com/opencost/opencost/blob/87b4c630de3d99bfc791c8409b565a14be685feb/pkg/kubecost/bingen.go#LL86C5-L86C97
which means that next time someone runs bingen the `DefaultCodecVersion` will drop back to version 17. This will cause a parsing failure on all windows. 

I saw this because I just ran bingen. Because I am merging that commit, only a couple of hours after this one I am just reverting it to 17. So no action required.

ameijer: thanks @Sean-Holcomb - I think I had been tinkering around with the CLI and mistakenly bumped it to 19. Thanks for the fix and explanation!

PR #1920, opened 2023-05-15
Parcs accum support ii

ameijer:
## What does this PR change?
* This PR Re-implements PARC accumulation support

* re-implementation of #1867 

* provides additional functionality

* Closes ...

## How was this PR tested?
* unit tests were written/updated, and tested on a real azure cluster to sanity check numbers

* no

*


pkg/kubecost/allocation.go
```diff
@@ -267,26 +291,77 @@ func (parc ProportionalAssetResourceCost) Key(insertByNode bool) string {
 
 type ProportionalAssetResourceCosts map[string]ProportionalAssetResourceCost
 
+func (parcs ProportionalAssetResourceCosts) Clone() ProportionalAssetResourceCosts {
+	cloned := ProportionalAssetResourceCosts{}
+
+	for key, parc := range parcs {
+		cloned[key] = parc.Clone()
```
nikovacevic: For values, I think you shouldn't need to clone here, and that may result in better performance. Cloning is required when you have pointers, not values, and want to avoid wacky state mutation bugs.

nikovacevic: I think Go does some fancy compiler stuff when it sees a for loop and values populating a map. But the Clone call probably breaks that.

pkg/kubecost/allocation.go
```diff
@@ -267,26 +291,77 @@ func (parc ProportionalAssetResourceCost) Key(insertByNode bool) string {
...
+		toInsert := ProportionalAssetResourceCost{
+			Node:                curr.Node,
+			Cluster:             curr.Cluster,
+			ProviderID:          curr.ProviderID,
+			CPUTotalCost:        curr.CPUTotalCost + parc.CPUTotalCost,
+			CPUProportionalCost: curr.CPUProportionalCost + parc.CPUProportionalCost,
+			RAMTotalCost:        curr.RAMTotalCost + parc.RAMTotalCost,
+			RAMProportionalCost: curr.RAMProportionalCost + parc.RAMProportionalCost,
+			GPUProportionalCost: curr.GPUProportionalCost + parc.GPUProportionalCost,
+			GPUTotalCost:        curr.GPUTotalCost + parc.GPUTotalCost,
```
nikovacevic, line 301: This is very nice

pkg/kubecost/allocation.go
```diff
@@ -267,26 +291,77 @@ func (parc ProportionalAssetResourceCost) Key(insertByNode bool) string {
...
+		toInsert.CPUPercentage = toInsert.CPUProportionalCost / toInsert.CPUTotalCost
+	}
+
+	if toInsert.GPUTotalCost > 0 {
+		toInsert.GPUPercentage = toInsert.GPUProportionalCost / toInsert.GPUTotalCost
+	}
+
+	if toInsert.RAMTotalCost > 0 {
+		toInsert.RAMPercentage = toInsert.RAMProportionalCost / toInsert.RAMTotalCost
+	}
```
nikovacevic, line 326: This looks good to me. Though it does complicate JSON encoding, another nice way to handle this would be to not even store the percentages statefully, having to keep them in sync with the proportional and total costs. They could just become method functions:

```go
func (a *Allocation) RAMPercentage() float64 {
    return a.RAMProportionalCost / a.RAMTotalCost
}
```

Certainly not saying that's a "requested change" but it's an option if keeping things in sync seems scary at all.

pkg/kubecost/allocation.go
```diff
@@ -1741,48 +1803,111 @@ func computeIdleCoeffs(options *AllocationAggregationOptions, as *AllocationSet,
 	return coeffs, totals, nil
 }
 
-func deriveProportionalAssetResourceCostsFromIdleCoefficients(idleCoeffs map[string]map[string]map[string]float64, totals map[string]map[string]float64, allocation *Allocation, options *AllocationAggregationOptions) (ProportionalAssetResourceCost, error) {
-	idleId, err := allocation.getIdleId(options)
-	if err != nil {
-		return ProportionalAssetResourceCost{}, fmt.Errorf("failed to get idle ID for allocation %s", allocation.Name)
-	}
+func deriveProportionalAssetResourceCosts(options *AllocationAggregationOptions, as *AllocationSet, shareSet *AllocationSet) error {
```
nikovacevic, line 1788: Much nicer

nikovacevic (approved):
I think this looks much nicer! Hope you feel good about it, too. And potentially clears up any concerns @michaelmdresser had about performance.

pkg/kubecost/allocation.go
ameijer: thanks, fixed!

pkg/kubecost/allocation.go
ameijer: thanks. I do feel better about this. And I think this re-write demonstrates a much better understanding of how this stuff works vs 6 weeks ago when you handed this off to me

pkg/kubecost/allocation.go
ameijer: I'm leaning towards keeping it here for debuggability.. certainly helped  me during this re-write

PR #1925, opened 2023-05-17
generate shared cost breakdown and add to allocation struct

nickcurie:
## What does this PR change?
* Adds shared cost breakdown struct to allocations

* https://github.com/kubecost/kubecost-cost-model/pull/1496

* Users can see a breakdown of their shared costs on `model/allocation` with the `includeSharedCostBreakdown` flag

* Closes https://kubecost.atlassian.net/browse/CORE-255

## How was this PR tested?
* Tested on `gcp dev-1`

* 

* v1.104


pkg/kubecost/allocation.go
```diff
@@ -1464,6 +1511,7 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 
 	// (8) Distribute shared allocations according to the share coefficients.
 	if shareSet.Length() > 0 {
+		scBreakdown := SharedCostBreakdowns{}
```
michaelmdresser: Shouldn't this be defined for each row? That would mean it should be instantiated within the `range aggSet.Allocations`, not before.

pkg/kubecost/allocation.go
```diff
@@ -844,6 +895,12 @@ func (a *Allocation) add(that *Allocation) {
 		a.ProportionalAssetResourceCosts.Add(that.ProportionalAssetResourceCosts)
 	}
 
+	// If both Allocations have SharedCostBreakdowns, then
+	// add those from the given Allocation into the receiver.
+	if a.SharedCostBreakdown != nil && that.SharedCostBreakdown != nil {
+		a.SharedCostBreakdown.Add(that.SharedCostBreakdown)
+	}
```
michaelmdresser: What if just one of the Allocations has a SCB? Shouldn't it still be added to the result?

pkg/kubecost/allocation.go
```diff
@@ -356,6 +357,55 @@ func (parcs ProportionalAssetResourceCosts) Add(that ProportionalAssetResourceCo
...
+	Name        string  `json:"name"`
+	TotalCost   float64 `json:"totalCost"`
+	CPUCost     float64 `json:"cpuCost,omitempty"`
+	GPUCost     float64 `json:"gpuCost,omitempty"`
+	RAMCost     float64 `json:"ramCost,omitempty"`
+	PVCost      float64 `json:"pvCost,omitempty"`
+	NetworkCost float64 `json:"networkCost,omitempty"`
+	LBCost      float64 `json:"loadBalancerCost,omitempty"`
+	//todo: what do we do if the shared allocation has a shared cost?
+	// SharedCost float64 `json:"sharedCost"`
```
michaelmdresser: Still a TODO?

pkg/kubecost/allocation.go
```diff
@@ -87,6 +87,7 @@ type Allocation struct {
 	// asset on which the allocation was run. It is optionally computed
 	// and appended to an Allocation, and so by default is is nil.
 	ProportionalAssetResourceCosts ProportionalAssetResourceCosts `json:"proportionalAssetResourceCosts"` //@bingen:field[ignore]
+	SharedCostBreakdown            SharedCostBreakdowns           `json:"sharedCostBreakdown"`            //@bingen:field[ignore]
```
michaelmdresser, line 90: @mbolt35 do bingen codecs have to be regenerated if we add an ignored field? My assumption is no, but I want to make sure.

michaelmdresser (approved):

pkg/kubecost/allocation.go
mbolt35: Nope, these are just transient fields though, so they won't be persisted. I'm guessing that was intended?

michaelmdresser: Yep, that's intended.

michaelmdresser: Thanks!

PR #1930, opened 2023-05-18
Proportional load bal costs

ameijer:
## What does this PR change?
* This PR adds support for tracking load balancer proportions. It also fixes some bugs with PARCs 

* none

* adds support for load balancer tracking to the allocation object

* Closes ...

## How was this PR tested?
* tested locally against cluster, validated numbers 

* None

*


pkg/kubecost/allocation.go
```diff
@@ -282,36 +310,41 @@ func (parcs ProportionalAssetResourceCosts) Clone() ProportionalAssetResourceCos
...
-			CPUProportionalCost: curr.CPUProportionalCost + parc.CPUProportionalCost,
-			RAMTotalCost:        curr.RAMTotalCost + parc.RAMTotalCost,
-			RAMProportionalCost: curr.RAMProportionalCost + parc.RAMProportionalCost,
-			GPUProportionalCost: curr.GPUProportionalCost + parc.GPUProportionalCost,
-			GPUTotalCost:        curr.GPUTotalCost + parc.GPUTotalCost,
+			Name:                         curr.Name,
+			Type:                         curr.Type,
+			Cluster:                      curr.Cluster,
+			ProviderID:                   curr.ProviderID,
+			CPUTotalCost:                 curr.CPUTotalCost,
```
ameijer: It's tough - I'm trying to accomplish this without needing to know whether this is an accumulation or not. In the case this is not an accumulation, let's say we have have 2 containers getting aggregated into pods. Each container costs 1/10th of CPU. We would expect, then, that the resulting pod would cost 2/10ths of a CPU, right? The CPU total cost stays the same, so we should not need to add the CPU total costs together. 

This breaks down with accumulation. Each time slice has both separate overall CPU costs AND allocated CPU costs. Let's say our pod ran for 3 hours, consuming 2/10ths of the CPU for all 3 hours. After accumulation across all 3 hours, we would expect the pod to still have a PARC on the CPU of 2/10ths. This same logic then breaks down. if we keep this logic and only add the numerators, then at the end of the 3 hours, our parc says that we consumed 6/10ths of the CPU, which is not correct.

nikovacevic: Ah yes, definitely an issue. It's tough, keeping the total cost on the individual allocations, isn't it? It'd almost be better to somehow track the totals separately. So does this current solution work for both acc and agg? Or just fixed for agg, but now broken again for acc?

nikovacevic: Does the TotalsStore help you at all here? Are you familiar with that?

nikovacevic: https://github.com/opencost/opencost/blob/develop/pkg/kubecost/totals.go#L13-L38

nikovacevic: If you have access to that, you may be able to just record the "local" proportional costs and look up the totals at the end.

pkg/kubecost/allocation.go
```diff
@@ -934,11 +972,38 @@ func (a *Allocation) add(that *Allocation) {
 	a.NetworkCostAdjustment += that.NetworkCostAdjustment
 	a.LoadBalancerCostAdjustment += that.LoadBalancerCostAdjustment
 
+	// Sum LoadBalancer Allocations
+	a.LoadBalancers = a.LoadBalancers.Add(that.LoadBalancers)
```
nikovacevic, line 1039: We probably wouldn't expect this to be _too_ expensive, but in environment with lots of load balancers, and lots of relationships between pods and load balancers, I wonder how heavy this gets. Allocating memory for maps within loops (i.e. the `Clone()` on line 985 happens for each and every allocation addition, and then allocating for map resizing on insert) is something to be careful with.

I bring it up here because we're no longer behind a flag, it seems. This happens on the main path for `Allocation.Add`. That said... I'm not terribly afraid of degraded performance here because this API is already slow, and not used by the UI anywhere. Just something to be mindful of.

nikovacevic (approved):

PR #1933, opened 2023-05-18
Ajay tripathy combined fixes

### LINKED ISSUE #2250: Support custom PV prices  that are determined by Storage Class
Let users specify storage class in CSV file. Use the specified storage class for determining PV hourly cost based on its Storage Class

EndTimeStamp,InstanceID,Region,AssetClass,InstanceIDField,InstanceType,MarketPriceHourly,Version
,,pv,metadata.name,storage_class_name,333,

We can specify PV names in CSV file

User will specify storage class in CSV file

AjayTripathy:
## What does this PR change?
* Fix on-prem PV pricing and data emission

* No

* Broader support for on-prem PVs

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/2250

## How was this PR tested?
* @ivankube spun up local clusters to support this

* Yes, we'll want to modify on-prem docs to show that this is supported.

* Yes, for 104


pkg/cloud/provider/csvprovider.go
```diff
@@ -308,10 +308,10 @@ func NodeValueFromMapField(m string, n *v1.Node, useRegion bool) string {
 		if mf[1] == "name" {
 			return toReturn + n.Name
 		} else if mf[1] == "labels" {
-			lkey := strings.Join(mf[2:], "")
+			lkey := strings.Join(mf[2:len(mf)], ".")
```
mbolt35, line 311: The `len(mf)` shouldn't be necessary.

pkg/metrics/kubemetrics.go
```diff
@@ -108,6 +108,9 @@ func InitKubeMetrics(clusterCache clustercache.ClusterCache, metricsConfig *Metr
 				metricsConfig:    *metricsConfig,
 			})
 		} else if opts.EmitKubeStateMetricsV1Only {
+			// We still need the kubecost_pv_info metric to look up storageclass on legacy clusters.
+			forceDisabled := []string{"kube_persistentvolume_capacity_bytes", "kube_persistentvolume_status_phase"}
```
mbolt35, line 112: These aren't KSM V1 Metrics?

ameijer (approved):

pkg/cloud/provider/csvprovider.go
AjayTripathy: Some kind of merge conflict happened here where the "." below got ommitted when merging from develop which I think removed the len(mf). Will remove.

pkg/metrics/kubemetrics.go
AjayTripathy: the v1Only is for KSM v1 metrics is to emit the metrics delta between v1 and v2 iirc. this preserves the old logical behavior since when flag was set the whole set of PV metrics would not be emitted previously. Was that behavior a bug before?

mbolt35 (approved):
LGTM

PR #1834, opened 2023-04-03
Run UI as USER 1001

AjayTripathy:
## What does this PR change?
* Run as user 1001

## Does this PR relate to any other PRs?
* No

## How will this PR impact users?
* Allows UI container to run as nonroot

## Does this PR address any GitHub or Zendesk issues?
* Closes https://github.com/opencost/opencost/issues/1824

## How was this PR tested?
* Will get tested by CI / CD building the container

## Does this PR require changes to documentation?
* No 

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next Opencost release? If not, why not?
* v103

PR #1934, opened 2023-05-19
Enable image to run as non-root

### LINKED ISSUE #1824: Opencost-ui container uses root user.

Opencost-ui container is using root user. It poses some potential vectors for attack and best practices is to launch a container as a non-root user.

Some kubernetes clusters may be configured with a policy preventing such containers from starting. Opencost API container is using non-root user.

https://github.com/opencost/opencost/blob/develop/Dockerfile
https://github.com/opencost/opencost/blob/develop/ui/Dockerfile

JorgeGar:
## What does this PR change?
* This PR allows Opencost UI to run as non-root by creating a user and giving it ownership over the folders `/var/www` and `/etc/nginx`.
It also removes the file `/etc/nginx/conf.d/default.conf` which was exposing an unneeded endpoint in port 80, which is not allowed to bind as non-root.

* https://github.com/opencost/opencost/pull/1834

* Existing users shouldn't be affected. Any user that wants to run the image as non-root must use user 1001.

* Closes https://github.com/opencost/opencost/issues/1824

## How was this PR tested?
* Build and run the image both locally and in a secured Kubernetes environment and it will get tested by CI / CD building the container

* No

* v1.103

### DIFF
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -11,15 +11,15 @@ COPY default.nginx.conf /etc/nginx/conf.d/
 COPY nginx.conf /etc/nginx/
 RUN rm -rf /etc/nginx/conf.d/default.conf
 
-RUN adduser 101 -g 1000 -D
-RUN chown 101:1000 -R /var/www
-RUN chown 101:1000 -R /etc/nginx
+RUN adduser 1001 -g 1000 -D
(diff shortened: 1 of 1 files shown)


JorgeGar:
I've created https://github.com/opencost/opencost/pull/1947 to fulfil DCO requirements.

PR #1941, opened 2023-05-22
Run builds locally before containerizing and add multi-arch support

michaelmdresser:
## What does this PR change?
* Supports my effort to deprecate architecture-specific build boxes maintained by Kubecost. Has the added benefit of speeding up the dev workflow for OpenCost contributors using ARM dev machines. Using native toolchains to cross-compile outside of Docker avoids huge build overhead induced by `buildx`'s use of QEMU.
* The switch to out-of-container CI means that I was able to add a cache for `npm` and Go, shortening overall CI completion time from ~8min to to as low as 1min45s

* Based on https://github.com/kubecost/kubecost-cost-model/pull/1505

* 

* `nix-shell -p just nodejs-18_x docker --run "cd ui && just build <tag>` succeeds
* `just build <tag>` succeeds
* This PR partially tests itself -- the CI workflow is updated to use the new `just` build and test commands.

* 

*


.github/workflows/pr.yaml
```diff
@@ -1,32 +1,79 @@
...
+
+      -
+        name: Install just
+        uses: extractions/setup-just@v1
+
+      -
+        name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.19'
```
ameijer: is there any way you can leave this off to use most recent go? I am just worried about keeping pinned versions up to date

.github/workflows/pr.yaml
```diff
@@ -1,32 +1,79 @@
...
+
+      # Saves us from having to redownload all modules
+      - name: Go Mod cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
```
ameijer, line 33: does a cache help us? aren't these just ephemeral builders?

.github/workflows/pr.yaml
cliffcolvin: I believe we can use `stable` here as long as there aren't version dependencies in the build.

ameijer: yes, stable might help keep us off the bleeding edge while still moving us forward. 

My view on versions is that if we can always just auto upgrade to latest stable, we can discover go compat issues slowly and during a dev build when it is easy to fix. If we pin, then upgrading to a new version becomes a risky thing because we are usually jumping multiple versions. 

I am interested in what we have been doing here to date

.github/workflows/pr.yaml
michaelmdresser: A cache is a massive speedup. Without the cache, the (indeed, yes) ephemeral builders have to re-download every one of our dependencies from public repos. By caching them (keyed on the `go.sum`s) we get to download them from an internal mirror, resulting in multi-minute savings.

.github/workflows/pr.yaml
michaelmdresser: Our `go.mod` does specify a Go version, but not because of any hard dependency. I'm definitely supportive of using latest stable here. I've updated this PR accordingly (and the KCM PR the same).

> I am interested in what we have been doing here to date

We've been using latest stable AFAIK -- all of our Dockerfiles are `golang:latest`.

ameijer (approved):
ship it 👍

nikovacevic (approved):

PR #1942, opened 2023-05-23
add utility functions for group requests

saweber:
This is for powering group requests, and adds the benefit of slightly decoupling utility services from request handlers.

## What does this PR change?
* Adds utility functions and structs for Allocation filters and CloudCost filters

## Does this PR relate to any other PRs?
* yes, https://github.com/kubecost/kubecost-cost-model/pull/1478/files

## How will this PR impact users?
* no effect for opencost, enables new kubecost functionality

## Does this PR address any GitHub or Zendesk issues?
* CORE-234

## How was this PR tested?
* existing + new automated tests

## Does this PR require changes to documentation?
* no

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next Opencost release? If not, why not?
* no permissions

PR #1952, opened 2023-05-25
add omitempty to filters to clean up group report responses

saweber:
## What does this PR change?
* adds omitempty to types primarily for group reports

* https://github.com/opencost/opencost/pull/1942

* remove empty fields from group report responses

* no

## How was this PR tested?
* manually

* no

* can't

### DIFF
--- a/pkg/filter/util/cloudcost.go
+++ b/pkg/filter/util/cloudcost.go
@@ -11,13 +11,13 @@ import (
 )
 
 type CloudCostFilter struct {
-	AccountIDs       []string `json:"accountIDs"`
-	Categories       []string `json:"categories"`
-	InvoiceEntityIDs []string `json:"invoiceEntityIDs"`
-	Labels           []string `json:"labels"`
-	Providers        []string `json:"providers"`
-	ProviderIDs      []string `json:"providerIDs"`
-	Services         []string `json:"services"`
+	AccountIDs       []string `json:"accountIDs,omitempty"`
+	Categories       []string `json:"categories,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/filter/util/cloudcost.go
```diff
@@ -11,13 +11,13 @@ import (
...
-	Providers        []string `json:"providers"`
-	ProviderIDs      []string `json:"providerIDs"`
-	Services         []string `json:"services"`
+	AccountIDs       []string `json:"accountIDs,omitempty"`
+	Categories       []string `json:"categories,omitempty"`
+	InvoiceEntityIDs []string `json:"invoiceEntityIDs,omitempty"`
+	Labels           []string `json:"labels,omitempty"`
+	Providers        []string `json:"providers,omitempty"`
+	ProviderIDs      []string `json:"providerIDs,omitempty"`
+	Services         []string `json:"services,omitempty"`
```
nealormsbee, line 20: Is this struct used for any responses other than the `groupReport` configuration (maybe cloud cost reports)?

The change to `omitempty` could affect other consumers if so. If this is localized entirely to `/getGroupReports`, then nevermind!

saweber: there are no external uses outside of group reports, so there is no impact. It is used internally inside the application, but since it is not treated as JSON, will have no impact.

nikovacevic (approved):
Will echo what Neal said---if we understand the impact this might have on compatibility, then no issue. That said, I don't know of anywhere these get JSON encoded/decoded, so probably fine?

saweber:
@nikovacevic Correct, this is currently not used outside group reports.

PR #1962, opened 2023-05-31
Fix ELV-102: interpolate missing cost and capacity data for nodes, when available

nikovacevic:
## What does this PR change?
* When there is an outage that causes missing data for certain OpenCost-related Prometheus metrics, it can cause costs and resource capacities of assets (e.g. nodes) to be correspondingly low. This corrects for that, when it can be detected in the process of accumulating data over time, leading to more correct figures and fixing apparent inaccuracies in figures like CPU cores and RAM GiBs.

* More accurate numbers when OpenCost has been down

* Fixes the node issues in https://kubecost.atlassian.net/browse/ELV-102

## How was this PR tested?
* New unit test, which failed before and passes now

Before:
```
Running tool: /usr/local/go/bin/go test -timeout 200s -run ^TestNode_Add_AccumulationInterpolation$ github.com/opencost/opencost/pkg/kubecost -v -count=1

=== RUN   TestNode_Add_AccumulationInterpolation
    /home/niko/code/kubecost/model/opencost/pkg/kubecost/asset_test.go:612: CPUCores: expected 2.000000, actual 1.333333
--- FAIL: TestNode_Add_AccumulationInterpolation (0.00s)
FAIL
FAIL	github.com/opencost/opencost/pkg/kubecost	0.005s
```

After:
```
Running tool: /usr/local/go/bin/go test -timeout 200s -run ^TestNode_Add_AccumulationInterpolation$ github.com/opencost/opencost/pkg/kubecost -v -count=1

=== RUN   TestNode_Add_AccumulationInterpolation
{"level":"debug","time":"2023-05-31T16:13:42-06:00","message":"Node.add: corrected CPU cost and capacity by interpolation for node-1 by 1.500000"}
{"level":"debug","time":"2023-05-31T16:13:42-06:00","message":"Node.add: corrected RAM cost and capacity by interpolation for node-1 by 1.500000"}
--- PASS: TestNode_Add_AccumulationInterpolation (0.00s)
PASS
ok  	github.com/opencost/opencost/pkg/kubecost	0.004s
```

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -9,7 +9,9 @@ import (
 	"time"
 
 	"github.com/opencost/opencost/pkg/log"
+	"github.com/opencost/opencost/pkg/util"
 	"github.com/opencost/opencost/pkg/util/json"
+	"github.com/opencost/opencost/pkg/util/stringutil"
 	"github.com/opencost/opencost/pkg/util/timeutil"
 )
 
@@ -1936,8 +1938,51 @@ func (n *Node) Add(a Asset) Asset {
 
 func (n *Node) add(that *Node) {
 	if n == nil {
-		n = that
 		return
+	}
+
+	// If we detect that we are accumulating a unique node, then we can make
+	// stronger assumptions later on that may allow us to correct for inaccuracies
+	// in accumulated data (e.g. cascading data pipelines) by interpolating
+	// correct data where it was briefly lost.
+	// For more detail, see ELV-102 https://kubecost.atlassian.net/browse/ELV-102
+	accumulatingUniqueNode := false
+	var thisCPUCores, thisRAMBytes float64
+	var thatCPUCores, thatRAMBytes float64
+	if n.Properties.ProviderID != "" && n.Properties.ProviderID == that.Properties.ProviderID {
+		accumulatingUniqueNode = true
+
+		// CPU cores and RAM bytes for each should match. If not, log a debug
+		// statement. If these values match, but the node resulting from the
+		// addition does not match, then we are missing data and can fix that
+		// by interpolation.
+		//
+		// e.g. if we are accumulating two hours of data, for the same node,
+		// which has 2 CPU cores and 8 GiB RAM, but Kubecost was down for the
+		// final 20m and the first 30m of the back-to-back hours, then without
+		// correction we would yield:
+		//  - (end-start) = 2 hours
+		//  - CPU = 1.3333 + 1 = 2.3333 core-hours == 1.16667 cores (58%)
+		//  - RAM = 5.3333 + 4 = 9.3333 gib-hours  == 4.66667 GiB (58%)
+		//
+		// We will also end up having only 58% of the actual RAM and CPU costs.
+		// No good! But we can detect this (by comparing capacities for these,
+		// before and after, which should never change) and scaling costs and
+		// capacities up to 100%.
+
+		thisCPUCores = n.CPUCores()
+		thisRAMBytes = n.RAMBytes()
+
+		thatCPUCores = that.CPUCores()
+		thatRAMBytes = that.RAMBytes()
+
+		if !util.IsApproximately(thisCPUCores, thatCPUCores) {
+			log.Debugf("Node.add: same node has different CPU core values: %s = %f and %f", n.Properties.Name, thisCPUCores, thatCPUCores)
+		}
+
+		if !util.IsApproximately(thisRAMBytes, thatRAMBytes) {
+			log.Debugf("Node.add: same node has different RAM GiB values: %s = %f and %f", n.Properties.Name, thisRAMBytes/stringutil.GiB, thatRAMBytes/stringutil.GiB)
+		}
 	}
 
 	props := n.Properties.Merge(that.Properties)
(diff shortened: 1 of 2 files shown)


nikovacevic:
Note that this will affect disks, too. And pods, though it's more difficult to interpolate resource allocation for a pod. (Would we even want to do this for pods? Would we use resource requests and ignore usage?)

At the very least, we should probably attempt a similar fix for disks.

cliffcolvin (approved):
LGTM; Nice find

michaelmdresser (approved):
Core logic seems correct. Does this hold up if one of the Windows is currently in progress? E.g. 
```
   win1   |      win2

 -------  |   ---now---   
```

pkg/kubecost/asset.go
```diff
@@ -2012,11 +2057,44 @@ func (n *Node) add(that *Node) {
...
 	if n.Overhead != nil && that.Overhead != nil {
-
 		n.Overhead.RamOverheadFraction = (n.Overhead.RamOverheadFraction*n.RAMCost + that.Overhead.RamOverheadFraction*that.RAMCost) / totalRAMCost
 		n.Overhead.CpuOverheadFraction = (n.Overhead.CpuOverheadFraction*n.CPUCost + that.Overhead.CpuOverheadFraction*that.CPUCost) / totalCPUCost
 		n.Overhead.OverheadCostFraction = ((n.Overhead.CpuOverheadFraction * n.CPUCost) + (n.Overhead.RamOverheadFraction * n.RAMCost)) / n.TotalCost()
 	}
+
+	// Check and correct values by interpolation.
+	// For more detail, see ELV-102 https://kubecost.atlassian.net/browse/ELV-102
+	if accumulatingUniqueNode {
```
ameijer, line 2067: kind of minor, but would something like addingUniqueNode work here as a var name? Does add know/care if it is part of an accumulation vs something else?

nikovacevic: I struggled with this name. (Got me!) If the node is unique, and we see two of them, then by definition we are accumulating---the same node can only be added with itself _over time_, as it cannot run twice at the same time. But `addingUniqueNode` is an equivalently good name. Anybody else want to help name this variable with us? :wink:

nikovacevic:
> Core logic seems correct. Does this hold up if one of the Windows is currently in progress?

@michaelmdresser It should, theoretically, but I am happy to add a unit test case! To make sure I understand... it'd be like the existing test case, except that the second hour would be "in progress," e.g.

```
   |
   |------      |  -----     |  => 30m + 25m = measured 55m out of expected 95m
   |                   ^                       (full 120m has not happened yet)
 __|____________|____________|
  0|           1|   NOW^    2|
```

nikovacevic:
I might leave this open until we decide on disks (strongly suggest doing) and pods (harder call) and maybe other things (like what? load balancers? cluster management?)

pkg/kubecost/asset.go
```diff
@@ -1936,10 +1938,53 @@ func (n *Node) Add(a Asset) Asset {
...
+		//  - (end-start) = 2 hours
+		//  - CPU = 1.3333 + 1 = 2.3333 core-hours == 1.16667 cores (58%)
+		//  - RAM = 5.3333 + 4 = 9.3333 gib-hours  == 4.66667 GiB (58%)
+		//
+		// We will also end up having only 58% of the actual RAM and CPU costs.
+		// No good! But we can detect this (by comparing capacities for these,
+		// before and after, which should never change) and scaling costs and
+		// capacities up to 100%.
+
+		thisCPUCores = n.CPUCores()
```
ameijer, line 1973: I see what you are doing here. At this point in the code, you can trust that thisCPUCores is accurate since the cpu core hours should line up with the node start end, and we are at 'maximum granularity', before expanding our windows and such and really starting the add operation. This lets you accurately describe the scale factor later on. It helped to think of thisCPUCores as a variable like cpuCoresPreAdd or something. 

It is after you adjust the windows and start/end times that the mismatching starts to happen since the existing logic made assumptions about start and end times which you are now interpolating here. 

reading over your comment 
```
                //If these values match, but the node resulting from the
		// addition does not match, then we are missing data and can fix that
		// by interpolation.
```

I don't see logic where you are enforcing the if condition that snippet started with - if the values don't match, you log an error but still continue the operation. Maybe comment needs updating? What action would you take if these don't match

ameijer (commented):
If I stop my node for the 40 mins out of the 120 mins window, will kubecost assume that it was running and give it cost even though I stopped it, or is there other logic to catch that scenario?

michaelmdresser:
> To make sure I understand... it'd be like the existing test case, except that the second hour would be "in progress," e.g.

Yes, precisely.

michaelmdresser:
> 

If I stop my node for the 40 mins out of the 120 mins window, will kubecost assume that it was running and give it cost even though I stopped it, or is there other logic to catch that scenario?


I profess unfamiliarity with how this works; do providers offer a mechanism to stop a node and its billing for a while, then restart it? And it would maintain the same `ProviderID`?

ameijer:
@michaelmdresser Here's an example sequence from AWS: 

State 1: running
image
State 2: stopped
image
State 3: running once more
image

Here's what AWS says about the stopped state: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Stop_Start.html

I am not sure what a stopped node looks like to k8s/prom, but I would assume it looks like a dead node

AjayTripathy:
> Would we use resource requests and ignore usage?

I believe that this is the current behavior. This seems fine but I think this is an improvement-- what's the lift to use this same interpolation for pods?


I might leave this open until we decide on disks (strongly suggest doing) and pods (harder call) and maybe other things (like what? load balancers? cluster management?)

+1 for doing on disks since it seems straightforward. I think loadbalancers also makes sense.

nikovacevic:
Thanks @AjayTripathy. The lift for Pods is probably not much more than what you see here (i.e. a few hours). I imagine it's pretty much exactly the same idea. Just want to be sensitive to the fact that we're not using UIDs on Pods, at least not universally, and so it's possible that a StatefulSet scaling up and down, or a pod with a non-unique name that is controlled in a way that brings it up and down could cause us to be very far off if we do this.

I can definitely carry this across to disks, pretty easily. But I do want to be sensitive Alex's and Michael's concerns, which we still need to clarify a bit more, I think. (E.g. does AWS charge for a stopped node, or do you stop paying for it?)

ameijer:
@nikovacevic AWS does not charge for stopped nodes: 

> When you stop an instance, we shut it down. We don't charge usage for a stopped instance, or data transfer fees, but we do charge for the storage for any Amazon EBS volumes. 

Not trying to hold up this PR or anything - just wanted to bring this in view. A stopped node in a k8s cluster... not sure how common that would be, but I do know people used stopped instances a lot outside of the k8s clusters

nikovacevic:
@ameijer not at all, that's very helpful. I've been meaning to return to this.

In light of that, where do we go from here? It seems that we're in a bit of a bind, because there are two ways we can be wrong:
1. Kubecost was down, so we miss some resource-hours and cost, and we DON'T merge this, so there's no corrective measure. The symptoms are as described in the bug:  a node might show up with < 100% of its actual cost or capacity; and over time it'll appear that the node shrunk and regained capacity over the window in which the downtime occurred.
2. We DO merge this, and we interpolate over a period of time when a node was stopped, so we overcharge.

Interestingly, in both cases the cost will be corrected by CUR reconciliation. Even more interestingly, case (2) above, in which a node is stopped and then restarted, actually isn't even supportable in Kubecost! A unique node can only have one start time and one end time by Kubecost's data model. So there's no way for us to do (2) correctly.

With that in mind, I'm weighing two thoughts in how to advise on a decision here:
1. How often is Kubecost down?
2. How often do users manually stop nodes in order to not pay for them, as opposed to scaling down and back up?

PR #1969, opened 2023-06-06
Nik/core 290

nik-kc:
## What does this PR change?
* Fixes an issue when inspecting shared costs with sharedNamespaces and/or sharedLabels set, where the shared costs would not be grouped properly.

* [Yup](https://github.com/kubecost/kubecost-cost-model/pull/1550).

* Inspected shared costs will be grouped correctly.

* [Addresses this issue](https://kubecost.atlassian.net/browse/CORE-290).

## How was this PR tested?
* Determined a set of sharedNamespaces and sharedLabels known to cause the above issue. Manually verified that all test cases were resolved with changes in this PR.

* Nope.

##
Before
Before
After
After

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -9,6 +9,7 @@ import (
 	"github.com/opencost/opencost/pkg/log"
 	"github.com/opencost/opencost/pkg/util"
 	"github.com/opencost/opencost/pkg/util/timeutil"
+	"golang.org/x/exp/slices"
 )
 
 // TODO Clean-up use of IsEmpty; nil checks should be separated for safety.
@@ -1045,6 +1046,8 @@ type AllocationAggregationOptions struct {
 	Reconcile                             bool
 	ReconcileNetwork                      bool
 	ShareFuncs                            []AllocationMatchFunc
+	SharedNamespaces                      []string
+	SharedLabels                          map[string][]string
 	ShareIdle                             string
 	ShareSplit                            string
 	SharedHourlyCosts                     map[string]float64
@@ -1539,7 +1542,12 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 					if alloc.SharedCostBreakdown == nil {
 						alloc.SharedCostBreakdown = map[string]SharedCostBreakdown{}
 					}
-					sharedCostName := sharedAlloc.generateKey(aggregateBy, options.LabelConfig)
+
+					sharedCostName, err := sharedAlloc.determineSharingName(options)
+					if err != nil {
+						return fmt.Errorf("failed to group shared costs: %s", err)
+					}
+
 					// check if current allocation is a shared flat overhead cost
 					if strings.Contains(sharedAlloc.Name, SharedSuffix) {
 						sharedCostName = "overheadCost"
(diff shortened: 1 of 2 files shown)


pkg/kubecost/allocation.go
```diff
@@ -1539,7 +1542,56 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
...
+					}
+					sharedCostName := sharedAlloc.generateKey(sharedAggregateBy, options.LabelConfig)
+
+					// get each value in the generated key, then reset the name
+					sharedCostNameValues := strings.Split(sharedCostName, "/")
+					sharedCostName = ""
+
+					// try to match to the first label
+					for _, sharedLabelValues := range options.SharedLabels {
+						for _, sharedCostLabel := range sharedCostNameValues {
```
nikovacevic: I don't think you exactly need (or want) to check each label value with each name because then you'll end up comparing label values with namespace values (as well as label values with unrelated label values).

e.g. if you are sharing by `app:cost-analyzer, env:test, namespace:kubecost` and you get the key `__unallocated__/test/default` you just want to compare:
1. `__unallocated__ == cost-analyzer` ? false
2. `test == test` ? true

This code would do something slightly different:
1. `cost-analyzer == __unallocated__` ? false
2. `cost-analyzer == test` ? false
3. `test == __unallocated__` ? false
4. `test == test` ? true

Not only is this more inefficient, it's prone to accidental matches, where you're comparing a label to a value the wrong value and might _accidentally match_ on it.

I think we just need a more linear checking of (first label matches first value?), (second label matches second value?) ... in that order. Does that make sense?

pkg/kubecost/allocation.go
```diff
@@ -1539,7 +1542,56 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
...
+						}
+						if len(sharedCostName) > 0 {
+							break
+						}
+					}
+
+					// if we didn't match to a label, try to match to a namespace
+					for _, sharedNamespace := range options.SharedNamespaces {
+						for _, sharedCostLabel := range sharedCostNameValues {
+							if sharedCostLabel == sharedNamespace {
```
nikovacevic: `sharedCostLabel` is a slightly misleading variable name, as it's not a label, but I don't think there's a logical issue here.

nikovacevic: Also, same issue here---you don't want to compare the shared namespaces to _all values_ here, right? Because some of those might be label values? We just want apples-to-apples comparisons.

nikovacevic (commented):
Close! Practically there. Let's just try to compare apples-to-apples with those labels and namespaces.

pkg/kubecost/allocation.go
```diff
@@ -1539,7 +1542,56 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
...
+						// try to match to the first label
+						for i, sharedLabelValues := range sharedLabels {
+							allocLabel := sharedCostNameValues[i]
+							if slices.Contains(sharedLabelValues, allocLabel) {
+								sharedCostName = allocLabel
+								break
+							}
+						}
+
+						// if we didn't match to a label, try to match to a namespace
```
nikovacevic: I think we'll do this even if we do match a label. Probably very minor, but might be something to try to fix nevertheless.

Have you considered making this whole chunk of code a function so that you can early return? E.g.
```go
func determineSharingName(sharedAlloc, options) (string, error)
```

Two advantages to that:
1. Easier to unit test just this little portion of logic
2. Can early return, instead of trying to break in funny ways

nikovacevic (commented):
If fixing that last little thing is a huge pain, let me know and I'll approve and then we can revisit.

pkg/kubecost/allocation.go
```diff
@@ -1984,7 +1987,12 @@ func deriveProportionalAssetResourceCosts(options *AllocationAggregationOptions,
 	return nil
 }
 
-func (a *Allocation) determineSharingName(options *AllocationAggregationOptions) string {
+func (a *Allocation) determineSharingName(options *AllocationAggregationOptions) (string, error) {
+	if a == nil {
+		return "", fmt.Errorf("determineSharingName called on nil Allocation")
+	} else if options == nil {
+		return "", fmt.Errorf("nil options passed into determineSharingName")
```
nikovacevic: Very small tweak---it would be a shame for this to blow up the entire `AggregateBy` process. Can we have this instead return `"unknown", nil`?

pkg/kubecost/allocation.go
```diff
@@ -1984,7 +1987,12 @@ func deriveProportionalAssetResourceCosts(options *AllocationAggregationOptions,
 	return nil
 }
 
-func (a *Allocation) determineSharingName(options *AllocationAggregationOptions) string {
+func (a *Allocation) determineSharingName(options *AllocationAggregationOptions) (string, error) {
```
nikovacevic, line 1990: Nice! This feels very tidy now, as a function.

pkg/kubecost/allocation_test.go
```diff
@@ -3214,3 +3214,200 @@ func Test_AggregateByService_UnmountedLBs(t *testing.T) {
...
+			"env": "env1",
+		},
+		Namespace: "namespace1",
+	}
+
+	// test non-nil allocation with nil options
+	name, err = alloc.determineSharingName(nil)
+	if err == nil {
+		t.Fatalf("determineSharingName: expected error; actual nil")
+	}
```
nikovacevic: If we make the suggested change above, then this will have to change to asserting `"unknown", nil`

pkg/kubecost/allocation_test.go
```diff
@@ -3214,3 +3214,200 @@ func Test_AggregateByService_UnmountedLBs(t *testing.T) {
 	spew.Config.DisableMethods = true
 	t.Logf("%s", spew.Sdump(set.Allocations))
 }
+
+func Test_DetermineSharingName(t *testing.T) {
```
nikovacevic, line 3218: Phenomenal. This should increase our confidence in the correctness of this change set immensely.

nikovacevic (approved):
Looks great! Just consider that one suggestion, and let me know when you're ready for a merge.

PR #1971, opened 2023-06-06
Implement v2.1 filtering logic for Allocation

michaelmdresser:
## What does this PR change?
No user-facing change.

Adds Allocation-specific supporting code to v2.1 filters. This PR does not actually switch internal types to use the new representation, just continues to build up the necessary scaffolding to make the switchover easier to implement and review.

## Does this PR relate to any other PRs?
* Requires (stacked on) https://github.com/opencost/opencost/pull/1970
* Code taken almost wholesale from https://github.com/opencost/opencost/pull/1762 (see #1970 for an explanation of why)

## How will this PR impact users?
* N/A

## Does this PR address any GitHub or Zendesk issues?
* https://kubecost.atlassian.net/browse/CORE-276

## How was this PR tested?
* Unit tests in this PR

## Does this PR require changes to documentation?
*

PR #1970, opened 2023-06-06
Add filter v2.1 scaffolding

michaelmdresser:
## What does this PR change?
No user-facing change!

This introduces just the scaffolding (AST, compiler, utils, etc.) from https://github.com/opencost/opencost/pull/1762 to a new package: `pkg/filter21`.

Implementation will roughly follow the following plan:

1. Scaffolding (this PR)
2. Allocation implementation for filter21 (https://github.com/opencost/opencost/pull/1971)
3. Switchover of Allocation types/query to filter21 Matcher interface (note to self: requires update to `pkg/util/allocationfilterutil` to add v1 -> v2.1 filters conversion)
5. Repeat 2+3 for Asset
6. Repeat 2+3 for CloudCost
N. Remove old filter v2 implementations from the codebase and rename `pkg/filter21` to `pkg/filter`

## Does this PR relate to any other PRs?
* Code taken almost wholly from https://github.com/opencost/opencost/pull/1762, updated for the latest version of OC, pieced out into multiple PRs, and avoiding a hard switchover of implementations until all of the necessary scaffolding is set up to minimize PR review fatigue.
* Foundation of https://github.com/opencost/opencost/pull/1971

## How will this PR impact users?
* N/A

## Does this PR address any GitHub or Zendesk issues?
* https://kubecost.atlassian.net/browse/CORE-276
* https://kubecost.atlassian.net/browse/CORE-275

## How was this PR tested?
* Unit tests

## Does this PR require changes to documentation?
*

PR #1982, opened 2023-06-09
Switch Allocation filters to v2.1

michaelmdresser:
## What does this PR change?
* #1970 and #1971 introduced an updated internal representation for filters that includes support for more complex filter expressions and provides the foundation for alias support which will be added in a future PR. This PR removes the now-defunct `AllocationFilterCondition` filter representation, replacing it with the new v2.1 AST.
* Note that `filterutil.go` comes primarily from #1762 with some additions
* The general principle is that we should be carrying around the _representation_ of a filter (`ast.FilterNode` and/or `filter21.Filter`) as long as possible, only compiling it down to a `Matcher` when we have to actually use the filter. By threading the filter through the code in this way, we can eventually gain support for more performant, filtered store querying by compiling the filter representation to a store query language.
* The v2.1 filter string language (e.g. `namespace:"kubecost"`) has no breaking change here, purely a backwards-compatible upgrade as covered in Bolt's superb commentary on #1762 . That immediately gets us "wildcards" in the form of the prefix and suffix syntax.
* The eventual plan is to remove `pkg/filter` and replace it with `pkg/filter21`, but we need to get through Assets and CloudCosts with v2.1 filters first

* Requires (stacked on) https://github.com/opencost/opencost/pull/1971
* Bolt's work in https://github.com/opencost/opencost/pull/1762 continues to be the foundation
* Required for https://github.com/kubecost/kubecost-cost-model/pull/1564

* N/A (the meaningful change to APIs will happen in the KCM PR)

* Part of https://kubecost.atlassian.net/browse/CORE-276

## How was this PR tested?
* Unit tests pass, which extensively cover parsing from nonstandard formats like SAML, reports, etc.
* Ran locally and made `/allocation` queries using the new filter syntax and observed expected results
* Loaded on a local cluster and clicked around the UI, things appear correct


pkg/kubecost/allocation.go
```diff
@@ -1135,7 +1147,11 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
 	// an empty slice implies that we should aggregate everything. See
 	// generateKey for why that makes sense.
 	shouldAggregate := aggregateBy != nil
-	shouldFilter := options.Filter != nil
+	shouldFilter := true
+	// TODO: Make sure this works as expected
```
nikovacevic: Still TO DO?

nikovacevic: Is this potentially something we could/should add to the interface? Something like `IsEmpty()`?

pkg/kubecost/allocation.go
```diff
@@ -1357,14 +1373,9 @@ func (as *AllocationSet) AggregateBy(aggregateBy []string, options *AllocationAg
...
 
-		skip := false
-
 		// (3) If the allocation does not match the filter, immediately skip the
 		// allocation.
-		if options.Filter != nil {
-			skip = !options.Filter.Matches(alloc)
-		}
-		if skip {
+		if !filter.Matches(alloc) {
```
nikovacevic, line 1381: That's nice. `Matches` returns `true` if `filter` is `nil`? Or we just know it won't be `nil` here?

nikovacevic: Nevermind, I see where we make sure it's not nil.

nikovacevic (approved):

pkg/kubecost/allocation.go
michaelmdresser: This was a poke I needed! Thank you. I'm wary of expanding the interface for logic that is only used by AggBy, so I moved this logic into a private helper func with unit tests.

PR #1985, opened 2023-06-14
Add Alias support to v2.1 Allocation filters

michaelmdresser:
## What does this PR change?
* Adds support for aliases in v2.1 filters via a compiler pass which transforms alias nodes into logically-equivalent non-alias nodes.

* Requires (stacked on) https://github.com/opencost/opencost/pull/1982
* Must merge with https://github.com/kubecost/kubecost-cost-model/pull/1572

* Added support for aliases (e.g. Product, Department) in v2.1 Allocation filters. Try it out with `filter=department:"foo"`.

* https://kubecost.atlassian.net/browse/CORE-278

## How was this PR tested?
* Unit tests
* WIP live tests, will be tracked in KCM PR

* Yes, TODO

### DIFF
--- a/pkg/filter21/ast/tree.go
+++ b/pkg/filter21/ast/tree.go
@@ -30,8 +30,11 @@ func (id *Identifier) Equal(ident Identifier) bool {
 
 // String returns the string representation for the Identifier
 func (id *Identifier) String() string {
-	if id.Field == nil {
+	if id == nil {
 		return "<nil>"
+	}
+	if id.Field == nil {
+		return "<nil field>"
 	}
 	s := id.Field.Name
 	if id.Key != "" {
--- a/pkg/filter21/ast/tree.go
+++ b/pkg/filter21/ast/tree.go
@@ -30,6 +30,9 @@ func (id *Identifier) Equal(ident Identifier) bool {
 
 // String returns the string representation for the Identifier
 func (id *Identifier) String() string {
+	if id.Field == nil {
+		return "<nil>"
+	}
 	s := id.Field.Name
 	if id.Key != "" {
 		s += "[" + id.Key + "]"
--- a/pkg/kubecost/allocationmatcher.go
+++ b/pkg/kubecost/allocationmatcher.go
@@ -3,9 +3,10 @@ package kubecost
 import (
 	"fmt"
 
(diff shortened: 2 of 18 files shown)


pkg/filter21/ast/tree.go
```diff
@@ -30,6 +30,9 @@ func (id *Identifier) Equal(ident Identifier) bool {
 
 // String returns the string representation for the Identifier
 func (id *Identifier) String() string {
+	if id.Field == nil {
```
nikovacevic: Worth also doing this if `id == nil`?

pkg/kubecost/allocationmatcher.go
```diff
@@ -65,11 +87,168 @@ func allocationSliceFieldMap(a *Allocation, identifier ast.Identifier) ([]string
...
+			labelOp,
+		),
+		ops.And(
+			ops.Not(ops.Contains(afilter.FieldLabel, aliasKey)),
+			ops.And(
+				ops.Contains(afilter.FieldAnnotation, aliasKey),
+				annotationOp,
+			),
+		),
+	), nil
```
nikovacevic, line 253: This is a neat translation of the old logic. (I still sort of hate the product idea haha---but very nice engineering effort here!)

nikovacevic (approved):

pkg/filter21/ast/tree.go
michaelmdresser: Go is the weirdest language, isn't it? I forgot about this behavior. https://go.dev/play/p/cMCUho0IyIl

Why, oh why, are you allowed to call methods on a typed nil?

Anyway, good callout -- I've added a guard.

pkg/kubecost/allocationmatcher.go
michaelmdresser: Thanks! Bolt and I spent a long conversation mulling this over, I'm glad you're pleased with it too.

As for the product idea, absolutely. For the sake of posterity (in case someone needs to connect the dots later), here are some links to past discussions:
- https://github.com/opencost/opencost/pull/1321#pullrequestreview-1039191007
- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1644
- https://github.com/opencost/opencost/pull/1457

PR #1991, opened 2023-06-20
Add custom S3 CSV billing export endpoint

klaasjand:
## What does this PR change?
* This PR adds support for custom S3 endpoints such as Minio for exporting CSV billing data.

* No

* Adds ability to use custom S3 endpoints for exporting billing data.

* No

## How was this PR tested?
* Manually via k8s deployment (via Helm chart).
* Tested with empty bucket, export file create.
* Tested with non-empty bucket, export file update.
* Tested with non-empty bucket, export file no updates needed.
* Tested with invalid S3 path.
* Tested with invalid secret key.
* Ran duration test.

* Yes, export path: see example in code, re-use AWS credential variables.

*


jwklijnsma (approved):

pkg/filemanager/filemanager.go
```diff
@@ -44,13 +46,91 @@ func NewFileManager(path string) (FileManager, error) {
 		return NewGCSStorageFile(path)
 	case strings.Contains(path, "blob.core.windows.net"):
 		return NewAzureBlobFile(path)
+	case strings.HasPrefix(path, "http://"), strings.HasPrefix(path, "https://"):
```
r2k1: With this change `s3` protocol will be assumed for any `http(s)` link.
It will require a breaking change for a potential extension in the future like adding a new destination type.

I would suggest instead using one of the 3 next approaches:
1) Introduce new protocol prefix string, like `customs3://endpoint.com/my-file.csv` or `minio://endpoint.com/my-file.csv`
2) Add domain filter like, minio.com (probably the least flexible option)
3) Extend existing `S3File` struct to support new destination. (I'm not familiar with minio, but it looks like you're using existing aws library and the same method calls.). Despite this being my preferred option, there might be potential challenges that I may not fully comprehend.

pkg/filemanager/filemanager.go
```diff
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/url"
 	"os"
+	gp "path"
```
r2k1: minor: I would rather change a variable name instead of creating an alias for a standard package. 

Makes code harder to understand.

pkg/filemanager/filemanager.go
```diff
@@ -44,13 +46,91 @@ func NewFileManager(path string) (FileManager, error) {
...
+	}
+
+	// Extract bucket and path from url
+	bucket, key, _ := strings.Cut(strings.TrimLeft(clPath, "/"), "/")
+
+	if bucket == "" || key == "" {
+		return nil, fmt.Errorf("invalid s3 path: %s", path)
+	}
+
+	const defaultRegion = "us-east-1"
```
r2k1: Do you need to hardcode the region?
I suspect this won't be correct for the majority of the people. If it's a required parameter I would rather fail with an error early and let the user fix it.

r2k1 (commented):
Overall I think it's a nice addition. I'd like just to see a non-conflicting way to configure it.

pkg/filemanager/filemanager.go
```diff
@@ -44,13 +46,91 @@ func NewFileManager(path string) (FileManager, error) {
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &CustomS3File{
+		s3Client: s3.NewFromConfig(cfg),
+		bucket:   bucket, // bucket
+		key:      key,    // path/to/file.csv
+	}, nil
```
r2k1: Can you just return `S3File{}` here? I don't see the difference.

pkg/filemanager/filemanager.go
klaasjand: I'm leaning towards option 1 because I'd rather don't want to mess up the current AWS implementation too much or add extra complexity to it. I would like to hear your thoughts on the following. Introduce a new protocol prefix string sort of like you mentioned, e.g. `customs3://fqdn:port/bucket-name/path/to/file.csv` and always enforce secure https so this translates to: `https://fqdn:port/bucket-name/path/to/file.csv`.

pkg/filemanager/filemanager.go
klaasjand: > Do you need to hardcode the region? I suspect this won't be correct for the majority of the people. If it's a required parameter I would rather fail with an error early and let the user fix it.

For the region I used this: https://github.com/minio/minio/discussions/15063 which is also described in the Minio documentation but both are based on using the AWS cli and not the SDK. I realize that's quite a bold assumption. After reading this: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/ it's probably better to make it customizable as there is already an environment variable present for that `AWS_REGION`.

pkg/filemanager/filemanager.go
klaasjand: > Can you just return `S3File{}` here? I don't see the difference.

You're right, my initial thought was to keep the original AWS S3 and the custom S3 code completely separated and have each their own Upload and Download implementation. But in retrospect it's probably better to reuse the code.

pkg/filemanager/filemanager.go
r2k1: Sounds good to me.

pkg/filemanager/filemanager.go
r2k1: I can see arguments for having it separate as well. 
I don't have a strong opinion on this. I think either way is fine.

pkg/filemanager/filemanager.go
klaasjand: After reviewing the AWS SDK docs for endpoint configuration I changed the implementation which made it a little bit more straightforward. See also: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/ 

We are using the v1 endpoint because OpenCost currently uses `github.com/aws/aws-sdk-go-v2/service/s3 v1.31.0`, from `v1.38.0` and above we have to switch to the v2 endpoint, see: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/#migration also added a comment about it in the code.

pkg/filemanager/filemanager.go
klaasjand: I removed the alias and renamed variables/arguments.

pkg/filemanager/filemanager.go
klaasjand: I changed the protocol prefix string/scheme to `alts3` (alternative S3) so `alts3://fqdn:port/bucket-name/path/to/file.csv` translates to `https://fqdn:port/bucket-name/path/to/file.csv`.

r2k1 (approved):
LGTM

r2k1:
@mattray Can you merge it?

mattray (approved):

PR #1992, opened 2023-06-20
Fix bug in Allocation metrics

### LINKED ISSUE #2184: "container_cpu_allocation" metrics drops below the amount of requested CPUs
**Describe the bug**  
"container_cpu_allocation" metrics drops below the amount of requested CPUs

**To Reproduce**  
1. `test-1` cluster, "thomasn-demo" namespace
2. Execute "kube_pod_container_resource_requests{resource="cpu", job="kubecost", namespace="thomasn-demo", container="cost-model"}" to see requests
3. Execute "container_cpu_allocation{namespace="thomasn-demo", container="cost-model"}` query for 6h
4. See the drops

**Expected behavior**  
container_cpu_allocation stays at least at the level of the request

**Screenshots**  
[container_cpu_allocation.pdf](https://github.com/kubecost/cost-analyzer-helm-chart/files/11335965/container_cpu_allocation.pdf)

thomasvn:
## What does this PR change?
* Fixes bug in the `container_cpu_allocation` and `container_memory_allocation_bytes` metrics. These metrics would occasionally only account for usage, and not requests. (screenshot below)

Screenshot 2023-06-16 at 9 41 18 PM

**Details:**

1. This bug occurred because the timestamps within the `req` and `used` vectorslices occasionally drifted too far apart.
2. Then `ApplyVectorOp()` would not synchronize the timestamps between these two vectors. 
3. This would create two elements within the resulting vectorslice even though CostModelEmitter only expects one ([code ref](https://github.com/opencost/opencost/blob/3f62fa0593692c49838f4721d131ffa36eddc91f/pkg/costmodel/metrics.go#L609-L614)).

Here are printed logs of `req`, `used`, and `allocation` from when a bad `container_cpu_allocation` metric was emitted.

```
REQ: Timestamp: 1686929350.000000, Value: 0.010000
USED: Timestamp: 1686929340.000000, Value: 0.005890
ALLOCATION: Timestamp: 1686929340.000000, Value: 0.005890
```

**To fix:**

Remove dependency on `ApplyVectorOp()`. This is possible because I've confirmed that the `req` and `used` slices will always only have one element within them.

* No

* Higher fidelity container_cpu_allocation metric.

* Fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/2184

## How was this PR tested?
* Before, my environment would see about one incorrect `container_cpu_allocation` metric each day.
* After running this change for 3 days, I am no longer seeing any incorrect `container_cpu_allocation` metrics.

* No

*

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -851,6 +851,11 @@ func findDeletedNodeInfo(cli prometheusClient.Client, missingNodes map[string]*c
 // getContainerAllocation takes the max between request and usage. This function
 // returns a slice containing a single element describing the container's
 // allocation.
+//
+// Additionally, the timestamp of the allocation will be the highest value
+// timestamp between the two vectors. This mitigates situations where
+// Timestamp=0. This should have no effect on the metrics emitted by the
+// CostModelMetricsEmitter
 func getContainerAllocation(req *util.Vector, used *util.Vector, allocationType string) []*util.Vector {
 	var result []*util.Vector
 
@@ -868,8 +873,11 @@ func getContainerAllocation(req *util.Vector, used *util.Vector, allocationType 
 		result = []*util.Vector{
 			{
 				Value:     math.Max(x1, y1),
-				Timestamp: req.Timestamp,
+				Timestamp: math.Max(req.Timestamp, used.Timestamp),
 			},
+		}
+		if result[0].Value == 0 && result[0].Timestamp == 0 {
+			log.Warnf("No request or usage data found during %s allocation calculation. Setting allocation to 0.", allocationType)
 		}
 	} else if req != nil {
 		result = []*util.Vector{
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -584,7 +584,6 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 				}
 				costs.CPUAllocation = getContainerAllocation(costs.CPUReq, costs.CPUUsed, "CPU")
 				costs.RAMAllocation = getContainerAllocation(costs.RAMReq, costs.RAMUsed, "RAM")
-
 				if filterNamespace == "" {
 					containerNameCost[newKey] = costs
 				} else if costs.Namespace == filterNamespace {
@@ -653,7 +652,6 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 			}
 			costs.CPUAllocation = getContainerAllocation(costs.CPUReq, costs.CPUUsed, "CPU")
 			costs.RAMAllocation = getContainerAllocation(costs.RAMReq, costs.RAMUsed, "RAM")
-
 			if filterNamespace == "" {
 				containerNameCost[key] = costs
 				missingContainers[key] = costs
@@ -830,9 +828,10 @@ func findDeletedNodeInfo(cli prometheusClient.Client, missingNodes map[string]*c
 	return nil
 }
 
-// getContainerAllocation returns a vectorslice of the max between `req` and
-// `used`. The function assumes the `req` and `used` vectorslices contain a
-// single element.
+// getContainerAllocation takes the max between request and usage. This function
+// returns a slice containing a single element describing the container's
+// allocation. This function assumes that the `req` and `used` slices both
+// contain a single element.
 func getContainerAllocation(req []*util.Vector, used []*util.Vector, allocationType string) []*util.Vector {
 	var result []*util.Vector
 	reqIsEmpty := len(req) == 0
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -500,11 +500,9 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 				// for the units of memory and CPU.
 				ramRequestBytes := container.Resources.Requests.Memory().Value()
 
-				// Because RAM (and CPU) information isn't coming from Prometheus, it won't
-				// have a timestamp associated with it. We need to provide a timestamp,
-				// otherwise the vector op that gets applied to take the max of usage
-				// and request won't work properly and will only take into account
-				// usage.
+				// Because information on container RAM & CPU requests isn't
+				// coming from Prometheus, it won't have a timestamp associated
+				// with it. We need to provide a timestamp.
 				RAMReqV := []*util.Vector{
 					{
 						Value:     float64(ramRequestBytes),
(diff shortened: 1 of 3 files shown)


AjayTripathy:
Wow, great job digging through this. I would love a simple unit test here.
@nikovacevic  can you look at this one very closely?

thomasvn:
@AjayTripathy Will do!

nikovacevic:
Gladly. Thanks for the contribution @thomasvn

pkg/costmodel/costmodel.go
```diff
@@ -828,32 +828,56 @@ func findDeletedNodeInfo(cli prometheusClient.Client, missingNodes map[string]*c
 	return nil
 }
 
+// getContainerAllocation takes the max between request and usage. This function
+// returns a slice containing a single element describing the container's
+// allocation. This function assumes that the `req` and `used` slices both
+// contain a single element.
 func getContainerAllocation(req []*util.Vector, used []*util.Vector, allocationType string) []*util.Vector {
```
nikovacevic: @thomasvn the fix makes sense, and the implementation makes sense — but I'm afraid that keeping this function signature leaves us open to possible misuse in the future by unsuspecting engineers (however unlikely). That is, if we suggest that the function requires a slice of Vectors, but internally we assume that it will have length == 1, and we don't return errors if that condition is not met, then we could see people use this function with a reasonable expectation of the old behavior, and have the function ignore all but the first number. (Also, I want to acknowledge that you've left a helpful comment, but people sometimes don't read comments.)

So, I'd propose we take this a step further in one of two ways:
1. Change the signature here to `getContainerAllocation(req *util.Vector, used *util.Vector, allocationType string) []*util.Vector` so that there can be no confusion. That will require changing the four places this is called, so that we only pass in that first datum, which will itself perhaps require some simple length checking before calling.
2. Change the signature here to `getContainerAllocation(req []*util.Vector, used []*util.Vector, allocationType string) ([]*util.Vector, error)` and return an error if a slice of length > 1 is provided. That will also require changing the four places this is called, to handle the error.

I think that both of these option will help us feel confident in the correctness of this solution, too. (That is, in addition to Ajay's requested unit test.) If we don't see any errors or issues with the new signature, then we can be confident that we won't get bitten by any use cases of which we were not aware.

nikovacevic:
P.S. Let me also echo Ajay here in saying that this is a very impressive catch!

pkg/costmodel/costmodel.go
thomasvn: @nikovacevic I agree with changing the function's arguments & signature! Way better than commenting assumptions. Thoughts on the following signature?

```
getContainerAllocation(req *util.Vector, used *util.Vector, allocationType string) *util.Vector {}
```

thomasvn: Nevermind the above suggestion I made. `CPUReq`, `CPUUsed`, and `CPUAllocation` are of type `[]*util.Vector`

https://github.com/opencost/opencost/blob/3f62fa0593692c49838f4721d131ffa36eddc91f/pkg/costmodel/costmodel.go#L93-L95

nikovacevic: Yeah, it's kind of a finicky decision, given the code context and history... I agree that returning `*util.Vector` would be better, but it might be tricky. As long as the function params are singular, I'd be happy! (Or, again, an `error`. Either one.)

thomasvn: @nikovacevic Made some changes, when you get a moment could you review again?

nikovacevic: Will do! Thank you.

thomasvn:
@AjayTripathy @nikovacevic Added in some unit tests and am open to feedback for improving those!

pkg/costmodel/costmodel.go
```diff
@@ -828,32 +848,62 @@ func findDeletedNodeInfo(cli prometheusClient.Client, missingNodes map[string]*c
...
+		}
+		y1 := used.Value
+		if math.IsNaN(y1) {
+			log.Warnf("NaN value found during %s allocation calculation for used.", allocationType)
+			y1 = 0.0
+		}
+		result = []*util.Vector{
+			{
+				Value:     math.Max(x1, y1),
+				Timestamp: math.Max(req.Timestamp, used.Timestamp),
```
thomasvn, line 888: @nikovacevic Curious to hear what you think about doing `math.Max(req.Timestamp, used.Timestamp)`. I wanted to remove the assumption that the caller would set `req.Timestamp = float64(time.Now().UTC().Unix())`.

pkg/costmodel/costmodel.go
```diff
@@ -582,8 +580,19 @@ func (cm *CostModel) ComputeCostData(cli prometheusClient.Client, cp costAnalyze
 					ClusterID:       clusterID,
 					ClusterName:     cm.ClusterMap.NameFor(clusterID),
 				}
-				costs.CPUAllocation = getContainerAllocation(costs.CPUReq, costs.CPUUsed, "CPU")
-				costs.RAMAllocation = getContainerAllocation(costs.RAMReq, costs.RAMUsed, "RAM")
+
+				if len(costs.CPUReq) > 0 && len(costs.CPUUsed) > 0 {
+					costs.CPUAllocation = getContainerAllocation(costs.CPUReq[0], costs.CPUUsed[0], "CPU")
+				} else {
+					log.Warnf("CPU Requests or Usage data missing for %s", containerName)
```
nikovacevic: I would set this to `Debugf` probably, unless we have a strong feeling about this log level.

nikovacevic: Also, I'm not quite sure that this is the right logic. Let's think through it... we probably want to `getContainerAllocation` if either one of those two is present, right? We might want to even keep the old way, whereby we _always_ call it.

So I'm wondering if we want something more like:
```go
var cpuReq, cpuUse *util.Vector
if len(costs.CPUReq) > 0 {
    cpuReq = costs.CPUReq[0]
}
if len(costs.CPUUsed) > 0 {
    cpuUse = costs.CPUUsed[0]
}
costs.CPUAllocation(cpuReq, cpuUse, "CPU")
```
... and likewise for RAM. That maintains the old patter, whereby `getContainerAllocation` is always called, no matter what. And it lets `getContainerAllocation` handle `nil` cases, which it does quite aptly.

What do you think of that suggestion @thomasvn ?

pkg/costmodel/costmodel.go
```diff
@@ -828,32 +848,62 @@ func findDeletedNodeInfo(cli prometheusClient.Client, missingNodes map[string]*c
...
-			r.Value = *y
+// getContainerAllocation takes the max between request and usage. This function
+// returns a slice containing a single element describing the container's
+// allocation.
+//
+// Additionally, the timestamp of the allocation will be the highest value
+// timestamp between the two vectors. This mitigates situations where
+// Timestamp=0. This should have no effect on the metrics emitted by the
+// CostModelMetricsEmitter
+func getContainerAllocation(req *util.Vector, used *util.Vector, allocationType string) []*util.Vector {
```
nikovacevic, line 871: This function looks good to me.

pkg/costmodel/costmodel.go
nikovacevic: That seems fine to me. The two places this data goes are:
1. Metric emission, in which timestamps are ignored; and
2. `GET /model/costDataModel`, which is practically unused.

I still think it's worth making a sensible decision here, but I don't think that there are any constraints that should make us prefer a different value.

The only small suggestion I might make is that the default value in the final `else` could similarly be zero, to indicate that there was no data at all. But again, I'm not worried about that case at all, really.

nikovacevic (commented):
Let me know what you think about that one big suggestion, especially if you have concerns. Once we get past that, I think we're good to merge.

pkg/costmodel/costmodel.go
thomasvn: > The only small suggestion I might make is that the default value in the final else could similarly be zero, to indicate that there was no data at all. But again, I'm not worried about that case at all, really.

Assuming there was full time series data in `[]*util.Vector`, it may be good to set the current timestamp? That way there would be indication at what point allocation could not be calculated due to missing requests and usage data?

```
Value:     0,
Timestamp: float64(time.Now().UTC().Unix()),
```

I'm good either way! As you mentioned this use case is no longer likely to happen.

pkg/costmodel/costmodel.go
thomasvn: @nikovacevic I'm in agreement with both suggestions, thanks! I'll make the changes now.

thomasvn: Just tested and pushed the changes!

nikovacevic (approved):
Alright I think this is good to go. Thanks @thomasvn !

PR #1995, opened 2023-06-21
Switch Assets to v2.1 filters

michaelmdresser:
## What does this PR change?
* Switches internal representation of Asset filters from `[]AssetMatchFunc` to `filter21.Filter`
  * Removes the `AssetMatchFunc` type
  * Adds a v1 asset filter -> `filter21.Filter` translation layer based on https://github.com/kubecost/kubecost-cloud/blob/52a34eb7381e50347dd01dd5270a7657f7761df4/internal/util/filterutil/filterutil.go#L168, which makes KCM's `pkg/asset/filter.go` obsolete
  * Adds necessary field support to the v2.1 filter parser to support Asset filters
  * Adds a `AssetMatchCompiler` to turn parsed Asset filters into an Asset Matcher
* Ultimately, this means that Asset-based APIs will be able to use v2.1 filter syntax for more complex queries https://github.com/kubecost/docs/blob/main/filters-api.md

* Must merge with https://github.com/kubecost/kubecost-cost-model/pull/1581
* Parallels https://github.com/opencost/opencost/pull/1982

* N/A (see KCM release note)

* https://kubecost.atlassian.net/browse/CORE-281

## How was this PR tested?
* Unit tests
* WIP live testing, see KCM PR for notes.

* Yes, will handle separately.

### DIFF
--- a/pkg/util/filterutil/filterutil.go
+++ b/pkg/util/filterutil/filterutil.go
@@ -457,6 +457,11 @@ func AssetFilterFromParamsV1(
 	}
 
 	if raw := qp.GetList(ParamFilterTypes, ","); len(raw) > 0 {
+		// Types have a special situation where we allow users to enter them
+		// capitalized or uncapitalized
+		for i := range raw {
+			raw[i] = strings.ToLower(raw[i])
+		}
 		filterOps = push(filterOps, filterV1SingleValueFromList(raw, assetfilter.FieldType))
 	}
 
--- a/pkg/util/filterutil/filterutil.go
+++ b/pkg/util/filterutil/filterutil.go
@@ -80,6 +80,24 @@ const (
 	ParamFilterTypes         = "filterTypes"
 )
 
+// ValidAssetFilterParams returns a list of all possible filter parameters
+func ValidAssetFilterParams() []string {
+	return []string{
+		ParamFilterAccounts,
+		ParamFilterCategories,
+		ParamFilterClusters,
+		ParamFilterLabels,
+		ParamFilterNames,
+		ParamFilterProjects,
+		ParamFilterProviders,
+		ParamFilterProviderIDs,
+		ParamFilterProviderIDsV2,
+		ParamFilterRegions,
+		ParamFilterServices,
+		ParamFilterTypes,
+	}
+}
+
 // AllocationPropToV1FilterParamKey maps allocation string property
 // representations to v1 filter param keys for legacy filter config support
 // (e.g. reports). Example mapping: "cluster" -> "filterClusters"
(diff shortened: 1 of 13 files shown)


pkg/util/filterutil/filterutil.go
```diff
@@ -65,8 +68,36 @@ const (
...
 	ParamFilterLabels      = "filterLabels"
 	ParamFilterServices    = "filterServices"
+
+	ParamFilterAccounts      = "filterAccounts"
+	ParamFilterCategories    = "filterCategories"
+	ParamFilterNames         = "filterNames"
+	ParamFilterProjects      = "filterProjects"
+	ParamFilterProviders     = "filterProviders"
+	ParamFilterProviderIDs   = "filterProviderIDs"
+	ParamFilterProviderIDsV2 = "filterProviderIds"
```
avrodrigues5, line 78: Can we not have a common filterProviderIDs for both v1 and v2? Is there a reason why this is different?

pkg/util/filterutil/filterutil.go
```diff
@@ -542,3 +731,13 @@ func toAllocationAliasOp(labelName string, filterValue string, wildcard bool) *a
 		},
 	}
 }
+
+// GetList provides a list of values from the first key if they exist, otherwise, it returns
+// the values from the second key.
+func GetList(primaryKey, secondaryKey string, qp mapper.PrimitiveMapReader) []string {
```
avrodrigues5, line 737: Where is this function called? I dont see it used anywhere Maybe im blind but thought of asking!

avrodrigues5 (approved):
More Like questions. Approved although i'm not maintainer.

pkg/util/filterutil/filterutil.go
michaelmdresser: It's used in `AssetFilterFromParamsV1`.

pkg/util/filterutil/filterutil.go
michaelmdresser: This is a good question, thank you. I pulled this code from the parallel KC cloud implementation (https://github.com/kubecost/kubecost-cloud/blob/194b58017aa632cfeb0ee6bd61804ad7a17d5349/internal/util/filterutil/filterutil.go#L61).

KCM doesn't appear to have had this "v2" param (https://github.com/kubecost/kubecost-cost-model/pull/1581/files#diff-a6a736ce66bb7870d4233a16dbdecdd398a25042f790a5663af4d67b6f5e4d41L46-L61).

@mbolt35 what's up with the v2 provider ID filter? should it be included in Opencost?

michaelmdresser: Talked with Bolt. `filterProviderIds` is only used in cloud right now, but I think we may as well add it here, it makes an eventual unification of filtering implementations between KC Cloud and Onprem easier. Pure "upgrade" in functionality with no breaking change as far as onprem is concerned.

Sean-Holcomb (approved):

PR #2001, opened 2023-06-26
fix asset query duration

ameijer:
## What does this PR change?
* fixes a bug I introduced where the duration of the query is set to the resolution

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


avrodrigues5 (approved):
LGTM. I see the old PR regression created and solution here!

teevans (approved):

PR #2011, opened 2023-06-29
Sean/lb private bingen

Sean-Holcomb:
## What does this PR change?
This PR updates bingen version tags for two fields.
Asset LoadBalancer field `Private` was added but did not get a bingen tag and bingen was not run. This would cause an issue the next time bingen was run for assets if the version tag was not added to `Private` then it would cause build failures on assets. This issue was caught at a point where there should be no repercussions 

Allocation new LoadBalancer field in 1.105 was given a tag for an allocation bingen version that went out in 1.104. Users that have that 1.104 bingen codec will have their Allocation data deleted by any future codec if this is not addressed. Users who have downloaded 1.105 RCs will have already lost their allocation data, and will lose it again with their next update, users that are still on the 1.104 bingen codec will not be impacted 

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/kubecost/kubecost_codecs.go
+++ b/pkg/kubecost/kubecost_codecs.go
@@ -13,12 +13,11 @@ package kubecost
 
 import (
 	"fmt"
+	util "github.com/opencost/opencost/pkg/util"
 	"reflect"
 	"strings"
 	"sync"
 	"time"
-
-	util "github.com/opencost/opencost/pkg/util"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/kubecost/kubecost_codecs.go
```diff
@@ -13,12 +13,11 @@ package kubecost
 
 import (
 	"fmt"
+	util "github.com/opencost/opencost/pkg/util"
```
mbolt35, line 16: Goland strikes again

ameijer (approved):

mbolt35 (approved):
LGTM

michaelmdresser (approved):

pkg/kubecost/kubecost_codecs.go
ameijer: @mbolt35 , that other PR I have for bingen fixes this and uses the gofmt standard (group by first/second/third party imports)

mbolt35: @ameijer I actually run go format against the results of the generated output, so it shouldn't ever be different:

https://github.com/kubecost/bingen/blob/09cd797f542c23a3c6ce1b1940f171dbad25dd16/pkg/generator/bingen.go#L66-L70

That is, unless I am mistaken about what `go/format` Format is actually doing :)

PR #2025, opened 2023-07-06
fix edge cases with AWS CN pricing file inconsistencies

saweber:
## What does this PR change?
* Fix an issue with AWS CN Pricing

* no

* keep KCM from crashing for AWS CN users

* Closes https://kubecost.atlassian.net/browse/BURNDOWN-153

## How was this PR tested?
* manually

* no

We haven't seen this issue for AWS outside of CN, but putting the fix in for non-CN pricing just in case.

The JSON object keys are different than what our code expects; some offer terms contain an hourly rate code that is unexpected. To handle this case, the logic is:

1) use the term that we expect
2) if that term is missing, and there is only one key in the underlying JSON object, use it
3) otherwise log a warning and use a default cost

JSON snippet:

The price dimension key name is ending with `ZQUBENBFS9`, and not the expected CN rate code `Q7UJUT2CE6`. I don't have an explanation why this is the case, and we don't have an AWS CN account to test with, from what I can tell.

```
"9FK48YK3M3ENNJRP" : {
        "9FK48YK3M3ENNJRP.KW44MY7SZN" : {
          "offerTermCode" : "KW44MY7SZN",
          "sku" : "9FK48YK3M3ENNJRP",
          "effectiveDate" : "2023-05-01T00:00:00Z",
          "priceDimensions" : {
            "9FK48YK3M3ENNJRP.KW44MY7SZN.ZQUBENBFS9" : {
              "rateCode" : "9FK48YK3M3ENNJRP.KW44MY7SZN.ZQUBENBFS9",
              "description" : "Global",
              "beginRange" : "0",
              "endRange" : "30",
              "unit" : "GB-Mo",
              "pricePerUnit" : {
                "CNY" : "0.0000000000"
              },
              "appliesTo" : [ "F6GP6WRNDGM8BWKY", "R83VXG9NAPDASEGN", "FJKE8U2RFEXU5GEH" ]
            }
          },
          "termAttributes" : {
            "Restriction" : "Limited SKU Usage"
          }
        }
      },```

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -1069,9 +1069,45 @@ func (aws *AWS) populatePricing(resp *http.Response, inputkeys map[string]bool) 
 						aws.Pricing[spotKey].OnDemand = offerTerm
 						var cost string
 						if _, isMatch := OnDemandRateCodes[offerTerm.OfferTermCode]; isMatch {
-							cost = offerTerm.PriceDimensions[strings.Join([]string{sku.(string), offerTerm.OfferTermCode, HourlyRateCode}, ".")].PricePerUnit.USD
+							priceDimensionKey := strings.Join([]string{sku.(string), offerTerm.OfferTermCode, HourlyRateCode}, ".")
+							dimension, ok := offerTerm.PriceDimensions[priceDimensionKey]
+							if ok {
+								cost = dimension.PricePerUnit.USD
+							} else {
+								// this is an edge case seen in AWS CN pricing files, including here just in case
+								// if there is only one dimension, use it, even if the key is incorrect, otherwise assume defaults
+								if len(offerTerm.PriceDimensions) == 1 {
+									for key, backupDimension := range offerTerm.PriceDimensions {
+										cost = backupDimension.PricePerUnit.USD
+										log.DedupedWarningf(5, "using:%s for a price dimension instead of missing dimension: %s", offerTerm.PriceDimensions[key], priceDimensionKey)
+										break
+									}
+								} else if len(offerTerm.PriceDimensions) == 0 {
+									log.DedupedWarningf(5, "populatePricing: no pricing dimension available for: %s.", priceDimensionKey)
+								} else {
+									log.DedupedWarningf(5, "populatePricing: no assumable pricing dimension available for: %s.", priceDimensionKey)
+								}
+							}
 						} else if _, isMatch := OnDemandRateCodesCn[offerTerm.OfferTermCode]; isMatch {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/cloud/aws/provider.go [hunk 1]
```diff
@@ -1069,9 +1069,45 @@ func (aws *AWS) populatePricing(resp *http.Response, inputkeys map[string]bool)
 						aws.Pricing[spotKey].OnDemand = offerTerm
 						var cost string
 						if _, isMatch := OnDemandRateCodes[offerTerm.OfferTermCode]; isMatch {
-							cost = offerTerm.PriceDimensions[strings.Join([]string{sku.(string), offerTerm.OfferTermCode, HourlyRateCode}, ".")].PricePerUnit.USD
+							priceDimensionKey := strings.Join([]string{sku.(string), offerTerm.OfferTermCode, HourlyRateCode}, ".")
```
saweber, line 1072: This could use a future refactor, but I was aiming to keep the current code structure to minimize testing and impact.

pkg/cloud/aws/provider.go
```diff
@@ -1069,9 +1069,45 @@ func (aws *AWS) populatePricing(resp *http.Response, inputkeys map[string]bool)
...
+									}
+								} else if len(offerTerm.PriceDimensions) == 0 {
+									log.DedupedWarningf(5, "populatePricing: no pricing dimension available for: %s.", priceDimensionKey)
+								} else {
+									log.DedupedWarningf(5, "populatePricing: no assumable pricing dimension available for: %s.", priceDimensionKey)
+								}
+							}
 						} else if _, isMatch := OnDemandRateCodesCn[offerTerm.OfferTermCode]; isMatch {
-							cost = offerTerm.PriceDimensions[strings.Join([]string{sku.(string), offerTerm.OfferTermCode, HourlyRateCodeCn}, ".")].PricePerUnit.CNY
+							priceDimensionKey := strings.Join([]string{sku.(string), offerTerm.OfferTermCode, HourlyRateCodeCn}, ".")
```
avrodrigues5, line 1092: Do you think this can be a private function instead of almost identical lines of code except the currency? Its not a blocker thou.

avrodrigues5 (approved):

pkg/cloud/aws/provider.go
saweber: It could be - to do the job well, it is going to take unwinding a lot of if CN / if not CN code in this file, which I think is best left for another day.

Furthermore - I don't want to these implementations together, when if anything, they are diverging.

nikovacevic (approved):

PR #2034, opened 2023-07-11
Fix the mismatch in sharing namespaces due to unmounted pv cost in each allocation

avrodrigues5:
## What does this PR change?
* Fixes the issue seen in https://github.com/kubecost/cost-analyzer-helm-chart/issues/2359, this was due to unmounted pv cost in each allocation that while computation is ignored both in totals and in summary allocation. As a result the sharing isnt proportional. These definitely fixes the problem but there's a pending question as to if the unmounted pv cost should even be a part of each allocation?

* None

* Get accurate result when custom shared resources and sharing idle by cluster or node 

* Closes ... https://kubecost.atlassian.net/browse/BURNDOWN-152 and subsequently https://github.com/kubecost/cost-analyzer-helm-chart/issues/2359

## How was this PR tested?
* Port-forwarding Niko's dev cluster as instructed and hitting end point 
1. localhost:9003/allocation/view?window=2023-07-07T00:00:00Z,2023-07-08T00:00:00Z&aggregate=namespace
Screen Shot 2023-07-11 at 9 36 35 AM

2. localhost:9003/allocation/view?window=2023-07-07T00:00:00Z,2023-07-08T00:00:00Z&aggregate=namespace&shareNamespaces=kube-system

Screen Shot 2023-07-11 at 9 37 23 AM

Also tried shareIdle by node via REST call
localhost:9003/allocation/view?window=2023-07-07T00:00:00Z,2023-07-08T00:00:00Z&aggregate=namespace&shareNamespaces=kube-system&shareIdle=true&idleByNode=true&cache=false

Screen Shot 2023-07-11 at 9 37 15 AM

The logs yield a sharing coefficient sum of 1.00XXX as expected 
Screen Shot 2023-07-11 at 9 21 44 AM

* No

* v1.105

### DIFF
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -858,12 +858,21 @@ func (a *Allocation) IsUnallocated() bool {
 }
 
 // IsUnmounted is true if the given Allocation represents unmounted volume costs.
+// Note: Due to change in https://github.com/opencost/opencost/pull/1477 made to include Unmounted
+// PVC cost inside namespace we need to check unmounted suffix across all the three major properties
+// to actually classify it as unmounted.
 func (a *Allocation) IsUnmounted() bool {
 	if a == nil {
 		return false
 	}
 
-	return strings.Contains(a.Name, UnmountedSuffix)
+	props := a.Properties
+	if props != nil {
+		if props.Container == UnmountedSuffix && props.Namespace == UnmountedSuffix && props.Pod == UnmountedSuffix {
+			return true
+		}
+	}
+	return false
 }
 
 // Minutes returns the number of minutes the Allocation represents, as defined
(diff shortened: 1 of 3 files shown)


AjayTripathy:
Incredible work! Any way we can leave a testing artifact here, unit test or otherwise?

avrodrigues5:
Hello @AjayTripathy , The numbers are mismatched because of isUnmounted ignoring the summary allocation resulted from [PR](https://github.com/opencost/opencost/pull/1477).  I am changing the definition of isunmounted itself and wouldnt need any unit test per se. 

I have raised a Jira ticket to investigate if the feature is actually working in [CORE-370](https://kubecost.atlassian.net/browse/CORE-370), To see if the feature is actually working as intended. Please see the explanation in the ticket. 

Irrespective i think this fix is needed if the actually case `where a PVC in a namespace is dynamically provisioning a PV, but the PVC is not bound to a pod.` is encountered. We included that cost under namespace would result in the problem stated by the user, and it should be included in the sharing coefficient calculation.

LMK if you have any opinions Ajay. Appreciate your feedback

avrodrigues5:
@nikovacevic I made the changes discussed. I verified the same end points and also match allocation end point and its values with summary when sharing resource.

nikovacevic (approved):
Thanks, @avrodrigues5 -- I think this looks good, based on conversations with you the last two days, and looking at the data.

PR #2047, opened 2023-07-18
fix the unmounted allocation caused by the offset duration leading to an unmounted PVC coefficient entry

avrodrigues5:
## What does this PR change?
* While building [pvcMap](https://github.com/opencost/opencost/blob/develop/pkg/costmodel/allocation_helpers.go#L1846) we offset PVC start time by resolution duration  . However this offset is not taken into consideration  while distribution pvc cost across pods. Leading to the duration entry causing an unmounted allocation.

* None

* no unmounted entry in each namespace.
* Have accurate PV cost for a pod 

* Closes ... [CORE-370](https://kubecost.atlassian.net/browse/CORE-370)

## How was this PR tested?
* Port-forwarding a NS in AWS dev-1 

Consider Niko's PVs 

Screen Shot 2023-07-18 at 12 23 26 PM

There should never be an unmount pvc because both pvs are claimed in namespace Niko. But before this fix

when following end point was run http://localhost:9003/allocation?window=2023-07-15T00:00:00Z,2023-07-16T00:00:00Z&filterContainers=__unmounted__ 

it has an unmounted entry for NIko

Screen Shot 2023-07-18 at 12 26 02 PM

After debugging the problem, it lied in PV shared coefficient function that caused an unmounted entry for Niko's namespace due to PVCMap being offset by 5minutes(offset duration used in KCM) and as a result leading to this unmounted allocation 

Screen Shot 2023-07-17 at 4 06 17 PM (1)

After the fix same end point http://localhost:9003/allocation?window=2023-07-15T00:00:00Z,2023-07-16T00:00:00Z&filterContainers=__unmounted__

Dint have any unmounted PVC entry for Niko namespace

* No

* v1.106

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -662,7 +662,7 @@ func (cm *CostModel) computeAllocation(start, end time.Time, resolution time.Dur
 	// split appropriately among each pod's container allocation.
 	podPVCMap := map[podKey][]*pvc{}
 	buildPodPVCMap(podPVCMap, pvMap, pvcMap, podMap, resPodPVCAllocation, podUIDKeyMap, ingestPodUID)
-	applyPVCsToPods(window, podMap, podPVCMap, pvcMap)
+	applyPVCsToPods(window, podMap, podPVCMap, pvcMap, resolution)
 
 	// Identify PVCs without pods and add pv costs to the unmounted Allocation for the pvc's cluster
 	applyUnmountedPVCs(window, podMap, pvcMap)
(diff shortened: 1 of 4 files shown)


nikovacevic (approved):
As far as I can tell, this checks out. The proof of its correctness is in the testing, which seems good too. Thank you, Alan!

nikovacevic:
Impressive find and fix.

avrodrigues5:
Thanks @nikovacevic , Please merge the PR as I'm not an OpenCost maintainer.

PR #2048, opened 2023-07-19
use the existing variable in error message

savar:
the error message was trying to use the variable/value which was not there and which was the reason for the error message in the first place

## What does this PR change?
* using the existing variable in an error message instead of the non-existent one

no

In case they had the error message before like that:
```
2023-07-18T11:47:05.702704554Z WRN unsupported cloud region ""
```
Now instead of the `""` empty string it should give a better information of the region in question.

no

## How was this PR tested?
manually running in a kubernetes cluster next to the official docker image binary.. sadly at that time neither (the original 1.105.0 version nor the self compiled one) showed the error message, so the real output is not clear,  but it looks like the variables are both strings, so it should work.

no

I will try to label it.

### DIFF
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -208,7 +208,7 @@ func getRegions(service string, subscriptionsClient subscriptions.Client, provid
 						if loc, ok := allLocations[displName]; ok {
 							supLocations[loc] = displName
 						} else {
-							log.Warnf("unsupported cloud region %q", loc)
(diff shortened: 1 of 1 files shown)


mattray (approved):

mattray:
@savar if you could amend your commit message for the DCO, we'll get this merged.

savar:
Sure.. done.. I hope the rebase didn't create an issue. If it did, please tell me, then I will force-push the old version again to not have that part inside.

PR #2049, opened 2023-07-19
ELV-145 stop querying for CPU and RAM breakdown by default

### LINKED ISSUE #1883: fmtQueryCPUModePct is costly, consider adding label matchers

nikovacevic:
## What does this PR change?
* Disables costly Prometheus queries for CPU and RAM mode breakdown. (Allows those features to be re-enabled by helm configuration.)

* Requires https://github.com/kubecost/cost-analyzer-helm-chart/pull/2448

* Less pressure on Prometheus

* Closes https://github.com/opencost/opencost/issues/1883
* Closes https://kubecost.atlassian.net/browse/ELV-145

## How was this PR tested?
* Manually, using logs and checking API responses

By default:
```
2023-07-19T20:57:12.324097149Z INF ELV-145: skipping node breakdown queries
2023-07-19T20:57:12.334288408Z INF ELV-145: skipping node breakdown queries
2023-07-19T20:57:12.462450038Z INF ELV-145: skipping node breakdown queries
```
Screenshot from 2023-07-19 14-58-24

With the new setting enabled:
```
2023-07-19T21:03:06.363100429Z INF ELV-145: running node breakdown queries
2023-07-19T21:03:06.554459558Z INF ELV-145: running node breakdown queries
2023-07-19T21:03:06.749878635Z INF ELV-145: running node breakdown queries
```
Screenshot from 2023-07-19 15-07-59

* Yes, I'll open a docs PR


pkg/costmodel/cluster.go
```diff
@@ -302,11 +308,17 @@ func ClusterDisks(client prometheus.Client, provider models.Provider, start, end
...
-				Breakdown: &ClusterCostsBreakdown{},
+				Breakdown: nil,
 				Local:     true,
 			}
+			if env.IsAssetModeBreakdownEnabled() {
+				diskMap[key].Breakdown = &ClusterCostsBreakdown{}
+			}
+		}
+
+		if diskMap[key].Breakdown != nil {
```
avrodrigues5, line 319: clarifying question: Should we check if key is found in the map here Niko?

avrodrigues5 (approved):
Although this needs an approval from OC maintainer!

pkg/costmodel/cluster.go
nikovacevic: Right above, we ensure it will exist. (Line 307, if it does not exist, then on 308 we create it.) I like your thoroughness, though, and am open to an extra safety check here.

avrodrigues5: I see in line 307... You good! thanks!

nikovacevic:
Currently, if we merge this and users upgrade, they will panic if they try to roll back to `v1.105` or earlier. This is because older versions do not have `nil` checks in the code, and this change will make those fields `nil` (and this change has the `nil` checks). I think I should probably keep the nil checks, but make the default value empty `&Breakdown{}` and not `nil`, just to be safer.

PR #2051, opened 2023-07-21
Add Sanity limits for persistent volumes

ghost:
## What does this PR change?
* It adds a sanity limit for PV byte usage, to preclude absurd inflated prices showing up. Currently, the Sanity Limit is set to 10 PetaBytes, but this can easily be changed, if appropriate. While a sanity limit is the current standard, it may be appropriate to consider a statistical outlier method, or an alternate clamping target.

* Negative

* Users should no longer encounter an error with inflated PV costs in some circumstances.

* Addresses [Zendesk 3130](https://kubecost.zendesk.com/agent/tickets/3130) and [Zendesk 3538](https://kubecost.zendesk.com/agent/tickets/3538)

## How was this PR tested?
* Locally

* Nope

* I would if it let me edit PR labels

### DIFF
--- a/pkg/costmodel/allocation_helpers.go
+++ b/pkg/costmodel/allocation_helpers.go
@@ -18,7 +18,10 @@ import (
 
 // This is a bit of a hack to work around garbage data from cadvisor
 // Ideally you cap each pod to the max CPU on its node, but that involves a bit more complexity, as it it would need to be done when allocations joins with asset data.
-const MAX_CPU_CAP = 512
+const CPU_SANITY_LIMIT = 512
+
+// Sanity Limit for PV usage, set to 10 PB, in bytes for now
+const PV_USAGE_SANITY_LIMIT_BYTES = 10737418240
 
 /* Pod Helpers */
 
@@ -231,7 +234,7 @@ func applyCPUCoresAllocated(podMap map[podKey]*pod, resCPUCoresAllocated []*prom
 			}
 
 			cpuCores := res.Values[0].Value
-			if cpuCores > MAX_CPU_CAP {
+			if cpuCores > CPU_SANITY_LIMIT {
 				log.Infof("[WARNING] Very large cpu allocation, clamping to %f", res.Values[0].Value*(thisPod.Allocations[container].Minutes()/60.0))
 				cpuCores = 0.0
 			}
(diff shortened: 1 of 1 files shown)


pkg/costmodel/allocation_helpers.go
```diff
@@ -18,7 +18,10 @@ import (
 
 // This is a bit of a hack to work around garbage data from cadvisor
 // Ideally you cap each pod to the max CPU on its node, but that involves a bit more complexity, as it it would need to be done when allocations joins with asset data.
-const MAX_CPU_CAP = 512
+const CPU_SANITY_LIMIT = 512
+
+// Sanity Limit for PV usage, set to 10 PB, in bytes for now
+const PV_USAGE_SANITY_LIMIT = 10737418240
```
nikovacevic: Very small nit -- can we add `_BYTES` to the end of the name here? Trying to get the whole team on board for adding units to names when we can, to avoid ambiguity.

pkg/costmodel/allocation_helpers.go
```diff
@@ -1818,7 +1821,13 @@ func applyPVBytes(pvMap map[pvKey]*pv, resPVBytes []*prom.QueryResult) {
...
 		}
 
-		pvMap[key].Bytes = res.Values[0].Value
+		pvBytesUsed := res.Values[0].Value
+		if pvBytesUsed >= PV_USAGE_SANITY_LIMIT {
+			pvMap[key].Bytes = pvBytesUsed
+		} else {
+			pvMap[key].Bytes = 0
+			log.Infof("[WARNING] PV usage exceeds sanity limit, clamping to zero")
+		}
```
nikovacevic: This looks backwards to me -- can we double-check? If the measured bytes exceed the limit, we want to clamp; if they do not exceed, we want to use the measured value, right?

pkg/costmodel/allocation_helpers.go
ghost: Absolutely, that's a sensible semantic change

nikovacevic (commented):
Just one thing to double-check here. (And I may be reading it backwards.)

pkg/costmodel/allocation_helpers.go
nikovacevic: Thank you!

pkg/costmodel/allocation_helpers.go
ghost: Yep, I accidentally swapped that, it's fixed now

nikovacevic (approved):

PR #2056, opened 2023-07-24
Fix PV usage sanity limit

nikovacevic:
## What does this PR change?
* Increases PV sanity limit from 10 GiB to 10 PiB

* https://github.com/opencost/opencost/pull/2051

* Fixes limiting feature

* No

## How was this PR tested?
* Manually (was seeing unexpected logs; now not)

* No

* Yes

### DIFF
--- a/pkg/costmodel/allocation_helpers.go
+++ b/pkg/costmodel/allocation_helpers.go
@@ -21,7 +21,12 @@ import (
 const CPU_SANITY_LIMIT = 512
 
 // Sanity Limit for PV usage, set to 10 PB, in bytes for now
-const PV_USAGE_SANITY_LIMIT_BYTES = 10737418240
+const KiB = 1024.0
+const MiB = 1024.0 * KiB
+const GiB = 1024.0 * MiB
+const TiB = 1024.0 * GiB
+const PiB = 1024.0 * TiB
+const PV_USAGE_SANITY_LIMIT_BYTES = 10.0 * PiB
 
 /* Pod Helpers */
 
@@ -1826,7 +1831,7 @@ func applyPVBytes(pvMap map[pvKey]*pv, resPVBytes []*prom.QueryResult) {
 			pvMap[key].Bytes = pvBytesUsed
 		} else {
 			pvMap[key].Bytes = 0
-			log.Infof("[WARNING] PV usage exceeds sanity limit, clamping to zero")
+			log.Warnf("PV usage exceeds sanity limit, clamping to zero")
 		}
 	}
 }


pkg/costmodel/allocation_helpers.go
```diff
@@ -21,7 +21,7 @@ import (
 const CPU_SANITY_LIMIT = 512
 
 // Sanity Limit for PV usage, set to 10 PB, in bytes for now
-const PV_USAGE_SANITY_LIMIT_BYTES = 10737418240
```
nikovacevic, line 24: `10737418240 = 10 * (1024 ^ 3)` which is 10 GiB

mattray (approved):

pkg/costmodel/allocation_helpers.go
nikovacevic: (And if you try to reach petabytes with an int it overflows.)

thomasvn: @nikovacevic I like the fixes for clarity here, thanks! So you're saying the following wouldn't overflow? Why not?

```
const PV_USAGE_SANITY_LIMIT_BYTES = 10.0 * PiB
```

nikovacevic: Correct. That doesn't overflow because it's a different data type, `float64`. Max `float64` is `1.7E+308` haha. Max `int` on the other hand is `2147483647`, which is only ~2GiB. (In fact... it's _precisely_ 1 less than 2GiB, which is `2147483648`! Why could that be? This is your math homework :wink: )

thomasvn: @nikovacevic Ah ok! `float64` type inferred at compile time because such a large number?

> Max int on the other hand is 2147483647, which is only ~2GiB. (In fact... it's precisely 1 less than 2GiB, which is 2147483648! Why could that be? This is your math homework 😉 )

Hmm ok ok need to dig deep here. My guess would be `int` is 32 bits? 1 bit reserved to indicate sign (+ or -). Max value of 31 bits is (2^31)-1 = 2147483647? (2^x)-1 because a simple example shows ```111 == 2^0 + 2^1 + 2^2 == 1 + 2 + 4 == 7 == (2^3)-1```

nikovacevic: Inferred because of `10.0` versus `10`, and the type of `PiB` is also `float64` for that same reason. Here's a fun "gotcha" example to play around with: https://go.dev/play/p/4nomozb9-t6

nikovacevic: And you pass your math assignment! Haha. (It's somewhat more complicated than this, but the gist is correct: a 32-bit integer can only support 2^32 unique numbers, or 4294967296 numbers -- which ends up being -2147483648, ..., 0, ... 2147483647 when encoded with [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement).)

PR #2060, opened 2023-07-28
Ignore missing map values

skl:
## What does this PR change?
* If I understand correctly, there are multiple levels of default pricing configs; as when a configmap is used to pass AWS bucket data, if default prices are not present then they're taken to be zero. Even though are actually AWS defaults already elsewhere, which is confusing.

* Not that I'm aware of

* I hope this is the right place to make the change to allow configmaps to be created without having to redefine all the default prices again.

* Not, it comes out of a [slack convo](https://cloud-native.slack.com/archives/C03D56FPD4G/p1690567548216929?thread_ts=1690557341.236339&cid=C03D56FPD4G)

## How was this PR tested?
* Unit testing

* Maybe, if you want to highlight that default values can be used if missed from custom pricing config, might be confusing though.

* Sorry not sure, I'm new here.


mattray:
Thanks for submitting this, with tests and everything! This looks good to me, I'll try to get someone from the Kubecost Engineering team to weigh-in since it's a change in behavior. Ideally we'll add a PR to https://github.com/opencost/opencost-website to mention this in the docs.

mattray:
Escalating to a P1 since pricing will be incorrect when you think you're using custom pricing and defaults are still used.

AjayTripathy:
@Sean-Holcomb I think you were recently looking at configmap stuff, is this a quick review for you?

Sean-Holcomb (commented):
As @skl mentioned this section of code is very confusing and not user friendly at all. I will try to walk through why this PR doesn't achieve what it sets out to do. 

In the original conversation you are talking about providing a config file via config map. This replaces the aws.json file that is used to hold the configs being used by OpenCost by changing the config path to the mount path of the config map. Since these values are the defaults and the original file is being replaced, all values do need to be populated or else there is nowhere else for them to come from. (This configuration approach is a problem as it can break other features). The change in this PR also would not make a difference because it deals with existing keys with no values, but the replaced aws.json file is just missing the keys entirely. 

The `UpdateFromMap` function you targeted here is a separate thing which updates the values in the aws.json file, whether it be the original one or mounted config map. As you can see it is passed a `map[string]string` which is required by the interface of  `ConfigMapWatcher` which can checks for changes in unmounted config maps. The `map[string]string` aspect of this is that the config map being watched must be key value. This is where the config map watcher in question is initialized:
https://github.com/opencost/opencost/blob/7b5b3df959264dc129cb9b174aeae63c788937b6/pkg/cloud/provider/provider.go#L66-L74
Notice the call to `p.UpdateConfigFromConfigMap(data) `

For all providers this is an access point to the function that you updated:

https://github.com/opencost/opencost/blob/7b5b3df959264dc129cb9b174aeae63c788937b6/pkg/cloud/aws/provider.go#L536-L538C2

To recap when you make an update to this secondary key/value config map. If you have custom pricing on, it should update the values in the custom pricing config map's aws.json file using values with existing keys, whether or not the value is empty. 

So what is this secondary config map? We can see in the snippet where the watcher is initialized that it is named by an env variable "PRICING_CONFIGMAP_NAME".

https://github.com/opencost/opencost/blob/7b5b3df959264dc129cb9b174aeae63c788937b6/pkg/env/costmodelenv.go#L209C1-L211C2

The default value is all that matters if we are using helm, but it could be set if you wanted to create it manually, which as of right now you would have to do with OpenCost but this is the Kubecost helm template for it
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml

Once again you will notice that if a value is not set in the config map there is just no key for it which means the change in this PR would not have an effect.

So what do we do about this?
As I mentioned above changing the config path env variable to point at a new `aws.json` file in a config-map is a big no no as it breaks cloud cost in open cost and also breaks things in our app. The ability to change this path is really just to enable local development and should not be part of any configuration flow for a deployment. Not doing this would leave the original `aws.json` file intact which would solve the problem you are trying to address.

Once this is done a template like the one I linked can be added to the helm chart which harnesses the existing `Values` that are already being used to populate the custom `aws.json` file.

Sean-Holcomb:
As mentioned

PR #2061, opened 2023-07-28
Fix shared cost coefficients for Summary Allocation API

nikovacevic:
## What does this PR change?
* Fix shared cost coefficients for Summary Allocation API by removing the filter keeping unmounted costs out of AllocationTotals. Also, create a new env var to allow use of a local AWS pricing object. 

* No

* Fixed shared costs

* Closes https://kubecost.atlassian.net/browse/BURNDOWN-142

## How was this PR tested?
* TODO (how are unit tests not catching this??)

* No

* Yes

### DIFF
--- a/pkg/kubecost/totals.go
+++ b/pkg/kubecost/totals.go
@@ -10,6 +10,13 @@ import (
 	"github.com/patrickmn/go-cache"
 )
 
+type CostPair struct {
+	Cost       float64
+	Adjustment float64
+}
+
+type CostPairSet map[string]CostPair
+
 // AllocationTotals represents aggregate costs of all Allocations for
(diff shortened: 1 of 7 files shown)


pkg/kubecost/totals.go
```diff
@@ -113,8 +113,8 @@ func ComputeAllocationTotals(as *AllocationSet, prop string) map[string]*Allocat
 	arts := map[string]*AllocationTotals{}
 
 	for _, alloc := range as.Allocations {
-		// Do not count idle or unmounted allocations
-		if alloc.IsIdle() || alloc.IsUnmounted() {
+		// Do not count idle allocations
+		if alloc.IsIdle() {
```
nikovacevic, line 126: This is the fix.

PR #2069, opened 2023-08-07
WIP: stop NaNs from propogating through opencost and kubecost

saweber:
## What does this PR change?
* Add SanitizeNaN functions to remove and prevent NaNs from breaking calculations and API responses due to NaN/JSON marshaling

* no

* stop NaNs

* See GTM-12

## How was this PR tested?
* unit tests

* worth noting that a rebuild will sanitize out existing NaNs

* I can't

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -279,6 +279,9 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	result.Errors = errors
 	result.Warnings = warnings
 
+	// Convert any NaNs to 0 to avoid JSON marshaling issues and avoid cascading NaN appearances elsewhere
+	result.SanitizeNaN()
+
 	return result, nil
 }
 
--- a/pkg/kubecost/allocation_test.go
+++ b/pkg/kubecost/allocation_test.go
@@ -3652,3 +3652,240 @@ func TestIsFilterEmptyFalse(t *testing.T) {
 		t.Errorf("matcher '%+v' should be not be reported empty but was", matcher)
 	}
 }
+
+func TestAllocation_SanitizeNaN(t *testing.T) {
+	tcName := "TestAllocation_SanitizeNaN"
+	alloc := getMockAllocation(math.NaN())
+	alloc.SanitizeNaN()
+	checkAllocation(t, tcName, alloc)
+}
+
+func checkAllocation(t *testing.T, tcName string, alloc Allocation) {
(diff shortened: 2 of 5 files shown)


pkg/kubecost/allocation_test.go
```diff
@@ -3652,3 +3652,247 @@ func TestIsFilterEmptyFalse(t *testing.T) {
...
+		RAMCost:                        f,
+		RAMCostAdjustment:              f,
+		SharedCost:                     f,
+		ExternalCost:                   f,
+		RawAllocationOnly:              getMockRawAllocationOnlyData(f),
+		ProportionalAssetResourceCosts: ProportionalAssetResourceCosts{"NaN": *getMockPARC(f)},
+		SharedCostBreakdown:            SharedCostBreakdowns{"NaN": getMockSharedCostBreakdown(f)},
+		LoadBalancers:                  LbAllocations{"NaN": getMockLbAllocation(f)},
+	}
+	return alloc
```
saweber, line 3744: I would like to convert these mock functions to use reflection (like the check function) if I have any extra time.

pkg/kubecost/allocation_test.go
```diff
@@ -3652,3 +3652,247 @@ func TestIsFilterEmptyFalse(t *testing.T) {
...
+
+func TestAllocation_SanitizeNaN(t *testing.T) {
+	tcName := "TestAllocation_SanitizeNaN"
+	alloc := getMockAllocation(math.NaN())
+	alloc.SanitizeNaN()
+	checkAllocation(t, tcName, alloc)
+}
+
+func checkAllocation(t *testing.T, tcName string, alloc Allocation) {
+	v := reflect.ValueOf(alloc)
```
saweber, line 3664: There is a lot of use of reflect in here to avoid repetitive code  - but only for tests.

pkg/costmodel/allocation.go
```diff
@@ -279,6 +279,9 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 	result.Errors = errors
 	result.Warnings = warnings
 
+	// Convert any NaNs to 0 to avoid JSON marshaling issues and avoid cascading NaN appearances elsewhere
+	result.SanitizeNaN()
```
saweber, line 283: I still need to investigate where else we need to add the SanitizeNaN call.

nikovacevic (approved):
LGTM. Phew, this one makes your eyes bleed! Haha.

PR #2071, opened 2023-08-09
Deprecate the "container_name" label in favor of "container" in ComputeAllocation queries

avrodrigues5:
## What does this PR change?
* In BYO prometheus, its hard dependency that user has prometheus relabelling block if in helm chart added to their [prometheus config ](https://github.com/kubecost/cost-analyzer-helm-chart/blob/06100016aaf0c46feb45fc853a4293c18e46d8a9/cost-analyzer/charts/prometheus/values.yaml#L1217C1-L1228C28). 
There shouldnt be any such dependency since kubernetes v1.14 onwards cadvisor only support container instead of container_name given via [1.14 release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md?plain=1#L985). This change ensure that if user config is missing the relabel our subquery fallback actually works.

* None

* Kubernetes clusters before v1.14 will be impacted, because the kubelet cAdvisor would still be emitting container_name. https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md?plain=1#L985

* Closes ... [Zendesk-4333](https://kubecost.zendesk.com/agent/tickets/4333) and Jira [CORE-330](https://kubecost.atlassian.net/browse/CORE-330)

## How was this PR tested?
Ensure both query returned same output
Screen Shot 2023-08-09 at 10 51 37 AM
Screen Shot 2023-08-09 at 10 51 18 AM

model/allocation before the fix and after the fix for unaccumulate and unaggregate allocation had same result, this is actually implied by the above test but just to be sure.

Before: 
Screen Shot 2023-08-09 at 10 53 16 AM

After:
Screen Shot 2023-08-09 at 11 20 10 AM

Non 10m request sizing option were seen too after the fix and disabling the recording rule

Screen Shot 2023-08-09 at 11 19 00 AM

* None

* v1.106

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -17,11 +17,11 @@ const (
 	queryFmtPodsUID                     = `avg(kube_pod_container_status_running{%s}) by (pod, namespace, uid, %s)[%s:%s]`
 	queryFmtRAMBytesAllocated           = `avg(avg_over_time(container_memory_allocation_bytes{container!="", container!="POD", node!="", %s}[%s])) by (container, pod, namespace, node, %s, provider_id)`
 	queryFmtRAMRequests                 = `avg(avg_over_time(kube_pod_container_resource_requests{resource="memory", unit="byte", container!="", container!="POD", node!="", %s}[%s])) by (container, pod, namespace, node, %s)`
-	queryFmtRAMUsageAvg                 = `avg(avg_over_time(container_memory_working_set_bytes{container!="", container_name!="POD", container!="POD", %s}[%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
-	queryFmtRAMUsageMax                 = `max(max_over_time(container_memory_working_set_bytes{container!="", container_name!="POD", container!="POD", %s}[%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
+	queryFmtRAMUsageAvg                 = `avg(avg_over_time(container_memory_working_set_bytes{container!="", container!="POD", %s}[%s])) by (container, pod_name, pod, namespace, instance, %s)`
+	queryFmtRAMUsageMax                 = `max(max_over_time(container_memory_working_set_bytes{container!="",  container!="POD", %s}[%s])) by (container, pod_name, pod, namespace, instance, %s)`
 	queryFmtCPUCoresAllocated           = `avg(avg_over_time(container_cpu_allocation{container!="", container!="POD", node!="", %s}[%s])) by (container, pod, namespace, node, %s)`
 	queryFmtCPURequests                 = `avg(avg_over_time(kube_pod_container_resource_requests{resource="cpu", unit="core", container!="", container!="POD", node!="", %s}[%s])) by (container, pod, namespace, node, %s)`
-	queryFmtCPUUsageAvg                 = `avg(rate(container_cpu_usage_seconds_total{container!="", container_name!="POD", container!="POD", %s}[%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
+	queryFmtCPUUsageAvg                 = `avg(rate(container_cpu_usage_seconds_total{container!="", container!="POD", %s}[%s])) by (container, pod_name, pod, namespace, instance, %s)`
 	queryFmtGPUsRequested               = `avg(avg_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!="", %s}[%s])) by (container, pod, namespace, node, %s)`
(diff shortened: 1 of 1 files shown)


saweber (approved):

nikovacevic (dismissed):
I won't lie... this makes me pretty nervous. But I think the idea makes sense.

cc @AjayTripathy @mbolt35 @michaelmdresser for visibility... do you guys see anything wrong with doing this?

michaelmdresser (dismissed):

mbolt35:
> I won't lie... this makes me pretty nervous. But I think the idea makes sense.
> 
> cc @AjayTripathy @mbolt35 @michaelmdresser for visibility... do you guys see anything wrong with doing this?

Why are we only changing this single query? Are we consistent here in all other queries?

avrodrigues5:
@jessegoodier @cliffcolvin  wanted to remove all mention of container_name to avoid NANs that user are seeing since if container_name is missing we might have NANs if we have divide by 0 down the pipeline. Making the change in same PR.

mbolt35:
I think this is problematic - Just a very light preliminary search for `container_name`: https://github.com/opencost/opencost/blob/a108e02325637fd6be8cfe7531c3459f57ae670e/pkg/costmodel/containerkeys.go#L160-L165

mbolt35:
I cannot stress enough that this type of change should be _very_ carefully reviewed. There are places in code we reference container_name, and we can't just remove things on a whim. This change will also effect the cloud agent. Please verify everything, no just a single metric comparison in prom.

jessegoodier:
> I cannot stress enough that this type of change should be _very_ carefully reviewed. There are places in code we reference container_name, and we can't just remove things on a whim. This change will also effect the cloud agent. Please verify everything, no just a single metric comparison in prom.

agreed. I believe we have a good method for testing. In the past I just put a proxy in front of prometheus and look for null result queries. 

I do know `kubecost_pod_network_ingress_bytes_total` and `kubecost_pod_network_egress_bytes_total`
have `pod_name` only.

cadvisor removed `_name` from `pod` and `container` back in k8s 1.16, which is no longer supported. 

Seems like removing `_name` from all our queries will make for more reliable/accurate queries.

jessegoodier:
https://github.com/kubernetes/kube-state-metrics/issues/501

mbolt35:
Yes, I think we should be looking to be consistent across the board, but these involve changes to code as well as the queries. Since we don't have an abstraction of a metrics provider, we don't have a lot of room to write in adapters for every specific metric. The slightest misstep here can have very negative overall consequences, so we need to verify a lot. I don't think this is the sort of risky change we want to make on a whim, ESPECIALLY because if users followed our "bring your own prometheus" documentation, we should point out the relabel instructions.

AjayTripathy:
> Yes, I think we should be looking to be consistent across the board, but these involve changes to code as well as the queries. Since we don't have an abstraction of a metrics provider, we don't have a lot of room to write in adapters for every specific metric. The slightest misstep here can have very negative overall consequences, so we need to verify a lot. I don't think this is the sort of risky change we want to make on a whim, ESPECIALLY because if users followed our "bring your own prometheus" documentation, we should point out the relabel instructions.

I agree we should not make this change without a deprecation/migration notice. Can we solve the initial problem with a relabel with a notice that this change is being made in v107 to give folks enough lead time?

michaelmdresser:
@avrodrigues5 I think discussion has become a bit tangled. I'd like to leave the discussion about the complex business of deprecating the `container_name` label in this PR. Would you please open a new PR with _only_ the CPU max subquery change (which we know is fairly safe) so that we can discuss it separately? That's just your original commit for this PR: https://github.com/opencost/opencost/pull/2071/commits/73f25b33586257c51f5c3160985bae84b6f9c7e7

I'm going to take the liberty of renaming this PR to the more general issue.

avrodrigues5:
> @avrodrigues5 I think discussion has become a bit tangled. I'd like to leave the discussion about the complex business of deprecating the `container_name` label in this PR. Would you please open a new PR with _only_ the CPU max subquery change (which we know is fairly safe) so that we can discuss it separately? That's just your original commit for this PR: [73f25b3](https://github.com/opencost/opencost/commit/73f25b33586257c51f5c3160985bae84b6f9c7e7)
> 
> I'm going to take the liberty of renaming this PR to the more general issue.

https://github.com/opencost/opencost/pull/2076

avrodrigues5:
> @avrodrigues5 I think discussion has become a bit tangled. I'd like to leave the discussion about the complex business of deprecating the `container_name` label in this PR. Would you please open a new PR with _only_ the CPU max subquery change (which we know is fairly safe) so that we can discuss it separately? That's just your original commit for this PR: [73f25b3](https://github.com/opencost/opencost/commit/73f25b33586257c51f5c3160985bae84b6f9c7e7)
> 
> I'm going to take the liberty of renaming this PR to the more general issue.

Thanks Michael, here you go https://github.com/opencost/opencost/pull/2076.

PR #2075, opened 2023-08-09
add ip field to load balancers

nickcurie:
## What does this PR change?
* Adds ip field to load balancers

* Yes https://github.com/kubecost/kubecost-cost-model/pull/1689

* Users can now see the IP associated with a load balancer

* Closes https://kubecost.atlassian.net/browse/SELFHOST-171

## How was this PR tested?
* Tested locally, and a unit test

* 

* v1.106

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -716,6 +716,7 @@ type LoadBalancer struct {
 	End        time.Time
 	Minutes    float64
 	Private    bool
+	Ip         string
 }
 
 func ClusterLoadBalancers(client prometheus.Client, start, end time.Time) (map[LoadBalancerIdentifier]*LoadBalancer, error) {
@@ -851,6 +852,12 @@ func ClusterLoadBalancers(client prometheus.Client, start, end time.Time) (map[L
 
 			hrs := (lb.Minutes * scaleFactor) / 60.0
 			lb.Cost += lbPricePerHr * hrs
+
+			if lb.Ip != "" && lb.Ip != providerID {
+				log.DedupedWarningf(5, "ClusterLoadBalancers: multiple IPs per load balancer not supported, using most recent IP")
+			}
+			lb.Ip = providerID
+
 			lb.Private = privateIPCheck(providerID)
 		} else {
 			log.DedupedWarningf(20, "ClusterLoadBalancers: found minutes for key that does not exist: %v", key)
--- a/pkg/kubecost/allocation.go
+++ b/pkg/kubecost/allocation.go
@@ -110,6 +110,8 @@ func (orig LbAllocations) Clone() LbAllocations {
 		newAllocs[key] = &LbAllocation{
 			Service: lbAlloc.Service,
 			Cost:    lbAlloc.Cost,
+			Private: lbAlloc.Private,
+			Ip:      lbAlloc.Ip,
 		}
 	}
 	return newAllocs
@@ -119,6 +121,7 @@ type LbAllocation struct {
 	Service string  `json:"service"`
 	Cost    float64 `json:"cost"`
 	Private bool    `json:"private"`
+	Ip      string  `json:"ip"` //@bingen:field[version=19]
 }
 
 func (lba *LbAllocation) SanitizeNaN() {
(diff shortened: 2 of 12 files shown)


ameijer:
@nickcurie , this LGTM but going to hold approval until the nil PV bug get chased down

nickcurie:
The PVAllocations bug exists on the develop branch too so I think its safe to say this PR did not cause that

ameijer (approved):
fair enough! ship it

pkg/kubecost/allocation.go
```diff
@@ -119,6 +119,7 @@ type LbAllocation struct {
 	Service string  `json:"service"`
 	Cost    float64 `json:"cost"`
 	Private bool    `json:"private"`
+	Ip      string  `json:"ip"`
```
Sean-Holcomb: This is a bingen type, you need a tag here, increment allocation version and run bingen again

Sean-Holcomb (changes requested):
One thing I noticed that is not in your change set is in LbAllocations.Clone() function the ip field and the private field are not cloned. Can you do a pass for anything like this in the eqauls() and clone() functions in the structs that you changed

pkg/kubecost/allocation.go
ameijer: didn't he already make the bingen changes? he would just have to run the bingen tag number on this specific field to ignore it?

nickcurie: Yeah, I definitely don't understand when to ignore vs. when to add to bingen and bump up the version

Sean-Holcomb: You only ignore if you do not want the field to be retained in filestorage. I don't think that is the case here.
Whenever you add a new field to a bingen struct you need add an ignore tag or increment the group version and add a version tag, then run bingen again.

If you look at the codec changes on this pr this field is read off the buffer without a version check, doing this will cause errors when unmarshalling binary that does not have the field.

pkg/kubecost/kubecost_codecs.go
```diff
@@ -5499,6 +5505,16 @@ func (target *LbAllocation) UnmarshalBinaryWithContext(ctx *DecodingContext) (er
 	e := buff.ReadBool() // read bool
 	target.Private = e
 
+	var g string
```
Sean-Holcomb: @nickcurie @ameijer

pkg/kubecost/allocation.go
ameijer: ah right. @nickcurie here is an example of us setting a bingen version on a field https://github.com/opencost/opencost/blob/develop/pkg/kubecost/allocation.go#L74

nickcurie:
@ameijer @Sean-Holcomb Just updated with the updated bingen stuff

ameijer (approved):

ameijer (changes requested):
took another look - we still need to update clone for these objects to ensure IP is copied

pkg/kubecost/asset.go
```diff
@@ -2526,6 +2528,10 @@ func (lb *LoadBalancer) add(that *LoadBalancer) {
 
 	lb.Cost += that.Cost
 	lb.Adjustment += that.Adjustment
+
+	if lb.Ip != that.Ip {
+		//TODO: should we add to an array here or just ignore?
```
ameijer, line 2533: I would add a warn log here and call it good

pkg/costmodel/cluster.go
```diff
@@ -851,6 +852,7 @@ func ClusterLoadBalancers(client prometheus.Client, start, end time.Time) (map[L
 
 			hrs := (lb.Minutes * scaleFactor) / 60.0
 			lb.Cost += lbPricePerHr * hrs
+			lb.Ip = providerID
```
ameijer, line 859: let's add another warning log here. If this lb object already has a non null IP, and that IP does not equal the provider ID, let's log a warning "warning - multiple IPs per load balancer not supported, picking most recent one" or something to that effect

ameijer (approved):

Sean-Holcomb (approved):

PR #2076, opened 2023-08-10
Fix subquery fallback to lookup container_cpu_usage_seconds_total metric on container instead of container_name 

avrodrigues5:
## What does this PR change?
* In BYO prometheus, its hard dependency that user has prometheus relabelling block if in helm chart added to their [prometheus config ](https://github.com/kubecost/cost-analyzer-helm-chart/blob/06100016aaf0c46feb45fc853a4293c18e46d8a9/cost-analyzer/charts/prometheus/values.yaml#L1217C1-L1228C28). 
There shouldnt be any such dependency since kubernetes v1.14 onwards cadvisor only support container instead of container_name given via [1.14 release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md?plain=1#L985). This change ensure that if user config is missing the relabel our subquery fallback actually works.

* None

* Kubernetes clusters before v1.14 will be impacted, because the kubelet cAdvisor would still be emitting container_name. https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md?plain=1#L985

* Closes ... [Zendesk-4333](https://kubecost.zendesk.com/agent/tickets/4333) and Jira [CORE-330](https://kubecost.atlassian.net/browse/CORE-330)

## How was this PR tested?
Ensure both query returned same output
Screen Shot 2023-08-09 at 10 51 37 AM
Screen Shot 2023-08-09 at 10 51 18 AM

model/allocation before the fix and after the fix for unaccumulate and unaggregate allocation had same result, this is actually implied by the above test but just to be sure.

Before: 
Screen Shot 2023-08-09 at 10 53 16 AM

After:
Screen Shot 2023-08-09 at 11 20 10 AM

Non 10m request sizing option were seen too after the fix and disabling the recording rule

Screen Shot 2023-08-09 at 11 19 00 AM

* None

* v1.106

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -84,7 +84,7 @@ const (
 	// the resolution, to make sure the irate always has two points to query
 	// in case the Prom scrape duration has been reduced to be equal to the
 	// ETL resolution.
-	queryFmtCPUUsageMaxSubquery = `max(max_over_time(irate(container_cpu_usage_seconds_total{container_name!="POD", container_name!="", %s}[%s])[%s:%s])) by (container_name, container, pod_name, pod, namespace, instance, %s)`
+	queryFmtCPUUsageMaxSubquery = `max(max_over_time(irate(container_cpu_usage_seconds_total{container!="POD", container!="", %s}[%s])[%s:%s])) by (container, pod_name, pod, namespace, instance, %s)`
 )
 
 // Constants for Network Cost Subtype


michaelmdresser (approved):
Approving because
1. This is a rare fallback code path
2. We plan to deprecate `container_name` in the long run, see https://github.com/opencost/opencost/pull/2071 
3. Tracing the code path fully reveals no hidden dependency on `container_name` which other code paths sometimes have (e.g. https://github.com/opencost/opencost/blob/a108e02325637fd6be8cfe7531c3459f57ae670e/pkg/costmodel/containerkeys.go#L160-L165, credit @mbolt35 for this reference)

cliffcolvin (approved):
LGTM

PR #2080, opened 2023-08-11
when possible, surface volume handle as provider id for PVs

ameijer:
## What does this PR change?
* a request by by the azure team for the resource ID to be surfaced via the API 

* ProviderID will no longer always be the name. In cases where the volume handle is available, that will be used as the provider id 

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -6,7 +6,7 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/opencost/opencost/pkg/cloud/provider"
+	cprovider "github.com/opencost/opencost/pkg/cloud/provider"
 	prometheus "github.com/prometheus/client_golang/api"
 	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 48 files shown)


pkg/costmodel/cluster.go
```diff
@@ -1436,10 +1441,6 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
 		}
 
 		diskMap[key].Cost = cost * (diskMap[key].Bytes / 1024 / 1024 / 1024) * (diskMap[key].Minutes / 60)
-		providerID, _ := result.GetString("provider_id") // just put the providerID set up here, it's the simplest query.
-		if providerID != "" {
-			diskMap[key].ProviderID = provider.ParsePVID(providerID)
-		}
```
nikovacevic: I think that this is going to break reconciliation. Sorry for taking so long on this review... I wanted to find the proof first.

nikovacevic: Here's my reasoning: Today, reconciliation of disks works by fetching CUR data, including so-called "provider ID" and cost, and matching that data with Asset data on this field. If today the match works, and I have no reason to believe it doesn't, then we cannot change the value here without also changing it in reconciliation. (Or else, this change works by virtue of the fact that it does not change the value of provider ID, in which case it is moot.)

Does that make sense? We could equivalently try to run this, and see if the provider ID values change -- and if they do, then we can check if reconciliation breaks.

nikovacevic: You might check with @Sean-Holcomb on cloud IDs and which fields are and are not available in the CUR, too. He may know more than I do here.

ameijer:
closing in favor of https://github.com/opencost/opencost/pull/2126

PR #1919, opened 2023-05-15
feat(labels): Add arch label to cost metrics

Pokom:
It's challenging today to display a breakdown of costs by chipset for workloads. This adds an `arch` label to all cost metrics so that queries can be written without joining to `kube_node_labels` to understand if the underlying architecture the workload is running on.

## What does this PR change?
* Adds a single label to the hourly_cost metrics

## Does this PR relate to any other PRs?
* 

## How will this PR impact users?
* 

## Does this PR address any GitHub or Zendesk issues?
- Closes #1916

## How was this PR tested?
* Tested against Azure, AWS and GCP nodes with and without arm nodepools

## Does this PR require changes to documentation?
* 

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next Opencost release? If not, why not?
*

PR #2093, opened 2023-08-16
Label entry to lookup node was missing architecture type

avrodrigues5:
## What does this PR change?
* [1919](https://github.com/opencost/opencost/pull/1919) PR introduced dependency to lookup node labels keys with architecture as a result of all node metrics' having architecture label in each node gauge vector metrics.

1. This PR fixes the creation of lookup key by adding architecture. This key will be later used in deletion of all node metrics.
2. For consistency i have added arch to `kubecost_node_is_spot` which was one of the remaining node metrics that did not have the new label `arch`

* None

* Actually fix node metrics without pod restart. 

* Closes ... [GTM-66](https://kubecost.atlassian.net/browse/GTM-66), [GTM-45](https://kubecost.atlassian.net/browse/GTM-45) and [GTM-54](https://kubecost.atlassian.net/browse/GTM-54)

## How was this PR tested?
* In latest RC for new nodes we see following logs  
Screen Shot 2023-08-15 at 3 13 03 PM

with this fix in node lookup  
Screen Shot 2023-08-15 at 5 25 53 PM
Screen Shot 2023-08-15 at 5 26 42 PM

when downsizing cluster to two node 
Screen Shot 2023-08-15 at 5 44 46 PM

Screen Shot 2023-08-15 at 5 46 07 PM

changes are accurately reflected in prometheus metrics which will solve all the above issues.

* None

* v1.106

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -187,7 +187,7 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider model
 		spotGv = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "kubecost_node_is_spot",
 			Help: "kubecost_node_is_spot Cloud provider info about node preemptibility",
-		}, []string{"instance", "node", "instance_type", "region", "provider_id"})
+		}, []string{"instance", "node", "instance_type", "region", "provider_id", "arch"})
 		if _, disabled := disabledMetrics["kubecost_node_is_spot"]; !disabled {
 			toRegisterGV = append(toRegisterGV, spotGv)
 		}
@@ -513,7 +513,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 
 				totalCost := cpu*cpuCost + ramCost*(ram/1024/1024/1024) + gpu*gpuCost
 
-				labelKey := getKeyFromLabelStrings(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID)
+				labelKey := getKeyFromLabelStrings(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID, node.ArchType)
 
 				avgCosts, ok := nodeCostAverages[labelKey]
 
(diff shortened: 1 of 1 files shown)


nik-kc (approved):

nikovacevic (approved):
Fantastic. Thanks @avrodrigues5! (And @nik-kc!)

avrodrigues5:
thanks @nikovacevic

thomasvn:
Great find @avrodrigues5 !! Could you clarify how the updates to the `kubecost_node_is_spot` metric in this PR affects the `node_total_hourly_cost` metric as shown in your screenshots?

avrodrigues5:
@thomasvn `kubecost_node_is_spot` had no implication on current bug you are right but this kubecost metric is one of the Six node metrics added in prometheus by kubecost. The original PR added arch for all of the other 5 metrics except the metrics we are speaking about. To have consistency i added it. 

Also if we dont add it https://github.com/avrodrigues5/opencost/blob/0d45eb7befa4ea9678bcf2cd7d670c0d69ea82e9/pkg/costmodel/metrics.go#L669-L715 we would need another labelKey to delete values in line 680... Prometheus silently drops `DeleteLabelValues` request if label values passed dont match the required label values.

thomasvn:
@avrodrigues5 Oh ok I think I understand now! The fix to the "ghost node" issue was your addition of `node.ArchType` to the following line:

https://github.com/opencost/opencost/blob/cb4a9a71598136c06e79afe6617563b9c6722d0b/pkg/costmodel/metrics.go#L516

This is because in order to [delete a metric from a Prometheus GaugeVec](https://github.com/opencost/opencost/blob/cb4a9a71598136c06e79afe6617563b9c6722d0b/pkg/costmodel/metrics.go#L674-L679), we need to key on all of that metric's labels. With the addition of PR #1919 we had added a new `arch` label to some of the GaugeVecs, but had not added it to `labelKey` used in `nodeSeen[labelKey]`.

Let me know if my understanding is correct here.

avrodrigues5:
Yes your understanding is correct @thomasvn .

PR #2107, opened 2023-08-21
GTM-63 sanitize label names before metric emission to prevent label name collisions

### LINKED ISSUE #2179: 500 error when requesting /metrics: error(s) occurred: * duplicate label names
---
# For reproducible bugs, please provide the following

**Describe the bug**  
500 error when requesting kubecost:9090/metrics:
```
duplicate label names
```

**To Reproduce**  

We are not sure what setting or environment variable is causing the issue. 

**Screenshots**  
```
>kubectl exec -i -t --namespace kubecost-agent \
deployment/kubecost-cost-analyzer -c cost-analyzer-frontend -- \
curl http://localhost:9003/metrics -v
*   Trying 127.0.0.1:9003...
* Connected to localhost (127.0.0.1) port 9003 (#0)
> GET /metrics HTTP/1.1
> Host: localhost:9003
> User-Agent: curl/7.88.1
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< Vary: Origin
< X-Content-Type-Options: nosniff
< Date: Fri, 21 Apr 2023 22:27:38 GMT
< Content-Length: 115
< 
An error has occurred while serving metrics:

2 error(s) occurred:
* duplicate label names
* duplicate label names
* Connection #0 to host localhost left intact
```

We have a debug log in slack if needed. It doesn't have anything useful from what I can tell.

nikovacevic:
## What does this PR change?
* Sanitize label names prior to metric emission so that label name collisions do not occur in Prometheus, at which point fatal errors are inevitable.

* No

* Users with label name collisions will no longer see the error `* duplicate label names`

* Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/2179
* Closes https://kubecost.atlassian.net/browse/GTM-63

## How was this PR tested?
* Unit tests
* Manually, by deploying the image and deploying a test pod with colliding label names

* No

* Yes

### DIFF
--- a/pkg/metrics/podlabelmetrics.go
+++ b/pkg/metrics/podlabelmetrics.go
@@ -7,36 +7,6 @@ import (
 )
 
 //--------------------------------------------------------------------------
-//  KubecostPodCollector
-//--------------------------------------------------------------------------
-
-// KubecostPodCollector is a prometheus collector that emits pod metrics
-type KubecostPodLabelsCollector struct {
-	KubeClusterCache clustercache.ClusterCache
-}
-
-// Describe sends the super-set of all possible descriptors of metrics
-// collected by this Collector.
-func (kpmc KubecostPodLabelsCollector) Describe(ch chan<- *prometheus.Desc) {
-	ch <- prometheus.NewDesc("kube_pod_annotations", "All annotations for each pod prefix with annotation_", []string{}, nil)
-}
-
-// Collect is called by the Prometheus registry when collecting metrics.
-func (kpmc KubecostPodLabelsCollector) Collect(ch chan<- prometheus.Metric) {
-	pods := kpmc.KubeClusterCache.GetAllPods()
-	for _, pod := range pods {
-		podName := pod.GetName()
-		podNS := pod.GetNamespace()
-
-		// Pod Annotations
-		labels, values := prom.KubeAnnotationsToLabels(pod.Annotations)
-		if len(labels) > 0 {
-			ch <- newPodAnnotationMetric("kube_pod_annotations", podNS, podName, labels, values)
-		}
-	}
-}
-
-//--------------------------------------------------------------------------
 //  KubePodLabelsCollector
 //--------------------------------------------------------------------------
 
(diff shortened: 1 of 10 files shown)


pkg/prom/metrics.go
```diff
@@ -104,3 +104,17 @@ func KubeAnnotationsToLabels(labels map[string]string) ([]string, []string) {
 func SanitizeLabelName(s string) string {
 	return invalidLabelCharRE.ReplaceAllString(s, "_")
 }
+
+// SanitizeLabels sanitizes all label names in the given map. This may cause
+// collisions, which is intentional as collisions that are not caught prior to
+// attempted emission will cause fatal errors. In the case of a collision, the
+// last value seen will be set, and all previous values will be overwritten.
+func SanitizeLabels(labels map[string]string) map[string]string {
```
nikovacevic, line 112: This is the function that fixes the labels.

pkg/metrics/podlabelmetrics.go
```diff
@@ -6,36 +6,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-//--------------------------------------------------------------------------
-//  KubecostPodCollector
```
nikovacevic, line 10: This was never referenced

AjayTripathy (approved):

AjayTripathy:
LGTM. Curious, what's your rough level of confidence this catches all the duplications?

nikovacevic:
> LGTM. Curious, what's your rough level of confidence this catches all the duplications?

Roughly? 95%. I didn't find any other references to maps of label in the `metrics` package.

pkg/prom/metrics.go
```diff
@@ -104,3 +104,17 @@ func KubeAnnotationsToLabels(labels map[string]string) ([]string, []string) {
...
+	response := make(map[string]string, len(labels))
+
+	for k, v := range labels {
+		response[SanitizeLabelName(k)] = v
```
thomasvn, line 116: @nikovacevic What're your thoughts on adding a WRN log here? While iterating, if we see that a key already exists in the map we can log the value that is about to be overwritten?

nikovacevic: I'm not opposed to doing that -- but I do want to be a bit careful not to make a really spammy log, as these will get run over and over and over. So I'd probably prefer to have a diagnostic than logs, but maybe that's being too picky.

thomasvn: Understood, let's leave as is for now!

PR #2112, opened 2023-08-22
improve auth error handling and response for AWS orphaned resources

saweber:
## What does this PR change?
* Improve error logic for AWS orphaned resources (volumes and addresses)

* No

* Users are less likely to see errors related to authentication and regions they are not allowed to view - these errors have been moved to info level logs, since it may be an intentional implementation decision and not an error.
* Orphaned resources will now prioritize returning valid volumes or addresses over returning errors.
* Prevents a panic if the incorrect AWS role name is configured when using AWS Keys

* Closes [...](https://kubecost.atlassian.net/browse/CORE-204)

## How was this PR tested?
* Manually, locally

* no

* can't

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/csv"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -15,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/aws/smithy-go"
 	"github.com/opencost/opencost/pkg/cloud/models"
 	"github.com/opencost/opencost/pkg/cloud/utils"
 	"github.com/opencost/opencost/pkg/kubecost"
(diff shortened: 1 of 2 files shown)


pkg/cloud/aws/provider.go
```diff
@@ -1559,8 +1561,26 @@ func (aws *AWS) getAllAddresses() ([]*ec2Types.Address, error) {
...
+				if errors.As(err, &awsErr) {
+					switch awsErr.ErrorCode() {
+					case "AuthFailure":
+						log.DedupedInfof(5, "Unable to get addresses for region %s due to AWS permissions, error message: %s", r, awsErr.ErrorMessage())
+						return
+					case "InvalidClientTokenId":
+						log.DedupedInfof(5, "Unable to get addresses for region %s due to AWS permissions, error message: %s", r, awsErr.ErrorMessage())
+						return
+					case "UnauthorizedOperation":
+						log.DedupedInfof(5, "Unable to get addresses for region %s due to AWS permissions, error message: %s", r, awsErr.ErrorMessage())
```
nikovacevic: Given that the body of each case is identical, we could just provide a list:
```go
case "AuthFailure", "InvalidClientTokenId", "UnauthorizedOperation":
```
Not important to change, but it might clean it up a bit.

pkg/cloud/aws/provider.go
```diff
@@ -1661,8 +1681,26 @@ func (aws *AWS) getAllDisks() ([]*ec2Types.Volume, error) {
 			// Query for first page of volume results
 			resp, err := aws.getDisksForRegion(context.TODO(), region, 1000, nil)
 			if err != nil {
-				errorCh <- err
-				return
+				var awsErr smithy.APIError
+				if errors.As(err, &awsErr) {
+					switch awsErr.ErrorCode() {
```
nikovacevic, line 1680: Ditto

nikovacevic (approved):

pkg/cloud/aws/provider.go
saweber: I learned something today! Thanks! And fixed.

PR #2114, opened 2023-08-22
Special case for Unallocated aliased fields when filtering

ghost:
Added special case for unallocated/unassigned aliased fields when filtering.

## What does this PR change?
* This PR modifies the ops returned by convertAliasFilterToLabelAnnotationFilter, to handle the special case where a user wants to examine unallocated/unassigned workloads within the aliased aggregation category.

* Nope

* If it was actually a regression, fixes the allocation view to so you can view unallocated workloads. If it wasn't a regression, added support for doing that with v2.1 filters to the backend.

* Closes [GTM-38](https://kubecost.atlassian.net/browse/GTM-38)

## How was this PR tested?
* Manually, against example failing queries used to replicate the original issue.

* I don't think so

* No, because it won't let me add labels.

### DIFF
--- a/pkg/kubecost/allocationmatcher.go
+++ b/pkg/kubecost/allocationmatcher.go
@@ -244,7 +244,9 @@ func convertAliasFilterToLabelAnnotationFilter(aliasKey string, filterValue stri
 		return nil, fmt.Errorf("unsupported op type '%s' for alias conversion", op)
 	}
 
-	return ops.Or(
+	// This handles the case where a label EXISTS/IS PRESENT for (is extant)
+	// for an aliased field. That's the primary case.
+	extantCaseNode := ops.Or(
 		ops.And(
 			ops.Contains(afilter.FieldLabel, aliasKey),
 			labelOp,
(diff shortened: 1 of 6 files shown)


michaelmdresser (commented):
Interesting fix! Is there a way to write a unit test for this? I know there are filter tests scattered around, I'm sorry I can't recall exactly the right place to look.

pkg/kubecost/allocationmatcher.go
```diff
@@ -256,5 +258,17 @@ func convertAliasFilterToLabelAnnotationFilter(aliasKey string, filterValue stri
 				annotationOp,
 			),
 		),
-	), nil
+	)
+	var node ast.FilterNode
+	if filterValue == "" || filterValue == "__unallocated__" {
+		node = ops.Or(
```
nikovacevic, line 272: Can we add a brief comment here about the non-extant case? Something that describes why we need these specific filter nodes for this specific condition? (It's described well in the PR description, but future authors may appreciate a brief comment in the code.)

nikovacevic (approved):
As we've discussed, this should work! Only ask is a small bit more commentary.

pkg/kubecost/allocationmatcher.go
ghost: Added, I'd meant to do that, woops.

nikovacevic: Excellent, looks great!

PR #1430, opened 2022-10-18
added current cluster id filter

AndrewChubatiuk:
Fixes #1431
## What does this PR change?
Introduces `CURRENT_CLUSTER_ID_FILTER_ENABLED` env variable which adds `<cluster_id_label>=<current_cluster_id>` filter to each prometheus query. It allows using shared prometheus for multiple Kubernetes clusters

## Does this PR relate to any other PRs?
* No

## How will this PR impact users?
* It just adds optional variable which by default doesn't change existing functionality

## How was this PR tested?
* It was tested in my EKS cluster with `CURRENT_CLUSTER_ID_FILTER_ENABLED` set to `false`, `true` and unset. Used UI for testing, so it means that tested `/allocation` endpoint only

## Does this PR require changes to documentation?
* `CURRENT_CLUSTER_ID_FILTER_ENABLED` variable can be added to documentation

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next Opencost release? If not, why not?
* Do not know which labels do you use

PR #2116, opened 2023-08-22
GCP Provider Cluster Filter Fix

mbolt35:
Unsure as to why this went untested, but this should correctly fix the query formatting.


nikovacevic (approved):

nikovacevic:
Introduced in https://github.com/opencost/opencost/pull/1430

PR #2117, opened 2023-08-23
[WIP] Add support for DefaultAzureCredentials

Davidsoff:
## What does this PR change?
* Add support for using [DefaultAzureCredentials](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-defaultazurecredential) for azure blob authentication. This would remove the need for the creation of secrets in kubernetes. 
* Start using the blob provider from the [Azure SDK for Go](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go).

* Not as far as I'm aware

* Allows use of [DefaultAzureCredentials](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-defaultazurecredential) for authentication to blob storage.

* None, as far as I'm aware

## How was this PR tested?
* Unit tests ran successfully. Other than that I'll need some help on figuring out how the config is read, and if I need to change anything else to make it do what I expect it to do.

* Yes, we'll need to let users know that OpenCost now supports DefaultAzureCredentials

* Not yet, this PR needs more work first


AjayTripathy:
So cool! @kaelanspatel I know you were investigating some of this-- are you the right person to review or maybe @Sean-Holcomb ?

mattray:
Thanks for submitting this PR! I'll raise it with Kubecost Engineering for review.

Sean-Holcomb:
First impression is that this is a great contribution, the Azure authorizer is kind of the problem child of the CSP authorizers because it is the only interface that does not provide service agnostic. Ideally would be able to change the interface to be service agnostic too, but it would be challenging to do so without breaking existing users.

Sean-Holcomb:
In terms of testing, one approach is to write an integration test like this https://github.com/opencost/opencost/blob/develop/pkg/cloud/azure/azurestorageintegration_test.go
pass criteria should be dependent on successful connection rather than contents of return. I would also want to build an image try both config methods.

mattray:
@r2k1 @babbageclunk any feedback on this from the Azure side? Do we want to update this and get it merged?

Sean-Holcomb:
@mattray I am working on some changes that we will allow us to merge this pr

r2k1:
I would suggest taking a look at

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential
https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#ChainedTokenCredential

Sean-Holcomb:
Closing in favor of https://github.com/opencost/opencost/pull/2363 which includes the commit from this PR along with some structural updates and renaming.

PR #2124, opened 2023-08-28
pkg/prom: fix dropped logging errors

alrs:
This fixes dropped `err` variables in `pkg/prom`.


michaelmdresser (approved):
Thanks for the contribution!

PR #2130, opened 2023-08-29
Resolve cross-platform opencost-ui container running as root

davidjericho:
## What does this PR change?
This fixes modifies the opencost-ui container so it'll run nginx as a non-root user by default. It is almost an exact copy of the contents from the original Dockerfile in the source directory.

CIS Kubernetes v1.23 advises in 5.2.7 to not run containers as root where not needed. Namespaces in K8S distros like RKE2 by default deny the ability to run as root. 

* Not that I've found.

* It shouldn't impact anyone with a default k8s cluster configuration. Those running with CIS and other compliance modes can run the container without interventions.

* No

## How was this PR tested?
* Following the OpenCost install instructions and integration tests noted in the contribuing guide. Work was done on RKE2 1.23 in CIS compliance mode.

* No

* No, as I can't see how to do this via the github editor.

### DIFF
--- a/ui/Dockerfile.cross
+++ b/ui/Dockerfile.cross
@@ -1,10 +1,17 @@
 FROM nginx:alpine
+
 COPY ./dist /var/www
 COPY default.nginx.conf /etc/nginx/conf.d/
 COPY nginx.conf /etc/nginx/
+COPY ./docker-entrypoint.sh /usr/local/bin/
+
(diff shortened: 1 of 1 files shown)


mattray (approved):

mattray:
The Dockerfile had been updated, but this was overlooked.
https://github.com/opencost/opencost/blob/develop/ui/Dockerfile

PR #2133, opened 2023-08-30
Protect config pricing against NaN and other invalid numerical values

nikovacevic:
## What does this PR change?
* Sanitizes and protects pricing information in config
* Adds extra NaN protection to use of CustomPricing config, because "NaN" can still be set from our own config, so a second check is required.

* No

* Invalid values will not reach configs, and so prevalence of NaN issues should be reduced.

* Partially addresses https://kubecost.atlassian.net/browse/BURNDOWN-228 (and any other NaN issue)

## How was this PR tested?
* New unit tests
* Manually sending requests to change config (see comment below for evidence)

* No

* Not yet -- might be 1.106, might be 1.107

### DIFF
--- a/pkg/cloud/models/models.go
+++ b/pkg/cloud/models/models.go
@@ -3,6 +3,7 @@ package models
 import (
 	"fmt"
 	"io"
+	"math"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


nikovacevic:
Previously, it was possible to get `NaN` for fields like `GPU`

Now, if you POST that you get an error:
```
POST /model/updateConfigByKey
{
  "GPU": "NaN"
}
```
```
{
  "code": 500,
  "status": "error",
  "data": {
    "provider": "custom",
    "description": "Default prices used to compute allocation between RAM and CPU. AWS pricing API data still used for total node cost.",
    "CPU": "0.031611",
    "spotCPU": "0.0066549999999999995",
    "RAM": "0.004237",
    "spotRAM": "0.000892",
    "GPU": "0.95",
    "spotGPU": "0.308",
    "storage": "0.00005479452",
    "zoneNetworkEgress": "0.01",
    "regionNetworkEgress": "0.01",
    "internetNetworkEgress": "0.143"
  },
  "message": "error updating provider config: error setting custom pricing field: invalid numeric value for field 'GPU': NaN"
}
```

And if you POST something valid, it will work -- and it will be formatted consistently for precision:
```
POST /model/updateConfigByKey
{
  "GPU": "0.87654321"
}
```
```
{
  "code": 200,
  "status": "success",
  "data": {
    "provider": "custom",
    "description": "Default prices used to compute allocation between RAM and CPU. AWS pricing API data still used for total node cost.",
    "CPU": "0.031611",
    "spotCPU": "0.0066549999999999995",
    "RAM": "0.004237",
    "spotRAM": "0.000892",
    "GPU": "0.876543210000",
    "spotGPU": "0.308",
    "storage": "0.00005479452",
    "zoneNetworkEgress": "0.01",
    "regionNetworkEgress": "0.01",
    "internetNetworkEgress": "0.143"
  }
}
```

pkg/cloud/models/models.go
```diff
@@ -215,24 +216,55 @@ func (cp *CustomPricing) GetSharedOverheadCostPerMonth() float64 {
...
+	}
+	if !allowNaN && math.IsNaN(num) {
+		return "", fmt.Errorf("expected a string representing a number; got 'NaN'")
+	}
 
+	// Format the numerical string we just parsed.
+	// - Format as a floating point number, as we received it.
+	// - Precision 12 because storage can be expressed that precisely.
+	// - Use a bit size of 64.
+	return strconv.FormatFloat(num, 'f', 12, 64), nil
```
michaelmdresser: Is storage the field requiring highest-precision today? Would we ever want this to be even higher precision than 12? (Or, vice-versa, lower precision?)

The documentation for `strconv.FormatFloat` says `The special precision -1 uses the smallest number of digits necessary such that ParseFloat will return f exactly.`. Why not use that? (I'm learning here, too).

pkg/cloud/models/models.go
```diff
@@ -215,24 +216,55 @@ func (cp *CustomPricing) GetSharedOverheadCostPerMonth() float64 {
...
-		return fmt.Errorf("Cannot set %s field value", name)
+		return fmt.Errorf("cannot set %s field value", name)
+	}
+
+	// If the custom pricing field is expected to be a string representation
+	// of a floating point number, e.g. a resource price, then do some extra
+	// validation work in order to prevent "NaN" and other invalid strings
+	// from getting set here.
+	switch strings.ToLower(name) {
+	case "cpu", "gpu", "ram", "spotcpu", "spotgpu", "spotram", "storage", "zonenetworkegress", "regionnetworkegress", "internetnetworkegress":
```
michaelmdresser, line 249: Alas...

michaelmdresser (approved):

pkg/cloud/models/models.go
nikovacevic: `-1` is a fantastic find. I'll use that. (I picked `12` to accord with the smallest number of the bunch, which was Storage.)

pkg/cloud/models/models.go
nikovacevic: I know... and we'll have to update this list if we add more :disappointed:

PR #2139, opened 2023-08-31
Provide ability to toggle load balancer PARC behavior

ameijer:
## What does this PR change?
* provides a way to call the API with different settings, allowing caller to build blended cost of both shared costs (single load balancer per AKS cluster) AND individual costs (1 external IP per service) 

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


ameijer:
cc @babbageclunk

nikovacevic (approved):

PR #2141, opened 2023-08-31
feat: athena configuration support for Catalog

### LINKED ISSUE #2140: Athena Catalog Configuration
Athena configuration does not allow using a custom Catalog (only allows the default aws data catalog)

We want to be able to specify our own custom catalog.

Dump the tables in the default catalog

**Additional context**
Add any other context, documentation links, or screenshots about the feature request here.

a7i:
## What does this PR change?
* Provide the ability to set custom Catalog name for Athena

* No

* Existing users and configuration will not be impacted since it's an optional filed
* New configurations that require a custom catalog can specify this new optional field

* Closes https://github.com/opencost/opencost/issues/2140

## How was this PR tested?
* Unit Test
* Our dev env of opencost installation running this fork

* Yes but unfortunately Opencost doesn't document any of the Athena configuration ([ref](https://github.com/opencost/opencost-website/blob/main/docs/configuration/AWS.md)) This is only documented for kubecost docs.

*


mattray:
Thanks for the PR! I'll raise it with Kubecost Engineering because it looks like it's not an OpenCost issue directly.

a7i:
> Thanks for the PR! I'll raise it with Kubecost Engineering because it looks like it's not an OpenCost issue directly.

@mattray I'm using opencost so it is an OpenCost issue directly in my case.

mattray:
@a7i what are you doing with Athena?

a7i:
> @a7i what are you doing with Athena?

We use it for querying Reserved Instances (RI) and CSP pricing of nodes. By default, OpenCost assumes that all schemas/tables are within the default `AwsDataCatalog` catalog but that is not the case for us. I have seen mentions of this in a kubecost Issue as well (kubecost/cost-analyzer-helm-chart#2290), and hence why I initially linked that.

But this would be a core feature support in OpenCost.

lmello:
Expanding a little bit on @a7i message, to explain our use case: 

Our cur Athena database is in another account and we have a catalog (using Athena and glue) shared with the account where opencost is running, this allow us to do cross account queries on the cur database.

This fix is required so we can specify the catalog name instead of using the AWS default one which does not have cur . 

We keep the default behavior in case the catalog is not specified.

Sean-Holcomb (approved):
This looks good, thank you for your contribution here

pkg/cloud/aws/athenaconfiguration.go
```diff
@@ -12,6 +12,7 @@ type AthenaConfiguration struct {
 	Bucket     string     `json:"bucket"`
 	Region     string     `json:"region"`
 	Database   string     `json:"database"`
+	Catalog    string     `json:"catalog""`
```
thomasvn, line 15: Looks like there's an extra `"` here for `"catalog""`. Does this affect the serialization/de-serialization of the JSON?

thomasvn: Nvm, this is fixed in #2170 !

PR #2145, opened 2023-09-04
pkg/costmodel: fix dropped errors

alrs:
This fixes two dropped `err` variables on calls to `getAllocatableVGPUs()`.


mattray:
Thanks for the PR! I've refered it to Kubecost Engineering for review.

alrs:
@mattray this was bit-rotting, so I've rebased against `develop`.

cliffcolvin (approved):
Thank you for the cleanup here. This looks good to me.

Issue #2144, opened 2023-09-01
Opencost quick setup does not work


tamalsaha:
**Describe the bug**
I am using the quick setup instructions:
https://www.opencost.io/docs/installation/install

It was working last week. It is not working today. The opencost-ui pod is in crashloop.

```
running with BASE_URL=/model
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2023/09/01 14:19:18 [emerg] 1#1: bind() to 0.0.0.0:80 failed (13: Permission denied)
nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)
```

https://raw.githubusercontent.com/opencost/opencost/develop/kubernetes/opencost.yaml

Needs a > 1024 port for opencost-ui.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Which version of OpenCost are you using?**
This may be the Kubecost release.

**Additional context**
Add any other context about the problem here. Kubernetes versions and which public clouds you are working with are especially important.

dwbrown2:
@mattray @teevans @cliffcolvin aware of any recent changes here?

dwbrown2:
@tamalsaha just to confirm are you using the following two commands?

```
helm install my-prometheus --repo https://prometheus-community.github.io/helm-charts prometheus \
  --namespace prometheus --create-namespace \
  --set prometheus-pushgateway.enabled=false \
  --set alertmanager.enabled=false \
  -f https://raw.githubusercontent.com/opencost/opencost/develop/kubernetes/prometheus/extraScrapeConfigs.yaml
```

```
[...]

PR #2148, opened 2023-09-06
Remove non-root user to restore UI

mattray:
## What does this PR change?
* Reverts OpenCost UI container to running as root, fixing https://github.com/opencost/opencost/issues/2144. I synchronized the `Dockerfile` and `Dockerfile.cross` so they've got the same content where appropriate. 

* https://github.com/opencost/opencost/pull/2130 brought in the original change.

* The OpenCost UI will run as root again, which was flagged as a security concern. We need to fix this properly, but 1.106 was already released.

## How was this PR tested?
* EKS amd64, Kind arm64, K3s arm64, K3s amd64

* 1.106.0 wasn't released for OpenCost, we'll do the 1.106.1 release instead.

### DIFF
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -9,16 +9,8 @@ FROM nginx:alpine
 COPY --from=builder /opt/ui/dist /var/www
 COPY default.nginx.conf /etc/nginx/conf.d/
 COPY nginx.conf /etc/nginx/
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Merging for now but we need to find a long term fix-- can we file an issue?

PR #2155, opened 2023-09-11
pkg/storage: defer Close() after error handling

alrs:
This fixes a `Close()` that was happening on a maybe-nil variable before error handling was done.


michaelmdresser (approved):
Thanks for the contribution!

PR #2165, opened 2023-09-15
csv export: override default max days (default: 90d)

### LINKED ISSUE #2164: csv export cost model: data query interval should be 1 day instead of 90 days
**Describe the bug**
OpenCost docs mention that CSV Export includes previous day data ([ref](https://www.opencost.io/docs/integrations/csv-export)).  The query [here](https://github.com/opencost/opencost/blob/130415a82778da984581923ec108ad23f286609c/pkg/costmodel/allocation.go#L299), however, requests 90d of data which is not allowed by AWS Managed Prometheus (by default).

```
"Error updating CSV: querying oldest sample: Prometheus communication error: 422 (Unprocessable Entity) URL: 'http://127.0.0.1:8080/workspaces/<REDACTED>/api/v1/query?query=min_over_time%28timestamp%28group%28node_cpu_hourly_cost%7B%7D%29%29%5B90d%3A1h%5D%29&time=1694721310', Request Headers: 'map[]', Headers: '{ X-Amzn-Requestid: [ 3e1177f2-3706-43f0-8495-412bd0e219ac ], Content-Length: [ 152 ], Vary: [ Accept-Encoding ], Date: [ Thu, 14 Sep 2023 19:55:10 GMT, Thu, 14 Sep 2023 19:55:10 GMT ], Server: [ amazon ], Connection: [ keep-alive ], Content-Type: [ text/plain; charset=utf-8 ] }', Body: '{\"status\":\"error\",\"errorType\":\"execution\",\"error\":\"expanding series: the query time range exceeds the limit (query length: 2160h5m0s, limit: 768h0m0s)\"}' Query: 'min_over_time(timestamp(group(node_cpu_hourly_cost{}))[90d:1h])'"}
```

**Expected behavior**
Only query 1 day of data

**Screenshots**

v105

**Additional context**
Kubernetes Version 1.26
AWS Managed Prometheus

a7i:
## What does this PR change?
* OpenCost docs mention that CSV Export includes previous day data ([ref](https://www.opencost.io/docs/integrations/csv-export)). The query [here](https://github.com/opencost/opencost/blob/130415a82778da984581923ec108ad23f286609c/pkg/costmodel/allocation.go#L299), however, requests 90d of data which is not allowed by AWS Managed Prometheus (by default). This allows the user to set the duration desired

* No

* for existing setup, the duration stays as 90d
* if env is set, it will use the duration as provided

* Closes #2164 

## How was this PR tested?
* Unit test and dev environment

* No

* yes

### DIFF
--- a/pkg/env/costmodelenv.go
+++ b/pkg/env/costmodelenv.go
@@ -106,6 +106,7 @@ const (
 	ExportCSVFile       = "EXPORT_CSV_FILE"
 	ExportCSVLabelsList = "EXPORT_CSV_LABELS_LIST"
 	ExportCSVLabelsAll  = "EXPORT_CSV_LABELS_ALL"
+	ExportCSVMaxDays    = "EXPORT_CSV_MAX_DAYS"
 )
 
 const DefaultConfigMountPath = "/var/configs"
@@ -126,6 +127,10 @@ func GetExportCSVLabelsAll() bool {
 
 func GetExportCSVLabelsList() []string {
 	return GetList(ExportCSVLabelsList, ",")
+}
+
+func GetExportCSVMaxDays() int {
+	return GetInt(ExportCSVMaxDays, 90)
 }
 
 // GetKubecostConfigBucket returns a file location for a mounted bucket configuration which is used to store
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -295,8 +295,9 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time, resolution time.Dur
 // it supposed to be a good indicator of available allocation data
 func (cm *CostModel) DateRange() (time.Time, time.Time, error) {
 	ctx := prom.NewNamedContext(cm.PrometheusClient, prom.AllocationContextName)
+	exportCsvDaysFmt := fmt.Sprintf("%dd", env.GetExportCSVMaxDays())
 
-	resOldest, _, err := ctx.QuerySync(fmt.Sprintf(queryFmtOldestSample, env.GetPromClusterFilter(), "90d", "1h"))
+	resOldest, _, err := ctx.QuerySync(fmt.Sprintf(queryFmtOldestSample, env.GetPromClusterFilter(), exportCsvDaysFmt, "1h"))
 	if err != nil {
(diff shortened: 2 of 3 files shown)


lmello (approved):

mattray:
Thanks for the PR (with tests)! I'm going to ping @r2k1 and his team since they wrote the original feature, hopefully he'll weigh in.

pkg/env/costmodelenv.go
```diff
@@ -106,6 +106,8 @@ const (
 	ExportCSVFile       = "EXPORT_CSV_FILE"
 	ExportCSVLabelsList = "EXPORT_CSV_LABELS_LIST"
 	ExportCSVLabelsAll  = "EXPORT_CSV_LABELS_ALL"
+
+	AllocationDateRange = "ALLOCATION_DATE_RANGE"
```
r2k1: `ALLOCATION_DATE_RANGE` sounds confusing.
It's also scoped to the csv export only. I would name it something like `EXPORT_CSV_*_DAYS`. `EXPORT_CSV_MAX_DAYS` or something like that.

pkg/env/costmodelenv.go
```diff
@@ -603,3 +605,9 @@ func GetRegionOverrideList() []string {
 
 	return regionList
 }
+
+func GetAllocationDateRange() string {
+	dateRange := GetInt64(AllocationDateRange, 90)
+
+	return fmt.Sprintf("%dd", dateRange)
+}
```
r2k1: I don't know how other configuration parameters are used.
But I think it would be easier to follow the logic if the function returns `int` or `time.Duration`.

r2k1 (commented):
Overall, LGTM. I would only ask to rename the ENV variable.

The 90 days selection was arbitrary, chosen for checking the oldest data in Prometheus. If this causes issues on some Prometheus setups, we can reduce it. I don't anticipate problems with existing deployments. However, reducing this period would affect the data amount exported on initial runs where opencost has been active for extended periods without CSV export.

An alternative is to execute multiple 30-day range queries. This may eliminate the need from users to configure an extra ENV variable.

r2k1 (approved):

mattray:
With @r2k1's approval I'm going to get this lined up for inclusion in 1.107. Thanks!

nikovacevic (approved):

PR #2174, opened 2023-09-19
update prom queries to use more accurate math

nickcurie:
## What does this PR change?
* Fixes a bug with `queryLocalStorageUsedAvg` and `queryLocalStorageUsedMax` where they were averaging/maxing disk usage across disks instead of summing the usages of the disks on a node and then averaging/maxing them

* https://github.com/kubecost/cost-analyzer-helm-chart/pull/2590

* Opencost/Kubecost have been incorrectly calculating cluster disk usage. Overall cost and capacity of Disks in the cluster are not affected, but features like and low disk right-sizing have been underestimating usage when making recommendations. We have fixed the calculation of usage, but incorrect sizing information may persist for up to a week after upgrading to a version of Opencost/Kubecost with the fix.

* Closes https://kubecost.atlassian.net/browse/GTM-86

## How was this PR tested?
* Tested on aws-test
  * Used the grafana pod in my namespace, exec'ed in and ran `df -h` to get the disk usage for the node, compared that value to the value given in prometheus from the query. Max query gave 7.121 GB, `df -h` reported 7.1 GB usage

* 

* v1.107


nickcurie:
@michaelmdresser I feel like my description of the issue is lacking a bit of detail. Any suggestions on how to make it more clear?

nik-kc (approved):

michaelmdresser:
@nickcurie 
"Opencost/Kubecost have been incorrectly calculating cluster PV and disk _usage_. Overall cost and capacity of PVs and Disks in the cluster is not affected, but features like PV right-sizing and disk right-sizing have been underestimating usage when making recommendations. We have fixed the calculation of usage, but incorrect sizing information may persist for up to a week after upgrading to a version of Opencost/Kubecost with the fix."

Something like that makes a good release note. Sometimes a rebuild is recommended, but I think in this case the blast radius is small enough that we can just let the bad data roll off.

For fellow reviewers, a good comparison point is a PR like this https://github.com/opencost/opencost/pull/1813 which fixes the ingestion of a particular usage field which was previously incorrect.

michaelmdresser (approved):

michaelmdresser:
EDIT: I was wrong about PVs -- we're still correct on PV stats

AjayTripathy:
curious-- what was the uniqueness in metrics that was causing avg not to work? @nickcurie

dwbrown2:
> curious-- what was the uniqueness in metrics that was causing avg not to work? @nickcurie

I believe it was metric duplication risk that caused us to use avg originally. @nickcurie any risk that you see for this?

nickcurie:
@dwbrown2 this latest commit should address those risks, I've added back in an average to normalize those potential duplicated metrics.
@AjayTripathy after talking to @nikovacevic the average was likely used to normalize any jobs that may have duplicated metrics like Webb mentioned. Adding a `sum` here will actually give a total amount of disk space used rather than the average disk usage across disks that could vary significantly across them (i.e. 2TB, 1TB, and 500GB disks would give us an average of 1.67TB and not a real total of 3.5TB)

michaelmdresser (approved):

PR #2177, opened 2023-09-20
fix: use spot label nodes from custom provider config

a7i:
## What does this PR change?
* Use `spotLabel` and `spotLabelValue` from custom provider config. We don't use Karpenter nor EKS but have our own self-managed Kubernetes clusters on AWS

* 

* For users that specify the `spotLabel` and `SpotLabelValue` in a custom provider, it will use those labels.

```json
  aws.json: |-
    {
        "athenaBucketName": "s3://...",
        "athenaCatalog": "..",
        "athenaDatabase": "..",
        "athenaProjectID": "...",
        "athenaRegion": "us-west-2",
        "athenaTable": "...",
        "athenaWorkgroup": "...",
        "description": "pricing model form Athena. AWS pricing API data still used for total node cost.",
        "provider": "aws",
        "spotLabel": "node-type",
        "spotLabelValue": "node-spot"
    }
```

* Closes ...

## How was this PR tested?
* Unit tests and local cluster

* No

*

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -666,6 +666,9 @@ func (k *awsKey) Features() string {
 // If the instance is a spot instance, it will return PreemptibleType
 // Otherwise returns an empty string
 func (k *awsKey) getUsageType(labels map[string]string) string {
+	if kLabel, ok := labels[k.SpotLabelName]; ok && kLabel == k.SpotLabelValue {
+		return PreemptibleType
+	}
 	if eksLabel, ok := labels[EKSCapacityTypeLabel]; ok && eksLabel == EKSCapacitySpotTypeValue {
 		// We currently write out spot instances as "preemptible" in the pricing data, so these need to match
 		return PreemptibleType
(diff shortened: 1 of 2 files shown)


lmello:
Adding a little bit more context on this PR: 
We were seeing the following error message in our logs: 
"Invalid Pricing Key us-west-2,c5.4xlarge,linux, "  notice the trailing comma. This PR fixes the trailing comma issue.

mattray:
Thanks for the PR following up on our conversation in #opencost! I've raised this with Kubecost Engineering to triage for inclusion in the next release (1.107).

Could you please give a little more context for the OpenCost version you found this and where you were experiencing this (just AWS us-west-2?) and any other environmental details you think might be relevant (Kubernetes version)?

a7i:
opencost version: v105
kubernetes: 1.26.7
We run our own plain vanilla Kubernetes on AWS (not using kOPS nor EKS) and we use Cluster-Autoscaler (not using Karpenter). As a result we use our own custom node labels (including spot labels).

This can be reproduced in all regions.

mattray:
If you're running 1.105.0 or 1.105.1 I highly recommend you upgrade to 1.105.2 or 1.106.1 because there is a bug with those releases where deleted nodes are not removed from cost calculations, creating inaccurate output.

nikovacevic (approved):

PR #2179, opened 2023-09-22
Add configuration option for AWS Athena pagination.

thomasvn:
## What does this PR change?

* Adds a configuration option for max results returned by an Athena query.
* Should be configurable via the `cloud-integration.json`. Requires further testing.

* Adds configurability for https://github.com/opencost/opencost/pull/2171
* Similar to https://github.com/opencost/opencost/pull/2141

* If the ingestion of CloudCosts results in Athena queries with result sets that are too large, this config can be tuned.

* None

## How was this PR tested?

* Not yet. Will need to tune the parameter and see how it affects queries in my deployment.

* Yes, will do.

*

### DIFF
--- a/pkg/cloud/aws/athenaconfiguration.go
+++ b/pkg/cloud/aws/athenaconfiguration.go
@@ -9,14 +9,15 @@ import (
 
 // AthenaConfiguration
 type AthenaConfiguration struct {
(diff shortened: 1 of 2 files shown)


thomasvn:
Not needed at this time. I will maybe return to this at a later point.

PR #2180, opened 2023-09-25
Add "mimir" to info log message

skl:
## What does this PR change?
* Add "mimir" to info log message

* No

* Improves log messaging for mimir users

## How was this PR tested?
* There don't appear to be any tests for these files.

* No

*


Pokom (approved):
Simple enough change and helps clarify when this error message is stated.

Possible improvement for a later time would be to check which type of client is being instantiated and skip the configuration all together, but that would require a bit of refactor.

cliffcolvin (approved):
Easy enough change. Thanks for adding this @Pokom.

PR #2184, opened 2023-09-28
Remove redundant `len` and `nil` check before loop

Juneezee:
## What does this PR change?

From the Go specification (https://go.dev/ref/spec#For_range):

> "1. For a nil slice, the number of iterations is 0."
> "3. If the map is nil, the number of iterations is 0."

`len` returns 0 if the slice or map is nil (https://pkg.go.dev/builtin#len). Therefore, having these checks before a loop is unnecessary:

- `len(v) != 0`
- `len(v) > 0`
- `v != nil`

No.

No user-facing change.

## How was this PR tested?
`go build ./...` and `go test ./...`

No.

No. External contributor cannot label PR.

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -1822,11 +1822,9 @@ func (aws *AWS) isDiskOrphaned(vol *ec2Types.Volume) bool {
 	}
 
 	// Do not consider volume orphaned if volume is attached to any attachments
-	if len(vol.Attachments) != 0 {
-		for _, attachment := range vol.Attachments {
-			if attachment.State == AttachedState {
-				return false
-			}
+	for _, attachment := range vol.Attachments {
+		if attachment.State == AttachedState {
+			return false
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


Juneezee (commented):
Tip: Hide the whitespace to make the diff easier to read

image

image

AjayTripathy:
Thank you for this PR. I suspect these hooks were left in the code to quickly add an 

else {
  // log error
}

since many of these seem to imply exceptional cases except for this https://github.com/opencost/opencost/pull/2184/files#diff-084f229ce009968d50ab3e5cbeb8ff6f42eb0fd08fdc7b9d027f6931d47ea483R1765

Juneezee:
@cliffcolvin Friendly ping :smiley:

AjayTripathy:
Hey @Juneezee I think we'd rather log something about these values being nil in an else clause, as I mentioned earlier, than drop the if altogether since these cases feel exceptional.

pkg/cloud/aws/provider.go
Juneezee, line 1: This one is redundant. I don't think we need a `log`.

pkg/costmodel/costmodel.go
Juneezee, line 1: This one is redundant. I don't think we need a `log`.

pkg/costmodel/metrics.go
Juneezee, line 1: This one is redundant. I don't think we need a `log`.

Juneezee (commented):
@AjayTripathy Updated. Please take a look again. Thanks!

mattray:
@AjayTripathy do you want to take a quick look?

PR #2187, opened 2023-10-01
use a default ratio to avoid issues with NaN ratios

AjayTripathy:
## What does this PR change?
* Uses some simple default ratios to stop a NaN from propagating throughout the application

* No

* Users will be able to set default pricing to zero if they so choose.

## How was this PR tested?
* Set CPU and RAM to zero in default pricing

* No

* Yes


pkg/costmodel/costmodel.go
```diff
@@ -1161,14 +1161,14 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
...
+				log.Warnf("cpuToRAMRatio[defaultCPU: %f / defaultRAM: %f] is NaN. Setting to 10.", defaultCPU, defaultRAM)
+				cpuToRAMRatio = 10
 			}
 
 			gpuToRAMRatio := defaultGPU / defaultRAM
 			if math.IsNaN(gpuToRAMRatio) {
-				log.Warnf("gpuToRAMRatio is NaN. Setting to 0.")
-				gpuToRAMRatio = 0
+				log.Warnf("gpuToRAMRatio is NaN. Setting to 100.")
+				gpuToRAMRatio = 100
```
lmello, line 1177: Is there any reason behind setting the ratio to 100, instead of other value?  
I understand that setting it to zero as it was before would result in NaN in other parts of the code, what would be the impact on the cost calculation by setting it to 100 by default (if the ratio is NaN) ?

AjayTripathy: These are basically magic constants as written today-- if the user doesn't supply the ratio we're just saying charge 1 GPU 100x as much as 1 GB as RAM.

lmello: thank you for the explanation.

cliffcolvin (approved):
LGTM

PR #2208, opened 2023-10-09
Use spot custom prices when spot node determined

legigor:
## What does this PR change?
* During the evaluation of the AKS cluster, I discovered that allocation costs significantly differ from what I expected.
* CostPerCPUHr for the allocation was 0.039, as well as other hourly costs
* In the pod's log lot of messages keep dropping:
  ```
  INF starting download retail price payload from "https://prices.azure.com/api/retail/prices?$skip=0&currencyCode='USD'&$filter=armRegionName+eq+%27centralus%27+and+armSkuName+eq+%27Standard_NC4as_t4_v3%27"
  ```

* When tested this URL with cURL, I got the empty response `[ [ ] ]`
* So, according to the code and comments, when no downloaded price data available, it uses default configured pricing data
* And here it comes that the spot node's cost used in allocation is the non-spot default price

More details not related to the given PR:

The problem is that Azure API is case-sensitive. Using `Standard_NC4as_T4_v3` in the query instead of `Standard_NC4as_t4_v3` returned the valid result. 

* It will significantly improve allocation cost precision for AKS users.

## How was this PR tested?
* Couldn't establish the proper unit test environment for the given function, so I tested it on my test Azure account.

* Nope, it fixes the bug in encapsulated logic.

* Yes, I'd recommend to include it to the nearest release


mattray:
Thanks for the PR! I'm going to refer this over to Kubecost Engineering to see about getting it into the next release.

AjayTripathy:
This won't make v107 but @cliffcolvin can we review once that's out the door and put in a patch?

nickcurie:
This change looks good to me. Lets make sure we rebase this with the latest opencost changes. cc: @cliffcolvin

cliffcolvin (approved):
LGTM and thanks @nickcurie for loading this up and doing some testing! I am going to resolve conflict and merge.

PR #2211, opened 2023-10-11
Implement Clone() for NodeOverhead so it shows up in accumulated results

michaelmdresser:
## What does this PR change?
* Implements `Clone()` for Node Overhead. Accumulated Asset API results will now be able to show Node Overhead.
* Adds small unit test for `Clone()` using `Equal()` which necessitated a change to `Equal()` for Asset properties and Breakdown.

For reviewers, copied from a commit message:
> The full test is of (*Node).Equal() which necessitated a change to
> equality for Breakdown and Properties.
> 
> It is odd to say that two things which are nil pointers of the same type
> are unequal (the old behavior). I think this is an acceptable change.
> The risk revolves around any functionality which expects Equal() to fail
> on two nil pointers -- I sincerely hope that nothing behaves this way.

* N/A

* Overhead will appear in accumulated Asset API responses

* Relates to https://kubecost.atlassian.net/browse/SELFHOST-811

## How was this PR tested?
* New unit test
* Ran KC locally and observed Overhead show up in accumulated Asset API responses -- it was not showing up before

* N/A

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -1589,6 +1589,10 @@ func (b *Breakdown) Clone() *Breakdown {
 
 // Equal returns true if the two Breakdowns are exact matches
 func (b *Breakdown) Equal(that *Breakdown) bool {
+	if b == nil && that == nil {
+		return true
+	}
+
 	if b == nil || that == nil {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


AjayTripathy:
so cool!

ameijer (approved):
LGTM

nikovacevic:
I can explain the `nil != nil` thing -- I have always treated `null` and other such values as "not equal to anything," which is how some paradigms do it. But now I actually think that's not appropriate, so I agree that `nil` should equal `nil` here. (For a nice justification see the accepted answer here https://stackoverflow.com/questions/75267/in-how-many-languages-is-null-not-equal-to-anything-not-even-null)

pkg/kubecost/asset.go
```diff
@@ -1889,6 +1893,32 @@ func (n *NodeOverhead) SanitizeNaN() {
...
+	}
+	if n != nil && other == nil {
+		return false
+	}
+	if n == nil && other == nil {
+		return true
+	}
+
+	// This is okay because everything in NodeOverhead is a value type.
+	return *n == *other
```
nikovacevic, line 1908: Ah so nice

nikovacevic: (Slightly tricky if that condition every changes... but these are tricky funcs to remember to update not matter what.)

nikovacevic (approved):

PR #2222, opened 2023-10-19
Open source CloudCost ingestion

Sean-Holcomb:
## What does this PR change?
This PR adds an ingestion pipeline and querying endpoints for CloudCost. This works by creating an in memory store where CloudCosts are held for querying. Queries are performed by combining results from multiple integrations which are each filtered and aggregated individually.

### API 
Included are several endpoints for working with CloudCosts, details on using these endpoints will be included in documentation.
Endpoints for querying CloudCost:
`/cloudCost`
`/cloudCost/view/graph`
`/cloudCost/view/totals`
`/cloudCost/view/table`
Endpoints for managing the Pipeline:
`/cloudCost/status`
`/cloudCost/rebuild`
`/cloudCost/repair`
Endpoints for managing configurations:
`/cloud/config/export`
`/cloud/config/enable`
`/cloud/config/disable`
`/cloud/config/delete`

### Other Changes
This PR also contains some changes to packaging a variable types here is a non exhaustive list of those changes
- CloudCost integrations now have a GetStatus() func that returns a  cloud.ConnectionStatus rather than GetConnectionStatus() which return a string. Other changes in the provider specific package are mainly repackaging and changes to the way Connection Status is updated, this addresses https://kubecost.atlassian.net/browse/SELFHOST-862
- CostMetricName consts now replace string consts of the same values, ParseCostMetricName is used in all places a string needs to be validated as a CostMetricName
-  CloudCostSet.Aggregate()  now interprets an []string{} to mean aggregate all with "" as the key
- CloudCostSetRange references to window have been removed as it is largely an unused field. The field has not been removed because of challenges with updating the bingen Codex. One option would be to remove it from the codex since that struct does not need to be binary encoded.
- CloudCostSetRange.Accumulate() has been moved to CloudCostSetRange.AccumulateAll() to make room for a Accumulate function that takes an AccumulateOption.

* 

* 

* Closes ...

## How was this PR tested?
* 

* Documentation for API and Configuration is in progress

*


pkg/cloud/config.go
```diff
@@ -17,6 +17,7 @@ type Config interface {
 type KeyedConfig interface {
 	Config
 	Key() string
+	Provider() string
```
Sean-Holcomb, line 20: Adding this function to the interface allows to removal of the GetProviderFromIntegration helper function which needed to be updated elsewhere

pkg/kubecost/cloudcostprops.go
```diff
@@ -156,10 +155,22 @@ func (ccp *CloudCostProperties) Intersection(that *CloudCostProperties) *CloudCo
...
+	CloudCostProviderProp,
+	CloudCostInvoiceEntityIDProp,
+	CloudCostAccountIDProp,
+	CloudCostCategoryProp,
+	CloudCostServiceProp,
+	CloudCostProviderIDProp,
+}
+
+// GenerateKey takes a list of properties and creates a "/" seperated key based on the values of the requested properties.
+// Invalid values are ignored with a warning. A nil input returns the default key, while an empty slice  returns the empty string
```
Sean-Holcomb, line 168: empty slice meaning aggregate all is a new feature here that has. There is a risk here of regression occuring else where because of this

pkg/proto/http.go
```diff
@@ -0,0 +1,209 @@
+package proto
```
Sean-Holcomb, line 1: This package was moved over, but not deleted in KCM to avoid even more files being pulled into these PRs

pkg/kubecost/cloudcost.go
```diff
@@ -371,9 +372,12 @@ func (ccs *CloudCostSet) Clone() *CloudCostSet {
...
 func (ccs *CloudCostSet) cloneSet() *CloudCostSet {
-	aggProps := make([]string, len(ccs.AggregationProperties))
-	for i, v := range ccs.AggregationProperties {
-		aggProps[i] = v
+	var aggProps []string
+	if ccs.AggregationProperties != nil {
+		aggProps = make([]string, len(ccs.AggregationProperties))
+		for i, v := range ccs.AggregationProperties {
+			aggProps[i] = v
+		}
```
Sean-Holcomb, line 380: This is an example of the second order effects of using []string{} to mean aggregate all. Clone was replacing nil with []string{} causing cloned cloudcostsets to aggregate all

nikovacevic (approved):

AjayTripathy:
@Sean-Holcomb mind signing DCO and syncing branch so we can merge?

PR #2251, opened 2023-10-27
Nil out overhead when adding nil to non-nil

michaelmdresser:
## What does this PR change?
* When adding a Node with nil overhead to one with non-nil overhead, the resulting overhead is now nil. This avoids hiding unknowns.

## How was this PR tested?
* New unit test


cliffcolvin (approved):
LGTM

PR #2255, opened 2023-10-30
check for nil struct before sanitizing

ameijer:
## What does this PR change?
* adds nil guards for sanitization. If the structs are nil we should not sanitize as it will through nil panic

* none

* prevents panics in edge cases 

* Closes ...

## How was this PR tested?
* covered by unit testing 

* 

*


cliffcolvin (approved):
LGTM

PR #2256, opened 2023-10-30
check for nil struct before sanitizing

ameijer:
cherry pick of #2255


ghost (approved):
Sure looks like it's doing exactly what you said it's doing

michaelmdresser (approved):

PR #2260, opened 2023-10-31
return err from query

Sean-Holcomb:
## What does this PR change?
* This PR address a bug introduced in https://github.com/opencost/opencost/pull/2222 which allows the big query queier to return a nil iterator without error

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/cloud/gcp/bigqueryquerier.go
+++ b/pkg/cloud/gcp/bigqueryquerier.go
@@ -2,6 +2,7 @@ package gcp
 
 import (
 	"context"
+	"fmt"
 
 	"cloud.google.com/go/bigquery"
 	"github.com/opencost/opencost/pkg/cloud"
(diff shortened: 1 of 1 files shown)


pkg/cloud/gcp/bigqueryquerier.go
```diff
@@ -49,9 +50,12 @@ func (bqq *BigQueryQuerier) Query(ctx context.Context, queryStr string) (*bigque
 	// If result is empty and connection status is not already successful update status to missing data
 	if iter == nil && bqq.ConnectionStatus != cloud.SuccessfulConnection {
 		bqq.ConnectionStatus = cloud.MissingData
-		return iter, nil
+	} else {
+		bqq.ConnectionStatus = cloud.SuccessfulConnection
```
saweber, line 54: For my understanding, how does this prevent the function from returning a nil iterator?

The downstream problem was returning a nil iterator with no error, which still seems to be a possibility here.

cliffcolvin (approved):
LGTM

pkg/cloud/gcp/bigqueryquerier.go
Sean-Holcomb: For context this line previously read
```
return query.Read(ctx)
}
```
I did it like this because it closely mirrors what it was before

PR #2271, opened 2023-11-01
Add Deployments to the breakdown drop-down

### LINKED ISSUE #2242: Deployments are missing from the OpenCost UI Breakdowns
If you set it in the URL it renders, we just need to add it to the choices.

mattray:
Fixes https://github.com/opencost/opencost/issues/2242


cliffcolvin (approved):
LGTM

PR #2273, opened 2023-11-01
Synced Date Range options

mattray:
They were inconsistent between the Allocation and Cloud Costs, made them the same and chose defaults that are generally useful in th OpenCost UI

Screenshot 2023-11-01 at 11 08 40 PM


cliffcolvin (approved):
LGTM

PR #2282, opened 2023-11-02
pkg/costmodel: fix dropped errors

alrs:
This picks up 12 dropped `err` variables in the `promparser` code found in `pkg/costmodel`.

### DIFF
--- a/pkg/costmodel/promparsers.go
+++ b/pkg/costmodel/promparsers.go
@@ -50,10 +50,7 @@ func GetPVInfo(qrs []*prom.QueryResult, defaultClusterID string) (map[string]*Pe
 	toReturn := make(map[string]*PersistentVolumeClaimData)
 
 	for _, val := range qrs {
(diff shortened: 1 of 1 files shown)


cliffcolvin (approved):
LGTM

cliffcolvin (dismissed):
@alrs on second thought here I've been asking around as this functionally changes the result. I believe this is a breaking change as clusterID isn't a required label. In the handled error case of this PR this will actually break when no clusterID is present. ClusterID being unset is an expected scenario for non-thanos (reading from local prom), and defaults to local cluster in that case.

cliffcolvin:
I think the best route forward would be to comment/doc this expected behavior and _ out the err returns.

alrs:
@cliffcolvin I've made the requested changes.

alrs:
@mattray @cliffcolvin any word?

nikovacevic (approved):

cliffcolvin (approved):

PR #2283, opened 2023-11-02
Making API and UI ports configurable

### LINKED ISSUE #2223: Expose OpenCost API and UI ports as configurable
Right now the UI is hard-coded to port 9090 and the API to 9003

https://github.com/opencost/opencost/blob/develop/ui/default.nginx.conf#L38
https://github.com/opencost/opencost/blob/develop/ui/default.nginx.conf#L59

If these were configurable we could run multiple instances of OpenCost in different namespaces for testing differences.

guresonur:
Fixes #2223 

## What does this PR change?
* It changes default.nginx.conf to default.nginx.conf.template in order to be able to configure nginx during startup with ports that are retrieved from the environment. in docker-entrypoint.sh script, envsubst utility is used to update API and UI ports with API_PORT and UI_PORT environment variables in nginx configuration. I have configured well known 9003 and 9090 ports as API and UI ports as defaults.

* No

* It won't have any impact if users do not configure API_PORT and UI_PORT explicitly.

* Closes #2223 

* Yes, users should be reminded now that they have option to set API and UI ports that will reflect in nginx. Also they need to be aware that Kubernetes manifests should be updated (service/port values in the case of custom ports supplied to nginx)


mattray:
Thanks for the PR, I'll go test it locally and report back

mattray:
Doesn't work with the API_PORT, given that's hard-coded in OpenCost. I'm investigating a supporting patch

mattray:
This patch was merged as part of https://github.com/opencost/opencost/pull/2366

Thanks!

PR #2293, opened 2023-11-06
Simplify prometheus queries

r2k1:
## What does this PR change?

Simplify prometheus queries.

Currently some of the metrics performs unnecessary calculations. Which are hard to understand and potentially are expensive to calculate.

Replace `avg_over_time(metric) * count_over_time(metric)` with `sum_over_time(metric)`

```
avg_over_time(metric) = sum_over_time(metric) / count_over_time(metric)
sum_over_time(metric) = avg_over_time(metric) * count_over_time(metric)
```

No impact is expected

## How was this PR tested?
Run locally, the request/usage data from API is within expected range.

No


mattray:
Thanks for the PR! This came up in Slack and @r2k1 sorted out the cause and potential solution. 
https://cloud-native.slack.com/archives/C03D56FPD4G/p1699302491784909

cliffcolvin (approved):
I ran these queries in prometheus manually and compared the results with no discrepancy. The changes here look good to me.

dwbrown2:
We should be very very careful with this change. I remember reviewing these queries early on and I believe they were used to handle 0/null cases. cc @AjayTripathy @nikovacevic

cliffcolvin:
> We should be very very careful with this change. I remember reviewing these queries early on and I believe they were used to handle 0/null cases. cc @AjayTripathy @nikovacevic

I requested their feedback on this.

nikovacevic:
If anything this should be safer (no division or multiplication that could cause weird null/NaN cases) but I agree that it's sensitive.

dwbrown2:
> If anything this should be safer (no division or multiplication that could cause weird null/NaN cases) but I agree that it's sensitive.

Do you remember rationale for initial implementation? I do at least remember that the `sum_over_time()` was definitely a supported function at the time.

nikovacevic:
I do not.

To me, the current implementation looks odd, and the change looks like a slight improvement. That said, there's always a risk to change. I could take it or leave it, but I consider it to be safe, likely.

AjayTripathy:
I think given the testing improvements @thomasvn has been working on-- we can run a very careful before and after on pod cost in the environment he's got for pod cost calculation? Thomas what are thoughts here?

thomasvn:
Yes, we have a nightly integration test validating a pod's CPU/RAM allocation costs! We can carefully monitor before/after this merge.

Although if the original implementation of this query was to handle an edge case in a user's environment or prometheus, I'm unsure if our tests would be capturing that specific edge case.

thomasvn:
@AjayTripathy According to git history, `count_over_time * avg_over_time` to calculate cpu/ram requests was part of the [initial commit](https://github.com/opencost/opencost/commit/6bd1bc3d9b74f1cb1a06e58f961754e4a26742a2). This may be a safe enhancement to the query?

pkg/costmodel/costmodel.go
```diff
@@ -144,39 +144,25 @@ const (
...
-						avg_over_time(container_memory_working_set_bytes{container!="", container!="POD", instance!="", %s}[%s] %s), "node", "$1", "instance", "(.+)"
-					), "container_name", "$1", "container", "(.+)"
-				), "pod_name", "$1", "pod", "(.+)"
-			)
-		) by (namespace, container_name, pod_name, node, %s)
-	)`
+					sum_over_time(container_memory_working_set_bytes{container!="", container!="POD", instance!="", %s}[%s] %s), "node", "$1", "instance", "(.+)"
+				), "container_name", "$1", "container", "(.+)"
+			), "pod_name", "$1", "pod", "(.+)"
+		)
```
thomasvn, line 159: @r2k1 What's the effect of changing this `count_over_time` to `sum_over_time`?

Additionally, is there any reason you chose to remove `sort_desc` from the query? Is it because the `queryCPUUsageStr` does not perform a sort?

r2k1: Performance and clarity.

`sort_desc` isn't used by any other query and have performance implications as well. There is also should be no reason to use it. I suspect it's just been left out after a manual query testing.

r2k1: I've seen many people got confused with the query in slack channel.

nikovacevic: @thomasvn to be clear, he's changing a (`count_over_time` * `avg_over_time`) to a `sum_over_time`. Changing a count, alone, to a sum would not make sense.

nikovacevic: For extra clarity here is old vs new:

### Old
```go
queryRAMUsageStr = `sort_desc(
	avg(
		label_replace(
			label_replace(
				label_replace(
					count_over_time(container_memory_working_set_bytes{container!="", container!="POD", instance!="", %s}[%s] %s), "node", "$1", "instance", "(.+)"
				), "container_name", "$1", "container", "(.+)"
			), "pod_name", "$1", "pod", "(.+)"
		)
		*
		label_replace(
			label_replace(
				label_replace(
					avg_over_time(container_memory_working_set_bytes{container!="", container!="POD", instance!="", %s}[%s] %s), "node", "$1", "instance", "(.+)"
				), "container_name", "$1", "container", "(.+)"
			), "pod_name", "$1", "pod", "(.+)"
		)
	) by (namespace, container_name, pod_name, node, %s)
)`
```

### New
```go
queryRAMUsageStr = `avg(
	label_replace(
		label_replace(
			label_replace(
				sum_over_time(container_memory_working_set_bytes{container!="", container!="POD", instance!="", %s}[%s] %s), "node", "$1", "instance", "(.+)"
			), "container_name", "$1", "container", "(.+)"
		), "pod_name", "$1", "pod", "(.+)"
	)
) by (namespace, container_name, pod_name, node, %s)`
```

thomasvn: @r2k1 Agree, thanks for the explanation

thomasvn: @nikovacevic Oh wow thanks! I hadn't noticed that it was a `count_over_time * avg_over_time`. That makes a lot of sense now.

PR #2296, opened 2023-11-07
Amortization For GCP Cloud Cost

Sean-Holcomb:
## What does this PR change?
This PR Introduces amortization for GCP Cloud Costs that have received credits from billing account global [Flexible CUDs](https://cloud.google.com/compute/docs/instances/committed-use-discounts-overview#spend_based) which do not have resource ids in the billing export. Additionally it attempts do a better job of not including credits and charges for [Resource-based CUDs](https://cloud.google.com/compute/docs/instances/committed-use-discounts-overview#resource_based) which do have the resource ids of the instances they are applied to.

### Flexible CUDs
Flexible CUD charges appear in the billing export as a variable number of line items per day that cannot be linked to credits distributed on individual line items, these line items may have credits of their own. This items in the billing export match the the values shown in the billing explorer in the GCP console. However there is not a straight forward way to distribute these costs to the credited line items when populating the amortized cost fields.

The work around is to determine the daily ratio of total-amount-payed/total-amount-credited, multiplying each line items amount credited by this ratio and adding it to the amortized cost metrics for that line item, while zeroing out the amortized cost metrics for the CloudCost that results from the Flexible CUD, results in a net-zero change in cost totals while, assigning the credited line item some proportion of the Flexible CUD spend for that day.

The limitation of this approach is that even with multiple Flexible CUD there will only be one ratio per day, if there are multiple Flexible CUD the amount determined by this algorithm will average the saving from all of them.

### Resource-based CUDs
These are standard CUDs that are applied to instances during specific window. In the billing export they have the resource id of the instance that they are being applied to with their charge, because of this these, line items were already being rolled into Cloud Costs. However there has not been any modification to which Cost Metrics these cost or their associated credit is charged to. To address this Credits and Charges for CUDs are only applied to Amortized Columns. These values are zeroed out for InvoicedCost, ListCost and NetCost as including them there would change the value of for the instance in a way that does not match those Cost Metrics' definitions.

### Query Changes
Additional select columns have been added to all Cloud Cost queries that break out credits into individual items, which help facilitate the above changes and the inclusion of the [cost_at_list](https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables/standard-usage#:~:text=or%20tag%20value.-,cost_at_list,-Float) column for populating the ListCost Cost Metric. cost_at_list is a relatively new column (added June 29, 2023), so if there is no value there, then the cost column is used instead as before.

* 

* GCP users with flexible CUDs should see more accurate reconciliation values. 
* GCP users with CUDS will see more accurate numbers for CloudCost for all Cost Metrics

* Closes ...

## How was this PR tested?
* This PR was tested against a billing export with a 1 year Flexible CUD

* No

*

### DIFF
--- a/pkg/cloud/gcp/cloudcost.go
+++ b/pkg/cloud/gcp/cloudcost.go
@@ -1,0 +1,100 @@
+package gcp
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/opencost/opencost/pkg/kubecost"
+)
+
+func IsK8s(labels map[string]string) bool {
+	if _, ok := labels["goog-gke-volume"]; ok {
+		return true
+	}
+
+	if _, ok := labels["goog-gke-node"]; ok {
+		return true
+	}
+
+	if _, ok := labels["goog-k8s-cluster-name"]; ok {
+		return true
+	}
+
+	return false
+}
+
+var parseProviderIDRx = regexp.MustCompile("^.+\\/(.+)?") // Capture "gke-cluster-3-default-pool-xxxx-yy" from "projects/###/instances/gke-cluster-3-default-pool-xxxx-yy"
+
+func ParseProviderID(id string) string {
+	match := parseProviderIDRx.FindStringSubmatch(id)
+	if len(match) == 0 {
+		return id
+	}
+	return match[len(match)-1]
+}
+
+func SelectCategory(service, description string) string {
+	s := strings.ToLower(service)
+	d := strings.ToLower(description)
+
+	// Network descriptions
+	if strings.Contains(d, "download") {
+		return kubecost.NetworkCategory
+	}
+	if strings.Contains(d, "network") {
+		return kubecost.NetworkCategory
+	}
+	if strings.Contains(d, "ingress") {
+		return kubecost.NetworkCategory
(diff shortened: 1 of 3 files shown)


pkg/cloud/gcp/cloudcost.go
```diff
@@ -0,0 +1,100 @@
...
+	if strings.Contains(d, "egress") {
+		return kubecost.NetworkCategory
+	}
+	if strings.Contains(d, "static ip") {
+		return kubecost.NetworkCategory
+	}
+	if strings.Contains(d, "external ip") {
+		return kubecost.NetworkCategory
+	}
+	if strings.Contains(d, "load balanced") {
```
michaelmdresser, line 59: Should this be `load balancer`?

michaelmdresser (approved):

pkg/cloud/gcp/cloudcost.go
Sean-Holcomb: If it was it would probably be `kubecost.LoadBalancer`. This function is old and just got moved. I can't say for sure what the descriptions were that contained this at this point but I trust it enought

PR #2299, opened 2023-11-08
Handle SUSE Rancher node special case

### LINKED ISSUE #2029: Incorrect cpu ram and gpu hourly cost on GPU nodes
**Describe the bug**
CPU RAM and GPU hourly cost metrics are much lower than they should be when a node has a GPU present. Found this issue in kubecost but also tested with opencost and we are seeing the same result.
See https://github.com/kubecost/cost-analyzer-helm-chart/issues/2418

This is in a Suse Rancher cluster where each node has the same `instance_type` of `rke2` which from the kubecost/opencost logs is not known and so pricing falls back to defaults. Specifying custom default prices exhibits the same unexpected behavior. Hourly costs reflect the expect cost values based on the custom pricing only when there is no gpu on the node. When the gpu is present the values are divided by a significant factor that seems related to the sum of the three prices.

Our guess is kubecost/opencost is doing something like summing the resource costs into a single machine cost but then unnecessarily dividing each cost by that total.

**To Reproduce**
See the kubecost issue https://github.com/kubecost/cost-analyzer-helm-chart/issues/2418 but install opencost instead of kubecost.

**Expected behavior**
GPU nodes should have hourly costs that match the specified prices.

**Screenshots**
E.g. from grafana, three nodes without gpu and one node with gpu, all using fallback default prices
image

1.105.0-rc.6 (b7736a7)

**Additional context**
Cluster is a Suse Rancher RKE2 cluster running on AWS EC2. GPU node in this instance is a `g4dn.xlarge` with one Nvidia Tesla T4 GPU.

ghost:
## What does this PR change?
* Adds a special case for SUSE Rancher in AWS, to skip ratio math that was adjusting instance pricing, courtesy of all SUSE Rancher nodes showing as "rke2", and thus forcing default pricing.
* **An important note** The draft PR currently includes a commit with replication code. The PR must not be merged until that commit is picked out. If it's not picked out, it will break opencost for anyone else. That's why this is a draft. I've left that commit in for now to enable others to test these changes, once people are happy, it can get picked out and we can move on.

* I sure hope not.

* Enables expected default pricing behavior for SUSE Rancher users

* Closes #2029 , and [BURNDOWN-222](https://kubecost.atlassian.net/browse/BURNDOWN-222)

## How was this PR tested?
* Locally, and in a dev cluster. **BUT**, and this is an important caveat, this was tested by deliberately overriding node details on a code level, to force the "rke2" instance type without having to actually spend money on GPU nodes. Ideally we'd like to see this also replicated in a real SUSE Rancher environ.

* I don't think so? Probably?

* Sure would if I could. Seriously, why can't I add labels?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 ui/.parcel-cache
 ui/.cache
 ui/dist
+ui/.env
 ui/node_modules/
 cmd/costmodel/costmodel
 cmd/costmodel/costmodel-amd64
--- a/pkg/cloud/azure/authorizer.go
+++ b/pkg/cloud/azure/authorizer.go
@@ -1,80 +1,60 @@
 package azure
 
 import (
-	"encoding/json"
 	"fmt"
 
-	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/opencost/opencost/pkg/cloud"
+	"github.com/opencost/opencost/pkg/util/json"
 )
 
-const AccessKeyAuthorizerType = "AzureAccessKey"
+const DefaultCredentialAuthorizerType = "AzureDefaultCredential"
 
+// Authorizer configs provide credentials from azidentity to connect to Azure services.
 type Authorizer interface {
 	cloud.Authorizer
-	GetBlobCredentials() (azblob.Credential, error)
+	GetCredential() (azcore.TokenCredential, error)
 }
 
 // SelectAuthorizerByType is an implementation of AuthorizerSelectorFn and acts as a register for Authorizer types
 func SelectAuthorizerByType(typeStr string) (Authorizer, error) {
 	switch typeStr {
-	case AccessKeyAuthorizerType:
-		return &AccessKey{}, nil
+	case DefaultCredentialAuthorizerType:
+		return &DefaultAzureCredentialHolder{}, nil
 	default:
 		return nil, fmt.Errorf("azure: provider authorizer type '%s' is not valid", typeStr)
 	}
 }
 
-type AccessKey struct {
-	AccessKey string `json:"accessKey"`
-	Account   string `json:"account"`
-}
+type DefaultAzureCredentialHolder struct{}
 
-func (ak *AccessKey) MarshalJSON() ([]byte, error) {
-	fmap := make(map[string]any, 3)
-	fmap[cloud.AuthorizerTypeProperty] = AccessKeyAuthorizerType
-	fmap["accessKey"] = ak.AccessKey
-	fmap["account"] = ak.Account
+func (dac *DefaultAzureCredentialHolder) MarshalJSON() ([]byte, error) {
+	fmap := make(map[string]any, 1)
+	fmap[cloud.AuthorizerTypeProperty] = DefaultCredentialAuthorizerType
+
 	return json.Marshal(fmap)
 }
 
-func (ak *AccessKey) Validate() error {
-	if ak.AccessKey == "" {
-		return fmt.Errorf("AccessKey: missing access key")
-	}
-	if ak.Account == "" {
-		return fmt.Errorf("AccessKey: missing account")
-	}
+func (dac *DefaultAzureCredentialHolder) Validate() error {
 	return nil
 }
 
-func (ak *AccessKey) Equals(config cloud.Config) bool {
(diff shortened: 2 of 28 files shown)


ian-lmcb:
Hi @CampfireRemnants, saw this pop up on my notifications per the reference to #2029.

I am wondering if there's a misinterpretation here as it sounds like the direction you're going is to fix rke2 instance pricing to use aws pricing which doesn't sound right to me (or sounds like an entire feature on its own, to detect and provide ec2 pricing for suserancher running on ec2). When we found this issue we were using SuseRancher in AWS to emulate a customer environment that uses SuseRancher on a VMware cluster and so it was intended and expected that custom pricing would be used for the cluster.

In our example we're seeing correctly the custom prices on rke2 instances without a gpu. Then on instances with a gpu all the prices go out of whack. So it sounds like there's a pricing issue for instances with gpu when custom prices are used, rather than an issue specific to suserancher or aws.

Possibly I may be misunderstanding the PR description

ghost:
@ian-lmcb So, this PR does still use the custom default prices, it pulls them from the config same as before; what we believe is happening here, is that because the instance type `rke2` isn't recognized by the costmodel (since it's not a "real" AWS instance type), it then falls into a backup flow for inferring pricing from the defaults, and the GPU count/CPU core count/RAM bytes reported by the Kubernetes API. That in turn has a lot of calculations  of ratios, etc. to modify those prices, and that's what seemed to be modifying the costs you were seeing emitted by a consistent ratio. What this does is skip over those ratio calculations that were throwing the reported numbers off when the instance type would indicate a SUSE Rancher cluster. Does that clarify the changes we made?

pkg/costmodel/aggregation.go
```diff
@@ -2275,15 +2275,15 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	// Otherwise it is computed at the cluster level. (Not relevant if idle
 	// is not included.)
 	idleByNode := qp.GetBool("idleByNode", false)
-	sharedLoadBalancer := qp.GetBool("sharelb", false)
+	// sharedLoadBalancer := qp.GetBool("sharelb", false)
```
AjayTripathy: what's going on here? Why are we removing the shared loadbalancer functionality for a special case?

ghost: Ah, thanks for the catch here--I believe that had been broken when this was first worked on, and got disabled back then so that it would run

AjayTripathy: Has that been fixed and these changes can be reverted?

cliffcolvin:
@CampfireRemnants can you take a look at Ajay's question above and we can make sure its resolved. Also we'll need to make sure and sign the commit for DCO to pass.

ghost:
The replication code that was being used to simulate a GPU environment has been picked back out--if anyone wanted to try running it themselves but hasn't gotten the chance yet, you should be able to pick it back in; This also took out the changes to run when the shared loadbalancer had been broken.

AjayTripathy (approved):

teevans (approved):

ian-lmcb:
Hi @CampfireRemnants, @AjayTripathy, I just want to let you know we're testing a suserancher environment that is running on a VMware vSphere data center and the node instance types we see here are not `rke2` but instead strings along the lines of `vsphere-vm.cpu-2.mem-4gb.os-ubuntu`.

So I reckon this fix will not work in this situation since the fix is applied only to instance type `rke2`.

I'm also confused about why the code is taking ratios in the first place when the default prices are already specified. I guess perhaps I misunderstand how the `defaultModelPricing` values are meant to be used. E.g. are the values not already per cpu, per unit of ram, per gpu prices?

The only documentation I've found for `defaultModelPricing` is in the example values.yaml here https://github.com/kubecost/cost-analyzer-helm-chart/blob/44428e49ed6acaab8fda180792db1f9bde436e11/cost-analyzer/values.yaml#L1339C143-L1339C143.

I would have thought that if I put in say 1.0 for the cpu cost then `node_cpu_hourly_cost` would become something like 1.0/730 = 0.00137 (using 730 hours per month) per cpu.

Looking back at #2029 it looks like I didn't see that for either the gpu and no-gpu node cases.

Is there documentation on how `defaultModelPricing` is meant to be used?

AjayTripathy:
If we detect that you're on a cloud provider, the default prices are used to generate ratios on splitting up node cost into constituent resource costs. If not, as should be the case in vSphere, we use those resource prices defined in defaultModelPricing directly. Are you seeing something different?

Issue #2290, opened 2023-11-05
Unused metrics


r2k1:
OpenCost currently emits several metrics, some of which remain unused. Notably, a subset of these unused metrics are high-cardinality ones, particularly those related to containers and pods This leads to an unnecessary increase in resource consumption.

Metrics in Question
1. kube_pod_container_resource_limits
2. kube_pod_container_resource_limits_memory_bytes
3. kube_pod_container_resource_limits_cpu_cores
4. kube_pod_container_status_restarts_total
5. kube_node_status_condition
6. kube_deployment_status_replicas_available
7. kube_deployment_spec_replicas
8. kube_persistentvolume_status_phase

mattray:
Thanks for taking the time to identify these. I'll escalate this to Kubecost Engineering to get them to verify this and hopefully we can clear these out.

r2k1:
I can remove the metrics. But I need a confirmation it's ok to do.

dwbrown2:
Interested to get input from others, but I know these are used by multiple downstream projects for cost monitoring. One idea would be to allow these to be disabled with a runtime flag?

r2k1:
Another potential metric is `kube_pod_status_phase`. Currently it emits 5 metrics per pod (1 per potential pod phase). This metrics is used in a single `queryGPURequestsStr` query. And I think it doesn't have an effect and can be removed. 

https://github.com/opencost/opencost/blob/bdb6c5df38613d845998b7f5423885cb36ae2fa1/pkg/costmodel/costmodel.go#L207

AjayTripathy:
Hey @r2k1 I'm supportive of making these optional behind a flag but on by default. How does that sound?

r2k1:
I managed to disable the metrics by adjusting opencost deployment:

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: opencost
  namespace: kube-system
spec:
  # ...
  template:
    # ...
    spec:
      # ...
      containers:
        - # ...
          name: opencost
          env:
            # ...
            - name: CONFIG_PATH
              value: "/etc/config"
          volumeMounts:
            - name: opencost-config-volume
              mountPath: /etc/config
              readOnly: true
      volumes:
        - name: opencost-config-volume
          configMap:
            name: opencost
            items:
              - key: metrics.json
                path: metrics.json
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: opencost
  namespace: kube-system
data:
  metrics.json: |
    {
[...]

PR #2312, opened 2023-11-16
Simplify GPU query

r2k1:
This PR addresses an optimization in the `GPURequest` query by removing its reliance on the `kube_pod_status_phase` metric. Unlike `CPURequest` and `RAMRequest`, `GPURequest` uniquely utilizes `kube_pod_status_phase`. However, the rationale behind this specific usage is unclear, as it lacks documentation in the git history, and the original author has no recollection of the reasoning.

I suspect it's been done to filter out not running pods. I think `node!=""` filter in `kube_pod_container_resource_requests` query should be sufficient to select only pods allocated to nodes and aligns with all other queries.

It's the only place where `kube_pod_status_phase` metric is used. It's one of the most resource-intensive metrics. It produces 5 series per pod.

With this change, this metric can be removed as a part of a fix in #2290.

No impact is expected.

* No


r2k1:
With this change request queries are almost identical:

```go
	queryRAMRequestsStr = `avg(
		label_replace(
			label_replace(
				sum_over_time(kube_pod_container_resource_requests{resource="memory", unit="byte", container!="",container!="POD", node!="", %s}[%s] %s)
				, "container_name","$1","container","(.+)"
			), "pod_name","$1","pod","(.+)"
		)
	) by (namespace,container_name,pod_name,node,%s)`
	queryCPURequestsStr = `avg(
		label_replace(
			label_replace(
				sum_over_time(kube_pod_container_resource_requests{resource="cpu", unit="core", container!="",container!="POD", node!="", %s}[%s] %s)
				, "container_name","$1","container","(.+)"
			), "pod_name","$1","pod","(.+)"
		)
	) by (namespace,container_name,pod_name,node,%s)`
	queryGPURequestsStr = `avg(
		label_replace(
			label_replace(
				sum_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!="", %s}[%s] %s) * %f, 
				"container_name","$1","container","(.+)"
			), "pod_name","$1","pod","(.+)"
		)
	) by (namespace,container_name,pod_name,node,%s)`
```

Almost, but not exactly. GPURequests suspiciously contains extra `* %f`. Where `%f` is `resolution.Hours()`.
It looks to me like there is an existing bug with GPU allocation.

mattray:
Thanks! I'm sending it to @cliffcolvin given all the other related PRs.

AjayTripathy:
@r2k1 can you share a little on how this was tested? Have you actually spun up GPU nodes to verify that we get an accurate count here?

AjayTripathy:
Friendly ping @r2k1 . Can we help you test this?

r2k1:
I failed to submit a comment for some reason.

Yes. I tested it. Spawn a cluster with GPU and compare results for allocation endpoint before and after change.
Didn't notice any difference for gpu allocation (or any other).

pkg/costmodel/costmodel.go
```diff
@@ -143,9 +143,8 @@ const (
 	queryRAMRequestsStr = `avg(
 		label_replace(
 			label_replace(
-				avg(
-					sum_over_time(kube_pod_container_resource_requests{resource="memory", unit="byte", container!="",container!="POD", node!="", %s}[%s] %s)
-				) by (namespace,container,pod,node,%s) , "container_name","$1","container","(.+)"
+				sum_over_time(kube_pod_container_resource_requests{resource="memory", unit="byte", container!="",container!="POD", node!="", %s}[%s] %s)
```
AjayTripathy, line 146: Did something get messed up in a merge here? I still think the avg wrapper is important

pkg/costmodel/costmodel.go
```diff
@@ -1683,11 +1678,11 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, cp costAnalyzerC
 
 	queryRAMAlloc := fmt.Sprintf(queryRAMAllocationByteHours, env.GetPromClusterFilter(), resStr, env.GetPromClusterLabel(), scrapeIntervalSeconds)
 	queryCPUAlloc := fmt.Sprintf(queryCPUAllocationVCPUHours, env.GetPromClusterFilter(), resStr, env.GetPromClusterLabel(), scrapeIntervalSeconds)
-	queryRAMRequests := fmt.Sprintf(queryRAMRequestsStr, env.GetPromClusterFilter(), resStr, "", env.GetPromClusterLabel(), env.GetPromClusterLabel())
+	queryRAMRequests := fmt.Sprintf(queryRAMRequestsStr, env.GetPromClusterFilter(), resStr, "", env.GetPromClusterLabel())
 	queryRAMUsage := fmt.Sprintf(queryRAMUsageStr, env.GetPromClusterFilter(), resStr, "", env.GetPromClusterLabel())
-	queryCPURequests := fmt.Sprintf(queryCPURequestsStr, env.GetPromClusterFilter(), resStr, "", env.GetPromClusterLabel(), env.GetPromClusterLabel())
+	queryCPURequests := fmt.Sprintf(queryCPURequestsStr, env.GetPromClusterFilter(), resStr, "", env.GetPromClusterLabel())
```
AjayTripathy, line 1740: As above

pkg/costmodel/costmodel.go
```diff
@@ -1683,11 +1678,11 @@ func (cm *CostModel) costDataRange(cli prometheusClient.Client, cp costAnalyzerC
 
 	queryRAMAlloc := fmt.Sprintf(queryRAMAllocationByteHours, env.GetPromClusterFilter(), resStr, env.GetPromClusterLabel(), scrapeIntervalSeconds)
 	queryCPUAlloc := fmt.Sprintf(queryCPUAllocationVCPUHours, env.GetPromClusterFilter(), resStr, env.GetPromClusterLabel(), scrapeIntervalSeconds)
-	queryRAMRequests := fmt.Sprintf(queryRAMRequestsStr, env.GetPromClusterFilter(), resStr, "", env.GetPromClusterLabel(), env.GetPromClusterLabel())
+	queryRAMRequests := fmt.Sprintf(queryRAMRequestsStr, env.GetPromClusterFilter(), resStr, "", env.GetPromClusterLabel())
```
AjayTripathy, line 1738: something weird happened here, think we should keep the change tightly scoped to GPU queries

pkg/costmodel/costmodel.go
```diff
@@ -161,9 +160,8 @@ const (
 	queryCPURequestsStr = `avg(
 		label_replace(
 			label_replace(
-				avg(
-					sum_over_time(kube_pod_container_resource_requests{resource="cpu", unit="core", container!="",container!="POD", node!="", %s}[%s] %s)
-				) by (namespace,container,pod,node,%s) , "container_name","$1","container","(.+)"
+				sum_over_time(kube_pod_container_resource_requests{resource="cpu", unit="core", container!="",container!="POD", node!="", %s}[%s] %s)
```
AjayTripathy, line 166: Same as above

pkg/costmodel/costmodel.go
```diff
@@ -181,14 +179,11 @@ const (
...
-					sum_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!="", %s}[%s] %s)
-					* %f
-				) by (namespace,container,pod,node,%s) , "container_name","$1","container","(.+)"
+				sum_over_time(kube_pod_container_resource_requests{resource="nvidia_com_gpu", container!="",container!="POD", node!="", %s}[%s] %s), 
+				"container_name","$1","container","(.+)"
 			), "pod_name","$1","pod","(.+)"
 		)
-	) by (namespace,container_name,pod_name,node,%s)
-	* on (pod_name, namespace, %s) group_left(container) label_replace(avg(avg_over_time(kube_pod_status_phase{phase="Running", %s}[%s] %s)) by (pod,namespace,%s), "pod_name","$1","pod","(.+)")`
+	) by (namespace,container_name,pod_name,node,%s)`
```
AjayTripathy, line 189: As above

AjayTripathy (changes requested):
It seems like the RAM and CPU queries changed here as well? Can we keep this one tightly scoped to GPU query changes? I am not sure if that was intentional or not-- this comment seems to indicate it was not? https://github.com/opencost/opencost/pull/2312#issuecomment-1815550850

pkg/costmodel/costmodel.go
r2k1: It's been intentional.

If you take a closer look at the queries.

```
avg(
  label_replace(
     avg (metric) by pod_name,
     pod_name, pod
  )
) by pod
```

Basically, it performs exactly the same aggregation twice. Once before renaming a label, and once after.

AjayTripathy (approved):

pkg/costmodel/costmodel.go
AjayTripathy: Ok, good catch thank you. @thomasvn since the core queries are changing would love if we took a close look on this in integration tests.

thomasvn: @AjayTripathy Yep! Our nightly integration test which validates a pod's allocation costs would catch any issues here.

PR #2313, opened 2023-11-17
Stop emitting unused metrics

### LINKED ISSUE #2290: Unused metrics

r2k1:
## What does this PR change?
Stop emitting metrics that are not queried by the opencost. Reduce resource consumption.

Hold of merging this PR until #2312 is resolved.
This PR removes `kube_pod_status_phase` which is still in use until the PR above is merged.
If PR is declined the `kube_pod_status_phase` should be removed from list of changes.

Some users could be impacted.

Closes #2290

## How was this PR tested?
Run locally with and without `EMIT_DEPRECATED_METRICS` env variable set. It toggles the specified metrics at `/metrics` endpoint.

I don't think the emitted metrics are described anywhere, but I'm not sure.


pkg/env/costmodelenv.go
```diff
@@ -42,6 +42,7 @@ const (
 
 	EmitPodAnnotationsMetricEnvVar       = "EMIT_POD_ANNOTATIONS_METRIC"
 	EmitNamespaceAnnotationsMetricEnvVar = "EMIT_NAMESPACE_ANNOTATIONS_METRIC"
+	EmitDeprecatedMetrics                = "EMIT_DEPRECATED_METRICS"
```
r2k1, line 46: Is it a good name? Any suggestions?

I assume there is more metrics that potentially can be deprecated in the future.
Reusing the same variable can be undesirable.

mattray: I'd prefer something like "EmitAdditionalMetrics" or "EmitNonDefaultMetrics" to indicate they're not used by OpenCost directly but they're available. "Deprecated" implies they'd be removed later, I don't think that's the intention, just disabling them for most users.

r2k1: I'm considering eventually removing them from the code, viewing it as a temporary safety measure in case of issues.

Opencost currently provides a limited selection of kube-state-metrics. I believe we should focus on including only those metrics essential for Opencost. For a more comprehensive set of metrics, users should directly utilize kube-state-metrics.

r2k1:
I disabled the metric scraping in Prometheus. And there is result:

Before:
```
max memory bytes
{container="opencost"} => 799576064 @[1701138475.918]
{container="prometheus-server"} => 951029760 @[1701138475.918]
cpu total seconds
{container="opencost"} => 303.54279 @[1701138476.656]
{container="prometheus-server"} => 118.409682 @[1701138476.656]
containers count
{} => 9721 @[1701138476.899]
node count
{} => 200 @[1701138477.246]

```

After:
```
max memory bytes
{container="opencost"} => 794161152 @[1701144307.661]
{container="prometheus-server"} => 730116096 @[1701144307.661]
cpu total seconds
{container="opencost"} => 208.483794 @[1701144307.956]
{container="prometheus-server"} => 45.930297 @[1701144307.956]
containers count
{} => 9724 @[1701144308.25]
node count
{} => 200 @[1701144308.548]
```

Note. CPU metric isn't accurate due to the different in the time window.

But the memory consumption went down from 950MB to 730MB.
Opencost memory remain unchanged at 800MB. But it's expected since the code changes hasn't been deployed. Only prometheus config to ignore the metric.

This total amount of series by metric name with the metric ignore:
image

AjayTripathy (approved):
Approving, but can we hold until https://github.com/opencost/opencost/pull/2312 is tested for regression?

PR #2317, opened 2023-11-17
Cloud Costs Kubernetesless

### LINKED ISSUE #2268: OpenCost without Kubernetes
With the introduction of [Cloud Costs](https://www.opencost.io/docs/#integrating-cloud-costs-and-configuration) there are users who may want to use OpenCost without tracking Kubernetes usage.

Deploying to Kubernetes may still be a requirement, but working without Prometheus could be supported.

mattray:
## What does this PR change?
* With the addition of the `KUBERNETES_ENABLED` and `CLOUD_COST_CONFIG_PATH` environment variables, you can now run OpenCost without Kubernetes (or Prometheus). The [/cloudCosts API](https://www.opencost.io/docs/integrations/api#cloudcost) is available, which means you could use this as a stand-alone API to your cloud provider billing.

no

* Users will be able to use the existing OCI image provided for Kubernetes deployments and simply pass new environment variables to it within Docker to run it wherever they want outside of Kubernetes. They could also build it on their workstation and point directly to the API.

`KUBERNETES_ENABLED=false CLOUD_COST_ENABLED=true CLOUD_COST_CONFIG_PATH='cloud-costs.json' ./costmodel`

and hit the API with

`curl -G http://localhost:9003/cloudCost -d window=7d -d aggregate=provider | jq`

* Closes https://github.com/opencost/opencost/issues/2268

## How was this PR tested?
* It was tested from the CLI and with Docker to ensure the `KUBERNETES_ENABLED=false` worked and on K3s and Kind to ensure that the Kubernetes functionality was unchanged. 
```
2023-11-17T15:28:06.366263035Z ??? Log level set to info
2023-11-17T15:28:06.366467786Z INF Starting cost-model version dev (HEAD)
2023-11-17T15:28:06.36652612Z INF Kubernetes enabled: true
...
```
* Unit tests were also added for the environment variable getters.

* It doesn't change the default behavior, but we'll want to document it as a potential new use model.

* Yes


AjayTripathy:
Super cool!

dwbrown2:
Do we need to each of the KUBERNETES_ENABLED and CLOUD_COST_ENABLED flags? Could OpenCost just dynamically enable/disable if the Kubernetes API is not present?

mattray:
@dwbrown2 `CLOUD_COST_ENABLED` was already built-in but has to be activated, so I kept that behavior for `KUBERNETES_ENABLED`. We probably could make either/both auto-detect the presence of credentials for cloud billing and if K8s is available. I could also see a use case where someone might only want Cloud Costs even if they were running on Kubernetes, so I feel there's value in keeping the environment variable.

mattray:
I can confirm that I've tested this with the current `develop` branch and this patch and the instructions in the [blog post](https://www.opencost.io/blog/k8sless-opencost) still work (substituting the pre-built containers with new ones built with this patch). The only change is now that https://github.com/opencost/opencost/pull/2366 is merged you run the following CLI for the UI
```
docker run -e API_SERVER=host.docker.internal -p 9090:9090 -t mattray/opencost-ui:dev01
```
note the API_SERVER environment variable.

Once this is merged and included in a release, we can add it to the documentation.

pkg/cloud/config/controller.go
```diff
@@ -47,8 +48,13 @@ type Controller struct {
 
 // NewController initializes an Config Controller
 func NewController(cp models.Provider) *Controller {
-	providerConfig := provider.ExtractConfigFromProviders(cp)
-	watchers := GetCloudBillingWatchers(providerConfig)
+	var watchers map[ConfigSource]cloud.KeyedConfigWatcher
+	if env.IsKubernetesEnabled() {
```
dwbrown2, line 52: Is this flag necessary? Could we not just display a "No K8s environment detected" in the UI/API if the K8s API isn't available?

mattray: It would require a substantial rewrite of the patch. In addition to needing to check for Kubernetes, the flag is used to short-circuit connecting to Prometheus and looking for the cloud-integration.json from the K8s config-file.

dwbrown2: Let's get more reviews and see what others think?

saweber: The ideal thing to do here would be to have a diagnostics page and endpoint, of which this would be one item, as well as the prometheus connection, cloud cost connections, etc, and that would be better suited to a separate issue.

saweber:
@mattray Have you looked into using these env vars to detect if we are in a k8s environment or not? If it could be done in the env.IsKubernetesEnabled() and keep a nice encapsulation, that would be ideal.
https://kubernetes.io/docs/tutorials/services/connect-applications-service/#environment-variables

mattray:
@saweber I've switched to detecting on KUBERNETES_PORT to determine if we're running on Kubernetes or not. It seems to work fine in my testing with Docker and K3s.

pkg/cloud/config/watcher.go
```diff
@@ -239,7 +239,13 @@ type MultiCloudWatcher struct {
 }
 
 func (mcw *MultiCloudWatcher) GetConfigs() []cloud.KeyedConfig {
-	multiConfigPath := path.Join(env.GetConfigPathWithDefault("/var/configs"), cloudIntegrationSecretPath)
+	var multiConfigPath string
+
+	if !env.IsKubernetesEnabled() {
```
saweber: super nit: flip this conditional for readability

saweber (approved):

mattray:
@saweber updated with your requested change. I'll merge if you're good

michaelmdresser (approved):
@mattray I'll approve in Steven's stead because I trust his review. CI is failing at the moment, though!

saweber:
LGTM, but please fix the compile error. :)

PR #2318, opened 2023-11-17
Sanitize ingested labels

cliffcolvin:
## What does this PR change?
* This PR sanitizes ingested labels that are stored prefixed by Prometheus with `label_`. It also stops ingesting labels that are Prometheus labels which aren't prefixed with `label_`.

* 

* Users will no longer need to prefix their queries by label with `label_` in front of their label key.
* Any old data would need to be re-ingested to have this change.

* Closes ...

## How was this PR tested?
* Unit test added for code coverage and accuracy.

* 

*

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1,6 +1,10 @@
 name: Develop PR - build test
 
 on:
+  push:
+    branches:
+      - develop
+
   pull_request:
     branches:
       - develop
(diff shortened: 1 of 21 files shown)


nikovacevic:
@cliffcolvin code looks good! Let me know when you've got testing notes, or if you need help with that, and I'll return for a definitive review.

ameijer (approved):
this LGTM. I reviewed the underlying implementation and this should accomplish exactly what we are looking for.

Cliff, it would be really easy to write a quick unit test here with some mock prom metrics just to ensure this behaves exactly the way you are looking for

PR #2329, opened 2023-11-22
SummaryAllocation support for serializing Properties and RawAllocationOnly fields

ghost:
## What does this PR change?
* Changes SummaryAllocation to no longer skip serializing Properties, and adds RawAllocationOnly struct, in order to support request sizing modals in KCM.

* Provides support for [Frontend](https://github.com/kubecost/kubecost-frontend/pull/549) for KCM

* It will let KC users using Kubecost Aggregator have fun little row modals

* Backend side of addressing [SELFHOST-948](https://kubecost.atlassian.net/browse/SELFHOST-948)

## How was this PR tested?
* Locally

* Don't believe so

* Nope, as github continues to not provide label options on opencost. I like to crack jokes about that sometimes.

### DIFF
--- a/pkg/kubecost/summaryallocation.go
+++ b/pkg/kubecost/summaryallocation.go
@@ -40,6 +40,10 @@ type SummaryAllocation struct {
 	ExternalCost           float64               `json:"externalCost"`
 	Share                  bool                  `json:"-"`
 	UnmountedPVCost        float64               `json:"-"`
+	// RawAllocationOnly added back in to the summary to support row modals.
+	// RawAllocationOnly is a pointer so if it is not present it will be
+	// marshalled as null rather than as an object with Go default values.
+	RawAllocationOnly *RawAllocationOnlyData `json:"rawAllocationOnly"`
 }
 
 // NewSummaryAllocation converts an Allocation to a SummaryAllocation by
@@ -69,6 +73,7 @@ func NewSummaryAllocation(alloc *Allocation, reconcile, reconcileNetwork bool) *
 		SharedCost:             alloc.SharedCost,
 		ExternalCost:           alloc.ExternalCost,
 		UnmountedPVCost:        alloc.UnmountedPVCost,
+		RawAllocationOnly:      alloc.RawAllocationOnly,
 	}
 
 	// Revert adjustments if reconciliation is off. If only network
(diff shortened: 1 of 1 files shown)


pkg/kubecost/summaryallocation.go
```diff
@@ -40,6 +40,10 @@ type SummaryAllocation struct {
 	ExternalCost           float64               `json:"externalCost"`
 	Share                  bool                  `json:"-"`
 	UnmountedPVCost        float64               `json:"-"`
+	// RawAllocationOnly added back in to the summary to support row modals.
+	// RawAllocationOnly is a pointer so if it is not present it will be
+	// marshalled as null rather than as an object with Go default values.
+	RawAllocationOnly *RawAllocationOnlyData `json:"rawAllocationOnly"`
```
michaelmdresser, line 46: `RawAllocationOnly` does not belong in `SummaryAllocation`, in my opinion.

RAO is only well-defined on unaccumulated, unaggregated Allocation data (let me know if an explanation would help) while `SummaryAllocation`'s main purpose is to serve accumulated and/or aggregated Allocation data with a small data footprint.

Why do you need this data here? If I'm remembering the feature request correctly, you should be able to get a per-request sizing-row max by calling the request sizing API with target utilization = 100%.

If you need per-container-instance (i.e. per pod/container tuple) max, please use the `/allocation` API with filters as it should already be able to serve that data and is designed for maximum-resolution information.

ghost: I got that request from James--it sounds like the row modal is doing a summary allocation request, and the missing items from the API are the RawAllocationOnly and Properties fields, for him to populate the row.

ghost: Given what you're saying about better API calls then, @jjarrett21 would it be better to swap to a different call for populating those modals, rather than adding the requisite information into the SummaryAllocations?

ghost: General FYI that I'm working with Austin on the front-end side, while James is on vacation, to see about pushing this through.

ghost: From conversation with Austin, it's sounding like there's an alternate path that wouldn't necessitate these changes to SummaryAllocation; if that's the case, then we can probably just close this out. I'm holding on confirmation

PR #2337, opened 2023-11-27
pkg/kubecost: fix clobbered errors

alrs:
This picks up two `err` variables that were being lost because of the use of `:=` inside a `for` loop.

### DIFF
--- a/pkg/kubecost/asset.go
+++ b/pkg/kubecost/asset.go
@@ -3841,6 +3841,7 @@ func (asr *AssetSetRange) InsertRange(that *AssetSetRange) error {
 	}
 
 	var err error
+	var as *AssetSet
 	for _, thatAS := range that.Assets {
 		if thatAS == nil || err != nil {
 			continue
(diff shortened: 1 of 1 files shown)


pkg/kubecost/asset.go
```diff
@@ -3852,7 +3853,7 @@ func (asr *AssetSetRange) InsertRange(that *AssetSetRange) error {
 			err = fmt.Errorf("cannot merge AssetSet into window that does not exist: %s", thatAS.Window.String())
 			continue
 		}
-		as, err := asr.Get(i)
+		as, err = asr.Get(i)
 		if err != nil {
 			err = fmt.Errorf("AssetSetRange index does not exist: %d", i)
```
michaelmdresser, line 3858: I think the right place to fix the clobbering is actually here, because this helpfully wraps extra context around the error.

pkg/kubecost/asset.go
```diff
@@ -3841,6 +3841,7 @@ func (asr *AssetSetRange) InsertRange(that *AssetSetRange) error {
 	}
 
 	var err error
+	var as *AssetSet
```
michaelmdresser, line 3844: Which makes this unnecessary.

alrs: The use of `:=` inside the `for` loop creates a new `err` variable that does not exist outside the loop where `err` is being returned.

michaelmdresser: You are correct. My apologies.

michaelmdresser (approved):
Thanks for the fix!

pkg/kubecost/asset.go
alrs: https://go.dev/play/p/JHopQtN4dnY

PR #2350, opened 2023-11-30
Cleaner logging in metrics.go

### LINKED ISSUE #2344: ERR Metric emission: failed to delete RAMAllocation with labels
**Describe the bug**
Getting few exceptions in the OpenCost Pod log.
```
2023-11-27T17:12:59.992546116Z ERR Metric emission: failed to delete RAMAllocation with labels: [default unmounted-pvs unmounted-pvs  ]
2023-11-27T17:12:59.992578074Z ERR Metric emission: failed to delete CPUAllocation with labels: [default unmounted-pvs unmounted-pvs  ] 
```

**To Reproduce**
1. Deploy OpenCost using [Getting started guide](https://www.opencost.io/docs/installation/install)
2. Observe the pod logs.

**Expected behavior**
No error should appear. 

**Screenshots**

Latest

**Additional context**
Kubernetes version - 1.27, AWS EKS.

wonko:
## What does this PR change?
* Rephrased some logging statements in metrics.go
* Made the `unmounted-pvs` name a constant and used it where needed
* Suppress deleting RAM/CPU/GPU metrics on the `unmounted-pvs` "pod"
* Changed a lot of the statements to Debug

* Better logging info, less non-actionable logs

* Closes https://github.com/opencost/opencost/issues/2344

## How was this PR tested?
* Local build, used image in my setup

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -733,7 +733,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 			for labelString, seen := range containerSeen {
 				if !seen {
 					labels := getLabelStringsFromKey(labelString)
-					if labels[1] == unmountedPVsContainer { // special "pod" to contain the mounted PVs - does not have RAM/CPU/...
+					if labels[1] == unmountedPVsContainer { // special "pod" to contain the unmounted PVs - does not have RAM/CPU/...
 						ok := cmme.RAMAllocationRecorder.DeleteLabelValues(labels...)
 						if !ok {
 							log.Debugf("Failed to delete RAMAllocation for container with labels: %v", labels)
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -421,7 +421,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 			var ok bool
 			defaultRegion, ok = util.GetRegion(nodeList[0].Labels)
 			if !ok {
-				log.DedupedWarningf(5, "Failed to locate default region")
+				log.DedupedWarningf(5, "Failed to read default region from labels on node %s", nodeList[0].Name)
 			}
 		}
 
@@ -471,7 +471,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 			// TODO: Pass CloudProvider into CostModel on instantiation so this isn't so awkward
 			nodes, err := cmme.Model.GetNodeCost(cmme.CloudProvider)
 			if err != nil {
-				log.Warnf("Metric emission: error getting Node cost: %s", err)
+				log.Warnf("Error getting Node cost: %s", err)
 			}
 			for nodeName, node := range nodes {
 				// Emit costs, guarding against NaN inputs for custom pricing.
@@ -535,24 +535,27 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 				// don't record cpuCost, ramCost, or gpuCost in the case of wild outliers
 				// k8s api sometimes causes cost spikes as described here:
 				// https://github.com/opencost/opencost/issues/927
-				if cpuCost < outlierFactor*avgCosts.CpuCostAverage {
+				cpuOutlierCutoff := outlierFactor * avgCosts.CpuCostAverage
+				if cpuCost < cpuOutlierCutoff {
 					cmme.CPUPriceRecorder.WithLabelValues(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID, node.ArchType).Set(cpuCost)
 					avgCosts.CpuCostAverage = (avgCosts.CpuCostAverage*avgCosts.NumCpuDataPoints + cpuCost) / (avgCosts.NumCpuDataPoints + 1)
 					avgCosts.NumCpuDataPoints += 1
 				} else {
-					log.Warnf("CPU cost outlier detected; skipping data point.")
+					log.Debugf("CPU cost outlier detected; skipping data point: %s had %f as cost, which is above %f.", nodeName, cpuCost, cpuOutlierCutoff)
 				}
-				if ramCost < outlierFactor*avgCosts.RamCostAverage {
+				ramOutlierCutoff := outlierFactor * avgCosts.RamCostAverage
+				if ramCost < ramOutlierCutoff {
 					cmme.RAMPriceRecorder.WithLabelValues(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID, node.ArchType).Set(ramCost)
 					avgCosts.RamCostAverage = (avgCosts.RamCostAverage*avgCosts.NumRamDataPoints + ramCost) / (avgCosts.NumRamDataPoints + 1)
 					avgCosts.NumRamDataPoints += 1
 				} else {
-					log.Warnf("RAM cost outlier detected; skipping data point.")
+					log.Debugf("RAM cost outlier detected; skipping data point: %s had %f as cost, which is above %f.", nodeName, ramCost, ramOutlierCutoff)
 				}
 				// skip redording totalCost if any constituent costs were outliers
-				if cpuCost < outlierFactor*avgCosts.CpuCostAverage &&
-					ramCost < outlierFactor*avgCosts.RamCostAverage {
+				if cpuCost < cpuOutlierCutoff && ramCost < ramOutlierCutoff {
 					cmme.NodeTotalPriceRecorder.WithLabelValues(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID, node.ArchType).Set(totalCost)
+				} else {
+					log.Debugf("CPU and RAM outlier detected, not recording node %s total cost %f", nodeName, totalCost)
 				}
 
 				nodeCostAverages[labelKey] = avgCosts
@@ -568,7 +571,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 			// TODO: Pass CloudProvider into CostModel on instantiation so this isn't so awkward
 			loadBalancers, err := cmme.Model.GetLBCost(cmme.CloudProvider)
 			if err != nil {
-				log.Warnf("Metric emission: error getting LoadBalancer cost: %s", err)
+				log.Warnf("Error getting LoadBalancer cost: %s", err)
 			}
 			for lbKey, lb := range loadBalancers {
 				// TODO: parse (if necessary) and calculate cost associated with loadBalancer based on dynamic cloud prices fetched into each lb struct on GetLBCost() call
@@ -644,7 +647,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 
 				parameters, ok := storageClassMap[pv.Spec.StorageClassName]
 				if !ok {
-					log.Debugf("Unable to find parameters for storage class \"%s\". Does pv \"%s\" have a storageClassName?", pv.Spec.StorageClassName, pv.Name)
+					log.Debugf("Unable to find parameters for storage class \"%s\". Pv \"%s\" might have an empty or invalid storageClassName.", pv.Spec.StorageClassName, pv.Name)
 				}
 				var region string
 				if r, ok := util.GetRegion(pv.Labels); ok {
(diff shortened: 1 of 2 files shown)


mattray:
@wonko thanks for the PR! This looks like pretty useful cleanup, I'll try to get one of the @opencost/opencost-maintainers to weigh in and I'll refer it to Kubecost Engineering.

pkg/costmodel/metrics.go
```diff
@@ -670,44 +673,44 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
+					log.Debugf("Removing metrics for %s, no data observed recently", labelString)
 					labels := getLabelStringsFromKey(labelString)
 
 					ok := cmme.NodeTotalPriceRecorder.DeleteLabelValues(labels...)
 					if ok {
-						log.Debugf("removed %s from totalprice", labelString)
+						log.Debugf("No data observed for node with labels %v, removed from totalprice", labels)
 					} else {
-						log.Errorf("FAILURE TO REMOVE %s from totalprice", labelString)
+						log.Debugf("No data observed for node with labels %v, but we could not remove it from totalprice", labels)
```
nikovacevic: This one needs to remain an `Errorf`, as its presence potentially indicates a major issue.

nikovacevic: If it's being changed for distasteful stylistic reasons (fair enough) then I'd recommend:
```go
log.Errorf("Failed to remove label set %v from metric node_total_hourly_cost.")
```

pkg/costmodel/metrics.go
```diff
@@ -670,44 +673,44 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
 					}
 					ok = cmme.NodeSpotRecorder.DeleteLabelValues(labels...)
 					if ok {
-						log.Debugf("removed %s from spot records", labelString)
+						log.Debugf("No data observed for node with labels %v, removed from spot records", labels)
 					} else {
-						log.Errorf("FAILURE TO REMOVE %s from spot records", labelString)
+						log.Debugf("No data observed for node with labels %v, but we could not remove it from spot records", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric kubecost_node_is_spot")
```

pkg/costmodel/metrics.go
nikovacevic: I would actually accept `Warnf` here, too, but probably nothing lower.

pkg/costmodel/metrics.go
```diff
@@ -670,44 +673,44 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
 					}
 					ok = cmme.CPUPriceRecorder.DeleteLabelValues(labels...)
 					if ok {
-						log.Debugf("removed %s from cpuprice", labelString)
+						log.Debugf("No data observed for node with labels %v, removed from cpuprice", labels)
 					} else {
-						log.Errorf("FAILURE TO REMOVE %s from cpuprice", labelString)
+						log.Debugf("No data observed for node with labels %v, but we could not remove it from cpuprice", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric node_cpu_hourly_cost")
```

pkg/costmodel/metrics.go
```diff
@@ -670,44 +673,44 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
 					}
 					ok = cmme.GPUPriceRecorder.DeleteLabelValues(labels...)
 					if ok {
-						log.Debugf("removed %s from gpuprice", labelString)
+						log.Debugf("No data observed for node with labels %v, removed from gpuprice", labels)
 					} else {
-						log.Errorf("FAILURE TO REMOVE %s from gpuprice", labelString)
+						log.Debugf("No data observed for node with labels %v, but we could not remove it from gpuprice", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric node_gpu_hourly_cost")
```

pkg/costmodel/metrics.go
```diff
@@ -670,44 +673,44 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
 					}
 					ok = cmme.GPUCountRecorder.DeleteLabelValues(labels...)
 					if ok {
-						log.Debugf("removed %s from gpucount", labelString)
+						log.Debugf("No data observed for node with labels %v, removed from gpucount", labels)
 					} else {
-						log.Errorf("FAILURE TO REMOVE %s from gpucount", labelString)
+						log.Debugf("No data observed for node with labels %v, but we could not remove it from gpucount", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric node_gpu_count")
```

pkg/costmodel/metrics.go
```diff
@@ -670,44 +673,44 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
 					}
 					ok = cmme.RAMPriceRecorder.DeleteLabelValues(labels...)
 					if ok {
-						log.Debugf("removed %s from ramprice", labelString)
+						log.Debugf("No data observed for node with labels %v, removed from ramprice", labels)
 					} else {
-						log.Errorf("FAILURE TO REMOVE %s from ramprice", labelString)
+						log.Debugf("No data observed for node with labels %v, but we could not remove it from ramprice", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric node_ram_hourly_cost")
```

pkg/costmodel/metrics.go
```diff
@@ -720,7 +723,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 					labels := getLabelStringsFromKey(labelString)
 					ok := cmme.LBCostRecorder.DeleteLabelValues(labels...)
 					if !ok {
-						log.Errorf("Metric emission: failed to delete LoadBalancer with labels: %v", labels)
+						log.Debugf("Failed to delete LoadBalancer with labels: %v", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric kubecost_load_balancer_cost")
```

pkg/costmodel/metrics.go
```diff
@@ -730,17 +733,21 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
-						log.Errorf("Metric emission: failed to delete RAMAllocation with labels: %v", labels)
-					}
-					ok = cmme.CPUAllocationRecorder.DeleteLabelValues(labels...)
-					if !ok {
-						log.Errorf("Metric emission: failed to delete CPUAllocation with labels: %v", labels)
-					}
-					ok = cmme.GPUAllocationRecorder.DeleteLabelValues(labels...)
-					if !ok {
-						log.Errorf("Metric emission: failed to delete GPUAllocation with labels: %v", labels)
+					if labels[1] == unmountedPVsContainer { // special "pod" to contain the unmounted PVs - does not have RAM/CPU/...
```
nikovacevic: Why are you proposing this conditional?

nikovacevic: Or perhaps this is meant to be `!=`, so that we stop attempting to delete the fake allocation data?

nikovacevic: As written, I would expect this to prevent deletion of real allocation label sets, but attempt to delete fake unmounted label sets.

pkg/costmodel/metrics.go
```diff
@@ -752,7 +759,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 					labels := getLabelStringsFromKey(labelString)
 					ok := cmme.PersistentVolumePriceRecorder.DeleteLabelValues(labels...)
 					if !ok {
-						log.Errorf("Metric emission: failed to delete PVPrice with labels: %v", labels)
+						log.Debugf("Failed to delete PVPrice for PV with labels: %v", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric pv_hourly_cost")
```

pkg/costmodel/metrics.go
```diff
@@ -764,7 +771,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 					labels := getLabelStringsFromKey(labelString)
 					ok := cmme.PVAllocationRecorder.DeleteLabelValues(labels...)
 					if !ok {
-						log.Errorf("Metric emission: failed to delete PVAllocation with labels: %v", labels)
+						log.Debugf("Failed to delete PVAllocation for PVC with labels: %v", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric pod_pvc_allocation")
```

pkg/costmodel/metrics.go
```diff
@@ -730,17 +733,21 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
-					if !ok {
-						log.Errorf("Metric emission: failed to delete CPUAllocation with labels: %v", labels)
-					}
-					ok = cmme.GPUAllocationRecorder.DeleteLabelValues(labels...)
-					if !ok {
-						log.Errorf("Metric emission: failed to delete GPUAllocation with labels: %v", labels)
+					if labels[1] == unmountedPVsContainer { // special "pod" to contain the unmounted PVs - does not have RAM/CPU/...
+						ok := cmme.RAMAllocationRecorder.DeleteLabelValues(labels...)
+						if !ok {
+							log.Debugf("Failed to delete RAMAllocation for container with labels: %v", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric container_memory_allocation_bytes")
```

pkg/costmodel/metrics.go
```diff
@@ -730,17 +733,21 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
+						}
+						ok = cmme.CPUAllocationRecorder.DeleteLabelValues(labels...)
+						if !ok {
+							log.Debugf("Failed to delete CPUAllocation for container with labels: %v", labels)
```
nikovacevic: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric container_cpu_allocation")
```

pkg/costmodel/metrics.go
```diff
@@ -730,17 +733,21 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
+						}
+						ok = cmme.GPUAllocationRecorder.DeleteLabelValues(labels...)
+						if !ok {
+							log.Debugf("Failed to delete GPUAllocation for container with labels: %v", labels)
+						}
+					} else {
+						log.Debugf("Did not try to delete RAM/CPU/GPU for fake '%s' container: %v", unmountedPVsContainer, labels)
```
nikovacevic, line 750: Same as above:
```go
log.Errorf("Failed to remove label set %v from metric container_gpu_allocation")
```

nikovacevic (changes requested):
The biggest issue here is definitely the conditional statement in `metrics.go` line 733. Please review that one.

pkg/costmodel/metrics.go
wonko: The "stylistic" reason is mainly to get rid of the capitals (makes the error look crazy important, but there's nothing the user can do) and the "metric emission: " in some places (the log lib gives the ability to add the component to the logs, i'd suggest to use that if you want to cleanly indicate which part of the codebase is emitting the loglines).  

As the log statement gives no hint of how to fix this as a user, it seemed informative at most for the developers, hence the `Debugf`. I've changed it to `Warnf`, but I'd strongly suggest to either make it `Infof` or give the user a hint of what to do with this warning (maybe add a refined wording of "costs might be inaccurate" to the warning).

This is exactly the reason I've made this PR: I got the same errors after installing opencost as the mentioned Pr, and had no idea if that was something that needed addressing. I had to get into the codebase to actually understand what was going on. There was no way that I as a user of the project could do anything about it.

(I fixed all other comments identically: updated the strings and changed to `Warnf`.)

pkg/costmodel/metrics.go
wonko: Totally my fault - tried to clean the code after validating, the statement got reversed wrongly. It is indeed a negated comparison I'd wanted there (`!=`). Changed it as it should be.

wonko (commented):
Thanks for the feedback, updated the PR accordingly!

pkg/costmodel/metrics.go
nikovacevic: I sympathize with your concern here. We share a desire to make the logs more relevant, useful, and actionable for users themselves. We have had an unfortunate incident in the past where these logs were not caught by the engineering team, resulting in the release of a nasty bug that affected accuracy, so we thought it prudent to make them extremely visible.

I support adding some verbiage like, "cost accuracy may be affected" -- perhaps
```go
log.Warnf("Failed to remove label set %v from metric node_total_hourly_cost. Failure to remove stale metrics may result in inaccurate data.")
```

The proper course of action upon receiving this log would be to reach out to the team by opening an issue.

That said, you're doing everyone a great service by eliminating the logging for unmounted PVs, which does seem to be extraneous and misleading. Thank you for that.

pkg/costmodel/metrics.go
```diff
@@ -730,17 +733,21 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
-						log.Errorf("Metric emission: failed to delete RAMAllocation with labels: %v", labels)
-					}
-					ok = cmme.CPUAllocationRecorder.DeleteLabelValues(labels...)
-					if !ok {
-						log.Errorf("Metric emission: failed to delete CPUAllocation with labels: %v", labels)
-					}
-					ok = cmme.GPUAllocationRecorder.DeleteLabelValues(labels...)
-					if !ok {
-						log.Errorf("Metric emission: failed to delete GPUAllocation with labels: %v", labels)
+					if labels[1] != unmountedPVsContainer { // special "pod" to contain the unmounted PVs - does not have RAM/CPU/...
```
nikovacevic: The last hyper-cautious change I'd like to recommend is a quick length check here. I expect that these are always of length 2 or greater, but it would be a shame to panic and take down the container if we find one that is not.

I.e. recommended change:
```go
if len(labels) >= 2 && labels[1] != unmountedPVsContainer {
```

nikovacevic (changes requested):
Just that one final tweak, and I will be happy to approve this. Thank you.

pkg/costmodel/metrics.go
```diff
@@ -720,7 +723,7 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 					labels := getLabelStringsFromKey(labelString)
 					ok := cmme.LBCostRecorder.DeleteLabelValues(labels...)
 					if !ok {
-						log.Errorf("Metric emission: failed to delete LoadBalancer with labels: %v", labels)
+						log.Warnf("Failed to remove label set %v from metric kubecost_load_balancer_cost: %v", labels)
```
nikovacevic: Ah, one more. It appears that we can remove the trailing `: %v` here, as there is only one variable being templated into the string.

nikovacevic (changes requested):

pkg/costmodel/metrics.go
wonko: done

pkg/costmodel/metrics.go
wonko: good catch, added it in

wonko:
- added the additional wording to the log messages
- added the `len` check on the conditional
- removed the left-over `%v`

nikovacevic (approved):

nikovacevic:
Thank you @wonko

lmello:
@mattray is this already in latest release ?

AjayTripathy:
This should now be on the latest release.

PR #2357, opened 2023-12-01
add profiling endpoints

ameijer:
## What does this PR change?
* adds /debug/ family of endpoints to allow opencost users to profile/tune memory usage

* none

* adds new endpoints. endpoints disabled by default

## How was this PR tested?
* ensured 404 when disabled, valid heap file returned when enabled on local run

* additional endpoints are added. we can use kubecost docs for debugging here (e.g., https://docs.kubecost.com/troubleshooting/app-memory) 

* 

cc @r2k1 @babbageclunk as per our discussion, should allow you to get heap dumps out of OC to see where the memory being used at scale

### DIFF
--- a/pkg/cmd/costmodel/costmodel.go
+++ b/pkg/cmd/costmodel/costmodel.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/http/pprof"
 	"time"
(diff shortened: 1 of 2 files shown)


cliffcolvin (approved):
LGTM! Thanks for this @ameijer

PR #2358, opened 2023-12-02
feat: allow adding prefix to prometheus query

alileza:
## What does this PR change?
* Adding support for configuring prefix for prometheus metric 

* nope

* there shouldn't be any behaviour changes unless the env var is set

## How was this PR tested?
* run it locally with both prometheus and opencost running

* probably? a new feature

*

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -17,54 +17,54 @@ const (
 	// upstream KSM has implementation change vs OC internal KSM - it sets metric to 0 when pod goes down
 	// VS OC implementation which stops emitting it
 	// by adding != 0 filter, we keep just the active times in the prom result
-	queryFmtPods                        = `avg(kube_pod_container_status_running{%s} != 0) by (pod, namespace, %s)[%s:%s]`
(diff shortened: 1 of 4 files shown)


AjayTripathy:
This feels like a pretty major change, can you share a little bit more about the motivation for this change?

alileza:
@AjayTripathy I wonder why you think it's major changes? there should be no effects unless your apply the env var. 

The issue simply in our orgs we segregate metrics per it's namespace, so we added `opencost_` prefix for all opencost /metrics endpoint. And it causing opencost didn't work as expected, because it still trying to query from those named metrics without `opencost_` prefix.

Also some of opencost expose metrics collide with kube-state-metrics, which this could help separate them as well

mattray:
Thanks for the PR, it is quite a substantial change so I'm looping in other @opencost/opencost-maintainers  for feedback

nikovacevic:
Seems like we first need to decide if this is a valid goal, or not. To inform that decision, we might explore whether this metric re-naming practice is a best practice, or something to be avoided in favor of keeping the original metric names. Who makes that call?

After we have that answer, I can take a pass and review whether or not this is implemented appropriately, which should be more-or-less straightforward (though admittedly very delicate).

PR #2363, opened 2023-12-04
Support for Azure Workload Identities

Sean-Holcomb:
## What does this PR change?
* Add support for using [DefaultAzureCredentials](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-defaultazurecredential) for azure blob authentication. This would remove the need for the creation of secrets in kubernetes.
* Start using the blob provider from the [Azure SDK for Go](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go). 
* Update the Azure Authorizer to be service agnostic like the Authorizers of the other providers.
replaces:  https://github.com/opencost/opencost/pull/2117

* Users will be able to configure Workload Identities to access CloudCosts

* Closes 

## How was this PR tested?
Image tested with Shared Key and Workload identity configurations for Azure Storage

* Yes it will

*

### DIFF
--- a/pkg/cloud/azure/storagebillingparser.go
+++ b/pkg/cloud/azure/storagebillingparser.go
@@ -37,7 +37,8 @@ func (asbp *AzureStorageBillingParser) ParseBillingData(start, end time.Time, re
 		return err
 	}
 
-	client, err := asbp.Authorizer.GetBlobClient(asbp.StorageConnection.getBlobURLTemplate())
+	serviceURL := fmt.Sprintf(asbp.StorageConnection.getBlobURLTemplate(), asbp.Account, "")
+	client, err := asbp.Authorizer.GetBlobClient(serviceURL)
 	if err != nil {
 		asbp.ConnectionStatus = cloud.FailedConnection
 		return err
--- a/pkg/cloud/azure/storagebillingparser.go
+++ b/pkg/cloud/azure/storagebillingparser.go
@@ -9,7 +9,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
 	"github.com/opencost/opencost/pkg/cloud"
 	"github.com/opencost/opencost/pkg/log"
 )
@@ -36,13 +37,13 @@ func (asbp *AzureStorageBillingParser) ParseBillingData(start, end time.Time, re
 		return err
 	}
 
-	containerURL, err := asbp.getContainer()
+	client, err := asbp.Authorizer.GetBlobClient(asbp.StorageConnection.getBlobURLTemplate())
 	if err != nil {
 		asbp.ConnectionStatus = cloud.FailedConnection
 		return err
 	}
 	ctx := context.Background()
-	blobNames, err := asbp.getMostRecentBlobs(start, end, containerURL, ctx)
+	blobNames, err := asbp.getMostRecentBlobs(start, end, client, ctx)
 	if err != nil {
 		asbp.ConnectionStatus = cloud.FailedConnection
 		return err
@@ -54,7 +55,7 @@ func (asbp *AzureStorageBillingParser) ParseBillingData(start, end time.Time, re
 	}
 
 	for _, blobName := range blobNames {
-		blobBytes, err2 := asbp.DownloadBlob(blobName, containerURL, ctx)
+		blobBytes, err2 := asbp.DownloadBlob(blobName, client, ctx)
 		if err2 != nil {
 			asbp.ConnectionStatus = cloud.FailedConnection
 			return err2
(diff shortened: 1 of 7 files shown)


nik-kc (approved):
Very nice! Love the update to Azure SDK for Go. Could this be extended to entirely replace `azure-storage-blob-go` throughout OC?

And someone better acquainted with CloudCosts can chime in regarding the CC side of this PR, but overall this looks great.

pkg/cloud/azure/storagebillingparser.go
```diff
@@ -101,32 +103,32 @@ func (asbp *AzureStorageBillingParser) parseCSV(start, end time.Time, reader *cs
...
-		// ListBlobs returns the start of the next segment; you MUST use this to get
-		// the next segment (after processing the current result segment).
-		marker = listBlob.NextMarker
-
 		// Using the list of months strings find the most resent blob for each month in the range
-		for _, blobInfo := range listBlob.Segment.BlobItems {
+		for _, blobInfo := range resp.Segment.BlobItems {
 			for _, month := range monthStrs {
-				if strings.Contains(blobInfo.Name, month) {
+				if strings.Contains(*blobInfo.Name, month) {
```
nikovacevic, line 136: Any chance we panic on these derefs?

pkg/cloud/azure/storageconnection.go
```diff
@@ -65,20 +44,25 @@ func (sc *StorageConnection) getBlobURLTemplate() string {
...
 	// read the body into a buffer
 	downloadedData := bytes.Buffer{}
-	_, err = downloadedData.ReadFrom(bodyStream)
+
+	_, err = downloadedData.ReadFrom(retryReader)
+	if err != nil {
+		return nil, err
+	}
+
+	err = retryReader.Close()
```
nikovacevic: Do we need to close the retry reader before we return the `ReadFrom` error?

nikovacevic (approved):
Couple of small questions, but otherwise looks good

pkg/cloud/azure/storageconnection.go
Sean-Holcomb: moved this to a 'defer' statement

pkg/cloud/azure/storagebillingparser.go
Sean-Holcomb: possible, did a little clean up around here

Sean-Holcomb:
@nik-kc that would be great, but will require a little more effort with removing some of the thanos code in package storage

chrisjohnson00:
Is there any place that this PR is added to the docs?
I don't see any mention here https://www.opencost.io/docs/configuration/azure about how to use a workload identity with OpenCost.
My normal way of doing it isn't working.  

IE:
- create an identity, assign the OpenCost role
- Create OpenCost service account, associate it with Federated Identity Credentials for my AKS cluster
- add the `azure.workload.identity/use=true` label to the deployment/pod

This results in the expected env vars being injected as well as the `azure-identity-token` mounted.

```
AZURE_CLIENT_ID:                       XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
AZURE_TENANT_ID:                       XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
AZURE_FEDERATED_TOKEN_FILE:            /var/run/secrets/azure/tokens/azure-identity-token
AZURE_AUTHORITY_HOST:                  https://login.microsoftonline.com/
```

Results in:

```
opencost 2024-02-08T00:54:03.805410641Z INF Failed to download pricing data: azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions//providers/Microsoft.Commerce/RateCard?%!f(MISSING)ilter=OfferDurableId+eq+%!M(MISSING)S-AZR-0003p%!+(MISSING)and+Currency+eq+%!U(MISSING)SD%!+(MISSING)and+Locale+eq+%!e(MISSING)n-US%!+(MISSING)and+RegionInfo+eq+%!U(MISSING)S%!&(MISSING)api-version=2015-06-01-preview: StatusCode=400 -- Original Error: adal: Refresh request failed. Status Code = '400'. Response body: {"error":"invalid_request","error_description":"Identity not found"} Endpoint http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&client_id=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX&resource=https%!A(MISSING)%!F(MISSING)%!F(MISSING)management.azure.com%!F(MISSING) 
```

Sean-Holcomb:
@chrisjohnson00 are you trying to set up CloudCost or RateCard?

chrisjohnson00:
> @chrisjohnson00 are you trying to set up CloudCost or RateCard?

@Sean-Holcomb 

Honestly, I don't know the difference between the two... there isn't much detail in the docs about what each is and when to use one vs the other.

I think RateCard, essentially I want to replace the default cost data with data from Azure.  I'm only interested in getting costs of the Kubernetes cluster itself, so I don't think I need CloudCost?

Sean-Holcomb:
Yes it does sound like you are looking for rate card. Unfortunately this PR is specifically for Cloud Cost integration. The RateCard Api has an entirely different configuration pathway and I am not sure that there is a way to run it when running OpenCost from the helm chart. There is also currently no plan to move the older "Pricing Source" integrations to the newer configuration method.

chrisjohnson00:
Thanks!

cmergenthaler:
@Sean-Holcomb So when using CloudCost we can basically ignore the `Failed to download pricing data` RateCard API error?

AjayTripathy:
> @Sean-Holcomb So when using CloudCost we can basically ignore the `Failed to download pricing data` RateCard API error?

That's fine yes.

PR #2365, opened 2023-12-05
feat: add export trigger endpoint

alileza:
## What does this PR change?
* It adds a new /export API to immediately trigger CSV export

nope

* This will help new users test CSV export feature, checking integration & format, rather than waiting for the `00:10`

nope

## How was this PR tested?
deploy on dev env, `curl -X POST [pod]:9003/export` and see the CSV export logs 

* Would be a nice addition to CSV export

*

### DIFF
--- a/pkg/cmd/costmodel/costmodel.go
+++ b/pkg/cmd/costmodel/costmodel.go
@@ -8,10 +8,10 @@ import (
 	"time"
 
 	"github.com/julienschmidt/httprouter"
-	"github.com/opencost/opencost/pkg/cloudcost"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/cors"
 
+	"github.com/opencost/opencost/pkg/cloudcost"
 	"github.com/opencost/opencost/pkg/costmodel"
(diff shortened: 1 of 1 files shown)


mattray:
@alileza this looks useful, thanks! I'll let @r2k1 comment on it since he provided the original CSV export functionality.

r2k1:
I'm keen to know about the specific use-case you're addressing with this approach.

I think current export logic initiates data export at the application's start. Which, in my view, should adequately address the issue of feature testing.

On a side note. It would be nice to have an endpoint that returns the file in the response.

alileza:
@r2k1 in my usecase its simply for testing purpose, since there's no way verify the CSV was uploaded to the the correct source (in my case S3), with which format, I'd have to wait for a day to gone by, just to see that there's an error on CSV processing.

r2k1:
Changing configuration would require a pod restart. And a pod restart would trigger the export.
I didn't need to wait a day for testing. I would just restart the opencost pod.

Was it caused by something outside of opencost, like permissions misconfiguration in s3?

PR #2367, opened 2023-12-05
Oracle Cloud Infrastructure (OCI) Provider

### LINKED ISSUE #2221: Oracle Cloud Infrastructure (OCI) Provider implementation
Create a new OCI provider implementation, so OpenCost may be used to estimate costs for workloads that run in OCI. This will benefit OCI customers or hybrid cloud users of OCI, and increase usage of OpenCost.

anders-swanson:
Oracle Cloud Infrastructure (OCI) provider
---------

## What does this PR change?
* Adds a new cloud service provider implementation for OCI.

* N/A

* Users running clusters on OCI will be able to access pricing data.

* Closes #2221 

## How was this PR tested?
* Tested on OCI with OKE and self-managed clusters.

* An informational section for the OCI provider should be added. Note that the OCI provider requires no additional configuration to function. 
https://github.com/opencost/opencost-website/pull/178


pkg/cloud/oracle/partnumbers/shape_part_numbers.json
mattray, line 1: Is this going to be updated frequently? Is there somewhere we could retrieve this if we needed to recreate/update it?

pkg/cloud/oracle/product.go
mattray, line 1: Perhaps we could comment in here the source of the part numbers and how we're going to maintain them?

pkg/cloud/oracle/provider.go
```diff
@@ -0,0 +1,309 @@
...
+	if err := o.ensurePricingData(); err != nil {
+		return "", 0.0, err
+	}
+	managementPlatform, _ := o.GetManagementPlatform()
+	if managementPlatform != managementPlatformOKE {
+		return "", 0.0, nil // Self-managed cluster.
+	}
+	o.DownloadPricingDataLock.Lock()
+	defer o.DownloadPricingDataLock.Unlock()
+	// TODO: Support lookup of cluster type, as BASIC_CLUSTER types are free.
```
mattray, line 245: Is this worth calling out in documentation?

pkg/cloud/oracle/region.go
```diff
@@ -0,0 +1,39 @@
+package oracle
+
+var oracleRegions = []string{
```
mattray, line 5: source of the list?

mattray (commented):
A few questions related to maintainability going forward, but it looks good to me.

mattray:
I'm not very familiar with adding a provider, @AjayTripathy or @Sean-Holcomb, could either of you review this please?

pkg/cloud/oracle/provider.go
anders-swanson: Yes, it is.

pkg/cloud/oracle/partnumbers/shape_part_numbers.json
anders-swanson: It's compiled from the cost estimator APIs, which are subject to change and could not be directly queried. When new shapes are added, this data may need to be recompiled.

mattray: Could we document recompiling that data? I just worry that this will go stale and we won't know how to fix it.

configs/oracle.json
```diff
@@ -0,0 +1,10 @@
+{
+  "provider":"Oracle",
+  "CPU": "0.015",
+  "RAM": "0.002",
+  "GPU": "2.00",
+  "storage": "0.00005479452",
+  "defaultLBPrice": "0.0113",
+  "internetNetworkEgress": "0.0085",
+  "oraclePricingEndpoint":"https://apexapps.oracle.com/pls/apex/cetools/api/v1/products"
```
Sean-Holcomb: Does this need to be configurable? When would you want to change it?

anders-swanson: Couple situations:
- You have custom pricing data you want to serve
- Your cluster is disconnected from the internet, or otherwise doesn't have access to this URI

Sean-Holcomb: Does Oracle provide an endpoint with custom pricing in the same format or would the user set it up themselves, If it is Oracle provided is there authorization associated or would the expectation be that they are using a kind of workload identity auth?

For the air gapped example, is the thought that they would have an alternate endpoint available on a vpn or it would just be disable?

As an explanation, I am hammering on this point because that config file is a bit of a mess and I don't want to increase dependencies on it. In the short term I would adding that URL as a const

anders-swanson: 1. Oracle has custom pricing, which requires authentication to access, as it is from your tenancy. Until this integration is done, custom pricing would have to be from a user-supplied URI.
2. For scenarios where the default URI is not accessible, users could provide their own endpoint.

Couple thoughts on the config file:
If the config file is too large, what would another means of providing external configurations?
OR, If there's another parameter we can reuse, that might also work.

mattray:
@anders-swanson OpenCost 1.108 is shipping with [Cloud Costs](https://www.opencost.io/blog/cloud-costs), do you want (in a separate PR) to add Oracle support for this too? 

Once @Sean-Holcomb or @AjayTripathy or one of the other @opencost/opencost-maintainers signs off on a review we'll merge for the next release

configs/oracle.json
AjayTripathy: @anders-swanson sorry for the slow responses here, we've been on a holiday schedule:) 

Many providers have huge config files-- we tokenize them and scan them to extract pricing for just the resources you're running. We should definitely not block this PR on that, though. Would be a great todo.

AjayTripathy: @Sean-Holcomb could we compromise by making this an environment variable? It will definitely be needed if the end user is for example airgapped.

AjayTripathy:
This basically LGTM @anders-swanson -- just see my note about swapping the configuration of the pricing information to an environment variable

AjayTripathy (approved):

configs/oracle.json
anders-swanson: Hi Ajay,
I can move this to an envrionment variable. Is there any concern with overloading environment variables with config values?

anders-swanson:
> @anders-swanson OpenCost 1.108 is shipping with [Cloud Costs](https://www.opencost.io/blog/cloud-costs), do you want (in a separate PR) to add Oracle support for this too?
> 

I may be able to integrate cloud costs in a follow-up PR.

AjayTripathy:
> > @anders-swanson OpenCost 1.108 is shipping with [Cloud Costs](https://www.opencost.io/blog/cloud-costs), do you want (in a separate PR) to add Oracle support for this too?
> 
> I may be able to integrate cloud costs in a follow-up PR.

Yep, happy to not block on that!

configs/oracle.json
Sean-Holcomb: overloaded env variables with default is preferable to this config file having additional fields added to it.

anders-swanson: Sounds good! It may be some time until I can implement the update due to holiday break, but know that it will be done.

AjayTripathy: hey @anders-swanson checking back in. Still on your radar? I'm hopeful this should be really quick!

anders-swanson: Hey, sorry for the delay, I've just returned from winter holidays. I plan to push out these changes tomorrow 🤞

mattray:
@anders-swanson you mentioned in Slack you're still working on getting approval on this. In the meantime could you get the DCO added to your commits?

anders-swanson:
> @anders-swanson you mentioned in Slack you're still working on getting approval on this. In the meantime could you get the DCO added to your commits?

Rebased for DCO, thanks!

pkg/cloud/oracle/provider.go
```diff
@@ -0,0 +1,314 @@
...
+// DownloadPricingData refreshes the RateCardStore pricing data.
+func (o *Oracle) DownloadPricingData() error {
+	o.DownloadPricingDataLock.Lock()
+	defer o.DownloadPricingDataLock.Unlock()
+	cfg, err := o.GetConfig()
+	if err != nil {
+		return err
+	}
+	if o.RateCardStore == nil {
+		url := os.Getenv("OCI_PRICING_URL")
```
AjayTripathy: Could you instead do this in https://github.com/opencost/opencost/blob/develop/pkg/env/costmodelenv.go and supply a default for pricing data that would work for most users?

AjayTripathy (changes requested):

AjayTripathy:
One quick note @anders-swanson then we should be good.

AjayTripathy (approved):

PR #2371, opened 2023-12-06
Add labels to cloud cost view table rows

Sean-Holcomb:
## What does this PR change?
* Adds labels to the cloudcost.ViewTableRow struct so that they can be displayed in pills on the CloudCost page.

* Update the response from `QueryViewTotals` on the cloudcost.ViewQuerier interface to a match the type it is converted to for response

* 

* Users will be able to see labels in CloudCost in the Cloud Cost view table

* Closes ...

## How was this PR tested?
* Tested in local deploys

* 

*

### DIFF
--- a/pkg/cloudcost/querier.go
+++ b/pkg/cloudcost/querier.go
@@ -29,7 +29,7 @@ const DefaultChartItemsLength int = 10
 // ViewQuerier defines a contract for return View types to the QueryService to service the View Api
 type ViewQuerier interface {
 	QueryViewGraph(ViewQueryRequest, context.Context) (ViewGraphData, error)
-	QueryViewTotals(ViewQueryRequest, context.Context) (*ViewTableRow, int, error)
+	QueryViewTotals(ViewQueryRequest, context.Context) (*ViewTotals, error)
 	QueryViewTable(ViewQueryRequest, context.Context) (ViewTableRows, error)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/cloudcost/querier.go
```diff
@@ -29,7 +29,7 @@ const DefaultChartItemsLength int = 10
 // ViewQuerier defines a contract for return View types to the QueryService to service the View Api
 type ViewQuerier interface {
 	QueryViewGraph(ViewQueryRequest, context.Context) (ViewGraphData, error)
-	QueryViewTotals(ViewQueryRequest, context.Context) (*ViewTableRow, int, error)
+	QueryViewTotals(ViewQueryRequest, context.Context) (*ViewTotals, error)
```
michaelmdresser, line 32: Nice

pkg/cloudcost/repositoryquerier.go
```diff
@@ -168,8 +171,14 @@ func (rq *RepositoryQuerier) QueryViewTable(request ViewQueryRequest, ctx contex
 		if err2 != nil {
 			return nil, fmt.Errorf("QueryViewTable: failed to retrieve cost metric: %w", err)
 		}
+		var labels map[string]string
+		if cloudCost.Properties != nil {
+			labels = cloudCost.Properties.Labels
+		}
```
michaelmdresser, line 177: Should `cloudCost.Properties.Labels` be full copied here instead of just pointer copied?

michaelmdresser (approved):

pkg/cloudcost/repositoryquerier.go
Sean-Holcomb: It is being pulled for binary for this purpose only so a full copy just seemed like an unnecessary memory bump

PR #2366, opened 2023-12-05
Expose and make the API and UI ports configurable 

mattray:
## What does this PR change?
* The API port and server and the UI port are all now configurable. The API port can be moved off of 9003 and the UI can be moved off of 9090 through environment variables. If you need to modify the API server in the `default.nginx.conf` you can set the `API_SERVER`.

## Does this PR relate to any other PRs?
* Continues the work started in https://github.com/opencost/opencost/pull/2283
* Enabled by https://github.com/opencost/opencost-helm-chart/pull/156

## How will this PR impact users?
* Ports are now configurable

## Does this PR address any GitHub or Zendesk issues?
* Closes https://github.com/opencost/opencost/issues/2223
* https://github.com/opencost/opencost/pull/2317 needs this to work

## How was this PR tested?
* Tested as a standard build with an unaware Helm chart
* Tested with the updated Helm chart with no changes
* Tested with the updated Helm chart with changed API and UI ports
* Verified with port-forwarding 9091 and 9004
* Test added to the code for the API_PORT

## Does this PR require changes to documentation?
* It will be part of the "running with Docker" documentation to be added eventually

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next OpenCost release? If not, why not?
* Yes, if we want to be able to ship the Docker version of OpenCost

PR #2373, opened 2023-12-08
update Dockerfile.cross with new default.nginx.conf.template

cliffcolvin:
## What does this PR change?
* adds the proper filename for nginx.conf.template to dockerfile.cross to fix issue with cross platform builds

* Fixes issue created from #2366 

* Allow users to use Dockerfile.cross to build cross platform images

* Closes ...

## How was this PR tested?
* Locally with docker buildx

* No

* 1.108

### DIFF
--- a/ui/Dockerfile.cross
+++ b/ui/Dockerfile.cross
@@ -5,7 +5,7 @@ ENV API_SERVER=0.0.0.0
 ENV UI_PORT=9090
 
 COPY ./dist /var/www
-COPY default.nginx.conf /etc/nginx/conf.d/
+COPY default.nginx.conf.template /etc/nginx/conf.d/default.nginx.conf.template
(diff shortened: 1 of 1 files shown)


nikovacevic (approved):
Can confirm that the file in question has that name locally. Do we want the name to still have `.template` at the destination location of the `COPY`? It likely doesn't matter at all, but it seems a bit odd.

cliffcolvin:
> Can confirm that the file in question has that name locally. Do we want the name to still have `.template` at the destination location of the `COPY`? It likely doesn't matter at all, but it seems a bit odd.

This was added from an external commit, and Matt approved it. I'll discuss with him later today to see. Just trying to stop breaking the build at this moment though. -- Making the change work, and can discuss if the change should stay this afternoon and understand the context around it.

PR #2376, opened 2023-12-08
aggregate by all if aggregate is empty

nickcurie:
## What does this PR change?
* 

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/cloudcost/queryservice_helper.go
+++ b/pkg/cloudcost/queryservice_helper.go
@@ -39,7 +39,7 @@ func ParseCloudCostRequest(qp httputil.QueryParams) (*QueryRequest, error) {
 	}
 
 	// if we're aggregating by nothing (aka `item` on the frontend) then aggregate by all
-	if aggregateBy != nil && len(aggregateBy) < 1 {
+	if len(aggregateBy) == 0 {
 		aggregateBy = []string{kubecost.CloudCostInvoiceEntityIDProp, kubecost.CloudCostAccountIDProp, kubecost.CloudCostProviderProp, kubecost.CloudCostProviderIDProp, kubecost.CloudCostCategoryProp, kubecost.CloudCostServiceProp}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cloudcost/queryservice_helper.go
```diff
@@ -38,6 +38,11 @@ func ParseCloudCostRequest(qp httputil.QueryParams) (*QueryRequest, error) {
 		aggregateBy = append(aggregateBy, prop)
 	}
 
+	// if we're aggregating by nothing (aka `item` on the frontend) then aggregate by all
+	if aggregateBy != nil && len(aggregateBy) < 1 {
```
michaelmdresser: `GetList()` returns `nil` if the string is empty https://github.com/opencost/opencost/blob/develop/pkg/util/mapper/mapper.go#L440

```suggestion
	if len(aggregateBy) == 0 {
```

should work

nickcurie: This fails the tests

nickcurie: `aggregateBy != nil` was what fixed it but probably caused the issue to not get solved

michaelmdresser: I think we should change the test; behavior isn't working as its supposed to so we're fixing a bug. I'd say that the unit test is wrong.

nikovacevic: So, we probably need to do the "all" thing and treat both `nil` and `{}` as "do not aggregate." Which might mean that we need a test change? Does that sound right?

nickcurie: ok let me nuke the test for now just to make sure this solves the problem

nickcurie: I'm not sure I'm clear on what "do not aggregate" means. Would that just return one line item or something?

michaelmdresser: By "do not aggregate" I believe niko means "return maximum-resolution lines", the same as a "full group by"

nickcurie: So what's `all` then? I thought that would be full group by

michaelmdresser: I believe "all" is supposed to produce a single, fully-totalled, line

michaelmdresser (approved):

PR #2385, opened 2023-12-11
Log and continue on a single record processing failures.

r2k1:
## What does this PR change?

"I'm experiencing an issue with multiple AKS clusters where opencost is returning a 500 error. The error message is `Error: unable to locate allocation totals for allocation,`. This message might be truncated and could contain more details, but I don't have access to that information (or any logs)

The current processing logic seems to halt opencost entirely when a single record fails. The proposed change is to ignore such failing records and allow processing to continue.

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -2555,8 +2555,7 @@ func (cm *CostModel) QueryAllocation(window kubecost.Window, resolution, step ti
 
 					if totals == nil {
 						log.Errorf("unable to locate asset totals for allocation %s", key)
-						return nil, fmt.Errorf("unable to locate allocation totals for allocation")
-
+						continue
 					}
 
 					parc.CPUTotalCost = totals.CPUCost
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -2554,7 +2554,7 @@ func (cm *CostModel) QueryAllocation(window kubecost.Window, resolution, step ti
 					}
 
 					if totals == nil {
-						log.Errorf("unable to locate asset totals for allocation %s", key)
+						log.Errorf("unable to locate asset totals for allocation %s, corresponding PARC is being skipped", key)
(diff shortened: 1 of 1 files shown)


r2k1:
@ameijer how do you think, is it something that is going to work?

I can't imagine how it can make things worse. But is it something that may help or just raises an error somewhere further in the processing?

Unfortunately, I can't replicate the issue. But I see no harm in this attempt to fix it.

pkg/costmodel/costmodel.go
```diff
@@ -2555,8 +2555,7 @@ func (cm *CostModel) QueryAllocation(window kubecost.Window, resolution, step ti
 
 					if totals == nil {
 						log.Errorf("unable to locate asset totals for allocation %s", key)
-						return nil, fmt.Errorf("unable to locate allocation totals for allocation")
-
+						continue
```
ameijer, line 2558: I think this is good @r2k1 . For others: this changes the behavior of a PARC for which we cannot locate totals for to simply not be included, rather than bubbling up an error. 

This doesn't fix whatever the underlying issue is with no totals in the store, but this at least lets us degrade gracefully

pkg/costmodel/costmodel.go
```diff
@@ -2555,8 +2555,7 @@ func (cm *CostModel) QueryAllocation(window kubecost.Window, resolution, step ti
 
 					if totals == nil {
 						log.Errorf("unable to locate asset totals for allocation %s", key)
```
ameijer:
```suggestion
						log.Errorf("unable to locate asset totals for allocation %s, corresponding PARC is being skipped", key)
```

ameijer:
happy to approve @r2k1 if you agree with my idea about making the error message a little clearer that we are going to skip the PARC return

ameijer (approved):
LGTM @r2k1 , I will work on getting this merged for you

cliffcolvin (approved):
LGTM

Issue #2243, opened 2023-10-26
node_total_hourly_cost inaccurate for some nodes in AWS


liorfranko:
**Describe the bug**
We're running opencost with default configurations, without customPricing, without CUR integration, and without spot-data-feed.
We need the prices based on AWS public prices.

I noticed that the metric node_total_hourly_cost is inaccurate for some nodes.
I compared two c5d.18xlarge nodes in zone us-east-1e in Virginia.
<img width="937" alt="Screen Shot 2023-10-26 at 11 31 04" src="https://github.com/opencost/opencost/assets/20726865/7b3125ee-47f6-4684-87a0-eafd04eed596">

Inaccurate node:
<img width="1660" alt="Screen Shot 2023-10-26 at 11 18 06" src="https://github.com/opencost/opencost/assets/20726865/a4af38ce-d60a-4029-a77b-ad4b1da35833">
```
kubectl describe nodes ip-10-207-136-27.ec2.internal
Name:               ip-10-207-136-27.ec2.internal
Roles:              <none>
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=c5d.18xlarge
                    beta.kubernetes.io/os=linux
                    failure-domain.beta.kubernetes.io/region=us-east-1
                    failure-domain.beta.kubernetes.io/zone=us-east-1e
                    intent=apps
                    k8s.io/cloud-provider-aws=7e43c8d478ffdd4cb3168691ff335629
                    karpenter.k8s.aws/instance-category=c
                    karpenter.k8s.aws/instance-cpu=72
                    karpenter.k8s.aws/instance-encryption-in-transit-supported=false
                    karpenter.k8s.aws/instance-family=c5d
                    karpenter.k8s.aws/instance-generation=5
                    karpenter.k8s.aws/instance-hypervisor=nitro
                    karpenter.k8s.aws/instance-local-nvme=1800
                    karpenter.k8s.aws/instance-memory=147456
                    karpenter.k8s.aws/instance-network-bandwidth=25000
                    karpenter.k8s.aws/instance-pods=737
                    karpenter.k8s.aws/instance-size=18xlarge
                    karpenter.sh/capacity-type=spot
                    karpenter.sh/initialized=true
                    karpenter.sh/provisioner-name=ds-infra-network-online-spot
                    karpenter.sh/registered=true
                    kubernetes.io/arch=amd64
[...]

PR #2386, opened 2023-12-12
Fix aws spot price with default config

sachin-rafay:
## What does this PR change?
* Currently, when user does not provide spot data feed but uses spot instances, the code picks hardcoded spot cpu cost. The solution here is to pick the public pricing for spot instances with default config (no spot data feed configured).

* I hope not.

* When user does not provide spot data feed but uses opencost on spot instances, public pricing will be picked.

* Closes ... https://github.com/opencost/opencost/issues/2243

## How was this PR tested?
* The changes were tested on AWS spot instances

* I hope not. When the user does not configure spot data feed, opencost should be expected to provide public pricing.

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -1321,22 +1321,34 @@ func (aws *AWS) createNode(terms *AWSProductTerms, usageType string, k models.Ke
 		}, meta, nil
 	} else if aws.isPreemptible(key) { // Preemptible but we don't have any data in the pricing report.
 		log.DedupedWarningf(5, "Node %s marked preemptible but we have no data in spot feed", k.ID())
-		// Throw error if public price is not found
-		if !publicPricingFound {
-			log.Errorf("Could not fetch public pricing data for spot instance \"%s\"", k.ID())
-			return nil, meta, fmt.Errorf("Could not fetch data for \"%s\"", k.ID())
+		if publicPricingFound {
+			// return public price if found
+			return &models.Node{
+				Cost:         cost,
+				VCPU:         terms.VCpu,
+				RAM:          terms.Memory,
+				GPU:          terms.GPU,
+				Storage:      terms.Storage,
+				BaseCPUPrice: aws.BaseCPUPrice,
+				BaseRAMPrice: aws.BaseRAMPrice,
+				BaseGPUPrice: aws.BaseGPUPrice,
+				UsageType:    PreemptibleType,
+			}, meta, nil
+		} else {
+			// return defaults if public pricing not found
+			return &models.Node{
+				VCPU:         terms.VCpu,
+				VCPUCost:     aws.BaseSpotCPUPrice,
+				RAMCost:      aws.BaseSpotRAMPrice,
+				RAM:          terms.Memory,
+				GPU:          terms.GPU,
+				Storage:      terms.Storage,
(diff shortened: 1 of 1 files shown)


pkg/cloud/aws/provider.go
```diff
@@ -1307,9 +1321,14 @@ func (aws *AWS) createNode(terms *AWSProductTerms, usageType string, k models.Ke
...
+			log.Errorf("Could not fetch public pricing data for spot instance \"%s\"", k.ID())
+			return nil, meta, fmt.Errorf("Could not fetch data for \"%s\"", k.ID())
+		}
 		return &models.Node{
+			Cost:         cost,
 			VCPU:         terms.VCpu,
-			VCPUCost:     aws.BaseSpotCPUPrice,
 			RAM:          terms.Memory,
 			GPU:          terms.GPU,
 			Storage:      terms.Storage,
```
thomasvn: In the case that a Spot Data Feed isn't available, I'm unsure if we should be creating a dependency on the AWS OnDemand Pricing Sheet?

thomasvn: Or maybe in the case that `!publicPricingFound`, we can set use defaults here instead of throwing an error?

sachin-rafay: Thanks for the review @thomasvn ! I have made the change to return defaults when public price is not available.

I have made one change while returning defaults though: I am returning both vCPU and RAM default price instead of just vCPU price like earlier. If you want me to change that, please let me know. Thanks!

mattray:
Thanks for the PR @sachin-rafay! @opencost/opencost-maintainers could we get a review to followup with @thomasvn's review? This seems pretty fundamental

pkg/cloud/aws/provider.go
```diff
@@ -1307,17 +1321,35 @@ func (aws *AWS) createNode(terms *AWSProductTerms, usageType string, k models.Ke
...
+				GPU:          terms.GPU,
+				Storage:      terms.Storage,
+				BaseCPUPrice: aws.BaseCPUPrice,
+				BaseRAMPrice: aws.BaseRAMPrice,
+				BaseGPUPrice: aws.BaseGPUPrice,
+				UsageType:    PreemptibleType,
+			}, meta, nil
+		} else {
+			// return defaults if public pricing not found
+			log.DedupedWarningf(5, "Could not find Node %s's public pricing info", k.ID())
```
thomasvn:
```suggestion
			log.DedupedWarningf(5, "Could not find Node %s's public pricing info, using default configured spot prices instead", k.ID())
```

Thoughts?

pkg/cloud/aws/provider.go
thomasvn: Thanks for this change. I'm in agreement here!

thomasvn:
These changes LGTM. We have some nightly integration tests which validate pod allocation costs scheduled on an OnDemand EKS node. I'll make sure that remains unchanged after merge of this PR.

thomasvn (approved):

JonCholas:
Any plan to merge this soon? this is blocking us to adopt kubecost for all our clusters

AjayTripathy (approved):

AjayTripathy:
Thank you so much for the review! @thomasvn and the PR @sachin-rafay . Merging now. Expect this in our release of 2.0 scheduled for late January.

PR #2396, opened 2023-12-18
Allocation properties: Add support for aliased labels

biancaburtoiu:
## What does this PR change?
This PR adds support for handling aliased labels in allocation properties. Specific to KC Cloud only.

https://github.com/kubecost/kubecost-cloud/pull/163

Users will be able to filter and aggregate leveraging default aliased label names. 

N/A

## How was this PR tested?
Local

No

*

### DIFF
--- a/pkg/kubecost/allocationprops.go
+++ b/pkg/kubecost/allocationprops.go
@@ -43,6 +43,10 @@ func (apt *AllocationProperty) GetAnnotation() string {
 
 // IsAliasedLabel returns true if the allocation property corresponds to an aliased label
 func (apt *AllocationProperty) IsAliasedLabel() bool {
+	if apt == nil {
+		return false
+	}
+
 	return *apt == AllocationDepartmentProp ||
 		*apt == AllocationEnvironmentProp ||
 		*apt == AllocationOwnerProp ||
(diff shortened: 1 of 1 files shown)


pkg/kubecost/allocationprops.go
```diff
@@ -41,6 +41,33 @@ func (apt *AllocationProperty) GetAnnotation() string {
 	return ""
 }
 
+// IsAliasedLabel returns true if the allocation property corresponds to an aliased label
+func (apt *AllocationProperty) IsAliasedLabel() bool {
+	return *apt == AllocationDepartmentProp ||
```
michaelmdresser: This will panic if `apt` is nil

biancaburtoiu: addressed, thanks for catching this!

michaelmdresser (approved):
I assume this comes from Cloud having columns dedicated to these aliases?

mbolt35 (approved):
Looks great for merge into the `bolt/allocation-property` branch.

mbolt35:
> I assume this comes from Cloud having columns dedicated to these aliases?

@michaelmdresser Yeah, this will also eventually go into one PR we'll submit to merge with opencost.

PR #2408, opened 2023-12-19
Enable (*FileStorage).ListDirectories to list symlinks which target directories

michaelmdresser:
## What does this PR change?
Enable `(*FileStorage).ListDirectories` to list symlinks which target directories

Done by making `DirFileToStargeInfo` follow symlinks to determine if the target is a directory, then returning the _unfollowed_ path as a valid directory so the path will be correctly constructed relative to the symlink, rather than relative to the target. The purpose is only to ensure that the target is in fact a directory.

- Required for https://github.com/kubecost/kubecost-cost-model/pull/2002

* N/A, nobody should currently be using symlinks

## How was this PR tested?
* Used in https://github.com/kubecost/kubecost-cost-model/pull/2002 to successfully ingest data "stored" in a symlinked path

### DIFF
--- a/pkg/storage/filestorage.go
+++ b/pkg/storage/filestorage.go
@@ -6,6 +6,7 @@ import (
 	gopath "path"
 	"path/filepath"
 
+	"github.com/opencost/opencost/pkg/log"
 	"github.com/opencost/opencost/pkg/util/fileutil"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


nik-kc (approved):

mbolt35:
This change looks fine to me! Will this support hard links as well? I think you can create a hard link in go via os.Link

My guess is that hard links auto-resolve, so the first IsDir() check will pass. Just curious if you had tested it. 

Otherwise, great enhancement!

mbolt35 (approved):
LGTM

michaelmdresser:
> This change looks fine to me! Will this support hard links as well? I think you can create a hard link in go via os.Link
> My guess is that hard links auto-resolve, so the first IsDir() check will pass. Just curious if you had tested it.

Great question. I just added a unit test for a symlink (which I should've done anyway). I duplicated it to make a hardlink test, but I get "Operation not permitted" when calling `os.Link` to make the hardlink. No clue why that is, but I'm going to leave it alone for now.

PR #2409, opened 2023-12-20
Make telemetry metrics respect disable state

michaelmdresser:
## What does this PR change?
* The `opencost_build_info`, `kubecost_http_requests_total`, `kubecost_http_response_time_seconds`, and `kubecost_http_response_size_bytes` will now respect the disabled metric configuration.

## Does this PR relate to any other PRs?
* 

## How will this PR impact users?
* 

## Does this PR address any GitHub or Zendesk issues?
* Closes ...

## How was this PR tested?
* By Jesse in a live env, disabling all of these metrics successfully using the config.

PR #2413, opened 2023-12-21
Pick #2409 for v1.108

michaelmdresser:
- Pick of #2409: 9287d2c


michaelmdresser:
@jessegoodier here's the pick into v1.108

PR #2416, opened 2023-12-23
Refactor Kubecost-specific environment variables to OPENCOST_ equivalents

### LINKED ISSUE #179: Add docs for KUBECOST_NAMESPACE env var on opencost exporter pod
As a **developer/engineer** I would like to know how to **overwrite the KUBECOST_NAMESPACE env variable** on the opencost exporter pod so the **pod not** necessarily tries to **execute in a pot. non existent namespace**. Having this in the docs would **also save time** and does **not enforce a developer/engineer** to search through **slack**.

As mentioned in [this](https://cloud-native.slack.com/archives/C03D56FPD4G/p1671114290408029) thread, I would like to extend the docs for better transparency. 
However, I'm not sure what the best place on the website would be? To me, no place sticks out here.
And, imho, also a contribution to the values.yaml of the [helm chart](https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/values.yaml) should be made. Wdyt?

mattray:
GetKubecostJobName->GetOpenCostJobName
GetKubecostNamespace->GetOpenCostNamespace
GetKubeostConfigBucket->GetOpenCostConfigBucket

## What does this PR change?
* Adds support for OPENCOST_NAMESPACE, OPENCOST_JOB_NAME, and OPENCOST_CONFIG_BUCKET without removing functionality of KUBECOST_* support. The defaults have changed, but if the environment variables were being used that shouldn't matter

* No

* Probably very few were actually using it, it wasn't used by the OpenCost Helm chart

* Fixes https://github.com/opencost/opencost-website/issues/179

## How was this PR tested?
* added unit tests and local build/test

* OpenCost usage was undocumented but noticed, need to verify values were being provided by Kubecost users

* Yes

### DIFF
--- a/pkg/clustercache/clustercache.go
+++ b/pkg/clustercache/clustercache.go
@@ -104,15 +104,15 @@ func NewKubernetesClusterCache(client kubernetes.Interface) ClusterCache {
 	batchClient := client.BatchV1().RESTClient()
 	pdbClient := client.PolicyV1beta1().RESTClient()
 
-	kubecostNamespace := env.GetKubecostNamespace()
-	log.Infof("NAMESPACE: %s", kubecostNamespace)
+	openCostNamespace := env.GetOpenCostNamespace()
+	log.Infof("NAMESPACE: %s", openCostNamespace)
 
 	kcc := &KubernetesClusterCache{
 		client:                     client,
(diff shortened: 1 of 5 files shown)


mattray:
Still valid, I can rebase if someone wants it merged.

PR #2423, opened 2023-12-28
Use node label instead of extracting from instance

### LINKED ISSUE #2281: No request or usage data found during CPU allocation calculation
**Describe the bug**
Warning in opencost container logs, warning are repeated every couple of minutes:

```
kubectl get pods -n opencost
opencost-844c646598-8c6d7   2/2     Running   0          80m

kubectl logs opencost-844c646598-8c6d7 opencost -n opencost
2023-11-02T10:32:36.413083616Z WRN No request or usage data found during CPU allocation calculation. Setting allocation to 0.
2023-11-02T10:32:36.413303319Z WRN No request or usage data found during CPU allocation calculation. Setting allocation to 0.
2023-11-02T10:32:36.41340512Z WRN No request or usage data found during CPU allocation calculation. Setting allocation to 0.
```

**To Reproduce**
1. Install opencost using helm on EKS or AKS
2. Check opencost container logs following helm deployment

Helm values file for EKS.

```
USER-SUPPLIED VALUES:
extraVolumes:
- configMap:
    name: opencost-aws
  name: custom-configs
nameOverride: opencost
opencost:
  customPricing:
    configPath: /tmp/custom-config
    configmapName: custom-pricing-model
    createConfigmap: true
    enabled: false
  exporter:
    defaultClusterId: iob-eks-1
    extraEnv:
      AWS_PRICING_URL: https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/eu-west-1/index.json
      CONFIG_PATH: /tmp/custom-config
    extraVolumeMounts:
    - mountPath: /tmp/custom-config
      name: custom-configs
    replicas: 2
    resources:
      limits:
        cpu: 1000m
        memory: 1Gi
      requests:
        cpu: 10m
        memory: 55Mi
  metrics:
    serviceMonitor:
      additionalLabels:
        lnrs.io/monitoring-platform: "true"
      enabled: true
  nodeSelector:
    kubernetes.io/os: linux
    lnrs.io/tier: system
  prometheus:
    internal:
      enabled: false
      namespaceName: monitoring
      port: 9090
      serviceName: kube-prometheus-stack-prometheus
    thanos:
      enabled: true
      internal:
        enabled: true
        namespaceName: monitoring
        port: 10902
        serviceName: thanos-query-frontend
  tolerations:
  - key: system
    operator: Exists
  ui:
    enabled: true
    ingress:
      annotations:
        alb.ingress.kubernetes.io/healthcheck-path: /-/ready
        alb.ingress.kubernetes.io/inbound-cidrs: 10.0.0.0/8,145.43.180.0/22
        alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS":443}]'
        alb.ingress.kubernetes.io/ssl-redirect: "443"
        alb.ingress.kubernetes.io/target-type: ip
        lnrs.io/zone-type: private
      enabled: true
      hosts:
      - host: <Redacted>
        paths:
        - /
      ingressClassName: alb-core-internal
    resources:
      limits:
        cpu: 1000m
        memory: 1Gi
      requests:
        cpu: 10m
        memory: 55Mi
podLabels:
  lnrs.io/k8s-platform: "true"
serviceAccount:
  annotations:
    eks.amazonaws.com/role-arn: <Redacted>
  automountServiceAccountToken: true
  create: true
```

**Expected behavior**
My question is whether these warning can be ignored or is there a more serious configuration issue? Can there warning be suppressed if this isn't an issue?

Version: quay.io/kubecost1/kubecost-cost-model:prod-1.106.2

**Additional context**
Public Clouds: EKS and AKS
EKS Server Version: v1.27.6-eks-f8587cb
AKS Server Version: v1.27.3

deepy:
Given a common cAdvisor setup the instance label will refer to the node by IP rather than hostname
This uses the `node` label instead, relying on a setup where that label is properly configured rather than an unexpected `instance` setup.

## What does this PR change?
* Switches the querying to the `node` label instead of relying on users configuring prometheus to add the node to the `instance` hostname

* Common monitoring setups will no longer break opencost (this also affects kubecost)
* If you're configuring prometheus directly you need the `relabel_configs` from below
* If you're using `ServiceMonitors` you need to make certain you have `honorLabels: true`

* Closes #2281

## How was this PR tested?
* On a local cluster
* In an IDE debugging
* (metrics breaking kubecost was confirmed in local cluster as well)

Yes, the following `relabel_configs` needs to be added:
```yaml
relabel_configs:
- action: labelmap
   regex: __meta_kubernetes_node_name
   replacement: node
```

* No access to labels


deepy:
image

mattray:
@deepy thanks for the PR! This seems relatively straightforward, but I'd like the @opencost/opencost-maintainers (especially @AjayTripathy) to weigh in on this.

deepy:
Absolutely, same here :-)
I'm currently running this setup live in AWS and while both the reports in the UI/API and the metrics are working fine and showing correctly results, I haven't fully validated every metric and every part of it.

And there's some places where I've left it as instance as everything resolved correctly when inspected under a debugger

Thanks for having a look at this!

mattray:
@opencost/opencost-maintainers anyone want to weigh in? This came from https://github.com/opencost/opencost/issues/2281

mbolt35:
This needs to be thoroughly tested across all providers. IIRC, there is a lot of nuance with this particular label.

AjayTripathy:
https://github.com/opencost/opencost/pull/2444 We're playing with making this configurable.

r2k1:
One note. Currently recommended way of installing opencost is using `latest` image tag.
This can be a potentially breaking change that will be automatically delivered to many users.

thomasvn:
@deepy I've recently also noticed an environment with cAdvisor returning an IP instead of hostname. This ended up causing issues with OpenCost computing Allocations.

@deepy Do you know what versions of the Kubelet or what versions of cAdvisor show this behavior? Many of my environments still show the "Instance" label as having the hostname. Regardless, I agree with this PR's approach and think we should give users an option on how they want to query cAdvisor here.

mattray:
@thomasvn what do you want to do with this PR?

deepy:
I don't think kubelet or cAdvisor versioning is what factors in here, I'm willing to bet a very small sum on it being configuration. Potentially just as simple as `honorLabels: true`

I'll put a little effort into another investigation at some point next week, but for now I think everything is in https://github.com/opencost/opencost/issues/2281#issuecomment-1837274896

rtnpro:
We, at Zapier, also faced similar issue, where opencost reported duplicate series for the same node, because of different (instance, node) values. This is similar to the bug reported here: https://github.com/opencost/opencost-helm-chart/issues/115

We have successfully patched opencost release v1.111.0 with the changes in this PR and it fixed the issue. After testing this change for more than a week in staging, we discovered no regression. We have successfully productionized this change. We are using AWS EKS.

Thanks @deepy for working on this fix.

It will be great if we could merge this fix to develop and release it.

ameijer:
going to announce the deprecation of the instance label in 1.114 release notes, paving the way for this to go in for 1.115

Mattes83:
I'd love to see this merged

thomasvn:
I've done a bit more research on this recently. I agree the root of the issue is likely due to the way Prometheus is configured to scrape cAdvisor metrics. @deepy could you tell me more about how your cAdvisor is configured? Most importantly, I think your cAdvisor config needs to have the following lines ([ref](https://github.com/prometheus-community/helm-charts/blob/720e7ec4e813034ea2605c64cef8c074821dfade/charts/prometheus/values.yaml#L922-L923))

```yaml
          - target_label: __address__
            replacement: kubernetes.default.svc:443
```

According to Prometheus docs ([ref](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config)), that `__address__` label eventually gets used for the `instance` label.

deepy:
@thomasvn every packaged setup I've seen so far with the exception of the one from kubecost referred in the linked issue has had `instance` as `host:port` and I think that alone is enough to justify merging this PR
But we've now switched to `alloy` and it like others explicitly set `node` to be the node name https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/charts/feature-cluster-metrics/templates/_cadvisor.alloy.tpl#L38

So we still have node like `ip-10-1-2-3.ec2.internal` with an instance like `10.1.2.3:10240`

ameijer:
@deepy we owe you an update! this PR has been the topic of much debate during the opencost community meetings. After this discussion, we decided against merging this PR since it would potentially cause duplicate/missing information for people without the proper relabeling configs. Not to worry though - @mbolt35 and @Sean-Holcomb are working on a refactor in https://github.com/opencost/opencost/pull/3031 that should address this concern. The new logic will first attempt to query the node label, and if it does not exist, will fall back to the instance label. Here's the specific code: https://github.com/opencost/opencost/blob/c651e2678d3ee20e06e86f36b6a623a5049ebd42/core/pkg/source/decoders.go#L434-L459 and https://github.com/opencost/opencost/blob/c651e2678d3ee20e06e86f36b6a623a5049ebd42/modules/prometheus-source/pkg/prom/metricsquerier.go#L558-L573 that should cover your use case, and will also prevent anyone still using instance labels from getting broken.

deepy:
@ameijer I can see what you're trying to do, but I think that you're about to potentially make a mistake in that case
`instance` is never going to work unless you've specifically configuration it work and from what I can tell opencost has been the only ones to do so

So taking a look at the 3 different scenarios that comes to mind, let's call them `node`, `instance`, and `default` for whether you've configured the node name to be in node, instance, or not configured it at all (i.e. no `node` and `instance` contains `ip:port`

This would only cover the 2 first instances, and for the third you'd get weird unexpected results that at a glance look like they might work but quickly break down when you start running the numbers
And the best indicator that it doesn't work? (that I know of) Was changed from WARN to DEBUG in https://github.com/opencost/opencost/pull/2715

Rather than falling back I'd suggest using a feature flag, then you could leave the default to the current functionality and still support this use-case
Another solution could be to build tooling that checks your metrics + opencost's to verify that everything is configured correctly

deepy:
(On that note, I don't know if the third scenario exists in any legitimate volume, but since I know `instance` is wrong for me I'd want to be sure there that opencost is not looking at it)

mbolt35:
@deepy

>  I can see what you're trying to do, but I think that you're about to potentially make a mistake in that case
instance is never going to work unless you've specifically configuration it work and from what I can tell opencost has been the only ones to do so

Can you expand on this a bit? `node` also doesn't work unless you add a configuration, right? Historically, the `instance` relabel has supported many installs of opencost without issue. We are adding support for a `new` label while continuing to support the existing method. Where is the mistake? 

If I look at the community prometheus helm chart, the configuration to set instance IS used, but not `node`:
https://github.com/prometheus-community/helm-charts/blob/720e7ec4e813034ea2605c64cef8c074821dfade/charts/prometheus/values.yaml#L922-L923

It seems reasonable to base our implementation on the default configurations set by the community prometheus chart. 

For what it's worth, If we were to throw away any backwards compatibility, making the change from an `instance` based configuration to a `node` based configuration makes the most sense, so I totally agree with the choice in a vacuum. However, it seems the `node` configuration is less prevalent than the `instance` configuration, albeit the desirability to support `node` does seem more optimal. Thus, we are preparing for this change by supporting both. This concedes a small compatibility issue (where having _both_ instance and node can aggregate separately before and after the change IFF the node configuration is added) in an effort to prioritize support for `node` without compromising the installs which leverage instance, and doesn't require older installs add a configuration. 

If the issue is that there is lack of reporting or diagnostics for this issue, then I agree with you 100%, but I'd consider that a different issue. I'm opposed to adding counter-intuitive flags which also require modifying configurations. Having good indicators when there is an issue is a great idea though.

deepy:
Oh no, this actually got pretty interesting, so for cAdvisor
`prometheus-community/prometheus` puts the node's name in `instance`
`prometheus-community/kube-prometheus-stack` puts the node's name in `node`

If I were to argue, I'd say `kube-prometheus-stack` is the one that gets it right, but that's besides the point, the point being that it's bizarre that the two charts produce different results and now I'm itching to create a PR that'll make a lot of people unhappy 😅 

So for the 3 default installs that's been mentioned in here, 2 have the node name in `node`. `grafana-k8s-monitoring` and `prometheus-community/kube-prometheus-stack`
While `prometheus-community/prometheus` has it in `instance` (and kubecost as well)

All my resistance against having a fallback comes from having a not-insignificant amount of time troubleshooting and debugging opencost
Which is why I think for those with setups where the `node` label is populated, having the fallback to `instance` in place is just a problem waiting to happen - and why I'd want to be able to disable it

Now I'm a very opinionated person, but in light of this, and how many places things can be configured incorrectly-
I think the real solution would be checking and verifying all the metrics/labels and throwing errors when the correct configuration can't be deduced

As for which configuration is more prevalent (node vs instance) I have no idea, but mostly everyone I know using the prometheus-community charts uses `kube-prometheus-stack` and not `prometheus`

I need to mull things over, but if you do implement the fallback then please throw a hard error if you end up retrieving a node name which contains a `:`, I think that'd prevent every misconfiguration I can think of

mbolt35:
@deepy Yikes! Honestly, we wouldn't be having this conversation at all if there weren't existing users with `instance` setups, so I definitely hear where you're coming from. I think this configuration value likely comes from a legacy version of the prometheus chart, which we used (at least 5 years ago) to base our metric query/parsing on. Unfortunately, there isn't a really good way to "migrate" metric data from one form to the other without breaking compatibility. 

> If I were to argue, I'd say kube-prometheus-stack is the one that gets it right

I agree given that it seems like the more modern chart uses a more well-defined label to represent the node (versus overwriting `instance`). We're definitely on the same page in terms of what we'd like to see long term :)

> All my resistance against having a fallback comes from having a not-insignificant amount of time troubleshooting and debugging opencost

I totally empathize, and I agree that there should absolutely be a mechanism for pinpointing the problem when there is a `ip:port` in the place of an expected node name. 

> Now I'm a very opinionated person, but in light of this, and how many places things can be configured incorrectly-
I think the real solution would be checking and verifying all the metrics/labels and throwing errors when the correct configuration can't be deduced

I really do appreciate your opinions and attention to this specific problem. It's been one of many adventures in debugging prometheus related gotchas. 

> I need to mull things over, but if you do implement the fallback then please throw a hard error if you end up retrieving a node name which contains a :, I think that'd prevent every misconfiguration I can think of

Originally, I was thinking of performing a dns lookup against the IP to determine the hostname IF that field was determined to be an `ip:port`, but I think that opens up a can of worms potentially, and while we could internally cache the ip->dns name values to prevent blasting the dns service, it still feels a bit too "cute" of a solution, and I think I like the hard error approach better. The error itself could even include the config update to make to avoid the problem. 

Let me know if you have any other ideas I can explore, thanks again for the help here!

mbolt35:
https://github.com/opencost/opencost/pull/3031/commits/2eb13005e2699c1eafeba0b35b63223ca905795a

deepy:
It might be a little less clear in the case of no `node` and ip:port in `instance, but I think you can simplify the check by changing it to a "is this a valid node name" rather than if it's an IP address/port combo

The scenario is that opencost is looking for a node name, and if it's unable to find a valid one then there's misconfiguration present (or configuration needs to be added)
So while checking if it's an IP:port gives the opportunity for a better error message, checking if it's a valid node name would cover other interesting situations as well

mbolt35:
> The scenario is that opencost is looking for a node name, and if it's unable to find a valid one then there's misconfiguration present (or configuration needs to be added) So while checking if it's an IP:port gives the opportunity for a better error message, checking if it's a valid node name would cover other interesting situations as well

This is actually not always the "wrong" thing - it's very possible to scrape a metric for a node that was removed. This happens very frequently with node autoscalers. If you're just talking about validating that the characters are valid for use with a node, I'm afraid that ip/port combos will fall under that regex.

deepy:
I'm talking about the latter yeah, simply verifying that the string is a valid name for a node
It's a good safeguard in general, whereas the `ip:port` one is specific to this case (`node` vs `instance`)

mbolt35:
It looks like Node names must follow DNS subdomain names: RFC 1123

* contain no more than 253 characters
* contain only lowercase alphanumeric characters, '-' or '.'
* start with an alphanumeric character
* end with an alphanumeric character

Ok, so it does look like this would exclude ip/ports. I'll update this logic accordingly! Thanks for the feedback.

mbolt35:
https://github.com/opencost/opencost/pull/3031/commits/cf4d510ad4dd1c27e37d7ce98a9b2c4296e2a402

ameijer:
Going to close this PR - looks like the spirit of this change has been incorporated into the forthcoming refactor PR as per @mbolt35 's code ref

PR #2429, opened 2024-01-02
Make FileStorage's Read and Write go through `flock()`

michaelmdresser:
## What does this PR change?
* All usage of FileStorage.Read() and FileStorage.Write() will go through the `flock()` syscall. No impact is expected on normal OpenCost usage.

* Pairs with https://github.com/kubecost/kubecost-cost-model/pull/2020

* N/A

## How was this PR tested?
* New unit test


mbolt35:
@michaelmdresser 

### General File Locking Dangers 
I *think* advisory locks are safe from the following scenario, but I wanted to make sure: 

_What happens when the process hangs/crashes while holding the lock? Restarting the process attempts to acquire lock and is forced to wait indefinitely?_

If failure to release the lock causes long wait times or hangs on crashes, we'll likely want to handle those cases (ie: signal handlers?) 

### Flock Cross-Platform Compatibility
`Flock` is not implemented in windows `syscall` - the quickest way to see this is to navigate to: https://pkg.go.dev/syscall and click the dropdown and select `windows/amd64`

image

If you search for `Flock` here, you'll find no such method. 

I would suggest we create a separate file containing the implementations for nix platforms and windows platforms.

`filelocks_windows.go`
```go
//go:build windows

func WriteLockedFD(f *os.File, data []byte) (int, error) {
    // implement using regular os.Write? 
}

// ReadLocked....
...
```
`filelocks.go`
```go
//go:build (linux || darwin)

// implementations using Flock
```

I wrote the above off the top of my head, so you may want to do a bit more digging on the "correct" way to do this -- I'm not certain, but you may be able to use the `unix` flag (which I think includes `darwin`). It's also worth looking into, but I think using the `_windows.go` file name AND the `//go:build windows` is belt and suspenders - it does appear that the go standard library uses the file name in most cases - I see no harm in using both mechanisms as long as they agree. 

Anytime `syscall` is used, it's very likely we'll have to address this cross-platform situation, especially in the opencost project. I don't think we have abandoned support for running on Windows nodes, and provided that there exist other packages that depend on opencost, I think we want to remain as cross-platform as possible.

With respect to flock vs the default go file-descriptor based locking, I would prefer making advisory locks be an opt-in solution rather than the default. I have a few hand-wavy (and possibly baseless) concerns: 
* I believe there are current scenarios today where there might be multiple processes reading/writing to the same file (not saying we _should_ be, but it's reasonable to expect there will be), and depending on frequency of those accesses, we may create some unexpected performance hangups. (Old > 1 replica deployments)
* 3+ years ago, we had a file access frequency bug that was churning the disc horribly (many many reads per second - accessing configs in our run loop, etc...). While I think most of the fixes that were made exist as `ProviderConfig`, it may be worth looking into how these changes interact there. The fixes addressed: 
  * Use a mutex to actually control concurrent reads/writes within the same process to establish better (deterministic) flow control -- clearly this only really applies to single process. 
  * Cache file contents in memory until a change had occurred. 
* Defaulting to advisory locks doesn't allow our API to use regular posix file locking (this may be a non-issue on modern nix kernels)

It's completely possible that I've exposed myself as an old man yelling at the wind, and the above are mostly paranoia. 

Additionally, I don't think it's worth over-complicating the simplified API we provide UNLESS advisory locks risk creating scenarios mentioned above (ie: crash fails to relinquish lock). I'm willing to bend towards the simplified approach of defaulting to `flock` if there are compelling reasons to do so. 

Lastly, I located this library which appears to have an implementation of flock for Windows, with some added cases handled for unix - https://github.com/gofrs/flock

Up to you whether or not we should include that dependency or use its implementation as a guide to ensure we cover the proper platform bases. 

**Bonus Fact**: Go SDK/Tooling leverages advisory file locking - https://github.com/golang/go/blob/master/src/cmd/go/internal/lockedfile/internal/filelock/filelock_unix.go

mbolt35 (changes requested):
@michaelmdresser 

### General File Locking Dangers 
I *think* advisory locks are safe from the following scenario, but I wanted to make sure: 

_What happens when the process hangs/crashes while holding the lock? Restarting the process attempts to acquire lock and is forced to wait indefinitely?_

If failure to release the lock causes long wait times or hangs on crashes, we'll likely want to handle those cases (ie: signal handlers?) 

### Flock Cross-Platform Compatibility
`Flock` is not implemented in windows `syscall` - the quickest way to see this is to navigate to: https://pkg.go.dev/syscall and click the dropdown and select `windows/amd64`

image

If you search for `Flock` here, you'll find no such method. 

I would suggest we create a separate file containing the implementations for nix platforms and windows platforms.

`filelocks_windows.go`
```go
//go:build windows

func WriteLockedFD(f *os.File, data []byte) (int, error) {
    // implement using regular os.Write? 
}

// ReadLocked....
...
```
`filelocks.go`
```go
//go:build (linux || darwin)

// implementations using Flock
```

I wrote the above off the top of my head, so you may want to do a bit more digging on the "correct" way to do this -- I'm not certain, but you may be able to use the `unix` flag (which I think includes `darwin`). It's also worth looking into, but I think using the `_windows.go` file name AND the `//go:build windows` is belt and suspenders - it does appear that the go standard library uses the file name in most cases - I see no harm in using both mechanisms as long as they agree. 

Anytime `syscall` is used, it's very likely we'll have to address this cross-platform situation, especially in the opencost project. I don't think we have abandoned support for running on Windows nodes, and provided that there exist other packages that depend on opencost, I think we want to remain as cross-platform as possible.

With respect to flock vs the default go file-descriptor based locking, I would prefer making advisory locks be an opt-in solution rather than the default. I have a few hand-wavy (and possibly baseless) concerns: 
* I believe there are current scenarios today where there might be multiple processes reading/writing to the same file (not saying we _should_ be, but it's reasonable to expect there will be), and depending on frequency of those accesses, we may create some unexpected performance hangups. (Old > 1 replica deployments)
* 3+ years ago, we had a file access frequency bug that was churning the disc horribly (many many reads per second - accessing configs in our run loop, etc...). While I think most of the fixes that were made exist as `ProviderConfig`, it may be worth looking into how these changes interact there. The fixes addressed: 
  * Use a mutex to actually control concurrent reads/writes within the same process to establish better (deterministic) flow control -- clearly this only really applies to single process. 
  * Cache file contents in memory until a change had occurred. 
* Defaulting to advisory locks doesn't allow our API to use regular posix file locking (this may be a non-issue on modern nix kernels)

It's completely possible that I've exposed myself as an old man yelling at the wind, and the above are mostly paranoia. 

Additionally, I don't think it's worth over-complicating the simplified API we provide UNLESS advisory locks risk creating scenarios mentioned above (ie: crash fails to relinquish lock). I'm willing to bend towards the simplified approach of defaulting to `flock` if there are compelling reasons to do so. 

Lastly, I located this library which appears to have an implementation of flock for Windows, with some added cases handled for unix - https://github.com/gofrs/flock

Up to you whether or not we should include that dependency or use its implementation as a guide to ensure we cover the proper platform bases. 

**Bonus Fact**: Go SDK/Tooling leverages advisory file locking - https://github.com/golang/go/blob/master/src/cmd/go/internal/lockedfile/internal/filelock/filelock_unix.go

michaelmdresser:
> What happens when the process hangs/crashes while holding the lock? Restarting the process attempts to acquire lock and is forced to wait indefinitely?

For crashes:
From `man flock.2`:

       Locks created by flock() are associated with an open file description  (see  open(2)).
       This  means  that  duplicate  file  descriptors  (created  by, for example, fork(2) or
       dup(2)) refer to the same lock, and this lock may be modified or released using any of
       these file descriptors.  Furthermore, the lock  is  released  either  by  an  explicit
       LOCK_UN  operation  on  any of these duplicate file descriptors, or when all such file
       descriptors have been closed.

When the process crashes, the OS will close all of its descriptors, so the "when all such file descriptors have been closed" case covers us.

If the process _hangs_ while holding the lock, we're going to be stuck on all r/w to the given file.

---

> Flock Cross-Platform Compatibility

> Anytime syscall is used, it's very likely we'll have to address this cross-platform situation, especially in the opencost project. I don't think we have abandoned support for running on Windows nodes, and provided that there exist other packages that depend on opencost, I think we want to remain as cross-platform as possible.

We discussed this. I'm crunched on time at the moment, so the PR now in a state where Windows will build (or be importable) fine but anything that goes through the locked r/w functions will error. I will implement Windows if I have time, we're unfortunately not able to use the Go project implementation because it is BSD licensed. The gofrs/flock implementation, as discussed, might leave a ton of .lock files around which we want to avoid.

pkg/util/fileutil/locks_unix.go
```diff
@@ -0,0 +1,103 @@
...
+func WriteLockedFD(f *os.File, data []byte) (int, error) {
+	// For the reasons outlined best in https://lwn.net/Articles/586904/ we're
+	// going to use flock() instead of fcntl() because we want a whole-file lock.
+	if err := syscall.Flock(int(f.Fd()), syscall.LOCK_EX); err != nil {
+		return 0, fmt.Errorf("unexpected error flock()-ing with EX: %w", err)
+	}
+
+	n, err := f.Write(data)
+	if err != nil {
+		return n, fmt.Errorf("writing data: %w", err)
```
ameijer, line 55: if an error writing the data, should we unlock?

pkg/util/fileutil/locks_unix.go
```diff
@@ -0,0 +1,103 @@
...
+// WriteLocked uses the flock() syscall to safely write to a file as long as
+// other users of the file are also using flock()-based access.
+//
+// WriteLocked will block until it gets lock access.
+//
+// For the reasons outlined best in https://lwn.net/Articles/586904/ this uses
+// flock() instead of fcntl(). The ability to lock byte ranges is not necessary
+// and flock() has better behavior.
+func WriteLocked(filename string, data []byte) (int, error) {
+	file, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE, 0600)
```
ameijer, line 63: to improve safety, should you open this in truncated mode? That way you guarantee that the contents of data will be the only thing in the file? is this achieved through some other mechanism here?

pkg/util/fileutil/locks_unix.go
michaelmdresser: Good point! done

pkg/util/fileutil/locks_unix.go
michaelmdresser: That's a good point. I _think_ calling `.Write()` also truncs but I don't' want to rely on that. Done.

michaelmdresser:
What I didn't discover here, but did through KCM unit testing, is that I need to be a little more careful with FD position. To be safe, functions have been updated to be very opinionated and only do full file R/W. Docstrings have been updated accordingly.

ameijer (approved):
LGTM

mbolt35 (approved):
LGTM!

PR #2432, opened 2024-01-03
Fix a test that relies on Prom library

michaelmdresser:
## What does this PR change?
We were assuming that the library always returns seconds, but it can now also return minutes (presumably if the seconds divide well). This is probably failing due to an update of that library.

* N/A

## How was this PR tested?
* Unit test was failing for me, now passes

### DIFF
--- a/pkg/kubecost/window_test.go
+++ b/pkg/kubecost/window_test.go
@@ -667,13 +667,17 @@ func TestWindow_DurationOffsetForPrometheus(t *testing.T) {
 	}
 
 	dur, off, err := w.DurationOffsetForPrometheus()
-	expDur := int(now.Sub(startOfToday).Seconds())
-	expDurStr := fmt.Sprintf("%ds", expDur)
 	if err != nil {
 		t.Fatalf("unexpected error: %s", err)
 	}
-	if dur != expDurStr {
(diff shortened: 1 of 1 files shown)


pkg/kubecost/window_test.go
```diff
@@ -667,13 +667,19 @@ func TestWindow_DurationOffsetForPrometheus(t *testing.T) {
...
 	}
-	if dur != expDurStr {
-		t.Fatalf(`expect: window to be "%s"; actual: "%s"`, expDurStr, dur)
+	// We can get a response in seconds OR minutes. Check seconds first as it
+	// is higher resolution.
+	expDurSec := int(now.Sub(startOfToday).Seconds())
+	expDurSecStr := fmt.Sprintf("%ds", expDurSec)
+	expDurMin := int(now.Sub(startOfToday).Minutes())
+	expDurMinStr := fmt.Sprintf("%dm", expDurMin)
+	if dur != expDurSecStr {
```
saweber: super nit: any particular reason to not AND the two conditions?

saweber (approved):

pkg/kubecost/window_test.go
michaelmdresser: No good reason at all -- fixed, thanks!

Sean-Holcomb (approved):

Issue #2309, opened 2023-11-16
"kubecost" packages should be refactored to "opencost"


mattray:
A substantial amount of code was recently open-sourced by Kubecost without changing the packages. Given there's nothing related to Kubecost other than they provided the code, the code should probably be refactored to map to functionality.

AjayTripathy:
+1, would greatly appreciate community support on this one. I don't think it's too hard.

lmello:
@mattray can you provide a list of which codes should be migrated ? I can have a look at this

lmello:
One thing to think about is how this could impact kubecost if they source those packages. It might require some work on their side also.

mattray:
I'll bring it up with @cliffcolvin to see how disruptive this will be and when we should time it.

mattray:
We're hoping to tackle this on the main `develop` branch right after 1.108 is released so we have plenty of time to deal with any downstream issues.

mattray:
Based off my quick read-through of the `pkg/kubecost` without understanding their full usage, here's a proposed first pass at where to move everything.

* `pkg/allocation` <- `allocation*, mock.go, summaryallocation*, totals.go, totals_json.go`
* `pkg/asset` <- `asset*, cloudusage.go, diff_test.go`
* `pkg/cloudcost/` <- `cloudcost*, costmetric.go`
* `pkg/util/json/` <- `json.go`
* `pkg/util/labelconfig/` <- `config.go`
* `pkg/util/pair/` <- `common.go` (rename to `pair.go`)
* `pkg/util/query/` <- `query.go`
* `pkg/util/window/` <- `window.go, coverage.go`

These I don't know, someone will need to decide if/where they belong:

* `bingen.go`
* `kubecost_codecs.go` (generated by bingen)
* `status.go` ETLStatus? Does this belong in OpenCost?

@opencost/opencost-maintainers from Kubecost please weigh in so we can move these early in this release cycle and deal with required downstream updates

mbolt35:
For the bingen specific generation, I'm pretty sure it uses the package name you target. ie: pkg/kubecost generates 'kubecost_codecs.go'

Bingen has a pretty soft requirement of keeping the models in the same package, so if we moved allocation, assets, etc to 'pkg/models' bingen would generate the 'models_codecs.go'

The bingen.go file should just go in the same package as the models. 

[...]

PR #2440, opened 2024-01-08
Opencost Core Refactor

mbolt35:
## Major Changes
This refactor is focusing on a few main concerns revolving around filters, legacy filters, and unification around the current implementation. It separates an easier to consume module for external projects which reference opencost for the models and/or utilities.  

### Opencost Core Module
Created a new go module under the `core/` directory which will now contain sharable opencost code:
  * Data Models: Any of the data model structs or support types are moved to `core`
  * Filters: All filter code is now part of `core`
  * Utility and Logging: Any utility types, logging, and other packages like `env` are moved to `core`
  * Abstraction separation from implementation - will eventually allow external implementations of opencost abstractions. 

### Filter Unification 
* The 2.1 Filter language will become to default standard for opencost querier services. There will be continued support for legacy http parameters for filters. 
* Moved all filter packages under `core/pkg/filter` -- legacy filtering support will temporarily be supported during this process under `core/pkg/filter/legacy` 
* **More work is required here!**

## Require Answers
* Is this the proper way to create a new go module in the same repository? A new `go.mod` was created for the module: `github.com/opencost/opencost/core` and the opencost `go.mod` uses: 
```
replace github.com/opencost/opencost/core => ./core
```
to ensure that both modules are still "versioned" together. However, it's unclear if this method of replacement is the correct convention for long term maintenance. 
  * When a git tag is created, will both modules leverage the same version? (that is currently the intent).


core/pkg/kubecost/window.go
```diff
@@ -28,8 +28,60 @@ var (
...
+
+	utcOffsetLock sync.Mutex
+	utcOffsetDur  *time.Duration
 )
 
+// get and cache the thanos offset duration.
+// TODO: Due to dependencies here, we have to drag a non-core config option into
+// TOOD: core scope. Any solution here would be a one-off until we can generalize
+// TODO: global configuration options.
+func thanosOffset() time.Duration {
```
mbolt35: As mentioned above, Thanos has some leaked implementation into the core module. To avoid expanding the scope of this refactor, I'm punting due to the impact being low. We'll address these infrequent cases next pass.

core/pkg/kubecost/window.go
```diff
@@ -28,8 +28,60 @@ var (
...
+// TODO: Same note as thanosOffset above - temporary work-around until more
+// TODO: generalized global configuration.
+func isThanosEnabled() bool {
+	return env.GetBool("THANOS_ENABLED", false)
+}
+
+// returns the configured utc offset as a duration
+// TODO: Same as the above options -- we should provide a one-time initialization configuration
+// TODO: for these values, or deprecate their use.
+func utcOffset() time.Duration {
```
mbolt35: This is another case where we've leaked opencost env configuration into our utilities. I think this can be solved next pass (it's also possible that UTC Offset configuration is no longer a valid config option).

core/pkg/util/promutil/promutil.go
```diff
@@ -0,0 +1,120 @@
+package promutil
```
mbolt35, line 1: This entire utility was lifted from opencost methods in the `prom` package (which remains in opencost). 

As of now, this is duplicated code, and the opencost dependents should leverage `promutil` instead. Must make this change before merge.

core/pkg/util/timeutil/timeutil.go
```diff
@@ -81,6 +83,40 @@ func DurationOffsetStrings(duration, offset time.Duration) (string, string) {
 	return DurationString(duration), DurationString(offset)
 }
 
+// ParseUTCOffset attempts to parse a UTC offset string. Returns a valid time.Duration
+// if error is nil.
+func ParseUTCOffset(offsetStr string) (time.Duration, error) {
```
mbolt35, line 88: There was an implementation of this nested within the `env` utility, but we've pulled it out to allow other utilities to leverage the same parsing mechanism.

Sean-Holcomb:
@bolt given everything is moving right now and this move is bound to break people, it seems like now would be a good time to remove the name kubecost from the pkg name. What do you think?

Sean-Holcomb:
> @bolt given everything is moving right now and this move is bound to break people, it seems like now would be a good time to remove the name kubecost from the pkg name. What do you think?

Not to say it should be this PR necessarily but flagging that both should land in the same release to prevent a double move for external users

core/pkg/kubecost/query.go
```diff
@@ -4,7 +4,7 @@ import (
 	"strings"
 	"time"
 
-	filter21 "github.com/opencost/opencost/pkg/filter21"
+	filter21 "github.com/opencost/opencost/core/pkg/filter"
```
Sean-Holcomb: Looks like this alias didn't removed

core/pkg/protocol/protocol.go
```diff
@@ -1,4 +1,4 @@
-package proto
+package protocol
```
Sean-Holcomb, line 1: no need for change but planting a flag, this package is duplicated in KCM and can be deleted from there

Sean-Holcomb (approved):

mattray:
> @bolt given everything is moving right now and this move is bound to break people, it seems like now would be a good time to remove the name kubecost from the pkg name. What do you think?

https://github.com/opencost/opencost/issues/2309 is the Issue

nikovacevic (approved):

core/pkg/protocol/protocol.go
mbolt35: I knocked it out there as well :)

core/pkg/opencost/opencost_codecs.go
```diff
@@ -457,6 +457,7 @@ func (target *Allocation) MarshalBinaryWithContext(ctx *EncodingContext) (err er
 	}
 	// --- [end][write][alias](LbAllocations) ---
 
+	buff.WriteFloat64(target.UnmountedPVCost) // write float64
```
mattray: is this coming from Kubecost?

mbolt35: No, this is being written from the struct definitions -- although I'm glad you spotted this. I think this field is supposed to be ignored. @ameijer Can I add the ignore to this changeset? Did I somehow stomp on your changes or are they still in-flight?

pkg/costmodel/allocation.go
```diff
@@ -4,16 +4,16 @@ import (
...
 
-	"github.com/opencost/opencost/core/pkg/kubecost"
 	"github.com/opencost/opencost/core/pkg/log"
 	"github.com/opencost/opencost/pkg/env"
 	"github.com/opencost/opencost/pkg/prom"
 )
 
 const (
-	// https://kubecost.atlassian.net/browse/BURNDOWN-234
+	// https://opencost.atlassian.net/browse/BURNDOWN-234
```
mattray: Pretty sure that one stays "kubecost"

mattray (approved):
Other than the 2 comments it looks like a global find and replace.

PR #2441, opened 2024-01-09
Add config option to not query for local disk cost

thomasvn:
## What does this PR change?
* Add config option to not query for local disk cost. Environment variable `ASSET_INCLUDE_LOCAL_DISK_COST`.
* By default, `ASSET_INCLUDE_LOCAL_DISK_COST=true`. Therefore by default there should be no change in functionality. To ignore a node's local disk cost (i.e. ephemeral storage), set `ASSET_INCLUDE_LOCAL_DISK_COST=false`.

* No

* By default, no impact.
* Users who opt in will no longer see line items on their "/assets" queries related to local disk costs.

* None

## How was this PR tested?

Building a custom image and running Kubecost. I inspected aggregator's database and also looked at the results of the following API request

```txt
http://localhost:9090/model/assets?window=1d&accumulate=true&filter=assetType:%22disk%22
```

### AWS Testing

When running the custom image, the API response only returns disks with ProviderIDs of `vol-*` or `local-pv*`. No longer seeing disks with ProviderID `ip-*`. Validated in CloudCosts that AWS does not charge for the local disk attached to the node.

Here is a sample [before.json](https://github.com/user-attachments/files/15506797/before.json) and [after.json](https://github.com/user-attachments/files/15506798/after.json).

### Azure Testing

When running the custom image, the API response only returns disks which start with `pvc-*`. No longer seeing disks which start with `aks-*`. When inspecting the CloudCosts API, it seems that there are certain scenarios where node disks do incur a charge. It is not yet recommended to enable this feature in Azure environments.

### GCP Testing

When running the custom image, the API response only returns disks which start with `pvc-*`. No longer seeing disks which start with `gke-*`. However, when inspecting the CloudCosts API I do believe that GCP charges for a node's local disk. It is not recommended to enable this feature in GCP environments.

* Yes

*


pkg/costmodel/cluster.go
```diff
@@ -205,13 +188,48 @@ func ClusterDisks(client prometheus.Client, provider models.Provider, start, end
...
+	// Only query metrics about local disk costs if the user has opted in. This
+	// is default disabled because most cloud providers do not charge unless you
+	// attach a storage volume to your instance. That cost is accounted for when
+	// querying the other prom metrics.
+	//
+	// Ref:
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/RootDeviceStorage.html
+	// https://learn.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#temporary-disk
+	// https://cloud.google.com/compute/docs/disks/local-ssd
+	resLocalStorageCost := []*prom.QueryResult{}
```
thomasvn, line 199: Initialized as an empty slice. If `ASSET_INCLUDE_LOCAL_DISK_COST=false`, then the logic below which iterates through these slices to add items to the diskmap will be a no-op.

AjayTripathy:
This in itself looks fine but needs careful testing when changing anything in the prometheus workflow.

mattray:
@thomasvn do you want this kept open or should it be closed?

thomasvn:
@mattray This is something I'd like to return to in the future. But I'm ok closing it for now to keep things tidy.

thomasvn:
Going to continue working & testing this PR now.

thomasvn:
@AjayTripathy @Sean-Holcomb Let me know what you think about these changes. Based on my research so far, I think I can only conclusively say that the AWS provider does not charge for local disk costs. For other providers it seems situational.

Also let me know if you think this is the appropriate place for the fix. Or if it should be fixed downstream in KCM.

Sean-Holcomb (approved):

PR #2443, opened 2024-01-10
Fix azure cluster management price

sachin-rafay:
## What does this PR change?
* Currently, cluster management cost for AKS is being reported as $0. The cluster management cost depends upon the tier of the AKS cluster. The tier of the AKS cluster can be changed. The change in this PR takes the cluster's name and resource group name as environment variables and calls Azure API to get the tier of the cluster every 15 minutes. The price according to the tier is then reported as cluster management cost.

* Not that I am aware of

* If the users configure Azure cluster name and resource group name as environment variables, they will be able to get cluster management cost of their AKS cluster. 

* Closes ...

## How was this PR tested?
* This PR was tested on an AKS cluster on Free, Standard and Premium tiers.

* Yes, the environment variable change may require change in documentation

### DIFF
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -925,25 +925,28 @@ func (az *Azure) DownloadPricingData() error {
 
 	envManagedClustersName := env.GetAzureClusterName()
 
-	err = az.refreshClusterManagementPricing(config, envResourceGroupName, envManagedClustersName)
-	if err != nil {
-		log.Errorf("error getting managed cluster cost: %s", err.Error())
-	}
+	if envResourceGroupName != "" && envManagedClustersName != "" {
 
-	// AKS Tier can change. So we need to refresh the tier information to get the cluster management cost.
-	go func() {
-		defer errs.HandlePanic()
-
-		for {
-			log.Infof("Azure AKS Tier Refresh scheduled in %.2f minutes.", AzureAksTierRefreshDuration.Minutes())
-			time.Sleep(AzureAksTierRefreshDuration)
-
-			err := az.refreshClusterManagementPricing(config, envResourceGroupName, envManagedClustersName)
-			if err != nil {
-				log.Errorf("error getting managed cluster cost: %s", err.Error())
-			}
+		err = az.refreshClusterManagementPricing(config, envResourceGroupName, envManagedClustersName)
+		if err != nil {
+			log.Errorf("error getting managed cluster cost: %s", err.Error())
 		}
-	}()
+
+		// AKS Tier can change. So we need to refresh the tier information to get the cluster management cost.
+		go func() {
+			defer errs.HandlePanic()
+
+			for {
+				log.Infof("Azure AKS Tier Refresh scheduled in %.2f minutes.", AzureAksTierRefreshDuration.Minutes())
+				time.Sleep(AzureAksTierRefreshDuration)
+
+				err := az.refreshClusterManagementPricing(config, envResourceGroupName, envManagedClustersName)
+				if err != nil {
+					log.Errorf("error getting managed cluster cost: %s", err.Error())
+				}
+			}
+		}()
+	}
 
 	return nil
 }
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -922,16 +922,10 @@ func (az *Azure) DownloadPricingData() error {
 	}
 
 	envResourceGroupName := env.GetAzureResourceGroupName()
-	if envResourceGroupName != "" {
-		config.AzureResourceGroupName = envResourceGroupName
-	}
 
 	envManagedClustersName := env.GetAzureClusterName()
-	if envManagedClustersName != "" {
-		config.AzureClusterName = envManagedClustersName
-	}
 
-	err = az.refreshClusterManagementPricing(config)
+	err = az.refreshClusterManagementPricing(config, envResourceGroupName, envManagedClustersName)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


mattray:
@sachin-rafay good catch! I'll get this into Kubecost Engineering to verify it's an issue across the board.

mattray:
@r2k1 have you observed this issue?

sachin-rafay:
Hi @mattray, @r2k1 Is there an issue with this fix? Do I need to change anything?

r2k1:
AKS indeed has a charge for managed clusters. But I don't know whether the opencost handles it

AjayTripathy:
Seems reasonable to me to add this. We'll get this reviewed!

sachin-rafay:
Can someone please review it? cc: @AjayTripathy

AjayTripathy:
cc @cliffcolvin can you help get this reviewed?

pkg/cloud/models/models.go
```diff
@@ -181,6 +181,8 @@ type CustomPricing struct {
 	AzureStorageContainer        string `json:"azureStorageContainer"`
 	AzureContainerPath           string `json:"azureContainerPath"`
 	AzureCloud                   string `json:"azureCloud"`
+	AzureResourceGroupName       string `json:"azureResourceGroupName"`
+	AzureClusterName             string `json:"azureClusterName"`
```
Sean-Holcomb: I would prefer to avoid routing additional values through through the custom pricing struct, and this does not seem necessary as the env variables can be checked at the beginning of the price checking process without saving them here.

pkg/env/costmodelenv.go
```diff
@@ -23,6 +23,8 @@ const (
 	AzureOfferIDEnvVar                   = "AZURE_OFFER_ID"
 	AzureBillingAccountEnvVar            = "AZURE_BILLING_ACCOUNT"
 	AzureDownloadBillingDataToDiskEnvVar = "AZURE_DOWNLOAD_BILLING_DATA_TO_DISK"
+	AzureResourceGroupEnvVar             = "AZURE_RESOURCE_GROUP_NAME"
+	AzureClusterNameEnvVar               = "AZURE_CLUSTER_NAME"
```
Sean-Holcomb, line 27: would it be possible to get these values from the kubernetes api without having them configured. At a glance the node label `kubernetes.azure.com/cluster` and the providerID seem to hold the relevant values, although there may be formatting challenges .

pkg/cloud/models/models.go
Sean-Holcomb: since they are Azure specific values I would prefer them to be properties of the Azure struct.

Sean-Holcomb (commented):
The general approach looks good for getting cluster pricing data, I just have a few qualms about the configuration and handling of values. 

I mentioned that you could possibly draw these values from the kubernetes API rather than configuring them via env variable. You have access to nodes in the cluster at the creation time of the Azure provider struct and could gather and set the values at some point around here https://github.com/opencost/opencost/blob/68d15105f958ed6ebb81af4c6a91f8ede0417ee2/pkg/cloud/provider/provider.go#L161

pkg/cloud/models/models.go
sachin-rafay: Thanks for the suggestion. Fixed it.

pkg/env/costmodelenv.go
sachin-rafay: Thanks for the suggestion. Fixed it.

sachin-rafay:
@Sean-Holcomb It seems like underscore is allowed in both resource group name and cluster name. My assumption was that AKS cluster name will not have underscore. Please give me more time to find another way to get either resource group or cluster name. If you have any pointers, can you please share?

pkg/env/costmodelenv.go
sachin-rafay: @Sean-Holcomb Thanks for the suggestion. Underscore is allowed in both resource group name and cluster name. So, it is not possible to parse using node labels. Can you please share another way to do it?

pkg/cloud/models/models.go
sachin-rafay: Thanks for the review @Sean-Holcomb . Fixed it. Please review.

pkg/cloud/azure/provider.go
```diff
@@ -916,6 +921,30 @@ func (az *Azure) DownloadPricingData() error {
 		}()
 	}
 
+	envResourceGroupName := env.GetAzureResourceGroupName()
+
+	envManagedClustersName := env.GetAzureClusterName()
+
+	err = az.refreshClusterManagementPricing(config, envResourceGroupName, envManagedClustersName)
```
Sean-Holcomb: Can we wrap all of this in an if statement so that we are not running it if these env variables are not set.

Sean-Holcomb (commented):
If you would still like to get this merged, can we write up some documentation on the values that need to be configured

pkg/cloud/azure/provider.go
sachin-rafay: Done

sachin-rafay:
@Sean-Holcomb I can update the doc. Can you please point me to the file I need to update? 

Docs change:
When using OpenCost for an AKS cluster, to get the cluster management cost, 2 environment variables need to be configured:
1. AZURE_RESOURCE_GROUP_NAME : the resource group in which the cluster was created
2. AZURE_CLUSTER_NAME : the name of the AKS cluster

Sean-Holcomb:
This would be the file https://github.com/opencost/opencost-website/blob/main/docs/configuration/azure.mdx

PR #2450, opened 2024-01-13
add breakdown struct to summary alloc

nickcurie:
## What does this PR change?
* 

* https://github.com/kubecost/kubecost-cost-model/pull/2048

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/kubecost/summaryallocation.go
+++ b/pkg/kubecost/summaryallocation.go
@@ -39,6 +39,7 @@ type SummaryAllocation struct {
 	SharedCost             float64               `json:"sharedCost"`
 	ExternalCost           float64               `json:"externalCost"`
 	Share                  bool                  `json:"-"`
+	SharedCostBreakdown    SharedCostBreakdowns  `json:"sharedCostBreakdown"`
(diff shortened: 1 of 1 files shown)


pkg/kubecost/summaryallocation.go
```diff
@@ -39,6 +39,7 @@ type SummaryAllocation struct {
 	SharedCost             float64               `json:"sharedCost"`
 	ExternalCost           float64               `json:"externalCost"`
 	Share                  bool                  `json:"-"`
+	SharedCostBreakdown    SharedCostBreakdowns  `json:"sharedCostBreakdown"`
```
ameijer, line 42: Nick, I assume you are including this because the allocations object isn't built by this function, and therefore have no way to return this to the FE? 

If so, my vote would be to wait until https://kubecost.atlassian.net/browse/SELFHOST-284 gets implemented, which would address this issue. I am not crazy about changing API behavior just to work around our own planning snafu that I introduced

nickcurie:
closing since we are not adding shared cost breakdown to summary allocation

PR #2456, opened 2024-01-17
Download Azure billing CSV to filesystem, instead of holding in memory

thomasvn:
## What does this PR change?
* Downloads the Azure billing CSV to the filesystem, instead of holding it in memory

* No

* For users with large Azure bills, this configuration should consume less memory.

* None

## How was this PR tested?

```sh
export CONFIG_PATH="/tmp/localrun/default"
export PROMETHEUS_SERVER_ENDPOINT="http://127.0.0.1:9080"
export CLOUD_COST_ENABLED="true"
mkdir -p $CONFIG_PATH/cloud-integration/
cp ~/cloud-integration.json $CONFIG_PATH/cloud-integration/cloud-integration.json

# Validated CloudCost ran as expected.
# Validated that Azure billing CSV files got downloaded to /tmp/localrun/default/db/cloudcost.
go run cmd/costmodel/main.go
```

* Not yet.

*


pkg/cloud/azure/storagebillingparser.go
```diff
@@ -56,17 +60,40 @@ func (asbp *AzureStorageBillingParser) ParseBillingData(start, end time.Time, re
...
-		blobBytes, err2 := asbp.DownloadBlob(blobName, client, ctx)
-		if err2 != nil {
-			asbp.ConnectionStatus = cloud.FailedConnection
-			return err2
-		}
-		err2 = asbp.parseCSV(start, end, csv.NewReader(bytes.NewReader(blobBytes)), resultFn)
-		if err2 != nil {
-			asbp.ConnectionStatus = cloud.ParseError
-			return err2
+		if env.IsAzureDownloadBillingDataToDisk() {
```
Sean-Holcomb: I don't see the need for this if here, if this is truly is an improvement there is no need for YA env variable. Lets just always do it like this.

pkg/cloud/azure/storagebillingparser.go
```diff
@@ -184,3 +211,38 @@ func (asbp *AzureStorageBillingParser) timeToMonthString(input time.Time) string
 	endOfMonth := input.AddDate(0, 1, -input.Day())
 	return startOfMonth.Format(format) + "-" + endOfMonth.Format(format)
 }
+
+// deleteFilesOlderThan7d recursively walks the directory specified and deletes
+// files which have not been modified in the last 7 days. Returns a list of
+// files deleted.
+func (asbp *AzureStorageBillingParser) deleteFilesOlderThan7d(localPath string) ([]string, error) {
```
Sean-Holcomb, line 207: This doesn't quite feel right to me, but I think it is fine for now. If we do end up supporting partitioned exports doing it this way will make thing easier.

Sean-Holcomb (approved):
I think it is best to get rid of the env variable and download the file for all users. Memory improvements should benefit everyone here

pkg/cloud/azure/storagebillingparser.go
thomasvn: Agree. Deleting files based off an arbitrary time cutoff is weird.

The goal though, was to make sure that local disk space doesn't fill up too quickly. With CloudCost running, we would eventually be downloading every Azure month-to-date bill to disk. We should clean up the files that we're likely no longer using?

pkg/cloud/azure/storageconnection.go
```diff
@@ -45,24 +46,32 @@ func (sc *StorageConnection) getBlobURLTemplate() string {
 	return "https://%s.blob.core.windows.net/%s"
 }
 
-func (sc *StorageConnection) DownloadBlob(blobName string, client *azblob.Client, ctx context.Context) ([]byte, error) {
```
thomasvn, line 48: @Sean-Holcomb I found no other references to `DownloadBlob()` so I've deleted it. Let me know if we need to bring it back.

pkg/cloud/azure/storagebillingparser.go
thomasvn: Made the changes! This is now the default functionality.

PR #2469, opened 2024-01-25
Fix 'go vet' warning "loop variable r captured by func literal"

mattray:
```
go vet ./...
pkg/cloud/aws/provider.go:1631:110: loop variable r captured by func literal 
pkg/cloud/aws/provider.go:1745:106: loop variable r captured by func literal
```

## What does this PR change?
* code quality

* no

* no

* no

## How was this PR tested?
* "go vet"

* no

* planning on adding "go vet" to test suite

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -1615,9 +1615,10 @@ func (aws *AWS) getAllAddresses() ([]*ec2Types.Address, error) {
 
 	// Get volumes from each AWS region
 	for _, r := range regions {
+		region := r // make a copy of r to avoid capturing loop variable
 		// Fetch IP address response and send results and errors to their
 		// respective channels
-		go func(region string) {
+		go func() {
 			defer wg.Done()
 			defer errs.HandlePanic()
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/aws/provider.go
```diff
@@ -1615,9 +1615,10 @@ func (aws *AWS) getAllAddresses() ([]*ec2Types.Address, error) {
 
 	// Get volumes from each AWS region
 	for _, r := range regions {
+		region := r // make a copy of r to avoid capturing loop variable
```
michaelmdresser, line 1618: I prefer doing `r := r` to avoid having two vars that may or may not point to the same thing, but that doesn't block this change. Good find

michaelmdresser (approved):

pkg/cloud/aws/provider.go
mbolt35: Go loop captures strike again! It stinks they're about to rollout a fix for this gotcha in 1.22 I think.

mbolt35 (approved):
Looks good to me!

PR #2474, opened 2024-01-26
avoid nil panic when seaching for GPU costs in non-existent nodes

ameijer:
## What does this PR change?
* When a node isn't returned from a prom query, just default its GPU costs to 0 

* none

* stops a crashloop

* Closes ...

## How was this PR tested?
* tested in kubecost dev cluster in ADOT namespace 

* 

*

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -619,8 +619,17 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 				if len(costs.GPUReq) > 0 {
 					// allocation here is set to the request because shared GPU usage not yet supported.
 					// if VPGUs, request x (actual/virtual)
-					vgpu, verr := strconv.ParseFloat(nodes[nodeName].VGPU, 64)
-					gpu, err := strconv.ParseFloat(nodes[nodeName].GPU, 64)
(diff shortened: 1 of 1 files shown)


pkg/costmodel/metrics.go
```diff
@@ -619,8 +619,17 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
...
 					// allocation here is set to the request because shared GPU usage not yet supported.
 					// if VPGUs, request x (actual/virtual)
-					vgpu, verr := strconv.ParseFloat(nodes[nodeName].VGPU, 64)
-					gpu, err := strconv.ParseFloat(nodes[nodeName].GPU, 64)
+					vgpu := 0.0
+					gpu := 0.0
+					var err, verr error
+					if matchedNode, found := nodes[nodeName]; found {
+						vgpu, verr = strconv.ParseFloat(matchedNode.VGPU, 64)
+						gpu, err = strconv.ParseFloat(matchedNode.GPU, 64)
```
michaelmdresser, line 627: do you think its worth throwing up a trace log if at least one of these fails to parse?

michaelmdresser (approved):

AjayTripathy (approved):

PR #2476, opened 2024-01-27
Make field map access concurrency-safe

michaelmdresser:
## What does this PR change?
* Fixes a rare panic caused by concurrent access to this map
```
fatal error: concurrent map writes

goroutine 4037 [running]:
github.com/opencost/opencost/core/pkg/filter/allocation.DefaultFieldByName({0x1041ec0f5, 0x7})
        /Users/delta/devel/kubecost/model/opencost/core/pkg/filter/allocation/parser.go:35 +0x198
github.com/opencost/opencost/core/pkg/filter/ops.asField[...]({0x1041ec0f5, 0x7})
        /Users/delta/devel/kubecost/model/opencost/core/pkg/filter/ops/ops.go:53 +0x7c
github.com/opencost/opencost/core/pkg/filter/ops.toFieldAndKey[...]({0x1041ec0f5, 0x7})
        /Users/delta/devel/kubecost/model/opencost/core/pkg/filter/ops/ops.go:156 +0xd0
github.com/opencost/opencost/core/pkg/filter/ops.identifier[...](...)
        /Users/delta/devel/kubecost/model/opencost/core/pkg/filter/ops/ops.go:160
```

* N/A

* Fixes a rare nil panic when filter fields are accessed concurrently

* N/A

## How was this PR tested?
* Error which was reliably occurring locally no longer occurs

### DIFF
--- a/core/pkg/filter/allocation/parser.go
+++ b/core/pkg/filter/allocation/parser.go
@@ -1,10 +1,6 @@
 package allocation
 
-import (
-	"sync"
-
-	"github.com/opencost/opencost/core/pkg/filter/ast"
-)
+import "github.com/opencost/opencost/core/pkg/filter/ast"
 
 // a slice of all the allocation field instances the lexer should recognize as
 // valid left-hand comparators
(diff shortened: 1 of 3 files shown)


mbolt35:
Major oversight on my side (as if a global map wasn't "bad" enough). This needs a broader solution that I can assist with. This pattern exists in a few places in opencost as well as some custom filter pipelines internally. Your solution would fix the issue, but it's got a lot of splash damage. 

I don't know why I made the initializer lazy, but it's definitely not worth a lock. There are a few places we need to change this, but it's better to just populate the map in `func init()` which is guaranteed to be called only once by a single thread. If we wanted to preserve the lazy initialization (which again, I don't really think it's necessary), we could introduce a `sync.Once`. 

Once initialized, concurrent reads from the map are fine (since there won't be any more writes). 

TLDR; no lock needed, just move initialization of the map to `func init()`. assets and cloud packages also need the same treatment.

michaelmdresser:
Much better than my hacky solution, good suggestion, I've implemented it

mbolt35 (approved):
LGTM!

PR #2477, opened 2024-01-27
pkg/prom: prune unused containerKey

alrs:
This removes the unused `containerKey` type and its unused functions.


mattray:
@alrs thanks for the PR. @AjayTripathy or any of the other @opencost/opencost-maintainers from Kubecost, is this from Kubecost?

michaelmdresser (approved):
Private functions and types. If Opencost continue to compile after their removal it should be fine.

PR #2388, opened 2023-12-15
feat: 2323 adding version and commit on opencost-ui logs

brito-rafa:
## What does this PR change?
* It adds version and commit on opencost-ui logs

## Does this PR relate to any other PRs?
* No

## How will this PR impact users?
* upon running `kubectl logs <pod> -c opencost-ui`, users will see the version and commit SHA of the container.

## Does this PR address any GitHub or Zendesk issues?
* Closes ... 2323

## How was this PR tested?
* See commnad below:
```
% docker run quay.io/brito_rafa0/opencost-ui:testa
Unable to find image 'quay.io/brito_rafa0/opencost-ui:testa' locally
testa: Pulling from brito_rafa0/opencost-ui
Digest: sha256:bd2d828ae209e4f314159821225783d61c3439e2f0b94339393ef0c19f86ae54
Status: Downloaded newer image for quay.io/brito_rafa0/opencost-ui:testa
running with BASE_URL=/model
Starting ui version 2323-ui-version-c2bc77c0-WIP (c2bc77c0) ## <-----------------------------
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
```
## Does this PR require changes to documentation?
* No

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next OpenCost release? If not, why not?
* Yes

PR #2479, opened 2024-01-29
Add version rendered at bottom of page.

### LINKED ISSUE #2323: OpenCost UI doesn't report version/build number
**Describe the bug**
When the costmodel container starts up it reports 
```
2023-11-21T20:19:37.749036463Z INF Starting cost-model version dev (HEAD)
```
where `HEAD` is the actual SHA when not building from source. This is super useful for rebuildling older releases if they weren't tagged properly.

mattray:
## What does this PR change?
* Adds the rendered string "OpenCost version: $VERSION ($HEAD)" to the footer of the OpenCost UI where environment variables can be used to configure what's in the footer.
Screenshot 2024-01-29 at 10 39 50 PM

* https://github.com/opencost/opencost/pull/2388 added it to the logging

* The UI will show which version is installed, making it easier for debugging.

* Closes https://github.com/opencost/opencost/issues/2323

## How was this PR tested?
* Built locally and deployed to K8s

* No

* It should be mentioned in the release notes as a useful debugging tip.


ui/docker-entrypoint.sh
```diff
@@ -9,9 +9,15 @@ else
     sed -i "s^{PLACEHOLDER_BASE_URL}^$BASE_URL^g" /var/www/*.js
 fi
 
+if [[ ! -z "$OPENCOST_FOOTER_CONTENT" ]]; then
+    sed -i "s^PLACEHOLDER_FOOTER_CONTENT^$OPENCOST_FOOTER_CONTENT^g" /var/www/*.js
+else
+    sed -i "s^PLACEHOLDER_FOOTER_CONTENT^OpenCost version:$VERSION ($HEAD)^g" /var/www/*.js
+fi
+
```
nealormsbee: Couple of questions on this. Sorry, I've been out of this codebase for quite a while.

1. So the developer can set a `$OPENCOST_FOOTER_CONTENT` env var, _or_ env vars for `$VERSION` and `$HEAD` to get a pre-formatted footer message? Are `$VERSION` and `$HEAD` guaranteed to be set? If they aren't guaranteed we might want a case to cover that, so that we don't wind up with `Opencost version: ()` by default.

2. Should `$HEAD` be something like `$COMMIT`? I guess whatever commit you're running from is technically the head of wherever you are...but idk.

3. Also, if it is typically set, does the `$VERSION` string contain a leading blank?  I think we want e.g. `version: 2.0` over `version:2.0`

nealormsbee (commented):
I can't technically approve these PRs because I'm not a maintainer 😅 
But I had a couple of questions!

ui/docker-entrypoint.sh
mattray: Thanks for the feedback!

1) The $VERSION and $HEAD are set with `--build-arg version=` and `--build-arg commit=` passed into the Dockerfile builds and `version` and `commit` have defaults, so they will be set.

2) Passed in at the build the same way as the cost-model build, which gets logging like
```
Starting cost-model version 1.108.0 (25d0064)
```

3) I'm making the change.

nealormsbee: Thanks for the explanation 😄

mattray: The package-lock.json doesn't work, I'm having to dissect the react dependencies, that's the cause for the hold up.

mattray:
@nealormsbee refactored and re-tested, seems good now.

nealormsbee (approved):
Looks good to me!

cliffcolvin (approved):

PR #2492, opened 2024-01-30
Fix pathing issue for multi-cloud water

Sean-Holcomb:
## What does this PR change?
* 

This PR addresses a bug where the Cloud Cost container is unable to find the location of the multi-cloud config file, which causes it to not run on those configurations. This has manifested in the nightly environment. 

The bug appears to have been introduced by https://github.com/opencost/opencost/pull/2317 which relies on the env variable "KUBERNETES_PORT" being set automatically. This does not appear to always happen which leads me to believe that we should not rely on it. 

This PR is hopefully a compromise between the old way path was determined and the ability to configure it. It leaves in place a check for the "CLOUD_COST_CONFIG_PATH" env variable, which if not set defaults to the standard secret path.

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/cloud/config/watcher.go
+++ b/pkg/cloud/config/watcher.go
@@ -3,7 +3,6 @@ package config
 import (
 	"fmt"
 	"io/ioutil"
-	"path"
 
 	"github.com/opencost/opencost/pkg/cloud"
 	"github.com/opencost/opencost/pkg/cloud/alibaba"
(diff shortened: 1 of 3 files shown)


pkg/cloud/config/watcher.go
```diff
@@ -239,31 +237,17 @@ type MultiCloudWatcher struct {
...
-		multiConfigPath = path.Join("/var", cloudIntegrationSecretPath)
-		exists, err = fileutil.FileExists(multiConfigPath)
-		if err != nil {
-			log.Errorf("MultiCloudWatcher:  error checking file at '%s': %s", multiConfigPath, err.Error())
-		}
-
-		// If config does not exist implies that this configuration method was not used
-		if !exists {
-			return nil
-		}
```
Sean-Holcomb, line 266: This block of code will not be necessary with 2.0 and was never necessary for opencost users

Sean-Holcomb:
Diagnosis may have been inaccurate, will reopen if necessary

PR #2494, opened 2024-01-31
workflow for opencost builds

cliffcolvin:
## What does this PR change?
*  The intent of this PR is to create the first stab at opencost release. This process has historically been tightly coupled with kubecost releases and we aim to create a separate release process so that features and fixes can be released at a cadence that is more in alignment with the community supporting opencost.

NEXT STEPS - 
- We need to automate the branch creation/tagging step
- We need to automate the helm update and checkin process

We can handle some of these things manually for now just need to get this moving forward.

* 

* No functional or frontend changes here, but the release process will be moved to opencost and the images will be hosted in ghcr as well as renamed to opencost.

* Closes ...

## How was this PR tested?
* In Progress

* 

*

### DIFF
--- a/.github/workflows/build-and-publish-release.yml
+++ b/.github/workflows/build-and-publish-release.yml
@@ -53,16 +53,19 @@ jobs:
           echo "IMAGE_TAG_UI_VERSION=ghcr.io/opencost/opencost-ui:${{ inputs.release_version }}" >> $GITHUB_ENV
           echo "IMAGE_TAG_QUAY=quay.io/kubecost1/kubecost-cost-model:${{ steps.sha.outputs.OC_SHORTHASH }}" >> $GITHUB_ENV
           echo "IMAGE_TAG_LATEST_QUAY=quay.io/kubecost1/kubecost-cost-model:latest" >> $GITHUB_ENV
-          echo "IMAGE_TAG_VERSION_QUAY=quay.io/kubecost1/kubecost-cost-model:prod-v${{ inputs.release_version }}" >> $GITHUB_ENV
+          echo "IMAGE_TAG_VERSION_QUAY=quay.io/kubecost1/kubecost-cost-model:prod-${{ inputs.release_version }}" >> $GITHUB_ENV
           echo "IMAGE_TAG_UI_QUAY=quay.io/kubecost1/opencost-ui:${{ steps.sha.outputs.OC_SHORTHASH }}" >> $GITHUB_ENV
           echo "IMAGE_TAG_UI_LATEST_QUAY=quay.io/kubecost1/opencost-ui:latest" >> $GITHUB_ENV
-          echo "IMAGE_TAG_UI_VERSION_QUAY=quay.io/kubecost1/opencost-ui:${{ inputs.release_version }}" >> $GITHUB_ENV
+          echo "IMAGE_TAG_UI_VERSION_QUAY=quay.io/kubecost1/opencost-ui:prod-${{ inputs.release_version }}" >> $GITHUB_ENV
      
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
       
       - name: Set up just
         uses: extractions/setup-just@v1
+      
+      - name: Install crane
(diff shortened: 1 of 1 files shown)


michaelmdresser (approved):

.github/workflows/build-and-publish-release.yml
```diff
@@ -0,0 +1,102 @@
+name: Build and Publish Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: "Version. Please DO NOT include the 'v' prefix"
+        required: true
```
michaelmdresser, line 8: I think @mattray had an idea that we could have a workflow like this trigger when a version tag appears on the repo -- this condition: https://github.com/kubecost/kubectl-cost/blob/a9d05d340121a823232b6fe05be6a3d35ea72c78/.github/workflows/build-release.yaml#L4-L6
```
on:
  push:
    tags:
      - 'v*'
```

Then all you have to do is add a `v.....` tag to the repo and the workflow will automatically kick off. If that's the long-term direction we want to go, it might be valuable to start there as well.

.github/workflows/build-and-publish-release.yml
```diff
@@ -0,0 +1,102 @@
...
+          just build '${steps.tags.outputs.IMAGE_TAG_VERSION_QUAY}'
+      - name: Build and push (multiarch) OpenCost UI
+        working-directory: ./opencost/ui
+        run: |
+          just build '${steps.tags.outputs.IMAGE_TAG_UI}'
+          just build '${steps.tags.outputs.IMAGE_TAG_UI_LATEST}'
+          just build '${steps.tags.outputs.IMAGE_TAG_UI_VERSION}'
+          just build '${steps.tags.outputs.IMAGE_TAG_UI_QUAY}'
+          just build '${steps.tags.outputs.IMAGE_TAG_UI_LATEST_QUAY}'
+          just build '${steps.tags.outputs.IMAGE_TAG_UI_VERSION_QUAY}'
```
michaelmdresser: `crane` is a great option here to avoid extra docker build work
Install in a separate step with
```yaml
      - name: Install crane
        uses: imjasonh/setup-crane@v0.1
```

and then you can do something like this

```sh
just build '${steps.tags.outputs.IMAGE_TAG}'
crane copy  '${steps.tags.outputs.IMAGE_TAG}' '${steps.tags.outputs.IMAGE_TAG_LATEST}'
crane copy  '${steps.tags.outputs.IMAGE_TAG}' '${steps.tags.outputs.IMAGE_TAG_VERSION}'
# ...etc
```

michaelmdresser (approved):
Still looks good!

.github/workflows/build-and-publish-release.yml
```diff
@@ -0,0 +1,102 @@
...
+          pushd ./opencost
+          echo "OC_SHORTHASH=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+          popd
+
+      - name: Set OpenCost Image Tags
+        id: tags
+        run: |
+          echo "IMAGE_TAG=ghcr.io/opencost/opencost:${{ steps.sha.outputs.OC_SHORTHASH }}" >> $GITHUB_ENV
+          echo "IMAGE_TAG_LATEST=ghcr.io/opencost/opencost:latest" >> $GITHUB_ENV
+          echo "IMAGE_TAG_VERSION=ghcr.io/opencost/opencost:${{ inputs.release_version }}" >> $GITHUB_ENV
```
michaelmdresser, line 50: A question for @mattray: do you want images tagged like `opencost:v1.2.3` or like `opencost:1.2.3`?

.github/workflows/build-and-publish-release.yml
cliffcolvin: ok, I can add a new pr with that, since I already put the tag out I need to get this one built. I'll throw up a second pr later.

.github/workflows/build-and-publish-release.yml
cliffcolvin: added

PR #2499, opened 2024-02-03
use `policy/v1` for `PodDisruptionBudget`

a7i:
## What does this PR change?
* uses policy/v1 for listing PDBs

* No

* policy/v1beta1 is no longer served as of 1.25

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/clustercache/clustercache.go
+++ b/pkg/clustercache/clustercache.go
@@ -9,7 +9,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/api/policy/v1beta1"
+	policyv1 "k8s.io/api/policy/v1"
 	stv1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


mattray:
@a7i thanks for opening this! Looking at the [Kubernetes API documentation](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125) I see this API was removed in 1.25 and available since 1.21. This means we should update 
https://www.opencost.io/docs/installation/install
and Kubecost should update
https://docs.kubecost.com/install-and-configure/install/environment#supported-kubernetes-versions
accordingly.

mattray (approved):
I want another @opencost/opencost-maintainers to merge it, but it looks good to me.

michaelmdresser (approved):
If I'm reading K8s notes correctly, `policy/v1` `PodDisruptionBudget` was added in K8s v1.21 (https://github.com/kubernetes/sig-release/blob/master/releases/release-1.21/release-notes/release-notes-draft.md). I think that's enough prior versions of support for us to roll forward with only supporting policy/v1 PDBs.

EDIT: Matt, I can't read. Thank you for doing the research.

michaelmdresser:
Hey @AjayTripathy -- heads up that this will bump KC's min K8s version from v1.20 to v1.21, but I think it might only affect Spot Commander.

PR #2500, opened 2024-02-04
Fix UI Build Command in CONTRIBUTING.md

ChihweiLHBird:
It's `just build` rather than `just build-ui`. Fixing it in the doc.


mattray (approved):

mattray:
@ChihweiLHBird good catch, thanks!

PR #2514, opened 2024-02-07
Correct GCP's default discount percentage

namm2:
## What does this PR change?
* Correct the GCP's default discount value to 0%, this should be the correct value when this field is not set, also all other clouds' configs are 0%.

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/cloud/gcp/provider.go
+++ b/pkg/cloud/gcp/provider.go
@@ -164,7 +164,7 @@ func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
 		return nil, err
 	}
 	if c.Discount == "" {
-		c.Discount = "30%"
+		c.Discount = "0%"
 	}
 	if c.NegotiatedDiscount == "" {
 		c.NegotiatedDiscount = "0%"


mattray:
@namm2 Good catch! I've confirmed the other providers have this set to zero. @opencost/opencost-maintainers any reason this may have been set differently?

nikovacevic:
Historically, we've set this to 30% intentionally to match [GCP's sustained use discount](https://cloud.google.com/compute/docs/sustained-use-discounts). Maybe product wants to chime in with regard to the best decision here. Do we assume sustained use or not? Or, even if we do assume sustained use, should we put the onus on the user to set this discount themselves?

mattray:
If we don't set it to 0%, could we add the rationale as a comment?

namm2:
AFAIU, SUD discount is variable based on the %usage and machine type (hence up to 30%) then the static value 30% is still not correct out-of-the-box. Also there's another Commited Use Discount which is variable too, I'd leave it to users to set their discount value than hard code it.

mbolt35:
@AjayTripathy this has been part of the product as long as I've been around. Would love your take here.

mattray:
@AjayTripathy you want to comment?

AjayTripathy:
We just take an average 30% discount because this is meant to be an estimate. This is user configurable so I'm not sure there's a compelling reason to change this right now.

namm2:
> We just take an average 30% discount because this is meant to be an estimate. This is user configurable so I'm not sure there's a compelling reason to change this right now.

Yes it's configurable but I mean the _default_ value should be 0% like other cloud providers, and it's up to end users to set this.

PR #2520, opened 2024-02-09
Add persistence for config controller

Sean-Holcomb:
## What does this PR change?
This PR adds persistence for the configuration controller which allows for configurations to be modified via handlers and have those changes persist beyond pod restarts.

At a high level the Config Controller is able to add its own configs. It can enable or disable configs from all sources, however it can only delete configs that it added. This limitation is because even if the controller deleted a config from a secret or the helm values it would just reappear, since it does not have control of that source.

Additionally it changes the way other configuration sources are handled such that if a config is no longer present from that source the corresponding status will be removed so that they can be deleted at their source.

* 

* No impact on OpenCost users

* Closes ...

## How was this PR tested?
* Added unit tests for Controller

* 

*


michaelmdresser (approved):
Nice tests. Looks like GitHub also found a fmt.Errorf called without the result being used.

PR #2521, opened 2024-02-09
Allow opencost-ui to run with a read-only root filesystem

kaitimmer:
## What does this PR change?

Allow to run with a read-only root file system.

This is important for people who need to run this in Kubernetes with a policy set that only allows pods that have this set:

```yaml
          securityContext:
            readOnlyRootFilesystem: true
```

No 

It does not directly affect users. The container continues to run with the same behavior as before. The only difference is, that now one can mount an emptyDir to /var/www and have the root filesystem read-only.

Setting this in the Helm charts as a default would be possible. I guess that would make sense from a security point of view.


mattray:
@kaitimmer thanks for the submission. I'll test it and verify it doesn't cause a regression

mattray (approved):
Works for me.

PR #2523, opened 2024-02-12
Ajay tripathy fix gpucost

AjayTripathy:
## What does this PR change?
* This needs to be a per-unit GPUcost, not a total

* https://github.com/opencost/opencost/pull/2516/files

* Fix GPU overcharging for nodes with multiple GPU

* Closes ...

## How was this PR tested?
* Manually

* No

* Yes, should be in next release and cherrypicked back

### DIFF
--- a/pkg/cloud/provider/csvprovider.go
+++ b/pkg/cloud/provider/csvprovider.go
@@ -279,7 +279,7 @@ func (c *CSVProvider) NodePricing(key models.Key) (*models.Node, models.PricingM
 			}
 		}
 		totalCost := hourly * float64(count)
-		node.GPUCost = fmt.Sprintf("%f", totalCost)
+		node.GPUCost = fmt.Sprintf("%f", hourly)
 		nc, err := strconv.ParseFloat(node.Cost, 64)
 		if err != nil {
 			log.Errorf("Unable to parse %s as float", node.Cost)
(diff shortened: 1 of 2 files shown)


pkg/cloud/provider/csvprovider.go
```diff
@@ -279,7 +279,7 @@ func (c *CSVProvider) NodePricing(key models.Key) (*models.Node, models.PricingM
 			}
 		}
 		totalCost := hourly * float64(count)
-		node.GPUCost = fmt.Sprintf("%f", totalCost)
+		node.GPUCost = fmt.Sprintf("%f", hourly)
```
michaelmdresser, line 282: If this has to be modified, doesn't the line a little below 
```
node.Cost = fmt.Sprintf("%f", nc+totalCost)
```

have to be modified?

michaelmdresser: Also, I don't understand why this isn't `totalCost` -- if there are multiple GPUs, shouldn't the cost be multiplied? Or is this because `GPUClassPricing` contains prices that total the GPUs on a node, rather than breaking them out individually?

AjayTripathy: No, don't think so. node.GPUCost is unit cost of a GPU; node.Cost is totalCost

AjayTripathy: This is the field that gets transformed into node_gpu_hourly_cost . Its sister queries are node_cpu_hourly_cost and node_ram_hourly_cost which are unit prices per 1cpu and 1GB RAM.

michaelmdresser: > No, don't think so. node.GPUCost is unit cost of a GPU; node.Cost is totalCost

Oh, huh, that's a little confusing; sorry I think I thought we were in a different part of the Opencost codebase. Now that I see we're in provider land, this makes more sense.

> This is the field that gets transformed into node_gpu_hourly_cost . Its sister queries are node_cpu_hourly_cost and node_ram_hourly_cost which are unit prices per 1cpu and 1GB RAM.

Perfect, this makes way more sense now.

michaelmdresser (approved):

PR #2534, opened 2024-02-14
More defensive approach to retrieving Node labels for a node

thomasvn:
## What does this PR change?
* Add NodeLabels one by one, instead of assigning the map of labels directly to the result.
* This protects against the case where there exist multiple `kube_node_labels` metrics for a single node.

* N/A

* Fixes a bug for users who have more than one `kube_node_labels` per node in their Prometheus environment.

* N/A

## How was this PR tested?
* Ran locally in an environment that _does not_ have duplicated `kube_node_labels`. Functioning as expected when querying the Assets API.

* N/A

*

### DIFF
--- a/pkg/costmodel/cluster_helpers.go
+++ b/pkg/costmodel/cluster_helpers.go
@@ -640,7 +640,12 @@ func buildLabelsMap(
 		// ingested label data. This removes the label_ prefix that prometheus
 		// adds to emitted labels. It also keeps from ingesting prometheus labels
 		// that aren't a part of the asset.
-		m[key] = result.GetLabels()
+		if _, ok := m[key]; !ok {
+			m[key] = map[string]string{}
+		}
+		for k, l := range result.GetLabels() {
+			m[key][k] = l
+		}
 	}
 	return m
 }


AjayTripathy (approved):

AjayTripathy:
This was really impactful. I'd like to get this change into the agent

pkg/costmodel/cluster_helpers.go
```diff
@@ -640,7 +640,12 @@ func buildLabelsMap(
 		// ingested label data. This removes the label_ prefix that prometheus
 		// adds to emitted labels. It also keeps from ingesting prometheus labels
 		// that aren't a part of the asset.
-		m[key] = result.GetLabels()
+		if _, ok := m[key]; !ok {
+			m[key] = map[string]string{}
+		}
+		for k, l := range result.GetLabels() {
+			m[key][k] = l
```
michaelmdresser, line 647: This is _way_ better, but will still result in one value for a `key` "winning" in the event of a conflict. It's not really possible to avoid that. I'm wondering if, in the future, there is a way for us to preserve conflicts instead of picking one winner silently. Not for this PR.

michaelmdresser (approved):
Really, really nice fix. How in the world did you discover it?

pkg/costmodel/cluster_helpers.go
thomasvn: Right, good catch! If there are two `kube_node_labels` metrics describing the same node and using the same labels, the second label read wins (example below). I can't think of a simple solution right now to resolve those conflicts, but we may be able to implement something similar to the `honor_labels` Prometheus config ([docs ref](https://prometheus.io/docs/prometheus/latest/configuration/configuration/)).

```
kube_node_labels{node="gke-qa-gcp1-default-pool-169fb9bb-0zz6", label_topology_gke_io_zone="us-east1-d"}
kube_node_labels{node="gke-qa-gcp1-default-pool-169fb9bb-0zz6", label_topology_gke_io_zone="us-east1-a"}
```

thomasvn:
@AjayTripathy @michaelmdresser Appreciate the reviews! I don't have write access so if either of you could merge that'd be great!

PR #2536, opened 2024-02-15
Support LB and PV item-level reconciliation

nikovacevic:
## What does this PR change?
* Adds Adjustment fields to LB and PV types in order to allow item-level reconciliation.

* Required for https://github.com/kubecost/kubecost-cost-model/pull/2179

* N/A

* N/A

## How was this PR tested?
* Manually

* No

* Yes

### DIFF
--- a/core/pkg/opencost/allocation.go
+++ b/core/pkg/opencost/allocation.go
@@ -122,11 +122,12 @@ func (orig LbAllocations) Clone() LbAllocations {
 }
 
 type LbAllocation struct {
-	Service string  `json:"service"`
-	Cost    float64 `json:"cost"`
-	Private bool    `json:"private"`
-	Ip      string  `json:"ip"`    //@bingen:field[version=19]
-	Hours   float64 `json:"hours"` //@bingen:field[version=21]
+	Service    string  `json:"service"`
+	Cost       float64 `json:"cost"`
+	Private    bool    `json:"private"`
+	Ip         string  `json:"ip"`         //@bingen:field[version=19]
+	Hours      float64 `json:"hours"`      //@bingen:field[version=21]
+	Adjustment float64 `json:"adjustment"` //@bingen:field[ignore]
 }
 
 func (lba *LbAllocation) SanitizeNaN() {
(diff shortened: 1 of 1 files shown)


nik-kc (approved):

core/pkg/opencost/allocation.go
```diff
@@ -122,11 +122,12 @@ func (orig LbAllocations) Clone() LbAllocations {
...
 
 type LbAllocation struct {
-	Service string  `json:"service"`
-	Cost    float64 `json:"cost"`
-	Private bool    `json:"private"`
-	Ip      string  `json:"ip"`    //@bingen:field[version=19]
-	Hours   float64 `json:"hours"` //@bingen:field[version=21]
+	Service    string  `json:"service"`
+	Cost       float64 `json:"cost"`
+	Adjustment float64 `json:"adjustment"` //@bingen:field[ignore]
```
michaelmdresser: Hey @mbolt35 when we add fields to bingen-serialized structs but the new field is ignored, does it still have to be added as the last field in the struct's list of fields? I recall this was a requirement for bingen fields in general but it may not apply to ignored fields.

ameijer: I would move to the end TBH just to avoid any chance of this

core/pkg/opencost/allocation.go
```diff
@@ -311,6 +312,7 @@ func (pvk *PVKey) FromString(key string) error {
 type PVAllocation struct {
 	ByteHours  float64 `json:"byteHours"`
 	Cost       float64 `json:"cost"`
+	Adjustment float64 `json:"adjustment"` //@bingen:field[ignore]
```
ameijer: I would move to the end, in line with @michaelmdresser 's comments above

nikovacevic: Good point, will do!

core/pkg/opencost/allocation.go
nikovacevic: Yep, good call out. I'll move it just to be safe.

nikovacevic: Done

core/pkg/opencost/allocation.go
nikovacevic: Done

ameijer (approved):
LGTM

core/pkg/opencost/allocation.go
mbolt35: @nikovacevic @ameijer @michaelmdresser I think the added caution is understandable, and I think it's perfectly fine to get into a habbit of just adding new things to the end. However, `@bingen:field[ignore]` actually prevents the fields from appearing on the struct definition that gets passed to the generator, so they can appear in any order. This is the only circumstance where order doesn't matter though

michaelmdresser (approved):

PR #2407, opened 2023-12-19
Revert "Revert "added a separate client for prometheus scrape config endpoint""

cliffcolvin:
Reverts opencost/opencost#2316

We temporarily reverted this change to accommodate time to fix downstream issues with this change. I am un-reverting this now to test and merge this.

PR #2567, opened 2024-02-26
use status/buildinfo endpoint to get status instead of status/config added variable to set static scrape interval

AndrewChubatiuk:
- Use status/buildinfo endpoint instead of status/config to provide support for VictoriaMetrics
- Added KUBECOST_SCRAPE_INTERVAL env variable to make scrape interval discovery optional

## What does this PR change?
* alternative approach to #2407 to enable VM support

* #2407

* 

* Closes ...

## How was this PR tested?
* 

* 

*


mattray:
@AndrewChubatiuk thanks, this looks useful to me. @opencost/opencost-maintainers anyone want to weigh in?

michaelmdresser (commented):
This looks straightforward. What is the difference between `api.Config` and `api.Buildinfo`? Is there any meaningful difference when using Prometheus?

AndrewChubatiuk:
@michaelmdresser VictoriaMetrics in a cluster mode has scrape config endpoint only on VMAgent component but to scrape metrics it should use VMSelect. It means that with a current approach to support VMAgent and VMSelect endpoints it's required to have two clients (which was introduced in #2407). To use a single client and keep it compatible with other Prometheus compatible solutions I've updated config endpoint (which is utilized to get a monitoring cluster status and scrape interval) to buildinfo and added ability to configure custom scrape interval manually

michaelmdresser:
That makes sense to me. However, I'm trying to make sure that this does not break functionality when using Prometheus. Is there a meaningful difference between `api.Config` and `api.Buildinfo` when using Prometheus? Have you tested this PR by running it against a Prometheus instance in addition to VictoriaMetrics?

AndrewChubatiuk:
yes, tested it with both prometheus and VM. there's no meaningful difference. left `api.Config` only in place, where it queries scrape interval, but the rest calls are just required to understand that prometheus is up and available.

michaelmdresser (approved):
OK, thanks for answering my questions

PR #2569, opened 2024-02-26
impl basic structs for plugins

ameijer:
## What does this PR change?
* roughs-in the basic structs that are going to be passed around for plugin costs

* 

* no impact, just defining the ways that costs will be passed around 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/core/pkg/model/custom_cost_request.go
+++ b/core/pkg/model/custom_cost_request.go
@@ -1,0 +1,20 @@
+package model
+
+import (
+	"time"
+
+	"github.com/opencost/opencost/core/pkg/opencost"
+)
+
+type CustomCostRequest struct {
+	TargetWindow *opencost.Window
+	Resolution   time.Duration
+}
+
+func (c *CustomCostRequest) GetTargetWindow() *opencost.Window {
+	return c.TargetWindow
+}
+
+func (c *CustomCostRequest) GetTargetResolution() time.Duration {
+	return c.Resolution
+}
--- a/core/pkg/model/custom_cost_response.go
+++ b/core/pkg/model/custom_cost_response.go
@@ -1,0 +1,241 @@
+package model
+
+import "github.com/opencost/opencost/core/pkg/opencost"
+
+type CustomCostResponse struct {
+	Metadata   map[string]string
+	Costsource string
+	Domain     string
+	Version    string
+	Currency   string
+	Window     opencost.Window
+	Costs      [][]*CustomCost
+	Errors     []error
+}
+
+type CustomCost struct {
+	Metadata       map[string]string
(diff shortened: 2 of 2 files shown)


nik-kc (approved):
Looks good to me!

cliffcolvin (approved):

core/pkg/model/custom_cost_request.go
michaelmdresser, line 1: I'm going to chime in for @mbolt35 here -- please don't name files with underscores unless taking advantage of the `_test.go` or `_<platform>.go` special behavior (https://stackoverflow.com/questions/25161774/what-are-conventions-for-filenames-in-go). 

This particular file name should behave fine, but we like to encourage avoiding underscores for consistency.

core/pkg/model/custom_cost_request.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/opencost/opencost/core/pkg/opencost"
+)
+
+type CustomCostRequest struct {
+	TargetWindow *opencost.Window
+	Resolution   time.Duration
+}
+
+func (c *CustomCostRequest) GetTargetWindow() *opencost.Window {
```
michaelmdresser, line 14: Why have getters if all fields are exported and you aren't copying them?

core/pkg/model/custom_cost_response.go
```diff
@@ -0,0 +1,241 @@
...
+	Costsource string
+	Domain     string
+	Version    string
+	Currency   string
+	Window     opencost.Window
+	Costs      []*CustomCost
+	Errors     []error
+}
+
+type CustomCost struct {
```
michaelmdresser, line 16: Would you add godoc strings at least to the structs explaining what they are? E.g. the FOCUS inspiration etc.

core/pkg/model/custom_cost_request.go
ameijer: fixed, thanks for tip

core/pkg/model/custom_cost_request.go
ameijer: for the RPC, I think they have to implement an interface. I have just added that interface definition

core/pkg/model/custom_cost_response.go
ameijer: will do. These will come in via the next PR

core/pkg/model/custom_cost_request.go
michaelmdresser: Ah, this probably answers the questions I just asked on the other PR. Perhaps we continue the discussion there; I'm curious about this requirement

Issue #2473, opened 2024-01-25
Possible scalability issue for clusters with large amount of resources.


lmello:
**Describe the bug**
we have a development clusters which has 100k+ kubernetes resources, our prod clusters are way bellow of that around 10k resources. 

In this specific development cluster opencost is getting an error while trying to retrieve the initial list of resources. 

```
I0125 16:42:24.260539       1 trace.go:205] Trace[2147448565]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.25.3/tools/cache/reflector.go:169 (25-Jan-2024 16:42:09.857) (total time: 14403ms):
Trace[2147448565]: ---"Objects listed" error:<nil> 14386ms (16:42:24.243)
Trace[2147448565]: [14.403251883s] [14.403251883s] END
```
Initial list of deployments in this specific cluster for kubernetes watchers usually could take up to 1 minute.  We know that based on other services that watch k8s events we have been running on this cluster. 

The following code is where all informers are started in opencost.
https://github.com/opencost/opencost/blob/develop/pkg/clustercache/clustercache.go#L112-L126 

There is usually some approaches to optimize that we successfully used before: 
* expose client-go flags
* use chunking
* provide ability to filter by label selector (this won't work in this case, because we would want to monitor the cost for all resources)
* look into optimising the informers vs shared informers 

We started the investigation today and are planning to first attempt to use chunking to reduce the load on etcd and prevent the failure. 

I am creating this issue to check at what number of resources was OpenCost and Kubecost were tested against or reported to support and document the issue we are facing. 

**To Reproduce**
1. Install opencost in a cluster with 100k resources
2. Check opencost logs
3. See error 
```
I0125 16:42:24.260539       1 trace.go:205] Trace[2147448565]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.25.3/tools/cache/reflector.go:169 (25-Jan-2024 16:42:09.857) (total time: 14403ms):
Trace[2147448565]: ---"Objects listed" error:<nil> 14386ms (16:42:24.243)
Trace[2147448565]: [14.403251883s] [14.403251883s] END
```

**Expected behavior**
No error should happen. 

**Screenshots**
NA
**Which version of OpenCost are you using?**
1.107

**Additional context**
NA

AjayTripathy:
[...]

PR #2570, opened 2024-02-27
Switch from JSON to protobuf for kube-api requests

r2k1:
## What does this PR change?

Switch from JSON to protobuf for kube-api requests. Improve performance. I assume it should also reduce load on kube-api.

On my test it:
- reduced memory allocation from 350MB to 300MB (for data fetching)
- reduced object allocation from 1.700.000 to 1.100.000 objects (for data fetching)
- max memory usage (in a test) reduced from 300MB to 240MB. (total for all app)
- added a new frame for each object type to a profiler, previously all of them were bundled together. (Now it visible pods are number one, followed by a ReplicaSet on my cluster)
- reduced CPU consumption for resource parsing from 520ms to 180

Memory allocation space diff:
Screenshot 2024-02-27 at 12 58 55

Related issue #2473

For the future references, my test setup:

AKS cluster with ~2500 pods.

```
package main

import (
	"fmt"
	"io"
	"net/http"
	"os"
	"runtime"
	"testing"
	"time"

	"github.com/opencost/opencost/pkg/cmd"
	"github.com/stretchr/testify/require"
)

func TestMemory(t *testing.T) {
	os.Setenv("KUBECONFIG", "/Users/r2k1/p/playground/performance-test/tc1.yaml")
	os.Setenv("PROMETHEUS_SERVER_ENDPOINT", "http://localhost:9092")
	os.Setenv("KUBERNETES_PORT", "443")
	os.Setenv("PPROF_ENABLED", "true")
	os.Setenv("CACHE_WARMING_ENABLED", "false")
	os.Setenv("DISABLE_AGGREGATE_COST_MODEL_CACHE", "true")
	os.Setenv("MAX_QUERY_CONCURRENCY", "5")

	go func() {
		err := cmd.Execute(nil)
		require.NoError(t, err)
	}()

	maxUsage := uint64(0)
	go trackMaxMemoryUsage(100*time.Microsecond, &maxUsage)
	time.Sleep(30 * time.Second)

	end := time.Now()
	start := end.Add(-time.Hour * 24)
	testFetch(t, fmt.Sprintf("http://localhost:9003/metrics"))
	testFetch(t, fmt.Sprintf("http://localhost:9003/allocation/compute?aggregate=namespace,controllerKind,controller,label:app,label:team,label:pod_template_hash&idleByNode=true&includeIdle=true&includeProportionalAssetResourceCosts=true&step=window&window=%s,%s", start.Format("2006-01-02T15:04:05Z"), end.Format("2006-01-02T15:04:05Z")))
	testFetch(t, fmt.Sprintf("http://localhost:9003/assets?window=%s,%s", start.Format("2006-01-02T15:04:05Z"), end.Format("2006-01-02T15:04:05Z")))
	t.Log("Max memory usage, KiB:", maxUsage/1024)
}


func testFetch(t *testing.T, url string) {
	startTime := time.Now()
	resp, err := http.Get(url)
	require.NoError(t, err)
	require.Less(t, resp.StatusCode, 300)
	require.GreaterOrEqual(t, resp.StatusCode, 200)
	data, err := io.ReadAll(resp.Body)
	require.NoError(t, err)
	t.Logf("%s, %v MiB, %s", time.Since(startTime), len(data)/1024/1024, url)
}

func trackMaxMemoryUsage(interval time.Duration, maxUsage *uint64) {
	var memStats runtime.MemStats

	for {
		runtime.ReadMemStats(&memStats)

		// Update maxUsage if the current HeapAlloc is greater
		if memStats.HeapAlloc > *maxUsage {
			*maxUsage = memStats.HeapAlloc
		}

		time.Sleep(interval)
	}
}
```

### DIFF
--- a/pkg/kubeconfig/loader.go
+++ b/pkg/kubeconfig/loader.go
@@ -17,7 +17,14 @@ func LoadKubeconfig(path string) (*rest.Config, error) {
 		loadingRules.ExplicitPath = path
 	}
 	loader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
-	return loader.ClientConfig()
+	config, err := loader.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+	config.UserAgent = "opencost"
+	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
+	config.ContentType = "application/vnd.kubernetes.protobuf"
+	return config, nil
 }
 
 // LoadKubeClient accepts a path to a kubeconfig to load and returns the clientset
--- a/pkg/kubeconfig/loader.go
+++ b/pkg/kubeconfig/loader.go
@@ -24,6 +24,8 @@ func LoadKubeconfig(path string) (*rest.Config, error) {
 		return nil, fmt.Errorf("loading kubeconfig: %w", err)
 	}
 	config.UserAgent = "opencost"
+	// use protobuf for faster serialization instead of default json
+	// https://kubernetes.io/docs/reference/using-api/api-concepts/#alternate-representations-of-resources
 	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
 	config.ContentType = "application/vnd.kubernetes.protobuf"
 	return config, nil
--- a/pkg/kubeconfig/loader.go
+++ b/pkg/kubeconfig/loader.go
@@ -1,6 +1,8 @@
 package kubeconfig
 
 import (
+	"fmt"
+
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
@@ -19,7 +21,7 @@ func LoadKubeconfig(path string) (*rest.Config, error) {
 	loader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
 	config, err := loader.ClientConfig()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("loading kubeconfig: %w", err)
 	}
 	config.UserAgent = "opencost"
 	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"


pkg/kubeconfig/loader.go
```diff
@@ -17,7 +17,14 @@ func LoadKubeconfig(path string) (*rest.Config, error) {
 		loadingRules.ExplicitPath = path
 	}
 	loader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
-	return loader.ClientConfig()
+	config, err := loader.ClientConfig()
+	if err != nil {
+		return nil, err
```
michaelmdresser: Would you please wrap this error?

pkg/kubeconfig/loader.go
```diff
@@ -17,7 +17,14 @@ func LoadKubeconfig(path string) (*rest.Config, error) {
...
+	}
+	config.UserAgent = "opencost"
+	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
+	config.ContentType = "application/vnd.kubernetes.protobuf"
```
michaelmdresser, line 30: Is there somewhere in the K8s documentation we can link to that explains these config values? E.g. what acceptable inputs are and what setting `ContentType = "application/vnd.kubernetes.protobuf"` means? I don't have the library in front of me, so if its already documented in the Go libraries godoc then that is plenty for me.

michaelmdresser (commented):
This is super cool! Thanks for the optimization numbers to help put the change in context. The decreased CPU usage is particularly exciting.

pkg/kubeconfig/loader.go
r2k1: https://kubernetes.io/docs/reference/using-api/api-concepts/#alternate-representations-of-resources

michaelmdresser (approved):

ameijer (approved):
neat!

pkg/kubeconfig/loader.go
mbolt35: Very very cool! I wonder why they haven't switched to this by default. Backwards compatibility I guess?

AjayTripathy (approved):

AjayTripathy:
Thank you Artur!

pkg/kubeconfig/loader.go
mbolt35: @r2k1 Just want to make sure that older versions of kubernetes will still work appropriately with this change. Is there any reason to expect versions of kubernetes on GKS, EKS, AKS may not work properly? 

Just want to make sure that we're covering our bases here.

r2k1: I'm not aware about any potential reasons why it would not work. I would expect all cloud services to use the same binary.

kube-state-metrics sets same values.

PR #2578, opened 2024-02-28
skip custom pricing if empty string

nickcurie:
## What does this PR change?
* Empty string values in custom pricing settings will be ignored in order to prevent the `/updateConfigByKey` endpoint from erroring 

* No

* Settings can be saved if a custom pricing value is set to empty string

* Closes https://kubecost.atlassian.net/browse/SELFHOST-1157

## How was this PR tested?
* Built an image and put it on gke dev-1, observed settings page saving when custom pricing values are set to empty string

* No

* v2.2?

### DIFF
--- a/pkg/cloud/models/models.go
+++ b/pkg/cloud/models/models.go
@@ -249,13 +249,18 @@ func SetCustomPricingField(obj *CustomPricing, name string, value string) error 
 	// from getting set here.
 	switch strings.ToLower(name) {
 	case "cpu", "gpu", "ram", "spotcpu", "spotgpu", "spotram", "storage", "zonenetworkegress", "regionnetworkegress", "internetnetworkegress":
-		// Validate that "value" represents a real floating point number, and
-		// set precision, bits, etc. Do not allow NaN.
-		val, err := sanitizeFloatString(value, false)
-		if err != nil {
-			return fmt.Errorf("invalid numeric value for field '%s': %s", name, value)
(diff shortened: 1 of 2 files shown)


nik-kc (approved):

pkg/cloud/models/models.go
```diff
@@ -249,13 +249,18 @@ func SetCustomPricingField(obj *CustomPricing, name string, value string) error
...
 	switch strings.ToLower(name) {
 	case "cpu", "gpu", "ram", "spotcpu", "spotgpu", "spotram", "storage", "zonenetworkegress", "regionnetworkegress", "internetnetworkegress":
-		// Validate that "value" represents a real floating point number, and
-		// set precision, bits, etc. Do not allow NaN.
-		val, err := sanitizeFloatString(value, false)
-		if err != nil {
-			return fmt.Errorf("invalid numeric value for field '%s': %s", name, value)
+		//what should we default to here?
+		if value == "" {
+			value = "1.0"
```
michaelmdresser: Instead of picking something, what if you just `return nil` here? We don't actually want to make an update because we've been asked to make an (apparently) meaningless update. `return nil` would avoid the error case while also not overwriting a config value with something the user did not expect.

nickcurie: Yeah this is a good call actually. I'll swap to this

michaelmdresser (approved):

PR #2579, opened 2024-02-28
Atm/plugin types

ameijer:
## What does this PR change?
* adds plugin interface definition, incorporates additional feedback into data structures

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


core/pkg/model/customcostresponse.go
```diff
@@ -2,60 +2,136 @@ package model
...
-	BilledCost     float32
-	AccountName    string
+	// provides metadata on the Custom CostResponse
+	// deliberately left unstructured
+	Metadata map[string]string
+	// the region that the resource was incurred
+	// corresponds to 'availability zone' of FOCUS
+	Zone string
+	// FOCUS billed Cost
+	BilledCost float32
```
michaelmdresser, line 45: Does FOCUS specify float32 instead of float64 for these fields? (Frankly, a non-lossy number type would be best, but that may not be realistic here)

core/pkg/plugin/plugin_interface.go
```diff
@@ -0,0 +1,55 @@
...
+	"net/rpc"
+
+	"github.com/hashicorp/go-plugin"
+	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/model"
+)
+
+// plugin interface
+type CustomCostSource interface {
+	GetCustomCosts(req model.CustomCostRequestInterface) []model.CustomCostResponse
```
michaelmdresser, line 15: Why does this take an interface instead of the straight-up struct? Do you anticipate a need to have implementation-dependent request structs?

core/pkg/plugin/plugin_interface.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// RPC impl
+type CustomCostRPC struct{ client *rpc.Client }
+
+func init() {
+	gob.Register(model.CustomCostRequest{})
+}
+func (c *CustomCostRPC) GetCustomCosts(req model.CustomCostRequestInterface) []model.CustomCostResponse {
```
michaelmdresser, line 24: Why not have this return an error and have the caller decide what to do with it? It'd entirely possible that the caller might want to change its behavior based on implementation-specific errors.

core/pkg/plugin/plugin_interface.go
```diff
@@ -0,0 +1,55 @@
...
+type CustomCostPlugin struct {
+	// Impl Injection
+	Impl CustomCostSource
+}
+
+func (p *CustomCostPlugin) Server(*plugin.MuxBroker) (interface{}, error) {
+	return &CustomCostRPCServer{Impl: p.Impl}, nil
+}
+
+func (CustomCostPlugin) Client(b *plugin.MuxBroker, c *rpc.Client) (interface{}, error) {
```
michaelmdresser, line 67: This signature is kind of funky; its on the `CustomCostPlugin` struct but doesn't use any of the fields?

michaelmdresser (commented):
Thanks for adding comments on the fields <3

core/pkg/model/customcostresponse.go
ameijer: focus doesn't specify anything other than 'numeric'. quote from the spec: 

> The FOCUS specification does not require a specific level of precision for numeric values. The level of precision required for a given column is determined by the provider and should be part of a data definition published by the provider.

i picked 32 bit since that seemed accurate enough, but open to a different type

core/pkg/plugin/plugin_interface.go
ameijer: frankly, I don't know, especially since the return value is happy with a concrete type. I don't foresee a need to have different underlying structs or anything like that, and would love to pass the struct right though. I sort of ran out of time here, thinking this is OK because it is only cast to an interface inside the calling method - all consumers of this API use structs and only structs. 

If you have any insights here, more than happy to fix this

core/pkg/plugin/plugin_interface.go
ameijer: I include errors in the custom cost response, to allow for graceful failure (maybe some costs had no errors, but others did). Would that approach be sufficient?

core/pkg/plugin/plugin_interface.go
ameijer: straight copy pasta from the go-plugin example. Here's what they have to say about it: 

> // Ignore MuxBroker. That is used to create more multiplexed streams on our
> // plugin connection and is a more advanced use case.
> func (GreeterPlugin) Client(b *plugin.MuxBroker, c *rpc.Client) (interface{}, error) {
> 	return &GreeterRPC{client: c}, nil
>  }

Could it be removed? Maybe. figured it was easier to just include it, and clean up if unused?

core/pkg/model/customcostresponse.go
michaelmdresser: I guess this LGTM then; people always say that money numbers should be represented in a better type than floating-point https://stackoverflow.com/questions/15726535/which-datatype-should-be-used-for-currency, but I'm guessing it gets very complicated for non-US currencies. Still, it'd be nice to multiply by 100 and store in cents.

core/pkg/plugin/plugin_interface.go
michaelmdresser: I think it might have to do with the `gob.Register` thing -- https://stackoverflow.com/questions/32676898/whats-the-purpose-of-gob-register-method it looks like you don't need to register if you aren't using interface types, which is why I find this particularly odd. We're in the wild west for me here though, so maybe there's some trick. I don't mind merging this as-is, then returning to it once we have a more concrete implementation we can test with

core/pkg/plugin/plugin_interface.go
michaelmdresser: I'm a little torn; my feeling is that if `c.client.Call` returns an error, then this method should also return an error.

core/pkg/plugin/plugin_interface.go
michaelmdresser: Very odd! Would you link out to the source in a comment? After that I don't mind at all; I wonder how it works.

ameijer: commented with link 👍

core/pkg/model/customcostresponse.go
ameijer: maybe we can leave as is then, and revisit after the first few plugins are implemented?

core/pkg/plugin/plugin_interface.go
ameijer: yup - ideas is to keep this easy to fix in the future without breaking a ton of stuff

core/pkg/plugin/plugin_interface.go
ameijer: so, I tried adding the error as an additional return value, however, that makes this no longer implement the interface. To avoid complicating things, I instead load the error into the existing field in the return struct. That should accomplish the goal of getting the user the actual error, while avoiding complicating the RPC interface. What do you think?

michaelmdresser: Gah! Makes sense, that's clear!

michaelmdresser (approved):

core/pkg/model/customcostresponse.go
michaelmdresser: For an async conversation (not blocking merge): @mbolt35 and @nikovacevic have you given any deeper thought to representations of currency?

PR #2599, opened 2024-03-01
Allow cost-model container to run with a read-only root filesystem

### LINKED ISSUE #1760: Pricing Configmap watcher unable to open file
**Describe the bug**
Whilst attempting to customise pricing configurations by adding the `pricing-configs` configmap, I ran into the following error:
```
ERROR UPDATING pricing-configs CONFIG: Failed to write file: open /models/gcp.json: permission denied
```

**To Reproduce**
1. Create and deploy a pricing configs configmap. E.g:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: "pricing-configs"
  namespace: opencost
data:
  description: "Default prices used to compute allocation between RAM and CPU. GCP pricing API data still used for total node cost."
  CPU: "0.031611"
  spotCPU: "0.006655"
  RAM: "0.004237"
  spotRAM: "0.000892"
  storage: "0.00005479452"
  zoneNetworkEgress: "0.01"
  regionNetworkEgress: "0.01"
  internetNetworkEgress: "0.12"
  projectID: "$GCP_PROJECT_ID"
  billingDataDataset: "$BQ_BILLING_DATASET"
```
2. Redeploy opencost deployment
3. See error above in the logs

**Expected behavior**
I expected the configmap watcher to populate /models/gcp.json.

latest (1.101.1)

**Additional context**
Kubernetes versions - 1.23
Public cloud - GCP (GKE)

kaitimmer:
## What does this PR change?
Same as in https://github.com/opencost/opencost/pull/2521
I want to be able to use a read-only root file system when running opencost. 

Yes, to https://github.com/opencost/opencost/pull/2521

It won't, but now you can mount an emptyDir to `/models` and set `readOnlyRootFilesystem: true` which is a requirement in our Kubernetes clusters.

IMHO this should fix: https://github.com/opencost/opencost/issues/1760


mattray:
Thanks for the PR, I'm going to need to test it before merging it. I thought we'd already dealt with these, but maybe it was just the opencost-ui container.

kaitimmer:
> Thanks for the PR, I'm going to need to test it before merging it. I thought we'd already dealt with these, but maybe it was just the opencost-ui container.

Yes, the last one was just for the UI. I missed this one at that time and only realized afterwards that the same problem exists here.

Dockerfile
```diff
@@ -43,12 +43,21 @@ LABEL org.opencontainers.image.url=https://opencost.io
...
+RUN mkdir -p /models
+
 ADD --chmod=644 ./THIRD_PARTY_LICENSES.txt /THIRD_PARTY_LICENSES.txt
-ADD --chmod=644 ./configs/default.json /models/default.json
-ADD --chmod=644 ./configs/azure.json /models/azure.json
-ADD --chmod=644 ./configs/aws.json /models/aws.json
-ADD --chmod=644 ./configs/gcp.json /models/gcp.json
-ADD --chmod=644 ./configs/alibaba.json /models/alibaba.json
-ADD --chmod=644 ./configs/oracle.json /models/oracle.json
+ADD --chmod=644 ./configs/default.json /opt/configs/models/default.json
```
lmello, line 50: both helm chart and users expect this to be in /models/ , people mount emptydir or ebs volumes to the /models/ directory.

this is a breaking change.

Also the code try to find the files in the /models/ directory, this changes the path where the files are stored but does not change the code path searching for the files. 

I am happy if you could find a solution to your requirement that is backwards compatible with the current use case.

kaitimmer:
Currently, I just do not have the time or need to fix this. So, I'll close this PR.

PR #2610, opened 2024-03-05
Atm/plugin ingestion

ameijer:
## What does this PR change?
* 

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/customcost/ingestor.go
+++ b/pkg/customcost/ingestor.go
@@ -1,19 +1,21 @@
 package customcost
 
 import (
-	"encoding/json"
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/hashicorp/go-plugin"
 	"github.com/opencost/opencost/core/pkg/log"
 	"github.com/opencost/opencost/core/pkg/model"
 	"github.com/opencost/opencost/core/pkg/opencost"
+	ocplugin "github.com/opencost/opencost/core/pkg/plugin"
+	"github.com/opencost/opencost/core/pkg/util/stringutil"
 	"github.com/opencost/opencost/core/pkg/util/timeutil"
 	"github.com/opencost/opencost/pkg/env"
+	"github.com/opencost/opencost/pkg/errors"
 )
 
 // IngestorStatus includes diagnostic values for a given Ingestor
@@ -22,7 +24,7 @@ type IngestorStatus struct {
 	LastRun  time.Time
 	NextRun  time.Time
 	Runs     int
-	Coverage opencost.Window
+	Coverage map[string]opencost.Window
 }
 
 // CustomCost IngestorConfig is a configuration struct for an Ingestor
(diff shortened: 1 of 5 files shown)


pkg/customcost/pipelineservice_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"time"
+
+	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/util/timeutil"
+	"github.com/opencost/opencost/core/pkg/version"
+)
+
+func TestPipelineService(t *testing.T) {
+	// establish temporary test assets dir
+	dir, err := os.MkdirTemp("", "ocplugin-*")
```
nik-kc: Could we update this to use `t.TempDir()`?

pkg/customcost/pipelineservice.go
```diff
@@ -81,23 +82,18 @@ func getRegisteredPlugins(configDir string, execDir string) (map[string]*plugin.
 		configs[name] = &plugin.ClientConfig{
 			HandshakeConfig: handshakeConfig,
 			Plugins:         pluginMap,
-			Cmd:             exec.Command(execDir+"/"+name+".ocplugin."+version.Architecture, config),
+			Cmd:             exec.Command(execDir+"/"+name+".ocplugin."+runtime.GOOS+"."+version.Architecture, config),
```
nik-kc: Could we pull out this duplicate string concatenation into a variable? Also, how is this string format determined?

nik-kc (approved):
Left a couple small notes, but overall looks good!

pkg/customcost/pipelineservice_test.go
ameijer: good thinking

pkg/customcost/pipelineservice.go
ameijer: I refactored it to use a fmt string, that hopefully cleans it up. Since the plugins run as binaries, the executable name depends on the OS and processor. The idea is that the person would get the binary built for their machine from here: https://github.com/opencost/opencost-plugins/releases/tag/v0.0.3

pkg/customcost/ingestor.go
```diff
@@ -54,116 +58,175 @@ type CustomCostIngestor struct {
...
-		return nil, fmt.Errorf("CustomCost: NewCustomCostIngestor: integration connot be nil")
+		return nil, fmt.Errorf("CustomCost: NewCustomCostIngestor: config connot be nil")
+	}
+	key := ""
+
+	for name := range plugins {
+		key += "," + name
 	}
 
+	key = strings.TrimPrefix(key, ",")
```
michaelmdresser, line 86: You make this key and then do nothing with it. What is it for?

pkg/customcost/ingestor.go
```diff
@@ -54,116 +58,175 @@ type CustomCostIngestor struct {
 	lastRun      time.Time
 	runs         int
 	creationTime time.Time
-	coverage     opencost.Window
+	coverage     map[string]opencost.Window
```
michaelmdresser, line 61: stringly-typed keys are something I'm not a huge fan of -- what is the key supposed to represent here? If its a plugin, why not have one CustomCostIngestor per plugin?

pkg/customcost/ingestor.go
```diff
@@ -54,116 +58,175 @@ type CustomCostIngestor struct {
...
+		ing.buildSingleDomain(start, end, domain)
+	}
+}
 
-	// // If already running, log that and return.
-	// if !ing.isRunning.CompareAndSwap(false, true) {
-	// 	log.Infof("CloudCost: ingestor: is already running")
-	// 	return
-	// }
+func (ing *CustomCostIngestor) buildSingleDomain(start, end time.Time, domain string) {
```
michaelmdresser, line 159: I feel that this should return an error -- I lean towards having the highest-possible-level component log errors

pkg/customcost/pipelineservice_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/opencost/opencost/core/pkg/util/timeutil"
+	"github.com/opencost/opencost/core/pkg/version"
+)
+
+func TestPipelineService(t *testing.T) {
+	// establish temporary test assets dir
+
+	dir := t.TempDir()
+
+	err := os.MkdirAll(dir+"/config", 0777)
```
michaelmdresser, line 22: For the future, its better to use
```go
import "path/filepath"

os.MkdirAll(filepath.Join(dir, "config"))
```

instead of string ops when constructing paths

PR #2612, opened 2024-03-05
Sth/plugin proto

Sean-Holcomb:
## What does this PR change?
This PR changes the way that Custom Costs are recieved

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


core/pkg/model/pb/messages.pb.go
```diff
@@ -0,0 +1,876 @@
...
+)
+
+const (
+	// Verify that this generated code is sufficiently up-to-date.
+	_ = protoimpl.EnforceVersion(20 - protoimpl.MinVersion)
+	// Verify that runtime/protoimpl is sufficiently up-to-date.
+	_ = protoimpl.EnforceVersion(protoimpl.MaxVersion - 20)
+)
+
+type CustomCostResponse struct {
```
michaelmdresser, line 24: If we're going to go to the effort of defining a nice proto schema for this data, why not just eliminate the other `CustomCostResponse` type entirely? Or is that an antipattern?

generate.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/zsh
```
michaelmdresser: For portability, I think `#!/usr/bin/env sh` or `#!/usr/bin/env bash` is better

core/pkg/model/pb/messages.pb.go
ameijer: Sean and I discussed this very thing. I was originally against it, but maybe that is worth revisiting

generate.sh
Sean-Holcomb: yeah sorry, I was just trying to make it work quickly on my machine. I will swap in one of those

core/pkg/model/pb/messages.pb.go
Sean-Holcomb: That is my position for the most part, There may be some value to having the other definition for data manipulation in opencost, but we could also just extend the proto struct with any helper functions we want. For kubecost, I want the transition to duck to come from the plugin as protobuf directly and go directly to duck schema and into the db

ameijer (approved):

PR #2613, opened 2024-03-06
Plugin queries

nik-kc:
## What does this PR change?
* 

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/core/pkg/version/version.go
+++ b/core/pkg/version/version.go
@@ -1,11 +1,14 @@
 package version
 
-import "fmt"
+import (
+	"fmt"
+	"runtime"
+)
 
 var (
 	Version      = "dev"
 	GitCommit    = "HEAD"
-	Architecture = "amd64"
+	Architecture = runtime.GOARCH
 )
 
 func FriendlyVersion() string {
--- a/core/pkg/version/version.go
+++ b/core/pkg/version/version.go
@@ -2,13 +2,11 @@ package version
 
 import (
 	"fmt"
-	"runtime"
 )
 
 var (
-	Version      = "dev"
-	GitCommit    = "HEAD"
-	Architecture = runtime.GOARCH
+	Version   = "dev"
+	GitCommit = "HEAD"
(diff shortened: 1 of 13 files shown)


core/pkg/version/version.go
```diff
@@ -1,11 +1,14 @@
...
+import (
+	"fmt"
+	"runtime"
+)
 
 var (
 	Version      = "dev"
 	GitCommit    = "HEAD"
-	Architecture = "amd64"
+	Architecture = runtime.GOARCH
```
ameijer: I like this better. Can you go ahead and remove this Architecture field, and update any refs to it to just use runtime.goarch?

pkg/customcost/querier.go
```diff
@@ -0,0 +1,164 @@
...
+	End         time.Time
+	AggregateBy []string
+	Step        time.Duration
+	Filter      filter.Filter
+}
+
+type CostTimeseriesResult struct {
+	TotalCost float32
+	Window    opencost.Window
+	Response  []model.CustomCostResponse
```
ameijer: mentioned in your other PR, which I think is obsolete(?), but we should have json field tags in these request/response objs

pkg/customcost/querier.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (q *Querier) QueryTotal(request CostTotalRequest, ctx context.Context) (*CostTotalResult, error) {
+	repo, queryTime, err := q.repoForTime(request.Start)
```
ameijer, line 58: I think you would need start and end here, right? if I want to know my total costs over the last week, this would only compute total costs for the first day of that week, if I am understanding this correctly?

pkg/customcost/querier.go
nik-kc: Yup, the other PR is absolute and I've closed it. I'll add the JSON field tags.

PR #2619, opened 2024-03-08
Nik/plugins

nik-kc:
## What does this PR change?
* 

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/customcost/types.go
+++ b/pkg/customcost/types.go
@@ -56,8 +56,6 @@ type CostTimeseriesResponse struct {
 	Timeseries []*CostResponse `json:"timeseries"`
 }
 
-// --------------
-
 func NewCostResponse(ccs *CustomCostSet) *CostResponse {
 	costResponse := &CostResponse{
 		Window:      ccs.Window,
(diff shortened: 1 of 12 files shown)


pkg/customcost/types.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return customCosts
+}
+
+func (cc *CustomCost) Add(other *CustomCost) {
+	cc.BilledCost += other.BilledCost
+	cc.ListCost += other.ListCost
+	cc.ListUnitPrice += other.ListUnitPrice
+	cc.UsageQuantity += other.UsageQuantity
```
ameijer, line 104: the only major thing I see in this PR is that when adding, we should take the intersection of the string properties. so if one domain != another domain, then the result is the domain is an empty string, that more closely matches our existing logic

ameijer: When this goes into the working branch, I will fix this unless @nik-kc hits this before the merge

ameijer (approved):
approving, but noting something that needs to be fixed

cliffcolvin (approved):

PR #2631, opened 2024-03-12
Create a filter implementation for K8s `runtime.Object`s

michaelmdresser:
## What does this PR change?
* Introduces a new filter parser and MatchCompiler implementation targeting the `runtime.Object` type in the K8s library
* Moves filter field strings to a common package to avoid duplication and drift in future filter implementations

* Not in Opencost (will be used by a downstream dependency to use Opencost-style filter strings for a new purpose)

* N/A

* N/A

## How was this PR tested?
* New unit tests

* No

### DIFF
--- a/core/pkg/filter/k8sobject/fields.go
+++ b/core/pkg/filter/k8sobject/fields.go
@@ -1,0 +1,18 @@
+package k8sobject
+
+import (
+	"github.com/opencost/opencost/core/pkg/filter/fieldstrings"
+)
+
+// K8sObjectField is an enum that represents K8sObject-specific fields that can
+// be filtered on.
+type K8sObjectField string
+
+const (
+	FieldNamespace      K8sObjectField = K8sObjectField(fieldstrings.FieldNamespace)
+	FieldControllerKind K8sObjectField = K8sObjectField(fieldstrings.FieldControllerKind)
+	FieldControllerName K8sObjectField = K8sObjectField(fieldstrings.FieldControllerName)
+	FieldPod            K8sObjectField = K8sObjectField(fieldstrings.FieldPod)
+	FieldLabel          K8sObjectField = K8sObjectField(fieldstrings.FieldLabel)
+	FieldAnnotation     K8sObjectField = K8sObjectField(fieldstrings.FieldAnnotation)
+)
--- a/core/pkg/filter/k8sobject/parser.go
+++ b/core/pkg/filter/k8sobject/parser.go
@@ -1,0 +1,14 @@
+package k8sobject
+
+import (
+	allocfilter "github.com/opencost/opencost/core/pkg/filter/allocation"
+	"github.com/opencost/opencost/core/pkg/filter/ast"
+)
+
+// NewK8sObjectFilterParser creates a new `ast.FilterParser` implementation.
+//
+// It currently just uses Allocation behavior (meaning it supports the same
+// fields Allocation does), but we may want to change that later.
+func NewK8sObjectFilterParser() ast.FilterParser {
+	return allocfilter.NewAllocationFilterParser()
+}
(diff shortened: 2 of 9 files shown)


mbolt35:
One very explicit thing you will need is to register the fields in the `ops` package either by adding it directly here: https://github.com/opencost/opencost/blob/8a4002a1c1f33412c198c7d8d774a80ac78e043b/core/pkg/filter/ops/ops.go#L28-L33
(this is likely the best choice -- just replace the `// typeutil.TypeOf[containerstats.ContainerStatsField](): containerstats.DefaultFieldByName,` line - heheheh)

If the fields are added outside of the opencost repo, you can use this in an init: https://github.com/opencost/opencost/blob/8a4002a1c1f33412c198c7d8d774a80ac78e043b/core/pkg/filter/ops/ops.go#L36

michaelmdresser:
TIL! Done, that point about stuff outside of the Opencost repo is neat as heck

core/pkg/filter/k8sobject/fields.go
```diff
@@ -0,0 +1,18 @@
...
+// be filtered on.
+type K8sObjectField string
+
+const (
+	FieldNamespace      K8sObjectField = K8sObjectField(fieldstrings.FieldNamespace)
+	FieldControllerKind K8sObjectField = K8sObjectField(fieldstrings.FieldControllerKind)
+	FieldControllerName K8sObjectField = K8sObjectField(fieldstrings.FieldControllerName)
+	FieldPod            K8sObjectField = K8sObjectField(fieldstrings.FieldPod)
+	FieldLabel          K8sObjectField = K8sObjectField(fieldstrings.FieldLabel)
+	FieldAnnotation     K8sObjectField = K8sObjectField(fieldstrings.FieldAnnotation)
```
mbolt35, line 17: Did go force you to add the type cast when you referred to the `fieldstrings` constants? 

ie:
```go
const FieldNamespace K8sObjectField = "namespace"
```
is fine, but 
```go
const FieldNamespace K8sObjectField = fieldstrings.FieldNamespace
```
barks at you that they're not the same type? 

Bah

core/pkg/filter/k8sobject/parser.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/opencost/opencost/core/pkg/filter/ast"
+)
+
+// a slice of all the allocation field instances the lexer should recognize as
+// valid left-hand comparators
+var k8sObjectFilterFields []*ast.Field = []*ast.Field{
+	ast.NewField(FieldNamespace),
+	ast.NewField(FieldControllerName, ast.FieldAttributeNilable),
```
mbolt35, line 11: Just wanted to point out (if you haven't noted field attributes yet) that `ast.FieldAttributeNilable` (or any attribute that might be needed) on a field is simply metadata that can be accessed in the compiler or code generator to help alleviate any special cases that may exist.

mbolt35: Attributes are just bitflags (sorry, I think I'm showing my age), so they're not super expansive, but likely more space than we'll ever need. I think Nilable starts at the 5th bit, and we're using an `int`, so it's expandable to another 27 custom attributes 🤷

core/pkg/filter/k8sobject/fields.go
michaelmdresser: Yeah, it barks; I'm glad it does, otherwise these enum-like things would be pretty useless.

core/pkg/filter/k8sobject/parser.go
michaelmdresser: Oh very neat, I just copied whatever Allocation was using for consistency; I _think_ it also makes sense here.

mbolt35 (approved):
LGTM

PR #2634, opened 2024-03-13
Adding additional check for EKS, using Kubelet

ghost:
## What does this PR change?
* Adds an additional check for EKS to check the Kubelet version 

* No

* Introduces an additional verification step just in case the `spec.ProviderID` is inaccurate. We've noticed anomalies with Karpenter-managed nodes, particularly missing certain expected labels.

* No

## How was this PR tested?
* I did not thoroughly test this; I simply duplicated the line from the Alibaba check since we are also verifying the Kubelet version there.

* No

* No, and no reason in particular.

### DIFF
--- a/.github/ISSUE_TEMPLATE/opencost-bug-report.md
+++ b/.github/ISSUE_TEMPLATE/opencost-bug-report.md
@@ -8,7 +8,7 @@ assignees: ''
 ---
 
 **Describe the bug**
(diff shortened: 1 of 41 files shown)


michaelmdresser (approved):
I think this makes sense, but I'd like to check with @avrodrigues5, @mbolt35, and/or @AjayTripathy on this.

avrodrigues5:
Looks fine for me too! Please rebase thou!

PR #2653, opened 2024-03-18
Add carbon estimation functionality to assets

kaelanspatel:
## What does this PR change?
Adds carbon estimation functionality to Opencost `/assets`, assigning a carbon cost in CO2e to each node/disk based off a lookup table.

Enabled by setting the `CARBON_ESTIMATES_ENABLED` env var.

No.

* Adds the `/assets/carbon` endpoint to Opencost, allowing users to see carbon costs for node and disk assets.

* Closes ...

## How was this PR tested?
* Locally against Prom by querying endpoint.
* On an existing install of Opencost by querying endpoint.

* Yes.

*

### DIFF
--- a/pkg/carbon/carbonassets.go
+++ b/pkg/carbon/carbonassets.go
@@ -17,6 +17,7 @@ var f embed.FS
 var carbonLookupNode map[interface{}]float64
 var carbonLookupDisk map[interface{}]float64
 
+// Opencost does not build network types
 var carbonValidInstanceTypes map[string]string
 var carbonValidRegions map[string]string
 
@@ -31,7 +32,7 @@ func init() {
 	reader := csv.NewReader(strings.NewReader(string(carbonData)))
 
 	// skip header
-	reader.Read()
+	_, err = reader.Read()
 	if err != nil {
 		log.Errorf("Error reading carbon lookup data: %s", err)
 		return
@@ -123,11 +124,10 @@ func RelateCarbonAssets(as *opencost.AssetSet) (map[string]CarbonRow, error) {
 			log.DedupedErrorf(10, "Cannot infer region information for asset '%s'", asset.GetProperties().ProviderID)
 		}
 
-		var foundCarbonCoeff bool
 		var carbonCoeff float64
 		switch asset.Type() {
 		case opencost.NodeAssetType:
-			carbonCoeff, foundCarbonCoeff = carbonLookupNode[struct {
+			carbonCoeff = carbonLookupNode[struct {
 				provider     string
 				region       string
 				instanceType string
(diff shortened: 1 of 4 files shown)


pkg/costmodel/handlers.go
```diff
@@ -27,31 +28,69 @@ func (a *Accesses) ComputeAssetsHandler(w http.ResponseWriter, r *http.Request,
...
+		http.Error(w, fmt.Sprintf("Error getting assets: %s", err), http.StatusInternalServerError)
+		return
+	}
+
+	carbonEstimates, err := carbon.RelateCarbonAssets(assetSet)
+
+	w.Write(WrapData(carbonEstimates, nil))
+}
+
+func (a *Accesses) computeAssetsFromCostmodel(window opencost.Window, filterString string) (*opencost.AssetSet, error) {
```
michaelmdresser, line 69: This is an excellent change regardless

pkg/carbon/carbonassets.go
```diff
@@ -0,0 +1,171 @@
...
+			case "Node":
+				carbonLookupNode[struct {
+					provider     string
+					region       string
+					instanceType string
+				}{
+					provider:     provider,
+					region:       region,
+					instanceType: instanceType,
+				}] = coeff
```
michaelmdresser: Using `interface{}` as the map key type and then manually defining a `struct` type every time you have to insert or look up a key isn't great. Would you please define these as proper package-level (private) structs? Further, I think there has to be a way to make the map key type not `interface{}`:
```go
type carbonLookupKeyDisk struct {
    provider string
    region string
}

var carbonLookupDisk map[carbonLookupKeyDisk]float64
```

pkg/carbon/carbonassets.go
```diff
@@ -0,0 +1,171 @@
...
+	carbonLookupDisk = make(map[interface{}]float64)
+
+	carbonValidInstanceTypes = make(map[string]string)
+	carbonValidRegions = make(map[string]string)
+
+	for _, carbonItem := range dat {
+
+		if coeff, err := strconv.ParseFloat(carbonItem[5], 64); err != nil {
+
+			log.DedupedWarningf(5, "Error setting carbon lookup item: malformed carbon cost '%s'", carbonItem[5])
```
michaelmdresser: In general, we should avoid the deduped log methods because they dedupe on the format string, not the full error, so we can swallow a ton of errors because of it. Since this is an `init()` I think it is fine to log at Error level without deduping, but I'd personally go further and `panic()` here because `init()` will only be called once, on load.

michaelmdresser: https://stackoverflow.com/questions/24790175/when-is-the-init-function-run

michaelmdresser (approved):

PR #2656, opened 2024-03-21
Sth/azure storage upgrade

Sean-Holcomb:
## What does this PR change?
This PR enables configuration of the Azure Storage implementation of the Storage interface via Workload Identities and other new integration methods. It does this by updating the azure SDK used and removing dependancies on the old SDKs.

There is also a small change to the S3 storage implementation to allow it to pass the newly written integration tests

* 

* Users will be able to integration with azure via WI for Storage

* Closes ...

## How was this PR tested?
* Added new integration tests to check the functionality of Bucket Storage type Storage implementations

* Documentation should be expanded to include explanation of how to configure with WI

*

### DIFF
--- a/pkg/storage/azurestorage.go
+++ b/pkg/storage/azurestorage.go
@@ -1,27 +1,29 @@
 package storage
 
-// Fork from Thanos S3 Bucket support to reuse configuration options
+// Fork from Thanos Azure Storage Bucket support to reuse configuration options
 // Licensed under the Apache License 2.0
-// https://github.com/thanos-io/thanos/blob/main/pkg/objstore/s3/s3.go
+// https://github.com/thanos-io/objstore/blob/main/providers/azure/azure.go
 
 import (
 	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"net"
 	"net/http"
-	"net/url"
 	"strings"
-	"sync"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/bloberror"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
 	"github.com/opencost/opencost/core/pkg/log"
 
-	"github.com/Azure/azure-pipeline-go/pipeline"
-	blob "github.com/Azure/azure-storage-blob-go/azblob"
-	"github.com/Azure/go-autorest/autorest/adal"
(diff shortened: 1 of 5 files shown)


pkg/storage/s3storage.go
```diff
@@ -462,17 +462,21 @@ func (s3 *S3Storage) List(path string) ([]*StorageInfo, error) {
...
 
+		name := trimName(object.Key)
+
+		// This sometimes happens with empty buckets.
+		if name == "" {
+			continue
+		}
+
 		stats = append(stats, &StorageInfo{
-			Name:    trimName(object.Key),
```
Sean-Holcomb, line 475: The issue here was that dirrectories were being included as empty string StorageInfo because they passed the initial empty string check but were subsequently blanked out by trimName(). Directories should not be returned by this function, so swapping the order of the function call and the check addresses the issue

pkg/storage/azurestorage.go
```diff
@@ -205,30 +181,29 @@ func NewAzureStorageWith(conf AzureConfig) (*AzureStorage, error) {
 		return nil, err
 	}
 
+	containerClient, err := getContainerClient(conf)
+	if err != nil {
+		return nil, err
```
michaelmdresser: Please wrap this error

pkg/storage/azurestorage.go
```diff
@@ -205,30 +181,29 @@ func NewAzureStorageWith(conf AzureConfig) (*AzureStorage, error) {
...
+	// Check if storage account container already exists, and create one if it does not.
 	ctx := context.Background()
-	container, err := createContainer(ctx, conf)
+	_, err = containerClient.GetProperties(ctx, &container.GetPropertiesOptions{})
 	if err != nil {
-		ret, ok := err.(blob.StorageError)
-		if !ok {
-			return nil, errors.Wrapf(err, "Azure API return unexpected error: %T\n", err)
+		if !bloberror.HasCode(err, bloberror.ContainerNotFound) {
+			return nil, err
```
michaelmdresser, line 194: Wrap

pkg/storage/azurestorage.go
```diff
@@ -296,14 +278,13 @@ func (b *AzureStorage) Write(name string, data []byte) error {
...
-	if _, err := blob.UploadStreamToBlockBlob(ctx, r, blobURL,
-		blob.UploadStreamToBlockBlobOptions{
-			BufferSize: len(data),
-			MaxBuffers: 1,
-		},
-	); err != nil {
+	blobClient := b.containerClient.NewBlockBlobClient(name)
+	opts := &blockblob.UploadStreamOptions{
+		BlockSize:   3 * 1024 * 1024,
+		Concurrency: 4,
```
michaelmdresser, line 285: Are these recommended defaults?

pkg/storage/azurestorage.go
```diff
@@ -640,28 +453,71 @@ func DefaultAzureTransport(config AzureConfig) (*http.Transport, error) {
...
+			},
+			Transport: &http.Client{Transport: dt},
+		},
+	}
+
+	// Use connection string if set
+	if conf.StorageConnectionString != "" {
+		containerClient, err := container.NewClientFromConnectionString(conf.StorageConnectionString, conf.ContainerName, opt)
+		if err != nil {
+			return nil, err
```
michaelmdresser: Please wrap the errors in here

michaelmdresser (approved):
Awesome tests

pkg/storage/azurestorage.go
Sean-Holcomb: ripped from the thanos implementation

Issue #2657, opened 2024-03-22
gpuCost computation error


ElieLiabeuf:
**Describe the bug**
gpuCost in /costDataModel is not computed correctly.

**To Reproduce**
I'm using g5.2xlarge instance on AWS with nvidia GPU operator v23.9.1.
Hourly instance cost is 1.3529600000

Specs:
```
{
   "memory":"32 GiB",
   "vcpu":"8",
   "gpu":"1"
}
```

Node has the following status.capacity:
```
apiVersion: v1
kind: Node
status:
  capacity:
    cpu: '8'
    memory: 32500272Ki
    nvidia.com/gpu: '1'
```

and the following labels:
```
apiVersion: v1
kind: Node
metadata:
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: g5.2xlarge
    beta.kubernetes.io/os: linux
    feature.node.kubernetes.io/cpu-cpuid.ADX: 'true'
    feature.node.kubernetes.io/cpu-cpuid.AESNI: 'true'
    feature.node.kubernetes.io/cpu-cpuid.AVX: 'true'
    feature.node.kubernetes.io/cpu-cpuid.AVX2: 'true'
    feature.node.kubernetes.io/cpu-cpuid.CLZERO: 'true'
    feature.node.kubernetes.io/cpu-cpuid.CMPXCHG8: 'true'
    feature.node.kubernetes.io/cpu-cpuid.FMA3: 'true'
    feature.node.kubernetes.io/cpu-cpuid.FP256: 'true'
    feature.node.kubernetes.io/cpu-cpuid.FXSR: 'true'
    feature.node.kubernetes.io/cpu-cpuid.FXSROPT: 'true'
    feature.node.kubernetes.io/cpu-cpuid.HYPERVISOR: 'true'
    feature.node.kubernetes.io/cpu-cpuid.IBPB: 'true'
    feature.node.kubernetes.io/cpu-cpuid.IBRS: 'true'
    feature.node.kubernetes.io/cpu-cpuid.IBRS_PREFERRED: 'true'
    feature.node.kubernetes.io/cpu-cpuid.IBRS_PROVIDES_SMP: 'true'
    feature.node.kubernetes.io/cpu-cpuid.LAHF: 'true'
    feature.node.kubernetes.io/cpu-cpuid.MCOMMIT: 'true'
    feature.node.kubernetes.io/cpu-cpuid.MOVBE: 'true'
[...]

PR #2658, opened 2024-03-22
Fix gpuCost computation error

ElieLiabeuf:
## What does this PR change?
Solve gpuCost computation error

Closes https://github.com/opencost/opencost/issues/2657

No

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -1095,7 +1095,7 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 				newCnode.GPU = fmt.Sprintf("%d", 0)
 			}
 			newCnode.VGPU = fmt.Sprintf("%d", q.Value())
-
+			gpuc = float64(q.Value())
 		} else if g, ok := n.Status.Capacity["k8s.amazonaws.com/vgpu"]; ok {
 			gpuCount := g.Value()
 			if gpuCount != 0 {


mattray:
@ElieLiabeuf thanks for the contribution. Could you sign your commit and fix the [DCO warning](https://github.com/opencost/opencost/pull/2658/checks?check_run_id=22990244391)? This looks relatively straightforward

ElieLiabeuf:
> @ElieLiabeuf thanks for the contribution. Could you sign your commit and fix the [DCO warning](https://github.com/opencost/opencost/pull/2658/checks?check_run_id=22990244391)? This looks relatively straightforward

It's fixed, thanks.

mattray:
@AjayTripathy you asked him to contribute this patch in https://github.com/opencost/opencost/issues/2657, could you review?

AjayTripathy:
@ElieLiabeuf this generally LGTM but I'm struggling to test this on my end without shared GPUs. Could you share how you tested and verified this fix?

In particular, could you share the results of /costDataModel after the fix along with node labels? I think that should be sufficient to get this merged.

AjayTripathy:
@thomasvn could you possibly pick up testing this as part of your GPU sprint?

thomasvn:
Sure! @ElieLiabeuf I'll take a look at this soon.

chipzoller:
Let's please get this in ASAP.

pkg/costmodel/costmodel.go
```diff
@@ -1095,7 +1095,7 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
 				newCnode.GPU = fmt.Sprintf("%d", 0)
 			}
 			newCnode.VGPU = fmt.Sprintf("%d", q.Value())
-
+			gpuc = float64(q.Value())
```
thomasvn, line 1098: I think this makes sense to me. This conditional was not previously setting `gpuc` which would later affect calculations of gpuToRam and RamToCpu cost ratios.

thomasvn: Just to confirm, you're currently hitting this bug because you have the following two labels on your node?

```yaml
    nvidia.com/gpu.count: '1'
    nvidia.com/gpu.replicas: '1'
```

thomasvn:
@ElieLiabeuf Could we try to merge this change into the `develop` branch instead of directly into the `v1.109` branch? A new PR may need to be created.

AjayTripathy:
@thomasvn if this is meaningful I may just ask you to take over the fix?

thomasvn:
@AjayTripathy Yes I can do that soon!

chipzoller:
Following up to see if we can get this in this week and start testing.

chipzoller:
After @thomasvn and I spent a few minutes looking at the code, it seems apparent that this logic will be reached if the node label `nvidia.com/gpu.replicas` is found in addition to there being any allocatable capacity for `nvidia.com/gpu` (the standard GPU resource). However, because this label is provided by GFD irrespective of whether any sharing strategy is in use (exclude one specific MIG configuration), getting this fix in is important since the calculations will be off any time GFD is deployed in the cluster. In cases where GFD and, therefore, this label are absent, the logic being patched here appears to be unaffected.

AjayTripathy:
In that case what do we need to merge @thomasvn ?

thomasvn:
We can close this PR now, as the fix will be available in https://github.com/opencost/opencost/pull/2862

PR #2663, opened 2024-03-22
Update the `/installInfo` endpoint

thomasvn:
## What does this PR change?
* Updates the `/installInfo` endpoint to query the k8s `pod.Spec` instead of `pod.Status`.
* We've found that `pod.Status` can sometimes show stale information. Specifically, it will state that a container is running an image which it is actually no longer running. More details here: https://github.com/kubernetes/kubernetes/issues/74081
* To reproduce, swap out one of your existing container images via `kubectl edit deployment/kubecost-cost-analyzer` then run the following two commands:
  * `kubectl describe pod -l app=cost-analyzer | grep "Image"`
  * `kubectl get pod kubecost-cost-analyzer-7d65b9479f-wtnnz -oyaml | grep "image"`

* N/A

* Some users will be impacted if they were using the `/installInfo` endpoint to grab container metadata such as its ImageID or its number of restarts. This PR removes that metadata since the `pod.Spec` endpoint does not return that info.
* This only impacts users of the `/installInfo` endpoint. Deleted fields are not referenced internally anywhere in OpenCost codebase.

* N/A

## How was this PR tested?

```sh
export CONFIG_PATH="/tmp/localrun/default"
export PROMETHEUS_SERVER_ENDPOINT="http://127.0.0.1:9080"
export KUBERNETES_PORT="localrun"
export KUBECOST_NAMESPACE=kubecost-amp-multi

go run cmd/costmodel/main.go
```

Then hitting `http://localhost:9003/installInfo` correctly returns the kubecost containers running in the namespace.

* N/A

*

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -1204,9 +1204,6 @@ type InstallInfo struct {
 type ContainerInfo struct {
 	ContainerName string `json:"containerName"`
 	Image         string `json:"image"`
-	ImageID       string `json:"imageID"`
-	StartTime     string `json:"startTime"`
-	Restarts      int32  `json:"restarts"`
 }
 
 func (a *Accesses) GetInstallInfo(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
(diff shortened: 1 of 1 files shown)


jessegoodier:
@thomasvn can we verify (or change this) to make sure that we are only looking at `KUBECOST_NAMESPACE` ?

This says `all`, which makes me nervous that we'll have inconsistent results:
https://github.com/opencost/opencost/blob/6866bc20d62e7dfea45c87ff5abbe7903fd12c26/pkg/costmodel/router.go#L1097

dwbrown2:
cc @nealormsbee who is now using this endpoint on FE.

thomasvn:
@jessegoodier Yes the code already does this! As shown, it filters for running pods in the Kubecost namespace with the label `app=cost-analyzer`. It currently limits to grabbing the first pod returned, so the only problem would be if multiple Kubecosts are running in a single namespace.

https://github.com/opencost/opencost/blob/d5d8c09afbf044100aacd4bd12f61f2fb061b283/pkg/costmodel/router.go#L1216-L1220

nealormsbee:
Thanks for the tag! Current changeset won't impact the Kubecost UI. We only leverage `image` and `containerName`.

cliffcolvin (approved):

michaelmdresser:
@cliffcolvin @thomasvn is this deserving of a v2.2 cherry-pick?

thomasvn:
@michaelmdresser Yes! Cherry-pick was performed here https://github.com/opencost/opencost/pull/2667.

PR #2675, opened 2024-03-28
Fix test results

Sean-Holcomb:
## What does this PR change?
* This PR fixes a failing test, comments in the code suggest that the test pass condition were incorrect so that is where the change was made
```
// Use daily if...
//  (1) daily is an option
// It is acceptable to query a range for which we only have partial data
```

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


cliffcolvin (approved):
thank you

PR #2686, opened 2024-04-04
Expose `GetKubecostContainers()`

nik-kc:
## What does this PR change?
* Separates out and exposes the functionality required to determine what images the various Kubecost pods are using.

* [KCM](https://github.com/kubecost/kubecost-cost-model/pull/2356).
* [Helm](https://github.com/kubecost/cost-analyzer-helm-chart/pull/3307).

* It won't.

* Nope.

## How was this PR tested?
* Verified that `/model/installInfo` (from which the code for `GetKubecostContainers()` was taken) returns the same response object.

* Nope.

### DIFF
--- a/core/pkg/log/log.go
+++ b/core/pkg/log/log.go
@@ -126,17 +126,6 @@ func Debugf(format string, a ...interface{}) {
 	log.Debug().Msgf(format, a...)
 }
 
-func DedupedDebugf(logTypeLimit int, format string, a ...interface{}) {
-	timesLogged := ctr.increment(format)
-
-	if timesLogged < logTypeLimit {
-		Debugf(format, a...)
-	} else if timesLogged == logTypeLimit {
-		Debugf(format, a...)
-		Debugf("%s logged %d times: suppressing future logs", format, logTypeLimit)
-	}
-}
-
 func Trace(msg string) {
 	log.Trace().Msg(msg)
 }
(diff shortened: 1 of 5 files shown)


core/pkg/log/log.go
```diff
@@ -126,6 +126,17 @@ func Debugf(format string, a ...interface{}) {
 	log.Debug().Msgf(format, a...)
 }
 
+func DedupedDebugf(logTypeLimit int, format string, a ...interface{}) {
```
michaelmdresser: I'd like to avoid increasing the use of `Deduped` log functions if it is at all possible. Swallowing dupes based on the format string makes these functions behave in unexpected ways. It is usually possible to achieve better behavior by just not logging in hot loops and instead accumulating error information outside of the loop to log at completion; this is superior deduping behavior.

pkg/costmodel/router.go
```diff
@@ -1211,23 +1211,46 @@ func (a *Accesses) GetInstallInfo(w http.ResponseWriter, r *http.Request, _ http
...
+}
+
+func GetKubecostContainers(kubeClientSet kubernetes.Interface) ([]ContainerInfo, error) {
+	pods, err := kubeClientSet.CoreV1().Pods(env.GetKubecostNamespace()).List(context.Background(), metav1.ListOptions{
+		LabelSelector: "app=cost-analyzer",
+		FieldSelector: "status.phase=Running",
+		Limit:         1,
+	})
+	if err != nil {
+		return nil, err
```
michaelmdresser: Please wrap this error

core/pkg/log/log.go
nik-kc: That's a valid point, I agree with your alternative solution.

michaelmdresser (approved):

PR #2687, opened 2024-04-04
cloud cost status returns empty slice instead of nil

Sean-Holcomb:
## What does this PR change?
* This PR updates the response from the `Status()` function to return an empty slice rather than nil when there are no statuses.

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


michaelmdresser (approved):
LGTM. I assume this helps the FE out?

cliffcolvin (approved):

PR #2700, opened 2024-04-12
Create a hash key when agg properties are not set for Cloud Cost

Sean-Holcomb:
## What does this PR change?
This PR updates the way keys are generated for CloudCostSet when no aggregation properties are included. This is primarily meant to change the way that CloudCostSet are keyed when unaggregated to bingen file. Users have observed that certain line items that are missing the ProviderID field are aggregated when the CloudCost set is saved which erases their label set. To preserve this label set it is being included the new key creation process.

The new key creation process simply creates a 64 bit hash from property set  including the labels sorted by key. The result is a fixed length key with a vanishingly low chance of collision. 

This change has resulted in generally smaller cloud cost bingen file size due to the fixed length key being shorter the concatenated key. So much so that it off set the increased number of Cloud Costs.

In testing on the billing data I observed small increases in the number of Cloud Costs. These are the percentage increases by provider:
AWS: 0.008%
GCP: 7.590%
Azure: 0.0%
These numbers were retrieved over a 30 day window

* 

* 

* Closes ...

## How was this PR tested?
* Tested with local runs and a unit test for the new hash key function

* 

*

### DIFF
--- a/core/pkg/opencost/cloudcostprops.go
+++ b/core/pkg/opencost/cloudcostprops.go
@@ -1,10 +1,14 @@
 package opencost
 
 import (
+	"encoding/hex"
 	"fmt"
+	"hash/fnv"
+	"sort"
 	"strings"
 
 	"github.com/opencost/opencost/core/pkg/log"
+	"golang.org/x/exp/maps"
 )
 
 type CloudCostProperty string
@@ -31,6 +35,7 @@ const (
 	CloudCostCategoryProp        string = "category"
 	CloudCostServiceProp         string = "service"
 	CloudCostLabelProp           string = "label"
+	CloudCostLabelSetProp        string = "labelSet"
 )
 
 func ParseCloudProperties(props []string) ([]CloudCostProperty, error) {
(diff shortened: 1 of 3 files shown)


nik-kc (approved):

core/pkg/opencost/cloudcostprops.go
```diff
@@ -283,3 +288,26 @@ func (ccp *CloudCostProperties) GenerateKey(props []string) string {
...
+	builder.WriteString(ccp.Service)
+	builder.WriteString(ccp.Category)
+
+	// Sort label keys before adding key/value pairs to the hash string
+	labelKeys := maps.Keys(ccp.Labels)
+	sort.Strings(labelKeys)
+	for _, k := range labelKeys {
+		builder.WriteString(k)
+		builder.WriteString(ccp.Labels[k])
+	}
```
michaelmdresser: Would you please add a comment explaining the bit about label erasure that motivates this code? You explained it well in the PR description

michaelmdresser (approved):

PR #2707, opened 2024-04-19
TypeUtil Enhancements 

mbolt35:
## What does this PR change?
* Leans on the standard library more for the `TypeOf[T]` functionality
* Adds `PackageOf[T]` which will return the fully qualified package for a `T` type
* Adds `PackageFromCaller(depth int)` to allow multi-depth call package name extraction.
* Adds `CurrentPackage()` which returns the fully qualified package of the calling function.
* Adds relevant unit tests for the old functionality as well as new. 

* It is purely utility and won't have any direct impact on users. 

## How was this PR tested?
* Added Unit Tests

### Use Cases
The main purpose for the enhancements to this core utility is to provide a very generic way of defining Tracers (which use the package name as a convention). 

for example: 
Assuming the base package: `github.com/opencost/opencost/core`, if I were to add a new source file in: `core/pkg/opencost/tracing.go` which contained the following contents:

```go 
package opencost 

import (
    "go.opentelemetry.io/otel"
    "github.com/opencost/opencost/core/pkg/util/typeutil"
)

// Identical to:
// var tracer = otel.Tracer("github.com/opencost/opencost/core/pkg/opencost") 
var tracer = otel.Tracer(typeutil.CurrentPackage())

func SomeInstrumentedFunction(ctx context.Context) {
    _, span := tracer.Start(ctx, "SomeInstrumentedFunction")
    defer span.End() 

    // expensive operation
}
```

You may want to fully abstract away the `typeutil` call by creating a factory function in another package. You could do this using `typeutil.PackageFromCaller(depth int) string`. For example, create a new package: `github.com/opencost/opencost/core/pkg/tracing` with the following in `core/pkg/tracing/tracing.go`:

```go 
package tracing 

import (
    "go.opentelemetry.io/otel"
    "go.opentelemetry.io/otel/trace"
    "github.com/opencost/opencost/core/pkg/util/typeutil"
)

// NewPackageTracer returns a new trace.Tracer implementation with the name set to the calling package
func NewPackageTracer() trace.Tracer {
    // instead of using typeutil.CurrentPackage() here (this would use "github.com/opencost/opencost/core/pkg/tracing")
    // we want to use the caller's package, so we use a custom depth of 2 with typeutil.PackageFromCaller(depth int) string
    // (2) Caller -> (1) NewPackageTracer() -> (0) typeutil.PackageForCaller()
    return otel.Tracer(typeutil.PackageFromCaller(2))
}
```

Then inside our `core/pkg/opencost/tracing.go` file, we could just simply use:

```go 
package opencost 

import (
    "github.com/opencost/opencost/core/pkg/tracing"
)

// Identical to:
// var tracer = otel.Tracer("github.com/opencost/opencost/core/pkg/opencost") 
var tracer = tracing.NewPackageTracer()

func SomeInstrumentedFunction(ctx context.Context) {
    _, span := tracer.Start(ctx, "SomeInstrumentedFunction")
    defer span.End() 

    // expensive operation
}
```

### DIFF
--- a/core/pkg/util/typeutil/typeutil.go
+++ b/core/pkg/util/typeutil/typeutil.go
@@ -51,11 +51,10 @@ func PackageFor(value any) string {
 	return t.PkgPath()
 }
 
-// CurrentPackage returns the package name of the caller. This is especially handy for automatically
-// generating package scoped tracing identifiers.
-func CurrentPackage() string {
+// PackageFromCaller returns the package name of the caller at the specified depth.
+func PackageFromCaller(depth int) string {
 	// get program counter for the first depth caller into this function
-	if pc, _, _, ok := runtime.Caller(1); ok {
+	if pc, _, _, ok := runtime.Caller(depth); ok {
 		f := runtime.FuncForPC(pc)
 		if f == nil {
 			return ""
@@ -88,4 +87,11 @@ func CurrentPackage() string {
 		return parentPkg + pkg[:dotIndex]
 	}
 	return ""
+}
+
+// CurrentPackage returns the package name of the caller. This is especially handy for automatically
+// generating package scoped tracing identifiers.
+func CurrentPackage() string {
+	// Depth is from: (0) Caller -> (1) CurrentPackage -> (2) PackageFromCaller
+	return PackageFromCaller(2)
 }
--- a/core/pkg/util/typeutil/typeutil.go
+++ b/core/pkg/util/typeutil/typeutil.go
@@ -3,15 +3,15 @@ package typeutil
 import (
 	"fmt"
 	"reflect"
+	"runtime"
+	"strings"
 )
 
 // TypeOf is a utility that can covert a T type to a package + type name for generic types.
 func TypeOf[T any]() string {
-	var inst T
 	var prefix string
 
-	// get a reflect.Type of a variable with type T
-	t := reflect.TypeOf(inst)
+	t := reflect.TypeFor[T]()
 
 	// pointer types do not carry the adequate type information, so we need to extract the
 	// underlying types until we reach the non-pointer type, we prepend a * each depth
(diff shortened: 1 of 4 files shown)


core/pkg/util/typeutil/typeutil.go
```diff
@@ -22,9 +22,94 @@ func TypeOf[T any]() string {
...
+	if t.PkgPath() == "" {
+		return prefix + name
 	}
 
 	// combine the prefix, package path, and the type name
-	return fmt.Sprintf("%s%s/%s", prefix, t.PkgPath(), t.Name())
+	return fmt.Sprintf("%s%s/%s", prefix, t.PkgPath(), name)
+}
+
+// TypeFor uses type inferencing to accepts a value and returns the fully qualified package
```
ameijer:
```suggestion
// TypeFor uses type inferencing to accept a value and returns the fully qualified package
```

ameijer (approved):
LGTM, this will cut down on magic strings. Thoughts on a convenience method to hardcode depth to 2? sounds like that would be a common setting in this paradigm and would avoid people wondering what to set that to.

mbolt35:
> LGTM, this will cut down on magic strings. Thoughts on a convenience method to hardcode depth to 2? sounds like that would be a common setting in this paradigm and would avoid people wondering what to set that to. 

That was what I was trying to work out with:

```
typeutil.CurrentPackage()
```

Which does hard code depth to 2. 

The issue is that it does depend on how many functions you jump in the call, so for other cases where you might need a deeper depth, I added the other func. Does it have a use outside of tracing? Ehh, my inner YAGNI voice is currently indifferent 😬

ameijer:
oh, totally missed the bottom part of that file. LGTM

michaelmdresser (approved):

PR #2713, opened 2024-04-23
Update CloudCost service categorization for AWS

thomasvn:
## What does this PR change?
* Update CloudCost service categorization for AWS. GuardDuty should not be classified as "Compute", but instead as "Other".

* N/A

* Fixes categorization of an AWS resource for the CloudCost API.
* For downstream apps (e.g. Kubecost), this also fixes the cost reconciliation process.

* N/A

## How was this PR tested?

Enabled AmazonGuardDuty in one of my environments (it's on a free trial, but still shows up in the CUR with a $0 charge).

Before this change, notice that AmazonGuardDuty was being categorized as "Compute" when it had a ProviderId of `i-*`. There are also cases when other AmazonGuardDuty services are categorized as "Network" or "Other" likely due to other columns in the CUR. Details here. [guardduty-cloudcost.txt](https://github.com/opencost/opencost/files/15180102/guardduty-cloudcost.txt)

After this change, all AmazonGuardDuty which has a ProviderID of `i-*` are now categorized as "Other". Even though runtime monitoring is being performed for each individual EC2 instance, we do not want to merge these costs into the cost of a node (which should solely be CPU + RAM + GPU). Details here. [guardduty-cloudcost-customimage.txt](https://github.com/opencost/opencost/files/15180119/guardduty-cloudcost-customimage.txt)

* N/A

*

### DIFF
--- a/pkg/cloud/aws/athenaquerier.go
+++ b/pkg/cloud/aws/athenaquerier.go
@@ -208,6 +208,10 @@ func SelectAWSCategory(providerID, usageType, service string) string {
 	// The node and volume conditions are mutually exclusive.
 	// Provider ID has prefix "i-"
 	if strings.HasPrefix(providerID, "i-") {
+		// GuardDuty has a ProviderID prefix of "i-", but should not be categorized as compute
+		if strings.ToUpper(service) == "AMAZONGUARDDUTY" {
+			return opencost.OtherCategory
+		}
 		return opencost.ComputeCategory
 	}
 	// Provider ID has prefix "vol-"


michaelmdresser (commented):
Conceptually LGTM but I'd like to see the results of a test before approving

ameijer:
@thomasvn , any other context information we can possibly use? trying to use a check on 'is guard duty' if possible

thomasvn:
@ameijer Not seeing anything else we could use to validate the service is `AmazonGuardDuty`. It seems the existing code mostly checks the `Service` name to determine its category:

https://github.com/opencost/opencost/blob/540b35727c6cb120aab07ad20d10ffe7e619d5ca/pkg/cloud/aws/athenaquerier.go#L218-L234

And it seems `Service` name is derived from this column of the CUR:

https://github.com/opencost/opencost/blob/540b35727c6cb120aab07ad20d10ffe7e619d5ca/pkg/cloud/aws/athenaintegration.go#L354

ameijer:
OK @thomasvn , after reviewing the code you linked, I think your approach is solid and in line with our existing approach. Would also love to see a live test as per @michaelmdresser 's request

AjayTripathy:
Are there services other than GuardDuty that target nodes? Should we not do the inverse and whitelist rows that are used for pricing instead of blacklisting GuardDuty? I am ok with a temporary fix here but we should think about future-proofing.

thomasvn:
> Are there services other than GuardDuty that target nodes?

@AjayTripathy Not that I know of, but it's certainly possible.

> Should we not do the inverse and whitelist rows that are used for pricing instead of blacklisting GuardDuty? I am ok with a temporary fix here but we should think about future-proofing.

This is something we considered, but we would need to be very confident that we have coverage of all the different `Service` names which would be considered Compute. Would definitely require a bit more rigorous testing. I'm open to doing this though! Change would look like:

```go
	if strings.HasPrefix(providerID, "i-") && strings.ToUpper(service) == "AMAZONEC2" {
		return opencost.ComputeCategory
	}
```

AjayTripathy:
Ok. Let's start with the p0 fix (blacklisting guardduty) which is the known issue, but file an issue to handle via whitelist and get that appropriately tested?

Sean-Holcomb (commented):
This looks good to me, my only question is if we can't just put this if above the if it is nested in. Are there providerId with "vol-" and service guard duty that we want classified as storage?

thomasvn:
@Sean-Holcomb I'm still currently testing in our environment, but I don't believe there is ever `Service=AmazonGuardDuty` and `ProviderID=vol-*`.

thomasvn:
@ameijer @Sean-Holcomb I've tested the following change in one of my environments. Details in the PR notes above. Let me know what you think and if we are good to approve/merge!

ameijer (approved):
thomas I reviewed the tests and it LGTM, sean is FTO today so @cliffcolvin might need to merge for us

Sean-Holcomb (approved):

PR #2717, opened 2024-04-25
demoted s3 commands to trace

nickcurie:
## What does this PR change?
* Demotes s3 command logs to trace (from debug)

* https://github.com/kubecost/kubecost-cost-model/pull/2389

* s3 command logs moved to trace

* 

## How was this PR tested?
* Local run

* No

* v2.3

### DIFF
--- a/pkg/storage/s3storage.go
+++ b/pkg/storage/s3storage.go
@@ -357,7 +357,7 @@ func (s3 *S3Storage) Read(name string) ([]byte, error) {
 // Exists checks if the given object exists.
 func (s3 *S3Storage) Exists(name string) (bool, error) {
 	name = trimLeading(name)
-	//log.Tracef("S3Storage::Exists(%s)", name)
+	log.Tracef("S3Storage::Exists(%s)", name)
 
 	ctx := context.Background()
 
@@ -410,7 +410,7 @@ func (s3 *S3Storage) Write(name string, data []byte) error {
 func (s3 *S3Storage) Stat(name string) (*StorageInfo, error) {
 	name = trimLeading(name)
 
-	//log.Tracef("S3Storage::Stat(%s)", name)
+	log.Tracef("S3Storage::Stat(%s)", name)
 	ctx := context.Background()
 
 	objInfo, err := s3.client.StatObject(ctx, s3.name, name, minio.StatObjectOptions{})
--- a/pkg/storage/s3storage.go
+++ b/pkg/storage/s3storage.go
@@ -347,7 +347,7 @@ func (s3 *S3Storage) FullPath(name string) string {
 func (s3 *S3Storage) Read(name string) ([]byte, error) {
 	name = trimLeading(name)
 
-	log.Debugf("S3Storage::Read(%s)", name)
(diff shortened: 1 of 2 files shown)


pkg/storage/s3storage.go
```diff
@@ -347,7 +347,7 @@ func (s3 *S3Storage) FullPath(name string) string {
 func (s3 *S3Storage) Read(name string) ([]byte, error) {
 	name = trimLeading(name)
 
-	log.Debugf("S3Storage::Read(%s)", name)
+	log.Tracef("S3Storage::Read(%s)", name)
```
michaelmdresser, line 350: I seem to recall this being useful for debugging, but maybe we have another higher-level log in Aggregator ingestion which outputs the file name also?

michaelmdresser (approved):

pkg/storage/s3storage.go
```diff
@@ -357,7 +357,7 @@ func (s3 *S3Storage) Read(name string) ([]byte, error) {
 // Exists checks if the given object exists.
 func (s3 *S3Storage) Exists(name string) (bool, error) {
 	name = trimLeading(name)
-	//log.Debugf("S3Storage::Exists(%s)", name)
+	//log.Tracef("S3Storage::Exists(%s)", name)
```
ameijer: uncomment this now that it is demoted

pkg/storage/s3storage.go
```diff
@@ -410,7 +410,7 @@ func (s3 *S3Storage) Write(name string, data []byte) error {
 func (s3 *S3Storage) Stat(name string) (*StorageInfo, error) {
 	name = trimLeading(name)
 
-	//log.Debugf("S3Storage::Stat(%s)", name)
+	//log.Tracef("S3Storage::Stat(%s)", name)
```
ameijer: uncomment

pkg/storage/s3storage.go
thomasvn: I believe it was most useful for ETL Utils. However we can add in some higher-level INF or DBG logging if needed.

ameijer:
@nickcurie this LGTM, why don't you sign off on the code and I will approve!

nickcurie:
@ameijer and merge too please :) (if you're a maintainer)

thomasvn:
@nickcurie In your testing, did you find that these S3 logs were too noisy at the DEBUG level? In multi-cluster architectures where S3 is your "source-of-truth" filesystem, it can sometimes be helpful to know how pods are reading & writing to that filesystem.

nickcurie:
@thomasvn yeah the s3 logs are super noisy during ingestion, and that was my reasoning to move them to trace. I think if we can somehow get them to either be less noisy or have a different log at the debug level that encapsulates what the (new) trace logs are doing that would be ideal. I'm not really sure how feasible that is though. I think if we feel strongly about them we could keep them as debug, or just keep the read as debug.

thomasvn:
@nickcurie Ok thanks for the context. Good with me! Let's move forward with this.

nikovacevic (approved):

PR #2719, opened 2024-04-28
initial pass of a whitelist that adds labels relevant for aggregation

AjayTripathy:
## What does this PR change?
* Support a label whitelist, but leave all labels on  

* No

* Lower prometheus memory impact.

## How was this PR tested?
* Unit tests, run manually on a GKE cluster

* Yes; we should document this feature when ready.

* Still mainly a proposal


AjayTripathy:
@cliffcolvin could we get a review on this one?

pkg/metrics/podlabelmetrics_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"testing"
+
+	"github.com/opencost/opencost/pkg/clustercache"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestWhitelist(t *testing.T) {
```
nikovacevic, line 12: Any way we could get an assertion here that could fail? Hard to call it a "test" if it can't fail. (Or perhaps I am being dumb and am missing that bit?)

pkg/metrics/metricsconfig.go
```diff
@@ -17,7 +17,9 @@ var (
 )
 
 type MetricsConfig struct {
-	DisabledMetrics []string `json:"disabledMetrics"`
+	DisabledMetrics    []string        `json:"disabledMetrics"`
+	UseLabelsWhitelist bool            `json:"useLabelsWhitelist,omitempty"`
+	LabelsWhitelist    map[string]bool `json:"labelsWhiteList,omitempty"`
```
nikovacevic, line 22: Where does this get configured? How would it be used?

pkg/metrics/podlabelmetrics.go
```diff
@@ -29,6 +37,40 @@ func (kpmc KubePodLabelsCollector) Describe(ch chan<- *prometheus.Desc) {
 	}
 }
 
+func (kpmc *KubePodLabelsCollector) UpdateControllerSelectorsCache() {
```
nikovacevic, line 40: Do we need anything Deployment-specific here? Or does the ReplicaSet case cover Deployments adequately? Looking over `pkg/costmodel/allocation.go` the only relevant metrics I see are:
- service_selector_labels
- deployment_match_labels
- statefulSet_match_labels

nikovacevic (approved):

pkg/metrics/podlabelmetrics_test.go
AjayTripathy: ah, this is my bad. good catch, I'll add an assertion.

pkg/metrics/podlabelmetrics.go
AjayTripathy: replicasets covers deployments!

pkg/metrics/metricsconfig.go
AjayTripathy: This is passed in a metricsConfigs JSON file https://github.com/kubecost/cost-analyzer-helm-chart/blob/16f11a314f43bb604a5d99bf3c92801c8a374b96/cost-analyzer/values.yaml#L3290

https://github.com/kubecost/cost-analyzer-helm-chart/blob/16f11a314f43bb604a5d99bf3c92801c8a374b96/cost-analyzer/templates/cost-analyzer-metrics-config-map-template.yaml#L11

thomasvn: Would it be helpful to be explicit in the variable name that these only apply to pod labels?

thomasvn (approved):

PR #2734, opened 2024-05-06
Prevent prefix from being applied twice to files

Sean-Holcomb:
## What does this PR change?
This PR addresses and issue with the PrefixBucketStorage, where absolute paths returned by its list function which include the prefix that it prepends will have the prefix prepended to them again when passed back to the storage. To address this before adding on the prefix there is now a check that ensures that the prefix is not already present in the path provided.

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/storage/prefixedbucketstorage.go
+++ b/pkg/storage/prefixedbucketstorage.go
@@ -32,7 +32,7 @@ func validPrefix(prefix string) bool {
 }
 
 func conditionalPrefix(prefix, name string) string {
-	if len(name) > 0 {
+	if len(name) > 0 && !strings.HasPrefix(name, prefix) {
 		return withPrefix(prefix, name)
 	}
 


cliffcolvin (approved):

thomasvn:
@Sean-Holcomb Nice defensive check here! Below, we're calling the List functions and passing the filepath with the prepended prefix:

https://github.com/opencost/opencost/blob/56098ed6c9d3ada5bc3e050bdf5ee368aff75543/pkg/storage/prefixedbucketstorage.go#L60-L68

Are you saying that there are other Storage implementations also prepending the prefix? If so could you help me understand where?

Sean-Holcomb:
@thomasvn I am saying that these functions are returning absolute paths which include the prefix, so if you listDir then listDir again you get the prefix prepended on an absolute path that already included the prefix 
```
2024-05-06T15:21:53.58606-07:00 DBG GCSStorage::List(xxxx-xxxx-45ae-xx-exxxxxxxxb893/xxxx-xxxx-45ae-xx-exxxxxxxxb893/federated/cluster-1/etl/bingen/allocations/1d)
```

PR #2760, opened 2024-05-28
GCP Cloud Costs resource name fallback

nik-kc:
## What does this PR change?
* Updates Cloud Cost ingestion for GCP to fall back to `resource.global_name` when `resource.name` is `null` for determining `ProviderID`.
* This is particularly relevant for Cloud SQL, Cloud Storage, and Cloud Logging, which very often have null `resource.name` values, resulting in `__unallocated__` `ProviderID` values.

* Nope.

* Users will no longer see `__unallocated__` in place of the correct `ProviderID`, especially for Cloud SQL, Cloud Storage, and Cloud Logging.

* Closes [GTM-286](https://kubecost.atlassian.net/browse/GTM-286).

## How was this PR tested?
* Tested new query in BigQuery, validated query results and data present in `resource.global_name`.
* To be tested in-cluster.

* Nope.

### DIFF
--- a/pkg/cloud/gcp/bigqueryintegration_types.go
+++ b/pkg/cloud/gcp/bigqueryintegration_types.go
@@ -113,6 +113,25 @@ func (ccl *CloudCostLoader) Load(values []bigquery.Value, schema bigquery.Schema
 			}
 
 			properties.ProviderID = ParseProviderID(resource)
+		case ResourceGlobalNameColumnName:
+			// skip if we already got ProviderID from resource.name, as resource.global_name is a fallback for when
+			// resource.name is null
+			if len(properties.ProviderID) > 0 {
+				continue
+			}
+
+			resourceGlobalNameValue := values[i]
+			if resourceGlobalNameValue == nil {
+				continue
+			}
+			resourceGlobalName, ok := resourceGlobalNameValue.(string)
+			if !ok {
+				log.DedupedErrorf(5, "error parsing GCP CloudCost %s: %v", ResourceGlobalNameColumnName, values[i])
+				properties.ProviderID = ""
+				continue
+			}
+
+			properties.ProviderID = ParseProviderID(resourceGlobalName)
 		case CostColumnName:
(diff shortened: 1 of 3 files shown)


teevans:
@nik-kc  - Any chance we can get some of those fine unit tests for this?

nik-kc:
> @nik-kc - Any chance we can get some of those fine unit tests for this?

As you wish!

pkg/cloud/gcp/bigqueryintegration_types.go
```diff
@@ -113,6 +113,25 @@ func (ccl *CloudCostLoader) Load(values []bigquery.Value, schema bigquery.Schema
...
+		case ResourceGlobalNameColumnName:
+			// skip if we already got ProviderID from resource.name, as resource.global_name is a fallback for when
+			// resource.name is null
+			if len(properties.ProviderID) > 0 {
+				continue
+			}
+
+			resourceGlobalNameValue := values[i]
+			if resourceGlobalNameValue == nil {
+				continue
```
nickcurie, line 126: I think we would want to add `properties.ProviderID = ""` here before continuing right? Just in case we hit the case where all `resource.global_name`'s are nil

nickcurie (changes requested):
One small thing, but otherwise this looks great!

pkg/cloud/gcp/bigqueryintegration_types.go
nik-kc: Good catch! You're absolutely right.

nickcurie (approved):
LGTM!

ameijer (approved):
LGTM

ameijer:
@mattray shouldn't my review be sufficient to merge this? Can we update the permissions there

Sean-Holcomb (approved):

mattray:
> @mattray shouldn't my review be sufficient to merge this? Can we update the permissions there

Fixed, sorry about that.

ameijer:
thank you @mattray !!

PR #2763, opened 2024-05-31
adds efficiency to return type

ameijer:
## What does this PR change?
* This modifies the summary allocation object to also return the efficiency field 

* https://github.com/kubecost/kubecost-cost-model/pull/2500

* Allows them to get additional information 

* Closes ...

## How was this PR tested?
* tested locally on test data 

* 

*

### DIFF
--- a/core/pkg/opencost/summaryallocation.go
+++ b/core/pkg/opencost/summaryallocation.go
@@ -89,7 +89,7 @@ func NewSummaryAllocation(alloc *Allocation, reconcile, reconcileNetwork bool) *
 	if sa.IsUnmounted() {
 		sa.UnmountedPVCost = sa.PVCost
 	}
-
+	sa.Efficiency = sa.TotalEfficiency()
 	return sa
 }
 
@@ -152,6 +152,8 @@ func (sa *SummaryAllocation) Add(that *SummaryAllocation) error {
 	sa.RAMCost += that.RAMCost
 	sa.SharedCost += that.SharedCost
 
+	// zero out the effieciency, as it is no longer valid
+	sa.Efficiency = sa.TotalEfficiency()
 	return nil
 }
 
@@ -174,6 +176,7 @@ func (sa *SummaryAllocation) Clone() *SummaryAllocation {
 		RAMCost:                sa.RAMCost,
 		SharedCost:             sa.SharedCost,
 		ExternalCost:           sa.ExternalCost,
+		Efficiency:             sa.Efficiency,
 	}
 }
 
--- a/core/pkg/opencost/summaryallocation.go
+++ b/core/pkg/opencost/summaryallocation.go
@@ -40,6 +40,7 @@ type SummaryAllocation struct {
 	ExternalCost           float64               `json:"externalCost"`
 	Share                  bool                  `json:"-"`
 	UnmountedPVCost        float64               `json:"-"`
+	Efficiency             float64               `json:"efficiency"`
 }
 
 // NewSummaryAllocation converts an Allocation to a SummaryAllocation by
--- a/core/pkg/opencost/summaryallocation.go
+++ b/core/pkg/opencost/summaryallocation.go
@@ -152,7 +152,6 @@ func (sa *SummaryAllocation) Add(that *SummaryAllocation) error {
 	sa.RAMCost += that.RAMCost
 	sa.SharedCost += that.SharedCost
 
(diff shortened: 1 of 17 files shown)


core/pkg/opencost/summaryallocation.go
```diff
@@ -40,6 +40,7 @@ type SummaryAllocation struct {
 	ExternalCost           float64               `json:"externalCost"`
 	Share                  bool                  `json:"-"`
 	UnmountedPVCost        float64               `json:"-"`
+	Efficiency             float64               `json:"efficiency"`
```
thomasvn, line 43: Is this necessary in the SummaryAllocation struct? What was the issue with the previous `sa.TotalEfficiency()`?

ameijer: It's a good point and worth discussing. We moved the calculation of this to the DB so that it can be sorted. That might not cover us for OC though. I added the field in here so that the DB and the go code can be on the same page and not do the same calculation in 2 places. I will implement a fallback mechanism, wait one

thomasvn:
Nice! Because you've added a new field to the SummaryAllocation struct, do we also need to update its methods? Like the `NewSummaryAllocation()`, `Add()`, `Clone()`, and `Equal()`? In case there are consumers of this struct aside from KCM.

core/pkg/opencost/summaryallocation_json.go
```diff
@@ -35,6 +35,16 @@ func (sa *SummaryAllocation) ToResponse() *SummaryAllocationResponse {
 		return nil
 	}
 
+	// if the efficiency has already been set,
+	// prefer that since it has been calculated in the DB
+	// and matches the sorting criteria more closely
```
thomasvn: Minor nit. Would it be best not to refer to a DB here? Generalize this to any consumer of the SummaryAllocation struct?

ameijer:
@thomasvn good thoughts. I made the changes... left out the update to Equals, though, since it is a derived property from things already being compared there. See what you think!

ameijer:
(also I will fix protobut issue, that is fixed in another PR)

core/pkg/opencost/summaryallocation.go
```diff
@@ -151,6 +152,8 @@ func (sa *SummaryAllocation) Add(that *SummaryAllocation) error {
 	sa.RAMCost += that.RAMCost
 	sa.SharedCost += that.SharedCost
 
+	// zero out the effieciency, as it is no longer valid
```
thomasvn: Is this comment still accurate? Looks like we're recalculating Efficiency after adding together the two SummaryAllocations

core/pkg/opencost/summaryallocation_json.go
```diff
@@ -35,6 +35,16 @@ func (sa *SummaryAllocation) ToResponse() *SummaryAllocationResponse {
 		return nil
 	}
 
+	// if the efficiency has already been set,
+	// prefer that since it has been calculated elsewhere
+	// and matches the sorting criteria more closely
+	efficiency := sa.Efficiency
+	if efficiency == 0 {
```
thomasvn, line 42: Because the OC code now instantiates `sa.Efficiency` when `NewSummaryAllocation()` is called, just want to double check that this condition is still working for you on the KCM side of things.

thomasvn (approved):
LGTM!

thomasvn:
cc @mbolt35 Just want to double check that updating the SummaryAllocation struct won't have any downstream consequences for the cost-model agent?

mbolt35:
> cc @mbolt35 Just want to double check that updating the SummaryAllocation struct won't have any downstream consequences for the cost-model agent?

I think `SummaryAllocation` is safe to make changes without updates to bingen. It does look like you're failing a protobuf validation test though. 
image

ameijer:
yeah @mbolt35 I didn't pin the protoc version in the CI, is fixed in another PR

core/pkg/opencost/summaryallocation.go
ameijer: you are correct! fixed

core/pkg/opencost/summaryallocation_json.go
ameijer: thanks Thomas - I just reviewed KCM and we use `result := new(opencost.SummaryAllocation)` as the ctor, so we are all covered!

cliffcolvin (approved):
LGTM

PR #2765, opened 2024-06-01
ENG-1706 Fix parsing of AWS OnDemand data for certain instance types

### LINKED ISSUE #2796: Opencost has wrong price for AWS p4d.24xlarge instance type starting 4/26/2024
**Describe the bug**
On the morning of 4/26 the metric `node_total_hourly_cost{job="opencost", instance_type="p4d.24xlarge"}` was reporting as 32.8, by 10AM PST it started reporting as 0.0320.

**To Reproduce**
Check the metric `node_total_hourly_cost{job="opencost", instance_type="p4d.24xlarge"}` on an EKS cluster with `p4d.24xlarge` nodes.

**Expected behavior**
Expected to get 32.8

**Screenshots**
Screenshot 2024-06-13 at 10 48 18 AM

Helm chart 1.38.1

**Additional context**
Using EKS.

thomasvn:
## What does this PR change?
* Fixes how OpenCost parses & applies AWS OnDemand pricing data for specific instance types.
* Specifically, instance types which show up twice in the pricing data. Once with `marketOption: OnDemand` and another time with `marketOption: CapacityBlock`. We are only concerned with the SKU associated with `OnDemand`.

* N/A

* If you have Cloud Billing Integration enabled to reconcile your OnDemand Pricing data, there is no impact.
* If using OnDemand Pricing data, this will fix pricing for AWS Instance Types which can be reserved via CapacityBlock (e.g. p4d.24xlarge, p5.48xlarge). This will only fix data from present time moving forward.

* Closes [ENG-1706](https://kubecost.atlassian.net/browse/ENG-1706)
* Fixes https://github.com/opencost/opencost/issues/2796

## How was this PR tested?
* Adds a unit test which validates that the code prioritizes `marketOption: OnDemand` even in the existence of `marketOption: CapacityBlock`
* Adds a regression test which downloads AWS OnDemand Pricing data and checks its schema. Throws a test failure if there is an unrecognized `marketOption` value. This would not necessarily indicate that a regression has occurred, but that we need to revisit our OnDemand pricing code to ensure it's still accurate.

I performed a manual mock test by relabeling a node's label to `node.kubernetes.io/instance-type: p5.48xlarge`.

**BEFORE**. 0.021

```
node_total_hourly_cost{arch="amd64",instance="ip-192-168-52-187.us-east-2.compute.internal",instance_type="p5.48xlarge",node="ip-192-168-52-187.us-east-2.compute.internal",provider_id="aws:///us-east-2b/i-0fef50fedefa4510e",region="us-east-2"} 0.02180604860305786
```

**AFTER**. 98.320

```
node_total_hourly_cost{arch="amd64",instance="ip-192-168-52-187.us-east-2.compute.internal",instance_type="p5.48xlarge",node="ip-192-168-52-187.us-east-2.compute.internal",provider_id="aws:///us-east-2b/i-0fef50fedefa4510e",region="us-east-2"} 98.3200896577759
```

* N/A

* N/A

### DIFF
--- a/justfile
+++ b/justfile
@@ -11,11 +11,14 @@ test-core:
     {{commonenv}} cd ./core && go test ./... -coverprofile=coverage.out
     {{commonenv}} cd ./core && go vet ./...
 
-
 # Run unit tests
 test: test-core
     {{commonenv}} go test ./... -coverprofile=coverage.out
     {{commonenv}} go vet ./...
+
+# Run unit tests and integration tests
+test-integration:
+    {{commonenv}} INTEGRATION=true go test ./... -coverprofile=coverage.out
 
 # Compile a local binary
 build-local:
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -1000,6 +1000,7 @@ func (aws *AWS) populatePricing(resp *http.Response, inputkeys map[string]bool) 
 
 				if product.Attributes.PreInstalledSw == "NA" &&
 					(strings.HasPrefix(product.Attributes.UsageType, "BoxUsage") || strings.Contains(product.Attributes.UsageType, "-BoxUsage")) &&
+					// product.Attributes.CapacityStatus == "Used" {
 					product.Attributes.CapacityStatus == "Used" &&
 					product.Attributes.MarketOption == "OnDemand" {
 					key := aws.KubeAttrConversion(product.Attributes.RegionCode, product.Attributes.InstanceType, product.Attributes.OperatingSystem)
@@ -1354,7 +1355,6 @@ func (aws *AWS) createNode(terms *AWSProductTerms, usageType string, k models.Ke
 	}
 	// Throw error if public price is not found
 	if !publicPricingFound {
-		log.Errorf("For node \"%s\", cannot find the following key in OnDemand pricing data \"%s\"", k.ID(), k.Features())
 		return nil, meta, fmt.Errorf("for node \"%s\", cannot find the following key in OnDemand pricing data \"%s\"", k.ID(), k.Features())
 	}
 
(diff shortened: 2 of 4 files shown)


Sean-Holcomb (approved):
Looks good, what is the run time on the unit test that hits the api?

thomasvn:
@Sean-Holcomb When run in Github's CI, it looks to be about 24s.

```
ok  	github.com/opencost/opencost/pkg/cloud/aws	24.640s	coverage: 14.7% of statements
ok  	github.com/opencost/opencost/pkg/cloud/azure	0.056s	coverage: 18.6% of statements
ok  	github.com/opencost/opencost/pkg/cloud/config	0.052s	coverage: 21.8% of statements
ok  	github.com/opencost/opencost/pkg/cloud/gcp	0.057s	coverage: 19.8% of statements
ok  	github.com/opencost/opencost/pkg/cloud/models	0.017s	coverage: 12.3% of statements
ok  	github.com/opencost/opencost/pkg/cloud/oracle	0.055s	coverage: 15.5% of statements
ok  	github.com/opencost/opencost/pkg/cloudcost	0.042s	coverage: 6.4% of statements
```

Sean-Holcomb:
That is a little long for a single unit test, and it is in fact an integration test. What do you think about adding 
```
if os.Getenv("INTEGRATION") == "" {
	t.Skip("skipping integration tests, set environment variable INTEGRATION")
}
```
to the top of the test, which makes it so that the test will not run unless that env variable isn't present. That way we aren't adding 20 seconds onto every build we do.

thomasvn:
@Sean-Holcomb I agree that 20seconds for each build may be too long. I've added in your recommended change, and also updated the `Justfile` to provide the option to run unit tests + integration tests.

Although these integration tests don't need to run for each PR commit, we should eventually establish a cadence to be running these integration tests.

jdaciuk:
@thomasvn after this merge the `node_total_hourly_cost` metric is indeed fixed, but aggregated cost queries to the opencost server are still incorrect (way to low). Any ideas on why that could be the case? Is the server not just using the hourly cost prometheus metrics every time it's hit?

thomasvn:
@jdaciuk Could you share the exact aggregated cost queries you're running against OpenCost?

One thing to note, is that this fix does not affect any historical data that has already been written into Prometheus. Could you try running the aggregated cost queries with a `window` which only includes new data from the time since you updated your container image?

jdaciuk:
@thomasvn this is my mistake somewhere downstream of the opencost query actually. The query results have been updated. Thanks again for the help here!

thomasvn:
@jdaciuk No problem, glad you found the fix!

PR #2768, opened 2024-06-03
[ENG-947] Window rounding, remove prom duration with offset logic

ameijer:
## What does this PR change?
* window parsing logic rounds all windows now, not just daily/weekly

* none

* When users submit a window, it will automatically get rounded to each window boundary

* Closes https://kubecost.atlassian.net/browse/ENG-947

## How was this PR tested?
* tested locally and via unit testing. A new unit test was added that enforces the rounding behavior

* 

*

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: install protobuf-go
         run: |
           go install github.com/golang/protobuf/protoc-gen-go@latest
-          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@1.3.0
+          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0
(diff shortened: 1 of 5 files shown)


.github/workflows/build-test.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: install protobuf-go
         run: |
           go install github.com/golang/protobuf/protoc-gen-go@latest
-          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0
```
ameijer, line 37: we needed to pin these to prevent CI failures, latest is mutable

nik-kc (approved):
:100:

nikovacevic (approved):
Amazingly simple change

PR #2778, opened 2024-06-07
Custom Cost Inconsistencies - ENG-2212

nik-kc:
## What does this PR change?
* Adds missing `accumulate` parameter parsing to the function that handles Custom Cost totals requests.
  * There were circumstances where Custom Cost `totals` and `timeseries` requests would be processed differently, even with the same request parameters. This was caused by downstream logic that would attempt to pick a value for `accumulate` when not explicitly set, which could set `accumulate` to a different value than what the user passed in.
* Defaults `accumulate` to `day`, for both `totals` and `timeseries`.

* Nope.

* Custom Cost totals and timeseries requests will be handled as expected.

* Closes [ENG-2212](https://kubecost.atlassian.net/browse/ENG-2212).

## How was this PR tested?
* Reproduced issue with nightly data, and validated that these changes resolved the issue.

* Nope.

### DIFF
--- a/pkg/customcost/queryservice_helper.go
+++ b/pkg/customcost/queryservice_helper.go
@@ -28,6 +28,8 @@ func ParseCustomCostTotalRequest(qp httputil.QueryParams) (*CostTotalRequest, er
 		return nil, err
 	}
 
+	accumulate := opencost.ParseAccumulate(qp.Get("accumulate", "day"))
+
 	var filter filter.Filter
 	filterString := qp.Get("filter", "")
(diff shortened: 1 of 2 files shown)


ameijer (commented):
@nik-kc any way I could convince you to unit test to look for both the default you chose and that totals and timeseries set accumulate in the first place? code itself LGTM

cliffcolvin (approved):
LGTM

ameijer (approved):

PR #2794, opened 2024-06-11
Handle partitioned and updating billing exports

Sean-Holcomb:
## What does this PR change?
This PR updates the Azure storage Cloud Cost to allow it to work with billing exports with various setting for partitioned and file overwrite. Until now only unpartitioned non-overwriting billing exports were supported. 

To support file overwrite we need to make sure that we are redownloading a blob if there have been updates to it since the last time we downloaded it. This is a simple of process of comparing the mod time of the blob and the local version if it has already been downloaded.

To support partitioned exports we look for the presence of a `manifest.json` file which contains a list of the blobs that contain the partitions, if this is located instead of adding the newest file for a month we add all blobs named in the manifest.

* 

* Closes ...

## How was this PR tested?
* This PR was tested via integration test against configs with unpartitioned non-replacement and partitioned replacement exports, resulting in the correct blobs being selected and valid cloud cost set being generated

* 

*

### DIFF
--- a/pkg/cloud/azure/storagebillingparser.go
+++ b/pkg/cloud/azure/storagebillingparser.go
@@ -3,6 +3,7 @@ package azure
 import (
 	"context"
 	"encoding/csv"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/cloud/azure/storagebillingparser.go
```diff
@@ -133,18 +141,20 @@ func (asbp *AzureStorageBillingParser) parseCSV(start, end time.Time, reader *cs
 	return nil
 }
 
-// getMostRecentBlobs returns a list of filepaths on the Azure Storage
+// getMostRecentBlobs returns a list of blobs in the Azure Storage
 // Container. It uses the "Last Modified Time" of the file to determine which
 // has the latest month-to-date billing data.
-func (asbp *AzureStorageBillingParser) getMostRecentBlobs(start, end time.Time, client *azblob.Client, ctx context.Context) ([]string, error) {
```
Sean-Holcomb, line 139: pass entire blob info object up rather than just the name to preserve mod time info which is used later on

nikovacevic (approved):

pkg/cloud/azure/storageconnection.go
```diff
@@ -80,11 +81,18 @@ func (sc *StorageConnection) StreamBlob(blobName string, client *azblob.Client)
...
-		log.DedupedInfof(3, "CloudCost: Azure: DownloadBlobToFile: file %v already exists, not downloading %v", localFilePath, blobName)
-		return nil
+func (sc *StorageConnection) DownloadBlobToFile(localFilePath string, blob container.BlobItem, client *azblob.Client, ctx context.Context) error {
+	blobName := *blob.Name
+	// Check if file already exists
+	if fileInfo, err := os.Stat(localFilePath); err == nil {
+		blobModTime := *blob.Properties.LastModified
+		// Check if the blob was last modified before the file was modified, indicating that the
+		// file is the most recent version of the blob
+		if blobModTime.Before(fileInfo.ModTime()) {
```
thomasvn, line 91: What happens if this is not true? In other words, if the Blob file has a more recent timestamp than our local file's timestamp? Does this mean we need to go fetch the updated CSV file from storage?

^^ I don't think that scenario would ever happen though.

thomasvn: Oops nevermind. After re-reading this, I see that we would proceed to download the updated Blob file! This comment is good to close.

pkg/cloud/azure/storagebillingparser.go
```diff
@@ -167,27 +177,89 @@ func (asbp *AzureStorageBillingParser) getMostRecentBlobs(start, end time.Time,
...
+			continue
+		}
+
+		// download manifest for the month
+		manifestBytes, err := asbp.DownloadBlob(*mostRecentManifest.Name, client, ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to retrieve manifest %w", err)
+		}
+
+		log.Infof(string(manifestBytes))
```
thomasvn:
```suggestion
```

Was this used for debugging?

pkg/cloud/azure/storagebillingparser.go
```diff
@@ -47,27 +48,29 @@ func (asbp *AzureStorageBillingParser) ParseBillingData(start, end time.Time, re
 	}
 	ctx := context.Background()
 	// Example blobNames: [ export/myExport/20240101-20240131/myExport_758a42af-0731-4edb-b498-1e523bb40f12.csv ]
```
thomasvn, line 51:
```suggestion
```

Comment here is probably no longer relevant?

nik-kc (approved):

pkg/cloud/azure/storagebillingparser.go
Sean-Holcomb: oops, yeah that is what that is

PR #2822, opened 2024-07-01
Adding support for Linode cloud

### LINKED ISSUE #2128: Support Linode
I'd like to be able to use OpenCost, but I use Linode, and you don't have hooks into their API as of yet.

I'd like Linode to be supported in the same way Scaleway are, preferrably.

I considered creating a custom config for it, but the pricing is too granular in comparison for me to be able to figure out.

**Additional context**
https://www.linode.com/docs/api/linode-types/#types-list

mhmxs:
## What does this PR change?
* Adds support for Linode cloud into the OpenCost project.

* No, this PR is independent and does not relate to any other PRs.

* Users will now be able to monitor and manage costs for Linode cloud clusters within the OpenCost platform.

* Closes https://github.com/opencost/opencost/issues/2128

## How was this PR tested?
* <s>Unit tests were added and run to verify the integration with Linode cloud.</s>
* Manual testing was performed by creating and managing Linode clusters to ensure accurate cost reporting.

* Yes, the documentation will need to be updated to include instructions for configuring and using Linode cloud support in OpenCost.

* No, because i don't have permission


mattray:
@mhmxs let us know when you're ready for review, but if you could fix the DCO that'll be needed too.

mhmxs:
@mattray Could you please help me figure out what are these warning means, and how to solve them (if i need to)?

```
WRN ETL: did not find allocations for asset key: lke195854-ctx/pvc-7489e92e12a3469f
```
:pray:

pkg/costmodel/allocation_helpers.go
```diff
@@ -1707,12 +1707,24 @@ func (cm *CostModel) applyNodesToPod(podMap map[podKey]*pod, nodeMap map[nodeKey
...
 // getCustomNodePricing converts the CostModel's configured custom pricing
 // values into a nodePricing instance.
-func (cm *CostModel) getCustomNodePricing(spot bool, providerID string) *nodePricing {
+func (cm *CostModel) getCustomNodePricing(spot bool, providerID, nodeName string) *nodePricing {
 	customPricingConfig, err := cm.Provider.GetConfig()
 	if err != nil {
 		return nil
 	}
 
+	if nodeName != "" {
```
mhmxs: @mattray Maybe i misunderstand something, but our pricing model doesn't fit to the default `customPricingConfig.CPU` model, so we need to find node price, because our `CostPerCPUHr` is different per node types. WDYT?

mhmxs: I found a way to avoid this change.

configs/linode.json
```diff
@@ -0,0 +1,18 @@
+{
+    "provider": "linode",
+    "description": "Linode estimates based on July 2024 advertised prices",
+    "linodeTokenSecret": "kube-system/linode",
+    "CPU": "0.0",
```
Sean-Holcomb, line 5: These should have some value

pkg/cloud/linode/provider.go
```diff
@@ -0,0 +1,195 @@
...
+				if cpricing.LinodeTokenSecret != "" && linodeToken == "" {
+					namespace := "default"
+					name := cpricing.LinodeTokenSecret
+					if secretName := strings.Split(cpricing.LinodeTokenSecret, "/"); len(secretName) > 1 {
+						namespace = secretName[0]
+						name = secretName[1]
+					}
+
+					ctx, cancel := context.WithTimeout(context.Background(), kubeAPITimeout)
+					if apiToken, err := l.KubeClient.CoreV1().Secrets(namespace).Get(ctx, name, v1.GetOptions{}); err != nil {
```
Sean-Holcomb, line 82: Historically we have mounted secrets to the container to pass values like this, can you walk me through your reasoning for passing in the KubeClient and retrieving the secret data in this way.

pkg/costmodel/cluster.go
```diff
@@ -218,7 +218,7 @@ func ClusterDisks(client prometheus.Client, cp models.Provider, start, end time.
 		queryLocalStorageUsedAvg := fmt.Sprintf(`avg(sum(avg_over_time(container_fs_usage_bytes{device=~"/dev/(nvme|sda).*", id="/", %s}[%s])) by (instance, device, %s, job)) by (instance, device, %s)`, env.GetPromClusterFilter(), durStr, env.GetPromClusterLabel(), env.GetPromClusterLabel())
 		queryLocalStorageUsedMax := fmt.Sprintf(`max(sum(max_over_time(container_fs_usage_bytes{device=~"/dev/(nvme|sda).*", id="/", %s}[%s])) by (instance, device, %s, job)) by (instance, device, %s)`, env.GetPromClusterFilter(), durStr, env.GetPromClusterLabel(), env.GetPromClusterLabel())
 		queryLocalStorageBytes := fmt.Sprintf(`avg_over_time(sum(container_fs_limit_bytes{device=~"/dev/(nvme|sda).*", id="/", %s}) by (instance, device, %s)[%s:%dm])`, env.GetPromClusterFilter(), env.GetPromClusterLabel(), durStr, minsPerResolution)
-		queryLocalActiveMins := fmt.Sprintf(`count(node_total_hourly_cost{%s}) by (%s, node)[%s:%dm]`, env.GetPromClusterFilter(), env.GetPromClusterLabel(), durStr, minsPerResolution)
```
Sean-Holcomb, line 221: What is the reasoning behind these prom query changes, I doesn't look like you have made changes that will process the additional labels

pkg/cloud/provider/provider.go
```diff
@@ -242,6 +244,16 @@ func NewProvider(cache clustercache.ClusterCache, apiKey string, config *config.
...
 			Config:           NewProviderConfig(config, cp.configFileName),
 		}, nil
+	case opencost.LinodeProvider:
+		log.Info("Found ProviderID starting with \"linode\", using Linode Provider")
+		return &linode.Linode{
+			KubeClient:       kubeClientset,
+			Clientset:        cache,
+			ClusterRegion:    cp.region,
+			ClusterAccountID: cp.accountID,
+			ClusterProjectID: cp.projectID,
```
Sean-Holcomb, line 254: What is Cluster ProjectID here

configs/linode.json
mhmxs: @Sean-Holcomb, I don't believe Linode can provide the specific pricing numbers we're looking for. I also find it challenging to see how other providers could do this because it's nearly impossible to represent CPU prices accurately in a single, static file.

In this case, the values are set to 0.0 in the config, but the provider's implementation fetches the actual pricing at runtime.

However, the Linode provider implementation currently has some limitations:
- It requires monitoring.coreos.com/v1.ServiceMonitor to function properly (see a related suggestion later).
- It does not support custom pricing.
- Cloud cost support is still under development.

An alternative solution could be adding a new function call here: https://github.com/opencost/opencost/blob/cc5773c1bc4af0838f0bc6f4a6177fe6505e12f6/pkg/costmodel/allocation_helpers.go#L1724. At this point, the node name is already known, so introducing a new function, such as NodePriceByName within the provider, should return the correct prices instead of relying on hard-coded dummy data.

What do you think of this approach?

pkg/cloud/linode/provider.go
mhmxs: @Sean-Holcomb So the main reason behind this solution is key rotation without downtime and extra operational actions. Linode doesn't fit into the static pricing model of OpenCost, so we fetch live pricing data from the provider. Because price should change, we fetch data in a loop, so I didn't see any reason to not refresh the token each time.

pkg/costmodel/cluster.go
mhmxs: This line of code would like to read `provider_id`: https://github.com/opencost/opencost/blob/cc5773c1bc4af0838f0bc6f4a6177fe6505e12f6/pkg/costmodel/cluster.go#L454 But without this change it is missing all the time.

pkg/cloud/provider/provider.go
mhmxs: ProjectID is a cluster IDish thing, exactly the cluster name reference from hostname (the closest data to cluster ID represented on the cluster)

pkg/costmodel/cluster.go
```diff
@@ -641,7 +641,7 @@ func ClusterNodes(cp models.Provider, client prometheus.Client, start, end time.
 	queryNodeRAMBytesAllocatable := fmt.Sprintf(`avg(avg_over_time(kube_node_status_allocatable_memory_bytes{%s}[%s])) by (%s, node)`, env.GetPromClusterFilter(), durStr, env.GetPromClusterLabel())
 	queryNodeGPUCount := fmt.Sprintf(`avg(avg_over_time(node_gpu_count{%s}[%s])) by (%s, node, provider_id)`, env.GetPromClusterFilter(), durStr, env.GetPromClusterLabel())
 	queryNodeGPUHourlyCost := fmt.Sprintf(`avg(avg_over_time(node_gpu_hourly_cost{%s}[%s])) by (%s, node, instance_type, provider_id)`, env.GetPromClusterFilter(), durStr, env.GetPromClusterLabel())
-	queryNodeCPUModeTotal := fmt.Sprintf(`sum(rate(node_cpu_seconds_total{%s}[%s:%dm])) by (kubernetes_node, %s, mode)`, env.GetPromClusterFilter(), durStr, minsPerResolution, env.GetPromClusterLabel())
+	queryNodeCPUModeTotal := fmt.Sprintf(`sum(rate(label_replace(node_cpu_seconds_total{%s}, "kubernetes_node", "$1", "node", "(.*)")[%s:%dm])) by (kubernetes_node, %s, mode)`, env.GetPromClusterFilter(), durStr, minsPerResolution, env.GetPromClusterLabel())
```
mhmxs, line 644: @Sean-Holcomb 
On Linode we don't always have `kubernetes_node`, but we always `node`. This change replace overrides `kubernetes_node` with `node` if it exists.

mhmxs:
> @mattray Could you please help me figure out what are these warning means, and how to solve them (if i need to)?
> 
> ```
> WRN ETL: did not find allocations for asset key: lke195854-ctx/pvc-7489e92e12a3469f
> ```
> 
> 🙏

@Sean-Holcomb Could you please take a look on this question?

What i don't understand is why this function doesn't take care on load balancer, storage or network prices?

shanduur:
Hi! Any updates on this?

.gitignore
Sean-Holcomb, line 1: please remove changes to this file

.dockerignore
Sean-Holcomb, line 1: please remove changes to this file

configs/linode.json
Sean-Holcomb: The point of this file is as fallback pricing, some value here is likely better then nothing. The solution you are suggesting is dependent on the existence of a pricing sheet that will need to be maintained, if that is being done dynamically, then what if the fetch fails? Does the node then have no cost?

pkg/cloud/provider/provider.go
Sean-Holcomb: If the goal is to get a cluster ID lets just save that as part of the object. The `ProjectID` nomenclature is GCP specific and is actually covered by `AccountID` in Cloud Agnostic terminology.

pkg/cloud/linode/provider.go
Sean-Holcomb: My issue is that you are changing the signature of `provider.NewProvider()` to pass in the `kubeClientSet` which is a very large dependency to push through here. 

Just to be clear, you are suggesting that accessing the secret this way allows you to swap out the secret value without the pod coming down which mounting the secret which does not allow.

Sean-Holcomb:
> > @mattray Could you please help me figure out what are these warning means, and how to solve them (if i need to)?
> > ```
> > WRN ETL: did not find allocations for asset key: lke195854-ctx/pvc-7489e92e12a3469f
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 🙏
> 
> @Sean-Holcomb Could you please take a look on this question?
> 
> What i don't understand is why this function doesn't take care on load balancer, storage or network prices?

This one appears to be about disk totals and from what I can tell is not something you should be concerned about.

configs/linode.json
```diff
@@ -0,0 +1,18 @@
...
+    "spotCPU": "0.0",
+    "RAM": "0.0",
+    "spotRAM": "0.0",
+    "GPU": "0.0",
+    "spotGPU": "0.0",
```
shanduur, line 10: We can estimate the values using e.g. Linear Regression, should be good enough for fallback values:

```suggestion
    "CPU": "0.0255",
    "RAM": "0.0060",
    "GPU": "0.6734",
```

shanduur:
Given that Richard is no longer working on this, I will create my own PR and include his work, then fix the comments.

PR #2858, opened 2024-08-05
Fix the issue where the prices for most Alibaba Cloud instance types are incorrect when the instance family generation number is greater than or equal to 7

qixiangyang:
Fix the issue where the prices for most Alibaba Cloud instance types are incorrect when the instance family generation number is greater than or equal to 7.

## What does this PR change?
* According to the customer service and official documentation of Alibaba Cloud, when the instance family generation of an instance is greater than or equal to 7, the system disk does not support "cloud_efficiency". In this case, "cloud_essd" should be used as the default system disk type parameter for queries.

* No

* Get the right price when calculate asset, saving or other result

* Closes ...

## How was this PR tested?
* add test unit

* No

* i want label this pr as "next release", just don't know how to do it


teevans:
Great catch! We'll have a look and see what we can do!

pkg/cloud/alibaba/provider.go
```diff
@@ -1148,6 +1148,24 @@ func getInstanceFamilyFromType(instanceType string) string {
...
+	re := regexp.MustCompile(`(\d+)`)
+	match := re.FindString(familyName)
+	if match != "" {
+		generation, err := strconv.Atoi(match)
+		if err == nil {
+			return generation
+		}
+	}
+	log.Warnf("unable to find the generation of the instance type %s, returning its default generation as 6", instanceType)
+	return 6
```
ameijer: is this safe? What specifically about generation 6 makes this suitable? at minimum, please use a const here to avoid the magic number. I would almost return -1 or something so that it clearly is a bogus value

pkg/cloud/alibaba/provider.go
```diff
@@ -1148,6 +1148,24 @@ func getInstanceFamilyFromType(instanceType string) string {
...
+func getInstanceFamilyGenerationFromType(instanceType string) int {
+	// FamilyName format: g7ne or g7 or r7 or r6e,
+	familyName := getInstanceFamilyFromType(instanceType)
+	re := regexp.MustCompile(`(\d+)`)
+	match := re.FindString(familyName)
+	if match != "" {
+		generation, err := strconv.Atoi(match)
+		if err == nil {
+			return generation
+		}
```
ameijer, line 1165: log error

pkg/cloud/alibaba/provider.go
```diff
@@ -982,7 +982,7 @@ func createDescribePriceACSRequest(i interface{}) (*requests.CommonRequest, erro
 		} else {
 			// When System Disk information is not available for instance family g6e, r7 and r6e the defaults in
 			// DescribePrice dont default rightly to cloud_essd for these instances.
```
ameijer: please update this comment to reflect new case

pkg/cloud/alibaba/provider.go
```diff
@@ -982,7 +982,7 @@ func createDescribePriceACSRequest(i interface{}) (*requests.CommonRequest, erro
...
-			if slices.Contains(alibabaDefaultToCloudEssd, node.InstanceTypeFamily) {
+			if slices.Contains(alibabaDefaultToCloudEssd, node.InstanceTypeFamily) || getInstanceFamilyGenerationFromType(node.InstanceType) > 6 {
```
ameijer, line 985: can you update `alibabaDefaultToCloudEssd` to exclude all >6 generation instance types, so that we don't have multiple conditions that would set this ESSD for gen 7+ instances?

pkg/cloud/alibaba/provider_test.go
```diff
@@ -112,6 +112,20 @@ func TestProcessDescribePriceAndCreateAlibabaPricing(t *testing.T) {
...
+				RegionID:           "cn-hangzhou",
+				PriceUnit:          "Hour",
+				MemorySizeInKiB:    "33554432KiB",
+				IsIoOptimized:      true,
+				OSType:             "Linux",
+				ProviderID:         "cn-hangzhou.i-test-01c",
+				InstanceTypeFamily: "g8a",
+			},
+			expectedError: nil,
+		},
```
ameijer, line 128: @qixiangyang, I'm concerned that simple checking no error won't ensure that this change works as you expect it to. 

Would you be able to enhance the test case to take in a literal request.CommonRequest that you can then do a deep equal on? Or even just a map of the expected query params. That way, this test would ensure your logic is actually applied in this case

ameijer (changes requested):
overall looks good, nice catch, a few small requests to improve the testing here and some documentation

ameijer:
also, please sign off on your code
image

pkg/cloud/alibaba/provider.go
```diff
@@ -1148,6 +1149,26 @@ func getInstanceFamilyFromType(instanceType string) string {
...
+	match := re.FindString(familyName)
+	if match != "" {
+		generation, err := strconv.Atoi(match)
+		if err != nil {
+			log.Errorf("unable to convert the generation of the instance type %s to integer", instanceType)
+		} else {
+			return generation
+		}
+	}
+	log.Warnf("unable to find the generation of the instance type %s, returning its default generation as 6", instanceType)
```
ameijer:
```suggestion
	log.Warnf("unable to find the generation of the instance type %s,", instanceType)
```

pkg/cloud/alibaba/provider.go
```diff
@@ -49,6 +49,7 @@ const (
 	ALIBABA_UNKNOWN_INSTANCE_FAMILY_TYPE       = "unknown"
 	ALIBABA_NOT_SUPPORTED_INSTANCE_FAMILY_TYPE = "unsupported"
 	ALIBABA_DISK_CLOUD_ESSD_CATEGORY           = "cloud_essd"
+	ALIBABA_DISK_CLOUD_EFFICIENCY_CATEGORY     = "cloud_efficiency"
```
ameijer: are you using this const anywhere? did you need it for your new test?

ameijer (changes requested):

pkg/cloud/alibaba/provider.go
qixiangyang: I'm sorry, I forgot to delete it. I did use this variable in the initial test.

ameijer (approved):
thank you @qixiangyang !

PR #2863, opened 2024-08-06
Remove extra group by args for query

Sean-Holcomb:
## What does this PR change?
* This PR fixes queries for Flexible CUD total credits and costs to return one row per day rather than one per day per Flexible CUD. Previously if multiple Flexible CUDs were present, for each day a random total credit and total cost were being selected, resulting in erroneous ratios being created. While the resulting rate from this update is not completely accurate as it is an average of all the CUD rates present in the bill, it will produce numbers that are correct against bill totals

* 

* 

* Closes ...

## How was this PR tested?
* Tested manually against BigQuery Integration

* 

*


mbolt35 (approved):
LGTM

PR #2868, opened 2024-08-12
doc: use markdown rather than HTML for header

glitched-w0rld:
This PR changes the README header to use markdown rather than HTML


AjayTripathy (approved):

AjayTripathy:
Thanks @DemoYeti !

PR #2879, opened 2024-08-22
Worker pkg Update

mbolt35:
## What does this PR change?
* Adds on the existing worker API:
* Added a way to pass a Runner `func(t)` with no return type 
* Added a collector group which only returns non-nil `*U` results. 
* Added ConcurrentDoWith, ConcurrentCollectWith, and ConcurrentRunWith variants which accept the worker pool size to use when running the workloads. 

## How was this PR tested?
* 90%+ coverage with unit tests

### DIFF
--- a/core/pkg/util/worker/worker.go
+++ b/core/pkg/util/worker/worker.go
@@ -9,6 +9,9 @@ import (
 	"github.com/opencost/opencost/core/pkg/collections"
 )
 
+// Runner is a function type that takes a single input and returns nothing.
+type Runner[T any] func(T)
+
 // Worker is a transformation function from input type T to output type U.
 type Worker[T any, U any] func(T) U
 
@@ -54,7 +57,7 @@ type queuedWorkerPool[T any, U any] struct {
 type ordered[T any, U any] struct {
 	workPool WorkerPool[T, U]
 	results  []U
-	wg       *sync.WaitGroup
+	wg       sync.WaitGroup
 	count    int
 }
 
@@ -129,7 +132,6 @@ func NewOrderedGroup[T any, U any](pool WorkerPool[T, U], size int) WorkGroup[T,
 	return &ordered[T, U]{
 		workPool: pool,
 		results:  make([]U, size),
-		wg:       new(sync.WaitGroup),
 		count:    0,
 	}
 }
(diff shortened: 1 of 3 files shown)


core/pkg/util/worker/worker.go
```diff
@@ -166,6 +168,95 @@ func (ow *ordered[T, U]) Wait() []U {
...
+// and ignores any nil results.
+type collector[T any, U any] struct {
+	workPool   WorkerPool[T, *U]
+	resultLock sync.Mutex
+	results    []*U
+	wg         sync.WaitGroup
+}
+
+// NewGroup creates a new WorkGroup implementation for processing a group of inputs in the order in which
+// they are pushed. Ordered groups do not support concurrent Push() calls.
```
ameijer: can you update comment?

core/pkg/util/worker/worker.go
```diff
@@ -166,6 +168,95 @@ func (ow *ordered[T, U]) Wait() []U {
...
+	ow.wg.Add(1)
+
+	go func() {
+		defer ow.wg.Done()
+		defer close(onComplete)
+
+		result := <-onComplete
+		if result != nil {
+			ow.resultLock.Lock()
+			ow.results = append(ow.results, result)
```
ameijer, line 247: neat. this feels powerful

core/pkg/util/worker/worker.go
```diff
@@ -166,6 +168,95 @@ func (ow *ordered[T, U]) Wait() []U {
...
 
+// noResultGroup is a WorkGroup implementation which arbitrarily pushes inputs to
+// a runner pool to be executed concurrently. This group does not collect results.
+type noResultGroup[T any] struct {
+	workPool WorkerPool[T, struct{}]
+	wg       sync.WaitGroup
+}
+
+// NewGroup creates a new WorkGroup implementation for processing a group of inputs in the order in which
+// they are pushed. Ordered groups do not support concurrent Push() calls.
```
ameijer: update comment

ameijer (commented):
code and tests look great. if you could just update comments to reflect key behavioral differences, then LGTM

core/pkg/util/worker/worker.go
mbolt35: definitely felt like a pattern that showed up a lot when using ConcurrentDo, so worth adding here

ameijer (approved):
LGTM

PR #2882, opened 2024-08-22
ENG-136: Exclude local storage provisioner by sig-storage-local-provisioner on node to show up in assets when param ASSET_INCLUDE_LOCAL_DISK_COST is set to false

avrodrigues5:
## What does this PR change?
*   sig-storage-local-provisioner is a storage provisioner that provisions a PV in a host node . This shows up as a cost in assets even when ASSET_INCLUDE_LOCAL_DISK_COST is set to false, which was introduced in [PR](https://github.com/opencost/opencost/pull/2441). The current PR fixes this by having a check to exclude creation of the diskMap which eventually is used to create disk asset when it has a prefix of "local-v-*". We are fortunate that the sig-storage-local-provisioner always provisions pv by this prefix . Here is the code reference [ref](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/b6f465027bd059e92c0032c81dd1e1d90e35c909/pkg/discovery/discovery.go#L410-L417)

* None

* Exclude assets from showing up as cost when ASSET_INCLUDE_LOCAL_DISK_COST is set to false

* Closes ... https://kubecost.atlassian.net/browse/ENG-136

## How was this PR tested?
* Putting it in a cluster that has this local storage with sig-storage-local-provisioner and the asset disk doesnt show up anymore 

Before 

Screenshot 2024-08-22 at 10 29 31 AM

After
Screenshot 2024-08-22 at 10 30 16 AM

* None

* v2.4

### DIFF
--- a/pkg/costmodel/cluster.go
+++ b/pkg/costmodel/cluster.go
@@ -273,10 +273,6 @@ func ClusterDisks(client prometheus.Client, cp models.Provider, start, end time.
 			continue
 		}
 
-		if !env.GetAssetIncludeLocalDiskCost() && strings.HasPrefix(volumeName, SIG_STORAGE_LOCAL_PROVISIONER_PREFIX) {
-			continue
-		}
-
 		key := DiskIdentifier{cluster, volumeName}
 		if _, ok := diskMap[key]; !ok {
 			diskMap[key] = &Disk{
@@ -544,6 +540,10 @@ func ClusterDisks(client prometheus.Client, cp models.Provider, start, end time.
 		if disk.ProviderID == "" {
 			disk.ProviderID = disk.Name
 		}
+	}
+
+	if !env.GetAssetIncludeLocalDiskCost() {
+		return filterSigStorageLocalProvisonerPVs(diskMap), nil
 	}
 
 	return diskMap, nil
@@ -1434,10 +1434,6 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
 			continue
 		}
 
-		if !env.GetAssetIncludeLocalDiskCost() && strings.HasPrefix(name, SIG_STORAGE_LOCAL_PROVISIONER_PREFIX) {
-			continue
-		}
-
 		if len(result.Values) == 0 {
 			continue
 		}
@@ -1471,10 +1467,6 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
 			continue
 		}
 
-		if !env.GetAssetIncludeLocalDiskCost() && strings.HasPrefix(name, SIG_STORAGE_LOCAL_PROVISIONER_PREFIX) {
-			continue
-		}
-
 		// TODO niko/assets storage class
 
 		bytes := result.Values[0].Value
@@ -1504,10 +1496,6 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
 		name, err := result.GetString("persistentvolume")
 		if err != nil {
 			log.Warnf("ClusterDisks: PV cost data missing persistentvolume")
-			continue
-		}
-
-		if !env.GetAssetIncludeLocalDiskCost() && strings.HasPrefix(name, SIG_STORAGE_LOCAL_PROVISIONER_PREFIX) {
 			continue
 		}
 
@@ -1575,10 +1563,6 @@ func pvCosts(diskMap map[DiskIdentifier]*Disk, resolution time.Duration, resActi
 				continue
 			}
 
-			if !env.GetAssetIncludeLocalDiskCost() && strings.HasPrefix(thatVolumeName, SIG_STORAGE_LOCAL_PROVISIONER_PREFIX) {
-				continue
-			}
-
 			thatClaimName, err := thatRes.GetString("persistentvolumeclaim")
 			if err != nil {
 				log.Debugf("ClusterDisks: pv claim data missing persistentvolumeclaim")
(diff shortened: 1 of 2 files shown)


pkg/costmodel/cluster.go
```diff
@@ -44,6 +45,11 @@ const (
 
 const MAX_LOCAL_STORAGE_SIZE = 1024 * 1024 * 1024 * 1024
 
+// When ASSET_INCLUDE_LOCAL_DISK_COST is set to false, local storage
+// provisioned by sig-storage-local-static-provisioner is excluded
+// by checking if the volume is prefixed by "local-pv-"
+const SIG_STORAGE_LOCAL_PROVISIONER_PREFIX = "local-pv-"
```
thomasvn: Would it be worthwhile adding this code ref in the comments here? https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/b6f465027bd059e92c0032c81dd1e1d90e35c909/pkg/discovery/discovery.go#L410-L417

thomasvn:
It's possible we need to add your logic to skip `local-pv-*` here as well. Thoughts?

https://github.com/opencost/opencost/blob/6a1ad9a3662d9cabb9a630b97724a382909e2bd3/pkg/costmodel/cluster.go#L489-L490

pkg/costmodel/cluster.go
```diff
@@ -263,6 +269,10 @@ func ClusterDisks(client prometheus.Client, cp models.Provider, start, end time.
 			continue
 		}
 
+		if !env.GetAssetIncludeLocalDiskCost() && strings.HasPrefix(volumeName, SIG_STORAGE_LOCAL_PROVISIONER_PREFIX) {
```
Sean-Holcomb: I am confused what this has to do with this env variable. The env variable is about including node attached disk where as it appears that these disks are PVs

Sean-Holcomb (commented):
Does this check have to be in so many places, aren't many of these functions calling eachother?

pkg/costmodel/cluster.go
avrodrigues5: PVs created from sig-storage-local-provisioner are volumes/disk carved for local storage disk of the node. [Sig-storage-local-static-provisioner ](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner) gives impression of a pv attached from a local disk of a node. 

So even though they are PVs, they're cost are enclosed in the node cost itself, similar to a node-attached disk. 

@thomasvn  wanted to skip the cost for this disks leveraging the bool added in [PR](https://github.com/opencost/opencost/pull/2441), correct me if im wrong here.

avrodrigues5:
> It's possible we need to add your logic to skip `local-pv-*` here as well. Thoughts?
> 
> https://github.com/opencost/opencost/blob/6a1ad9a3662d9cabb9a630b97724a382909e2bd3/pkg/costmodel/cluster.go#L489-L490

No THomas we dont need it... THe code blob doesnt create a diskMap entry rather just updates existing diskMap with storage class. So no need!

avrodrigues5:
> Does this check have to be in so many places, aren't many of these functions calling eachother?

Sadly, @Sean-Holcomb all 6 places need it.. Wherever there is an entry to DiskMap when doesn't exist , if we miss the continue statement if will show up in asset. I'm open to opinion here. 

https://github.com/opencost/opencost/blob/86d0fa1c74e6c4cc89a8a7a04fee9e2a6214aa99/pkg/costmodel/cluster.go#L1478

For example if we miss check here, the asset disk will have local-pv entry with only size. This happens in other places too. So all 6 need a continue when there's a prefix of "local-pv-"

pkg/costmodel/cluster.go
avrodrigues5: Done!

pkg/costmodel/cluster.go
thomasvn: @avrodrigues5 Yes this is correct! They will show up as k8s PVs, but under the hood sig-storage-local-provisioner is creating those PVs using the node's local disk.

thomasvn (approved):
LGTM! I'd also say this is a low risk change because it only applies when users set `ASSET_INCLUDE_LOCAL_DISK_COST=false` (default is true).

pkg/costmodel/cluster.go
Sean-Holcomb: Okay but if they are showing up as PV's in the k8s api, then they are showing up as PV's in the prom metrics. Which means it has nothing to do with that env variable

Sean-Holcomb:
> > Does this check have to be in so many places, aren't many of these functions calling eachother?
> 
> Sadly, @Sean-Holcomb all 6 places need it.. Wherever there is an entry to DiskMap when doesn't exist , if we miss the continue statement if will show up in asset. I'm open to opinion here.
> 
> https://github.com/opencost/opencost/blob/86d0fa1c74e6c4cc89a8a7a04fee9e2a6214aa99/pkg/costmodel/cluster.go#L1478
> 
> For example if we miss check here, the asset disk will have local-pv entry with only size. This happens in other places too. So all 6 need a continue when there's a prefix of "local-pv-"

That makes sense,  to continue pushing on the idea of this change being added in one spot, could we loop through the disk map one it is complete and remove items that have that prefix?

pkg/costmodel/cluster.go
avrodrigues5: @Sean-Holcomb I'm unsure what you want me to do with respect to the environment variable? Please let me know what is expected.

Great suggestion on filtering diskmap at the end before returning. I have made changes and updated the test around it too.

Sean-Holcomb (approved):

PR #2893, opened 2024-08-28
Update path in swagger.json

wuminorb:
## What does this PR change?
Document for API (swagger.json) 

no

Correction of Information

no

## How was this PR tested?
Tested with curl

Yes, only require documentation changes

No. I am not sure which release should include this PR


AjayTripathy:
@nikovacevic aren't both endpoints valid?

nikovacevic:
@AjayTripathy looks like we don't use the `/compute` endpoint any more: https://github.com/opencost/opencost/blob/develop/pkg/cmd/costmodel/costmodel.go#L53-L54

nikovacevic (approved):

nikovacevic:
Well, wait. We do have another router here, with `/compute`: https://github.com/opencost/opencost/blob/develop/pkg/costmodel/router.go#L1738

nikovacevic:
Okay, I can confirm that both are valid.

AjayTripathy:
If both are valid should we really be removing one @nikovacevic ?

nikovacevic:
To ask "should we remove one" indicates that we document both. We don't, though -- we only document the `/compute` endpoint, so the `/allocation` API is undocumented (but identical).

That said, we reference `/allocation/compute` elsewhere (e.g. in Postman docs) so it seems like, if we're only going to document one, then the `compute` version seems to be the one we want to keep.

(Ideally, we'll probably want to clean up, dedupe, and potentially version these APIs at some point, so that it's more clear what users should be using.)

AjayTripathy:
@nikovacevic can we get this merged if you support the change? I don't think the quality gate is super relevant here.

nikovacevic:
I originally approved, then revoked after I understood what was going on. I don't know what's right here. Which API do we want Opencost users to use, `/allocation` or `/allocation/compute`? They do the same thing.

Seems like the `swagger.json` is properly documented, already, and this PR can be closed -- but technically the docs can go either way.

AjayTripathy:
@wuminorb any thoughts? Seems like both urls should work?

cliffcolvin:
> @wuminorb any thoughts? Seems like both urls should work?

@AjayTripathy we haven't heard back here, thoughts on moving this forward?

wuminorb:
hi @AjayTripathy @nikovacevic and @cliffcolvin , sorry, I forgot this PR. 

As `/allocation` is the endpoint that shown in document. And it's also more simple and straight forward. `/allocation/compute` may kind of confusion as storage/network costs also included. My suggestions is that we can use `/allocation` in document and swagger ui. And maybe we can keep the router for now, add some comments/annotation to tell people this is deprecated,  in case someone is using `/allocation/compute`. What do you think?

cliffcolvin:
I agree, working on merging. Thanks for the contribution here @wuminorb, will get this merged quickly.

PR #2902, opened 2024-09-03
Update OCI provider cost metadata

anders-swanson:
## What does this PR change?
* Adds updated oci shapes and regions

* N/A 

* Users all current oci regions and instance shapes will now get accurate pricing

* N/A

## How was this PR tested?
* Check pricing information for new shapes, region

* N/A

* N/A, not tied to a specific release or issue.


anders-swanson:
FYI @Sean-Holcomb, adding OCI metadata for new regions/instance types that have been rolled out since the last OCI provider update.  Also saw cloud costs was updated, thanks!!!

cliffcolvin:
@anders-swanson thank you very much for this contribution! Could I trouble you to help write some unit test coverage for these changes? We are working to enforce 50% + coverage on new code. Right now this PR is at 21.2% and would be great to have some level of validation on these static sources when we allow updates to them.

anders-swanson:
@cliffcolvin, for sure, I added a test to validate the static region slice using the OCI SDK 👍 this will ensure any junk values don't make it in the slice.

The JSON file is already validated as well-formed by ratecard_test.go

anders-swanson:
The coverage gate still fails even with an added test. Could you explain how the coverage is calculated? This should be 100% coverage of the newly added code.

AjayTripathy:
@cliffcolvin could you help with above question so we can get this merged?

cliffcolvin (approved):
I don't have a great answer on why the coverage percentage is what its at right now, but this is valuable to the oci integration and coverage looks good by the naked eye. Merging.

PR #2896, opened 2024-08-29
handle aws generated tags

Sean-Holcomb:
## What does this PR change?
This PR updates AWS CUR integration to ingest AWS generated tags in addition to user generated tags. To do this columns with the AWS label prefix are gathered in a separate slice. With this separate list of columns the prefix can be properly handled and saved to cloud cost labels. This PR includes this addition for both Athena and S3 integrations.

## Does this PR relate to any other PRs?
* 

## How will this PR impact users?
* Users will now be able to see AWS generated labels in Cloud Costs
Screenshot 2024-08-29 at 1 08 42 PM

## Does this PR address any GitHub or Zendesk issues?
* Closes ENG-150

## How was this PR tested?
* This PR was tested by running cloud cost and aggregator locally and ensuring that AWS label `aws_eks_cluster_name` was being populated in cloud cost.
`curl --location --globoff 'http://localhost:9004/cloudCost?window=yesterday&filter=label[aws_eks_cluster_name]%3A%22dev-1%22&aggregate=label%3Aaws_eks_cluster_name'`
```
      "cloudCosts": {
                    "dev-1": {
                        "properties": {
                            "labels": {
                                "aws_created_by": "AssumedRole:AROAUKXXWPGDYLPOEB2GE:AutoScaling",
                                "aws_eks_cluster_name": "dev-1",
                                "eks_cluster_name": "dev-1"
                            }
                        },
                        "window": {
                            "start": "2024-08-27T00:00:00Z",
                            "end": "2024-08-28T00:00:00Z"
                        },
                        "listCost": {
                            "cost": 28.5624505521,
                            "kubernetesPercent": 1
                        },
                        "netCost": {
                            "cost": 28.5624505521,
                            "kubernetesPercent": 1
                        },
                        "amortizedNetCost": {
                            "cost": 28.5624505521,
                            "kubernetesPercent": 1
                        },
                        "invoicedCost": {
                            "cost": 28.5624505521,
                            "kubernetesPercent": 1
                        },
                        "amortizedCost": {
                            "cost": 28.5624505521,
                            "kubernetesPercent": 1
                        }
                    }
                },
                "window": {
                    "start": "2024-08-27T00:00:00Z",
                    "end": "2024-08-28T00:00:00Z"
                },
                "aggregationProperties": [
                    "label:aws_eks_cluster_name"
                ]
            }
```

Because this PR deals with integrations it will need to by pass unit testing gates

## Does this PR require changes to documentation?
* 

[...]

PR #2916, opened 2024-09-10
Refactor and add testing for Athena and S3 row parsing

Sean-Holcomb:
## What does this PR change?
* This PR adds in test for the S3 and Athena integration. To do so the row parsing of the integrations is broken out into its own unit testable values. Additionally many of the strings that were previously in use I replaced with constants as well as standardizing existing constants.
* Fixes bug on S3 integration where amortized and net amortized costs were being swapped

* adds test coverage for https://github.com/opencost/opencost/pull/2896

* Closes ...

## How was this PR tested?
* Tested with unit tests

* 

*


ameijer:
@Sean-Holcomb LMK when you have test coverage and I will review!

Sean-Holcomb:
@ameijer I am not sure what is going on with the quality gate here but I covered every line of code in the new functions. The rest of the code is not unit testable, which is the point of this PR.

Sean-Holcomb:
@ameijer I have added test cases that give full coverage on the functions that I added so that this code could be tested. If you don't want to review as is I will remove the constants that I added and put back the literals that I replaced to make the quality gate pass

ameijer (approved):
LGTM

PR #2936, opened 2024-09-26
fix release script, make relative repo for better testing in fork

cliffcolvin:
Signed-off-by: Cliff Colvin <ccolvin@kubecost.com>

## What does this PR change?
* Updates release script to work properly (image name 1.111.0 instead of 1.111).
* Updates release script to pull repo action is run in (so it works on a fork).
* Updates release script to use repo owner in the image name so forks can test.

* NA

* Allows for releases to work correctly.

* Closes ...

## How was this PR tested?
* in cliffcolvin/opencost
* 
Screenshot from 2024-09-26 12-27-18

* NA

* NA

### DIFF
--- a/.github/workflows/build-and-publish-release.yml
+++ b/.github/workflows/build-and-publish-release.yml
@@ -62,7 +62,6 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v4
         with:
-          repository: 'opencost/opencost'
           ref: '${{ steps.branch.outputs.BRANCH_NAME }}'
           path: ./opencost
 
--- a/.github/workflows/build-and-publish-release.yml
+++ b/.github/workflows/build-and-publish-release.yml
@@ -57,7 +57,7 @@ jobs:
         id: branch
         run: |
           VERSION_NUMBER=${{ steps.version_number.outputs.RELEASE_VERSION }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-publish-release.yml
```diff
@@ -57,12 +57,11 @@ jobs:
         id: branch
         run: |
           VERSION_NUMBER=${{ steps.version_number.outputs.RELEASE_VERSION }}
-          echo "BRANCH_NAME=v${VERSION_NUMBER%.*}" >> $GITHUB_ENV
+          echo "BRANCH_NAME=v${VERSION_NUMBER%.*}" >> $GITHUB_OUTPUT
```
ameijer, line 60: where do you reference BRANCH_NAME?

cliffcolvin: in checkout repo right below this. 

The intent and i can add comments here is that you push tag v1.111.0 for example:

1. We make RELEASE_VERSION lose the v for reference later.
2. We do add v back and lose the last .x (patch number) and use this for BRANCH_NAME.
3. Checkout branch name so we are on the right branch when we build.
4. Image tag uses RELEASE_VERSION (dropped v, and full release number so in this case 1.111.0)

ameijer (approved):

PR #2944, opened 2024-10-06
Support openshift in cluster prometheus with kube-rbac-proxy enabled

mittal-ishaan:
## What does this PR change?
* Adds support for the ability to query Prometheus which has [kube-rbac-proxy](https://catalog.redhat.com/software/containers/rhacm2/kube-rbac-proxy-rhel8/611d4320eb34b73652492bb8) enabled. It does so by grabbing the serviceaccount token and using it as the authorization token when communicating with Prometheus.
This is particularly important in OpenShift environments which already come with a Prometheus installed.

* https://github.com/kubecost/cost-analyzer-helm-chart/pull/3700 
* https://github.com/opencost/opencost-helm-chart/pull/229
But this needs to be merged first

* Openshift users will be able use in-cluster prometheus while installing opencost without needing to have another prometheus

* Closes ...

## How was this PR tested?
* Built the opencost image with this and tried in the openshift cluster to install opencost with in-cluster prometheus. Opencost was successfully able to query it.
For openshift cluster:
If we set `ENABLE_KUBE_RBAC_PROXY` env var to true:
Kubecost ->
Screenshot from 2024-10-23 20-25-02
opencost -> 
Screenshot from 2024-10-23 20-40-56
If we set `ENABLE_KUBE_RBAC_PROXY` env var to false:
Kubecost ->
Screenshot from 2024-10-23 20-26-49
Opencost ->
Screenshot from 2024-10-23 20-43-04
For not openshift cluster(gcp):
when using external prometheus with kubecost and setting `ENABLE_KUBE_RBAC_PROXY` to false
image
deploying opencost:
image

* Yes, we need to 

update the documentation on how users can set kubeRBACProxy environment variable using helm values so as to use this.

*


pkg/prom/prom.go
```diff
@@ -374,6 +378,21 @@ type PrometheusClientConfig struct {
 
 // NewPrometheusClient creates a new rate limited client which limits by outbound concurrent requests.
 func NewPrometheusClient(address string, config *PrometheusClientConfig) (prometheus.Client, error) {
+
+	var tlsCaCert *x509.CertPool
+	// get the bearer token and ca cert from serviceaccount if we have kube rbac proxy enabled
+	if env.IsKubeRbacProxyEnabled() {
+		restConfig, err := restclient.InClusterConfig()
+		if err != nil {
+			log.Fatalf("Failed to get in-cluster config: %s", err)
```
thomasvn: Probably inappropriate to perform a `log.Fatalf` here? This would terminate the whole application. We should probably just return the error? The error message should be descriptive stating that `ENABLE_KUBE_RBAC_PROXY` was set to true but we couldn't retrieve the Bearer Token or Cert?

mittal-ishaan: my bad here, thanks for pointing this out. updating it

pkg/env/costmodelenv.go
```diff
@@ -72,6 +72,7 @@ const (
 	MultiClusterBearerToken       = "MC_BEARER_TOKEN"
 
 	InsecureSkipVerify = "INSECURE_SKIP_VERIFY"
+	EnabledKubeRbacProxyEnabled = "ENABLE_KUBE_RBAC_PROXY"
```
thomasvn:
```suggestion
	EnableKubeRbacProxy = "ENABLE_KUBE_RBAC_PROXY"
```

pkg/prom/prom.go
```diff
@@ -374,6 +378,21 @@ type PrometheusClientConfig struct {
 
 // NewPrometheusClient creates a new rate limited client which limits by outbound concurrent requests.
 func NewPrometheusClient(address string, config *PrometheusClientConfig) (prometheus.Client, error) {
+
+	var tlsCaCert *x509.CertPool
+	// get the bearer token and ca cert from serviceaccount if we have kube rbac proxy enabled
+	if env.IsKubeRbacProxyEnabled() {
+		restConfig, err := restclient.InClusterConfig()
+		if err != nil {
+            log.Errorf("ENABLE_KUBE_RBAC_PROXY was set to true but failed to get in-cluster config: %s", err)
```
thomasvn: Potential indentation helpful here?

pkg/prom/prom.go
```diff
@@ -384,6 +403,7 @@ func NewPrometheusClient(address string, config *PrometheusClientConfig) (promet
 		TLSHandshakeTimeout: config.TLSHandshakeTimeout,
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: config.TLSInsecureSkipVerify,
+			RootCAs:            tlsCaCert,
```
thomasvn, line 407: Could you show some test results where you've run OpenCost with `ENABLE_KUBE_RBAC_PROXY=true` and `ENABLE_KUBE_RBAC_PROXY=false`. Just want to make sure a `nil` tlsCaCert is ok here.

mittal-ishaan: sure let me update the pr description for it

pkg/prom/prom.go
```diff
@@ -374,6 +378,21 @@ type PrometheusClientConfig struct {
...
+	// get the bearer token and ca cert from serviceaccount if we have kube rbac proxy enabled
+	if env.IsKubeRbacProxyEnabled() {
+		restConfig, err := restclient.InClusterConfig()
+		if err != nil {
+            log.Errorf("ENABLE_KUBE_RBAC_PROXY was set to true but failed to get in-cluster config: %s", err)
+		}
+		config.Auth.BearerToken = restConfig.BearerToken
+		tlsCaCert, err = certutil.NewPool(`/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`)
+		if err != nil {
+            log.Errorf("ENABLE_KUBE_RBAC_PROXY was set to true but failed to load service-ca.crt: %s", err)
```
thomasvn: Open question. If we are unable to get the cluster's CA cert but the user specifies `ENABLE_KUBE_RBAC_PROXY=true`, should we gracefully exit the program here? In other words, should we be returning the error here as opposed to just logging it and moving on?

mittal-ishaan: Exiting the program due to this will make it unusable if the user is not able to get the cluster CA cert for some reason, while there is a way that the user will still be able to successfully deploy kubecost or opencost if this happens. It can be possible that the user sets `INSECURE_SKIP_VERIFY` env var to be true, this will make the ca cert value not needed.

thomasvn (approved):

cliffcolvin:
@mittal-ishaan thank you for this very well documented PR.

cliffcolvin (approved):
LGTM, ThomasN has also approved here, testing notes look great. Will merge once tests and build passes.

Issue #1110, opened 2022-03-11
Job costs attributed to the wrong namespace


mac-chaffee:
I have a strange bug where an old job (which doesn't use that many resources) in one namespace is showing up in kubecost with really high costs and in the wrong namespace.

Here is the job appearing in kubecost associated with the "monitoring" namespace (that's where kubecost is running, for the record)
<img width="1516" alt="image" src="https://user-images.githubusercontent.com/7581860/157976521-bcc1f546-62ab-4af7-9b73-f232c1270709.png">

But that job is actually in the "translator-dev" namespace:

```
$ kubectl get job -A | grep nn-load-node-normalization-loader-con-0-job
translator-dev   nn-load-node-normalization-loader-con-0-job     1/1           15m        15d
```

Here is the job:

```
$ kubectl get job -n translator-dev nn-load-node-normalization-loader-con-0-job -o yaml
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    meta.helm.sh/release-name: nn-load
    meta.helm.sh/release-namespace: translator-dev
  creationTimestamp: "2022-02-24T13:20:56Z"
  labels:
    app.kubernetes.io/managed-by: Helm
  name: nn-load-node-normalization-loader-con-0-job
  namespace: translator-dev
  resourceVersion: "76419107"
  uid: 05ba3fe3-f34f-472d-8a8c-607845c26ac7
spec:
  backoffLimit: 6
  completions: 1
  parallelism: 1
  selector:
    matchLabels:
      controller-uid: 05ba3fe3-f34f-472d-8a8c-607845c26ac7
  template:
    metadata:
      creationTimestamp: null
      labels:
        controller-uid: 05ba3fe3-f34f-472d-8a8c-607845c26ac7
        job-name: nn-load-node-normalization-loader-con-0-job
      name: nn-load-node-normalization-loader-con-0-job
    spec:
      containers:
      - args:
        - GeneProtein.txt
        command:
        - /code/run_loader.sh
        image: renciorg/r3_nodenorm:2.0.5
        imagePullPolicy: IfNotPresent
        name: loader-container
        resources:
          limits:
            cpu: "2"
            memory: 10Gi
          requests:
            cpu: 250m
            memory: 1Gi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /code/run_loader.sh
          name: loader-config
          subPath: run_loader_conflation.sh
        - mountPath: /data
[...]

PR #2961, opened 2024-10-18
Configurable Labels in Allocation API for restricted Prometheus environments 

jlopes-randoli:
## What does this PR change?
* Changes allocation API queries to allow for custom labels for namespace, container and pod in environments that modify the default labels for metrics and are restrictive about changing the default configuration.

* No

* Allows for additional configuration if required, but otherwise does not affect default settings

* No

## How was this PR tested?
* Integration testing on environment with modified and restricted metric labels

* Most likely, to address the new functionality for anyone dealing with this particular use case.

*


AjayTripathy:
Hi @jlopes-randoli thank you for this PR. These look like some changes in the core library which can sometimes be hard to test. Can you share more on your testing methodology and how we can reproduce this environment?

jlopes-randoli:
@AjayTripathy 

This was mainly tested in an Openshift environment using their default managed Prometheus instance with user workload monitoring enabled. Similar to the issue described [here](https://www.opencost.io/docs/configuration/gcp#gcp-managed-prometheus) on GCP Managed Prometheus, I encountered the same issue in Openshift managed Prometheus instance which also modifies the metric labels generated and used by Opencost to have an exported_ prefix. (e.g exported_namespace, exported_pod, exported_container). This caused almost none of the expected controllers or costs to properly appear in the allocation cost report. My testing methodology involved testing the allocation API endpoints in my environment and confirming that the previous functionality of the allocation API had been restored, and that no previous functionality had been impacted by the changes. The environment can be recreated on an Openshift cluster with a minimum version of 4.14 (the version these changes were tested on) that has user workload monitoring enabled. ([docs](https://docs.openshift.com/container-platform/4.17/observability/monitoring/enabling-monitoring-for-user-defined-projects.html#enabling-monitoring-for-user-defined-projects%29))

AjayTripathy:
Thank you. We are a little backed up on these reviews until post kubecon, but @cliffcolvin let's get assigned to take a look then. I actually know @thomasvn worked on something similar before.

thomasvn:
@jlopes-randoli I’ve encountered a similar `exported_*` label prefix issue before too.

The best solution, in my experience, isn’t to make these labels configurable in OpenCost queries but to set [`honor_labels=true`](https://github.com/opencost/opencost/blob/30e5d9a27f41d00c5ced614006e0ac53d20ff657/kubernetes/prometheus/extraScrapeConfigs.yaml#L3) in Prometheus’s scrape config ([docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)).

This is because when `honor_labels=false`, labels are only prefixed with `exported_` if there’s a clash between server-side label operations and the scraped metric labels. For example, `kube_pod_container_resource_requests` might have `exported_namespace`, while `container_memory_working_set_bytes` keeps the original `namespace`. The behavior depends on the Prometheus server configuration.

jlopes-randoli:
@thomasvn Hey Thomas! Thanks for reviewing the PR.

Part of the issue that we encountered with that workaround for our particular use case was because of the restrictions on OpenShift managed Prometheus and the usage of a ServiceMonitor for scraping.

This is from their documentation regarding the restrictions in place:
	
_When Prometheus scrapes user workload targets that expose a namespace label, the system stores this label as exported_namespace. This behavior ensures that the final namespace label value is equal to the namespace of the target pod. You cannot override this default configuration by setting the value of the honorLabels field to true for PodMonitor or ServiceMonitor objects._

I've tried other workarounds including looking into a flag that is supposed to override this default behaviour, but was told by Red Hat support that this is also not a viable option to fix this issue. Because of these restrictions and the way the operator overrides these changes to the configuration is the reason that we've gone with this approach.

thomasvn:
@jlopes-randoli You mention that you're unable to modify the PodMonitor and ServiceMonitor. I'm curious what parts of the managed Prometheus you are able to modify? For example are you able to edit any relabeling rules?

Here are previous examples where editing Prometheus's config helped to solve the `exported_*` issue:

- https://github.com/opencost/opencost/issues/1110
- https://www.opencost.io/docs/configuration/gcp#gcp-managed-prometheus

thomasvn:
@jlopes-randoli As I've been creating a reproducible environment, a few questions came up in the process. How does your OpenShift environment scrape OpenCost?

- Does your OpenShift environment define it via an `extraScrapeConfig` as documented [here](https://raw.githubusercontent.com/opencost/opencost/develop/kubernetes/prometheus/extraScrapeConfigs.yaml) and [here](https://opencost.io/docs/installation/prometheus/)? I _believe_ doing so would be another workaround to the problem.
- Or is it scraping via the [`kubernetes-service-endpoints`](https://github.com/prometheus-community/helm-charts/blob/aff760443a80091775c422184160f09acaa0b641/charts/prometheus/values.yaml#L935-L950) scrapeconfig?
- Additionally, does your OpenShift environment run a kube-state-metrics deployment? I'm asking because this emits metrics which may duplicate the metrics OpenCost emits.

jlopes-randoli:
@thomasvn

For your first question, yes we are currently using the extraScrapeConfigs in our environment, not the service endpoints.

Also, for the second it is running a kube-state-metrics deployment. Would this cause any issues?

thomasvn:
@jlopes-randoli Having kube-state-metrics running _may_ result in some issues. For example take a look at the below screenshot.

Here I have both opencost and kube-state-metrics being scraped by Prometheus. We can see that the metric `kube_pod_container_resource_requests` is being emitted by both opencost and kube-state-metrics, and is therefore duplicated. Also notice that some labels are being prefixed with `exported_` due to the kubernetes-service-endpoints config ([ref](https://github.com/prometheus-community/helm-charts/blob/fe79476450db94f79d65c4863d63484c26a12037/charts/prometheus/values.yaml#L935-L950)).

DuplicateMetric

@jlopes-randoli Could you port-forward into Prometheus and let me know if you're seeing the same behavior in your environment?

jlopes-randoli:
@thomasvn hey Thomas, just wanted to check in if you're still taking a look at this and if you have anymore questions or suggestions

Also, I did the test you asked for and didn't find any duplicate metrics from kube-state-metrics, just the ones generated from the cost job.

thomasvn:
Hey @jlopes-randoli, thanks for following up! I’m just trying to get a clearer picture of what’s happening in your environment to see if there’s a better solution.

Can we double-check that you’re not seeing duplicate metrics between the `opencost` and `kubernetes-service-endpoints` scrape configs? As far as I can tell, the `opencost` config (below) uses `honor_labels: true` and has no `relabel_configs`, so it shouldn’t produce any `exported_*` labels.

```yaml
extraScrapeConfigs: |
  - job_name: opencost
    honor_labels: true
    scrape_interval: 1m
    scrape_timeout: 10s
    metrics_path: /metrics
    scheme: http
    dns_sd_configs:
    - names:
      - opencost.opencost
      type: 'A'
      port: 9003
```

Could you share screenshots of the following Prometheus queries to help check for duplicates and pinpoint where `exported_*` might be coming from?

```
kube_pod_container_resource_requests{container="opencost", resource="cpu"}
kube_pod_container_status_running{container="opencost"}
kube_pod_labels{pod="opencost-6bdc8b658f-rpssw"}
```

jlopes-randoli:
Hey @thomasvn, here are the screenshots you asked for.

thomasvn:
Thanks @jlopes-randoli! I'm seeing a lot of `exported_*` labels here. And these are metrics coming from the Prometheus ScrapeConfig `job=insights-opencost`.

Can you share more about how that ScrapeConfig is configured? Would you be able to share its YAML?

jlopes-randoli:
Hey @thomasvn , here's the current situation. Because of the amount of time that has passed, I no longer have access to the original environment I was testing this with and will have to recreate this issue in another environment. 

Also, because of the recent refactor that was merged I will have to significantly modify my PR to work with the changes made.

What I can do before I try to make changes to the PR, I will recreate this environment and attempt again to use the workarounds you suggested. It's been a while since I was actively working on this issue (internally we used our own fork of Opencost with these changes to resolve the issue and have now moved on to bundling our own Prometheus instance with our agent), so I will need to confirm again the reason that the extraScrapeConfigs solution was not sufficient. 

Did you ever end up getting your own environment up and running to recreate the issue? We were using OpenShift with user workload monitoring (https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/user-workload-monitoring.md) which is the default approach recommended for getting Prometheus metrics in an OpenShift cluster. In these docs they mention a bit about their decision to restrict all metrics to use "exported_" metrics:

"honor_labels controls how prometheus handles conflicts between labels already present in scraped data and the labels that prometheus would attach server side. If honor_labels is set to true labels from the scraped data are kept and server-side labels are ignored. If set to false the conflicts are resolved renaming the scraped data to exported_<label-name>. This is exactly what we want for user workloads, because we want to avoid the users possibly remapping the values of labels as we base our tenancy model on this. We have two options here that can be done in prometheus-operator, either discard that object completely in the same way as we do with filesystem access right now or create the respective objects and change the honorLabel value to false."

So while I'm 100% sure that OpenCost will never work using the ServiceMonitor method, I'm open to investigating again if it's possible to use additionalScrapeConfigs to work-around this.  I remember attempting to and it did not resolve the issue but it's possible I need to look more deeply into how to do this properly with the restrictions the OpenShift operator has on modifying it.

I think regardless of whether this PR ends up merging or not, if we are able to find an alternative work-around, there definitely needs to be a documentation update because I think if OpenCost has so many difficulties running out of the box with the monitoring set-up that OpenShift pushes as the default, this should definitely be documented for future users that will encounter this issue.

thomasvn:
Hi @jlopes-randoli. Thanks for the update, and all of that context. I was not able to reproduce the issue using the standard installation path on AWS (https://opencost.io/docs/configuration/aws), since this recommends using `extraScrapeConfigs` with `honor_labels=true`. I have not yet tested in an OpenShift environment, with OpenShift's built in monitoring namespace and ServiceMonitors.

When possible, I'll try to set up an OpenShift environment and reproduce there. I agree that even if a PR does not get merged, we should document the appropriate configurations/workarounds required here.

jlopes-randoli:
Hello @thomasvn I had the chance to do a deeper investigation and attempt the additionalScrapeConfigs solution again and have a better idea of what problems are preventing that solution.

The Prometheus instance that OpenShift automatically deploys is handled by the Prometheus operator. So when I initially attempted to resolve this, I tried the method as described [here](https://github.com/prometheus-operator/prometheus-operator/tree/main/example/additional-scrape-configs) to reference a secret with the additionalScrapeConfigs and modify the Prometheus CRD to use this configuration. This method didn't work, so I attempted to use a [ScrapeConfig](https://prometheus-operator.dev/docs/developer/scrapeconfig/) to accomplish this. This method also didn't work and this is when I realized that the CRD changes I was adding were getting overwritten and removed.

The reason for this is OpenShift does not deploy the Prometheus Operator directly but they have another operator they've developed for deploying and managing the entire monitoring stack: [Cluster Monitoring Operator](https://github.com/openshift/cluster-monitoring-operator)

The problem then is that I cannot edit the Prometheus CRD and configuration directly. I can only configure the configmap which the Cluster Monitoring Operator provides as detailed [here](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/monitoring/config-map-reference-for-the-cluster-monitoring-operator#prometheusk8sconfig). The configuration they provide for modifying the Prometheus instance is limited and doesn't allow for configuring additional scrape configs.

peatey:
hey all this has been kicking around for e minute do we mind closing it out?

PR #125, opened 2019-07-12
Update RAM/CPU allocation description

dwbrown2:

PR #2963, opened 2024-10-18
ENG-2914: Introduce GPUUsageMax, isShared and GPU device information

kubecost-pb:
## What does this PR change?
Introduces GPUUsageMax, isGPUShared and GPU Device information

https://github.com/kubecost/kubecost-cost-model/pull/2897
https://github.com/kubecost/kubecost-integration-tests/pull/127

* 

* Closes ENG-2914, ENG-2955

## How was this PR tested?
* Tested locally on dev machine

* 

*


core/pkg/opencost/allocation.go
```diff
@@ -100,9 +100,64 @@ type Allocation struct {
 	// UnmountedPVCost is used to track how much of the cost in PVs is for an
 	// unmounted PV. It is not additive of PVCost() and need not be sent in API
 	// responses.
-	UnmountedPVCost   float64 `json:"-"`                 //@bingen:field[ignore]
-	GPURequestAverage float64 `json:"gpuRequestAverage"` //@bingen:field[version=22]
-	GPUUsageAverage   float64 `json:"gpuUsageAverage"`   //@bingen:field[version=22]
+	UnmountedPVCost             float64        `json:"-"`                 //@bingen:field[ignore]
+	DeprecatedGPURequestAverage float64        `json:"gpuRequestAverage"` //@bingen:field[version=22]
+	DeprecatedGPUUsageAverage   float64        `json:"gpuUsageAverage"`   //@bingen:field[version=22]
```
mbolt35: I would change these fields to the following:

```go
deprecatedGPURequestAverage float64        `json:"-"` //@bingen:field[version=22]
deprecatedGPUUsageAverage   float64        `json:"-"`   //@bingen:field[version=22]
```

We don't want these fields to be exposed at all outside of the package (since they're now deprecated). Also, if at all possible, we don't want these fields to appear in the JSON encoding either.

core/pkg/opencost/allocation.go
```diff
@@ -174,8 +229,9 @@ func (lba *LbAllocation) SanitizeNaN() {
 // then this type would be unnecessary and its fields would go into the regular Allocation
 // and not in the AggregatedAllocation.
 type RawAllocationOnlyData struct {
-	CPUCoreUsageMax  float64 `json:"cpuCoreUsageMax"`
-	RAMBytesUsageMax float64 `json:"ramByteUsageMax"`
+	CPUCoreUsageMax  float64  `json:"cpuCoreUsageMax"`
+	RAMBytesUsageMax float64  `json:"ramByteUsageMax"`
+	GPUUsageMax      *float64 `json:"gpuUsageMax"` //@bingen:field[version=23]
```
mbolt35, line 234: I think this is fine, but could create some annoying patterns to double-nil check for a valid `GPUUsageMax` -- Just noting that if we find ourselves writing:

```go
if ra != nil && ra.GPUUsageMax != nil {
    //  ...
}
```
we may want to consider a helper func. Don't see any evidence of that yet, so not sure it's necessary now.

core/pkg/opencost/allocation.go
```diff
@@ -963,17 +1037,17 @@ func (a *Allocation) RAMEfficiency() float64 {
...
 	}
 
-	if a.GPURequestAverage > 0 && a.GPUUsageAverage > 0 {
-		return a.GPUUsageAverage / a.GPURequestAverage
+	if a.DeprecatedGPURequestAverage > 0 && a.DeprecatedGPUUsageAverage > 0 {
+		return a.DeprecatedGPUUsageAverage / a.DeprecatedGPURequestAverage
 	}
 
-	if a.GPUUsageAverage == 0.0 || a.GPUTotalCost() == 0.0 {
+	if a.DeprecatedGPUUsageAverage == 0.0 || a.GPUTotalCost() == 0.0 {
```
mbolt35: Do we need to keep any of this logic related to the old fields? I think it would make the most sense to not have any use of the deprecated fields, and only use the new fields. If this has been done on another PR, then ignore this feedback :)

core/pkg/opencost/allocation_json.go
```diff
@@ -84,8 +85,8 @@ func (aj *AllocationJSON) BuildFromAllocation(a *Allocation) {
 	aj.CPUCostIdle = formatFloat64ForResponse(a.CPUCostIdle)
 	aj.CPUEfficiency = formatFloat64ForResponse(a.CPUEfficiency())
 	aj.GPUCount = formatFloat64ForResponse(a.GPUs())
-	aj.GPURequestAverage = formatFloat64ForResponse(a.GPURequestAverage)
-	aj.GPUUsageAverage = formatFloat64ForResponse(a.GPUUsageAverage)
+	aj.GPURequestAverage = formatFloat64ForResponse(a.DeprecatedGPURequestAverage)
+	aj.GPUUsageAverage = formatFloat64ForResponse(a.DeprecatedGPUUsageAverage)
```
mbolt35: Feels reasonable to drop these from the JSON output.

mbolt35 (changes requested):
It does seem like we should do a bit more cleanup of the old field use. With this "swap" change + migration, I think we are safe to remove any logic related to these fields, including the removal from the JSON encoding as well. In terms of the logic, everything looks great!

I also realize there are a few related PRs, so I may be missing some context - hit me up on slack if you want to go over some of this.

core/pkg/opencost/allocation.go
```diff
@@ -198,8 +255,16 @@ func (r *RawAllocationOnlyData) Equal(that *RawAllocationOnlyData) bool {
...
 	}
-	return util.IsApproximately(r.CPUCoreUsageMax, that.CPUCoreUsageMax) &&
+	cmpResult := util.IsApproximately(r.CPUCoreUsageMax, that.CPUCoreUsageMax) &&
 		util.IsApproximately(r.RAMBytesUsageMax, that.RAMBytesUsageMax)
+
+	if r.GPUUsageMax != nil && that.GPUUsageMax != nil {
+		cmpResult = cmpResult && util.IsApproximately(*r.GPUUsageMax, *that.GPUUsageMax)
+	} else {
+		cmpResult = false
+	}
```
mbolt35: So, I think you would still want equality to be true if both GPUUsageMax fields are nil? ie: If both fields are nil, we can just leave `cmpResult` alone.

```
if r.GPUUsageMax != nil && that.GPUUsageMax != nil {
    cmpResult = cmpResult && util.IsApproximately(*r.GPUUsageMax, *that.GPUUsageMax)
} else if !(r.GPUUsageMax == nil && that.GPUUsageMax == nil) {
    cmpResult = false
}
```

mbolt35: Great unit test case :)

core/pkg/opencost/allocation.go
```diff
@@ -2587,13 +2676,13 @@ func (a *Allocation) SanitizeNaN() {
...
+	if math.IsNaN(a.DeprecatedGPURequestAverage) {
+		log.DedupedWarningf(5, "Allocation: Unexpected NaN found for DeprecatedGPURequestAverage name:%s, window:%s, properties:%s", a.Name, a.Window.String(), a.Properties.String())
+		a.DeprecatedGPURequestAverage = 0
 	}
-	if math.IsNaN(a.GPUUsageAverage) {
-		log.DedupedWarningf(5, "Allocation: Unexpected NaN found for GPUUsageAverage name:%s, window:%s, properties:%s", a.Name, a.Window.String(), a.Properties.String())
-		a.GPUUsageAverage = 0
+	if math.IsNaN(a.DeprecatedGPUUsageAverage) {
+		log.DedupedWarningf(5, "Allocation: Unexpected NaN found for DeprecatedGPUUsageAverage name:%s, window:%s, properties:%s", a.Name, a.Window.String(), a.Properties.String())
+		a.DeprecatedGPUUsageAverage = 0
```
mbolt35: Don't think we care anymore about the deprecated fields. Safe to remove.

core/pkg/opencost/allocation.go
```diff
@@ -3570,3 +3660,14 @@ func (asr *AllocationSetRange) Clone() *AllocationSetRange {
...
+func migrateAllocation(as *Allocation, fromVersion uint8, toVersion uint8) {
+	if fromVersion == toVersion {
+		return
+	}
+
+	if fromVersion == 22 && toVersion >= 23 {
+		as.GPUAllocation.GPUUsageAverage = &as.DeprecatedGPUUsageAverage
+		as.GPUAllocation.GPURequestAverage = &as.DeprecatedGPURequestAverage
+	}
+}
```
mbolt35: 🙌 amazing

core/pkg/opencost/allocation.go
mbolt35: You can just rename these and re-run bingen.

kubecost-pb:
> It does seem like we should do a bit more cleanup of the old field use. With this "swap" change + migration, I think we are safe to remove any logic related to these fields, including the removal from the JSON encoding as well. In terms of the logic, everything looks great!
> 
> I also realize there are a few related PRs, so I may be missing some context - hit me up on slack if you want to go over some of this.

@mbolt35 Fixed. Please re-review and let me know if I missed any

core/pkg/opencost/allocation.go
kubecost-pb: Fixed

core/pkg/opencost/allocation_json.go
kubecost-pb: Fixed. I actually changed these to use the new fields. Let me know what do you think..

core/pkg/opencost/allocation.go
kubecost-pb: Fixed. Used new fields.

core/pkg/opencost/allocation.go
kubecost-pb: Done. Re-running bingen didnt do anything in `bingen_codecs.go`, is that expected?

core/pkg/opencost/allocation.go
```diff
@@ -3570,3 +3703,14 @@ func (asr *AllocationSetRange) Clone() *AllocationSetRange {
...
+
+func migrateAllocation(as *Allocation, fromVersion uint8, toVersion uint8) {
+	if fromVersion == toVersion {
+		return
+	}
+
+	if fromVersion == 22 && toVersion >= 23 {
+		as.GPUAllocation.GPUUsageAverage = &as.deprecatedGPUUsageAverage
+		as.GPUAllocation.GPURequestAverage = &as.deprecatedGPURequestAverage
+	}
```
mbolt35: Awesome!

core/pkg/opencost/allocation_json.go
```diff
@@ -84,8 +85,8 @@ func (aj *AllocationJSON) BuildFromAllocation(a *Allocation) {
 	aj.CPUCostIdle = formatFloat64ForResponse(a.CPUCostIdle)
 	aj.CPUEfficiency = formatFloat64ForResponse(a.CPUEfficiency())
 	aj.GPUCount = formatFloat64ForResponse(a.GPUs())
-	aj.GPURequestAverage = formatFloat64ForResponse(a.GPURequestAverage)
-	aj.GPUUsageAverage = formatFloat64ForResponse(a.GPUUsageAverage)
+	aj.GPURequestAverage = formatFloat64ForResponse(a.deprecatedGPURequestAverage)
+	aj.GPUUsageAverage = formatFloat64ForResponse(a.deprecatedGPUUsageAverage)
```
mbolt35: Not sure about these fields? Should we get the values here from the new data, or rather, we should remove/omit these fields from this `AllocationJSON` object and add a new structure representing the new data field?

kubecost-pb: We already added a new structure representing new data in Ln #125.

This fields, I can just delete since we already have "-" in the json for ln 29-30.. 

Thoughts @mbolt35

mbolt35 (approved):
Looks great!

pkg/costmodel/allocation.go
```diff
@@ -66,6 +67,8 @@ const (
 	queryFmtLBActiveMins                = `count(kubecost_load_balancer_cost{%s}) by (namespace, service_name, %s)[%s:%s]`
 	queryFmtOldestSample                = `min_over_time(timestamp(group(node_cpu_hourly_cost{%s}))[%s:%s])`
 	queryFmtNewestSample                = `max_over_time(timestamp(group(node_cpu_hourly_cost{%s}))[%s:%s])`
+	queryFmtIsGPuShared                 = `avg(avg_over_time(kube_pod_container_resource_requests{container!="", node != "", pod != "", container!= "", unit = "integer",  %s}[%s])) by (container, pod, namespace, node, resource)`
+	queryFmtGetGPuInfo                  = `avg(avg_over_time(DCGM_FI_DEV_DEC_UTIL{container!="",%s}[%s])) by (container, pod, namespace, device, modelName, UUID)`
```
thomasvn, line 71:
```suggestion
	queryFmtIsGPUShared                 = `avg(avg_over_time(kube_pod_container_resource_requests{container!="", node != "", pod != "", container!= "", unit = "integer",  %s}[%s])) by (container, pod, namespace, node, resource)`
	queryFmtGetGPUInfo                  = `avg(avg_over_time(DCGM_FI_DEV_DEC_UTIL{container!="",%s}[%s])) by (container, pod, namespace, device, modelName, UUID)`
```

Very minor. If you accept this change, you'll also have to update the references to this variable.

pkg/costmodel/allocation_helpers.go
```diff
@@ -614,12 +621,13 @@ func applyRAMBytesUsedMax(podMap map[podKey]*pod, resRAMBytesUsedMax []*prom.Que
 	}
 }
 
-func applyGPUUsageAvg(podMap map[podKey]*pod, resGPUUsageAvg []*prom.QueryResult, podUIDKeyMap map[podKey][]podKey) {
+// same func is used for both GPUUsageAvg and GPUUsageMax
+func applyGPUUsage(podMap map[podKey]*pod, resGPUUsageAvgOrMax []*prom.QueryResult, podUIDKeyMap map[podKey][]podKey, mode string) {
```
thomasvn, line 625: Could this instead be broken up into 4 functions, for better clarity? It currently handles results from Prometheus queries `resGPUsUsageAvg`, `resGPUsUsageMax`, `resIsGpuShared`, `resGetGPUInfo`.

pkg/costmodel/allocation_helpers.go
```diff
@@ -642,15 +650,66 @@ func applyGPUUsageAvg(podMap map[podKey]*pod, resGPUUsageAvg []*prom.QueryResult
...
+				} else {
+					thisPod.Allocations[container].GPUAllocation.GPUUsageAverage = &res.Values[0].Value
+				}
+			case GpuUsageMaxMode:
+				thisPod.Allocations[container].RawAllocationOnly.GPUUsageMax = &res.Values[0].Value
+			case GpuIsSharedMode:
+				// if a container is using a GPU and it is shared, isGPUShared will be true
+				// if a container is using GPU and it is NOT shared, isGPUShared will be false
+				// if a container is NOT using a GPU, isGPUShared will be null
+				if res.Metric["resource"] == "nvidia_com_gpu_shared" {
```
thomasvn, line 675: @chipzoller Just want to double check with you. Is this the primary way or only way you can request a shared (specifically time spliced) GPU?

```yaml
resources:
  limits:
    nvidia.com/gpu.shared: "1" 
  requests:
    nvidia.com/gpu.shared: "1" 
```

chipzoller: This isn't technically the only way a shared GPU can be requested, but it is the only way to disambiguate a shared versus non-shared GPU request when looking at the Pod spec itself. Yes, it represents an opinionated approach we want to take (to start), but it's predictable and simpler. The sharing method cannot be specified via the static device plugin today. This is a gap that will be addressed by Dynamic Resource Allocation (DRA) in future versions of Kubernetes.

thomasvn: Ok great! Thanks for clarifying.

PR #2978, opened 2024-11-06
Fix: fetching for azure nodes that are not spot instances  

barmettlerl:
## What does this PR change?
* It enables to fetch non spot node instance costs from azure
* It refactors the `NodePricing` function of the azure provider, make it easier to read.

* No

* It will enable to fetch non spot node instance from azure. So less errors containing `failed to retrieve spot retail pricing` 

* No, I thought instead of opening a issue i can fix it directly ;) 

## How was this PR tested?
* I let it run on my cluster, the produced metrics corresponded to the metrics read by opencost. An example, where the fetch before didn't work was for me [this resource](https://prices.azure.com/api/retail/prices?$skip=0&currencyCode=%27USD%27&$filter=armRegionName+eq+%27switzerlandnorth%27+and+armSkuName+eq+%27Standard_D8d_v5%27). 
* I let the existing test run -> and they worked well. I think they worked well, because they didn't covered this usecase. 

* It does not.

* Should be included into the next release

### DIFF
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -291,7 +291,7 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 	spotPrice := ""
 	for _, item := range pricingPayload.Items {
 		if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
-			if  !strings.Contains(strings.ToLower(item.SkuName), " spot") {
+			if !strings.Contains(strings.ToLower(item.SkuName), " spot") {
 				spotPrice = fmt.Sprintf("%f", item.RetailPrice)
 			} else {
 				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
@@ -300,6 +300,10 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 	}
 
 	log.DedupedInfof(5, "done parsing retail price payload from \"%s\"\n", pricingURL)
+
+	if spot && spotPrice != "" {
+		return spotPrice, nil
+	}
 
 	if retailPrice == "" {
 		return retailPrice, fmt.Errorf("Couldn't find price for product \"%s\" in \"%s\" region", skuName, region)
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -290,8 +290,9 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 	retailPrice := ""
 	for _, item := range pricingPayload.Items {
 		if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
-			// if spot is true SkuName should contain "spot, if it is false it should not
 			if spot == strings.Contains(strings.ToLower(item.SkuName), " spot") {
+				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
+			} else {
 				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
 			}
 		}
@@ -936,11 +937,6 @@ func convertMeterToPricings(info commerce.MeterInfo, regions map[string]string, 
 		return nil, nil
 	}
 
-	if strings.Contains(meterSubCategory, "Cloud Services") || strings.Contains(meterSubCategory, "CloudServices") {
-		// This meter doesn't correspond to any pricings.
-		return nil, nil
-	}
-
 	if strings.Contains(meterCategory, "Storage") {
 		if strings.Contains(meterSubCategory, "HDD") || strings.Contains(meterSubCategory, "SSD") || strings.Contains(meterSubCategory, "Premium Files") {
 			var storageClass string = ""
@@ -1092,11 +1088,6 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
 func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetadata, error) {
 	az.DownloadPricingDataLock.RLock()
 	defer az.DownloadPricingDataLock.RUnlock()
-	pricingDataExists := true
-	if az.Pricing == nil {
-		pricingDataExists = false
-		log.DedupedWarningf(1, "Unable to download Azure pricing data")
-	}
 
 	meta := models.PricingMetadata{}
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/azure/provider.go
```diff
@@ -290,9 +290,10 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 	retailPrice := ""
 	for _, item := range pricingPayload.Items {
 		if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
-			// if spot is true SkuName should contain "spot, if it is false it should not
 			if spot == strings.Contains(strings.ToLower(item.SkuName), " spot") {
 				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
+			} else {
+				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
```
Sean-Holcomb, line 297: this seems to negate the need for this conditional

pkg/cloud/azure/provider.go
```diff
@@ -936,11 +937,6 @@ func convertMeterToPricings(info commerce.MeterInfo, regions map[string]string,
 		return nil, nil
 	}
 
-	if strings.Contains(meterSubCategory, "Cloud Services") || strings.Contains(meterSubCategory, "CloudServices") {
-		// This meter doesn't correspond to any pricings.
-		return nil, nil
-	}
-
```
Sean-Holcomb: Is there a reason that this was removed?

pkg/cloud/azure/provider.go
```diff
@@ -1092,11 +1088,6 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
 func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetadata, error) {
 	az.DownloadPricingDataLock.RLock()
 	defer az.DownloadPricingDataLock.RUnlock()
-	pricingDataExists := true
-	if az.Pricing == nil {
-		pricingDataExists = false
-		log.DedupedWarningf(1, "Unable to download Azure pricing data")
-	}
```
Sean-Holcomb, line 1099: from what I can tell the removal of this nil check is fine with the locks in place

pkg/cloud/azure/provider.go
Sean-Holcomb: The main issue with this change is that regardless of whether the node in question is spot or ondemand the pricing that comes out have this is whatever came last which could be spot or ondemand

pkg/cloud/azure/provider.go
```diff
@@ -1106,54 +1097,56 @@ func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetada
...
-		if n, ok := az.Pricing[spotFeatures]; ok {
-			log.DedupedInfof(5, "Returning pricing for node %s: %+v from key %s", azKey, n, spotFeatures)
-			if azKey.isValidGPUNode() {
-				n.Node.GPU = "1" // TODO: support multiple GPUs
-			}
-			return n.Node, meta, nil
+		featureString = fmt.Sprintf("%s,spot", featureString)
+	}
+
+	if n, ok := az.Pricing[featureString]; ok {
```
Sean-Holcomb, line 1127: The main issue here is that you have broken non-spot feature strings here which previously would have been "region,instance,ondemand" now they are just "region,instance"

pkg/cloud/azure/provider.go
```diff
@@ -1106,54 +1097,56 @@ func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetada
...
-			}
-			spotNode := &models.Node{
-				Cost:      spotCost,
+		return n.Node, meta, nil
+	}
+
+	cost, err := getRetailPrice(region, instance, config.CurrencyCode, isSpot)
+
+	if err != nil {
+		log.DedupedWarningf(5, "failed to retrieve spot retail pricing: %s", err)
```
Sean-Holcomb: this warning is now scoped for non-spot nodes too

pkg/cloud/azure/provider.go
```diff
@@ -1106,54 +1097,56 @@ func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetada
...
+	} else {
+		gpu := ""
+		if azKey.isValidGPUNode() {
+			gpu = "1"
```
Sean-Holcomb: I think we would want to use azKey.GetGPUCount()

pkg/cloud/azure/provider.go
Sean-Holcomb: can you re do this so that in the case of a spot instance it looks for a spot sku first and if there is not one it returns ondemand and for ondemand it should never be able to return a spot sku

Sean-Holcomb (changes requested):
It seems that you are attempting to make pricing for spot nodes fall back to the pricing for equivalent ondemand node types more effectively, however I am a little concerned with the way the algorithm is being changed here.

 Previously it was:
1. Check pricing for spot key
2. attempt to query spot pricing
3. Check pricing for ondemand key
4.  Use default pricing

Now it is:
1. Check Pricing for spot key
2. attempt to query spot pricing (may return on demand pricing)
3. Use default pricing

If you make the changes I have suggested then we should be able to merge this

pkg/cloud/azure/provider.go
barmettlerl: Makes total sense, don't know why this change was in the PR

pkg/cloud/azure/provider.go
barmettlerl: No, I will restore this. Might be due some changes between my fork and the dev on opencost

pkg/cloud/azure/provider.go
barmettlerl: true

pkg/cloud/azure/provider.go
barmettlerl: that's true, I'll change that

pkg/cloud/azure/provider.go
barmettlerl: I added a check for `az.Pricing == nil`.

pkg/cloud/azure/provider.go
```diff
@@ -1092,68 +1091,71 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
-		if n, ok := az.Pricing[spotFeatures]; ok {
-			log.DedupedInfof(5, "Returning pricing for node %s: %+v from key %s", azKey, n, spotFeatures)
-			if azKey.isValidGPUNode() {
-				n.Node.GPU = "1" // TODO: support multiple GPUs
-			}
-			return n.Node, meta, nil
+		featureString = fmt.Sprintf("%s,%s,spot", region, instance)
+	} else {
+		featureString = azKey.Features()
+	}
```
Sean-Holcomb, line 1125:
```suggestion
		features := strings.Split(featureString, ",")
		region := features[0]
		instance := features[1]
		featureString = fmt.Sprintf("%s,%s,spot", region, instance)
	}
```

pkg/cloud/azure/provider.go
```diff
@@ -1092,68 +1091,71 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
 	config, _ := az.GetConfig()
 
-	// Spot Node
 	slv, ok := azKey.Labels[config.SpotLabel]
 	isSpot := ok && slv == config.SpotLabelValue && config.SpotLabel != "" && config.SpotLabelValue != ""
+
+	features := strings.Split(azKey.Features(), ",")
+	region := features[0]
+	instance := features[1]
+	var featureString string
```
Sean-Holcomb, line 1120:
```suggestion
	featureString := azKey.Features()
```

pkg/cloud/azure/provider.go
```diff
@@ -290,7 +290,6 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 	retailPrice := ""
```
Sean-Holcomb, line 290:
```suggestion
	retailPrice := ""
	spotPrice := ""
```

pkg/cloud/azure/provider.go
```diff
@@ -290,7 +290,6 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
...
 	for _, item := range pricingPayload.Items {
 		if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
-			// if spot is true SkuName should contain "spot, if it is false it should not
 			if spot == strings.Contains(strings.ToLower(item.SkuName), " spot") {
 				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
 			}
```
Sean-Holcomb:
```suggestion
			if  !strings.Contains(strings.ToLower(item.SkuName), " spot") {
				spotPrice = fmt.Sprintf("%f", item.RetailPrice)
			} else {
				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
			}
```

Sean-Holcomb: I was thinking that you could save both the ondemand and spot prices then outside of this loop do something like this.
```
if spot && spotPrice != "" {
	retailPrice = spotPrice
}
```
This way if the spot price is not found then it defaults to the ondemand price from the api

Sean-Holcomb (commented):
I think you are getting close here, I added in some suggestion that I believe will get you closer to your original intention with this PR

pkg/cloud/azure/provider.go
barmettlerl: I agree something like that would be on spot :)

barmettlerl: I've added something like that

pkg/cloud/azure/provider.go
```diff
@@ -1092,68 +1098,69 @@ func (az *Azure) AllNodePricing() (interface{}, error) {
...
-			gpu := ""
-			if azKey.isValidGPUNode() {
-				gpu = "1"
-			}
-			spotNode := &models.Node{
-				Cost:      spotCost,
+		return n.Node, meta, nil
+	}
+
+	cost, err := getRetailPrice(region, instance, config.CurrencyCode, isSpot)
```
Sean-Holcomb, line 1135: looks like one of my previous suggestions broke the code

Sean-Holcomb (changes requested):

pkg/cloud/azure/provider.go
barmettlerl: ah, yes indeed. I missed that. Sorry

Sean-Holcomb (approved):
Good work!

cliffcolvin:
Thanks for your contribution @barmettlerl! Got this merged in!

PR #2989, opened 2024-11-15
Update azure.json

### LINKED ISSUE #1923: Azure node pricing seems odd. 1 Core == 20 GB of memory.
For Azure nodes cost of the node is based on the fact 1 CPU Core == 20 GB of Memory. This feels odd.

Based on this information https://github.com/opencost/opencost/tree/develop/configs

Azure:
```json
{
    "CPU": "0.03900",
    "RAM": "0.001917",
}
```

Alibaba:
```json
{
    "CPU": "0.031611",
    "RAM": "0.004237"
}
```

AWS:
```json
{
    "CPU": "0.031611",
    "RAM": "0.004237"
}
```

Default:
```json
{
    "CPU": "0.031611",
    "RAM": "0.004237"
}
```

GCP:
```json
{
    "CPU": "0.031611",
    "RAM": "0.004237"
}
```

The Azure is the odd one in the list. For all other clouds default ratio is 1 Core = 7.4GB of Memory.

A typical node has 4GB of RAM for each core. For such nodes 83% of the node cost will be attributed to the CPU and 17% to the memory. It means an azure cluster cost primarily dictated by the CPU.

For comparison, in other clouds only 65% of similar node cost is attributed to the CPU.

I couldn't find any information on how these config files were generated. Is there any way to validates it's correct?
It can be correct, but feels odd.

AjayTripathy:
## What does this PR change?
* Adjust Azure CPU/GPU/RAM ratios such that RAM is 1/GB, CPU is 10, GPU is 60

* No

* Adjusts Azure defaults to more reasonable numbers

* Closes https://github.com/opencost/opencost/issues/1923

## How was this PR tested?
* Proposal; needs testing still

* Yes; we need to document these ratios


configs/azure.json
```diff
@@ -3,8 +3,8 @@
     "description": "Azure estimates based on April 2019 advertised prices",
     "CPU": "0.03900",
     "spotCPU": "0.007764",
-    "RAM": "0.001917",
-    "GPU": "0.0428925",
+    "RAM": "0.00400",
+    "GPU": "0.01828925",
```
r2k1: This made GPU cheaper than a CPU core

AjayTripathy: Thanks for the catch. I think I meant to make it 0.18..

AjayTripathy:
@r2k1 sorry for the late response and thanks for the review; PTAL when you can.

bmiguel-teixeira:
Hi @AjayTripathy 

Qiick question. How are you guys extrapolating these "unit" costs from the general pricing? Or even these 1/10/60 rations?

PR #2998, opened 2024-11-20
Revert "reduce memory consumption (#2725)"

cliffcolvin:
Reverts opencost/opencost#2725

Issues reported where we are leaking pointers based on this change. Causes issues with stability in clusters that would have a possible duplicate annotations. 
I will get a task on our side to own working on the bigger memory change from Artur and try and get that merged. This one needed to be reverted to fix critical bug though.


mbolt35 (approved):
LGTM

PR #3002, opened 2024-11-22
More error logging when parsing response from Prometheus

thomasvn:
## What does this PR change?
* More error logging when parsing response from Prometheus.
* This is an important debugging utility in the event that OpenCost cannot parse Prometheus responses. An inability to parse Prometheus responses typically results in a non-functioning Allocations and Assets API.

Example log below:

```
Error parsing Prometheus response: 'value' field does not exist in data result vector. Query: 'avg(node_total_hourly_cost{}) by (node, cluster, provider_id)[3h:10m]'. Response: 'map[metric:map[cluster:REDACTED node:REDACTED provider_id:REDACTED]]'
```

* N/A

* Discussed above

* N/A

## How was this PR tested?
* N/A

* N/A

* N/A


pkg/prom/result_test.go
thomasvn, line 1: I didn't feel it was necessary to add unit tests, since this PR only modifies functions which return `error`. However I've added some here to pass SonarQube gate.

Let me know if you think we should keep or remove the unit tests.

nickcurie: Definitely think we should keep unit tests! Thanks for doing that

kubecost-pb (approved):

pkg/prom/result.go
```diff
@@ -16,52 +16,48 @@ var (
...
-func DataPointFormatErr(query string) error {
-	return fmt.Errorf("Improperly formatted datapoint from Prometheus fetching query '%s'", query)
+func DataPointFormatErr(query string, promResponse interface{}) error {
+	return fmt.Errorf("Error parsing Prometheus response: improperly formatted datapoint. Query: '%s'. Response: '%+v'", query, promResponse)
 }
 
-func MetricFieldDoesNotExistErr(query string) error {
-	return fmt.Errorf("Metric field does not exist in data result vector fetching query '%s'", query)
+func MetricFieldDoesNotExistErr(query string, promResponse interface{}) error {
+	return fmt.Errorf("Error parsing Prometheus Response: 'metric' field does not exist in data result vector. Query: '%s'. Response: '%+v'", query, promResponse)
```
nickcurie: let's lowercase this first "response" just to stay consistent with other errors

nickcurie (approved):
LGTM, just one small thing. Thanks for doing this!

pkg/prom/result.go
thomasvn: Oops, great catch. Thanks!

mbolt35 (approved):
Looks great - thanks @thomasvn !

thomasvn:
@mbolt35 Could you merge for me? I don't have access.

PR #3004, opened 2024-11-25
removing node exporter health check

mittal-ishaan:
## What does this PR change?
* Remove Node Exporter health check from the diagnostics API response because it is not critical to Opencost metrics and can cause confusion with false alarms about node-exporter

* NO

* The prometheus metrics diagnostics api response will not have node exporter health check

* Closes ... https://kubecost.atlassian.net/browse/ENG-3087

## How was this PR tested?
* ran opencost and kubecost locally and hit `diagnostics/prometheusMetrics` . The node exporter health check was not present

* NO

*


pkg/prom/diagnostics.go
```diff
@@ -78,13 +78,14 @@ var diagnosticDefinitions map[string]*diagnosticDefinition = map[string]*diagnos
...
 		Description: "Determine if metrics from Kubecost are available during last 5 minutes.",
 	},
-	NodeExporterDiagnosticMetricID: {
-		ID:          NodeExporterDiagnosticMetricID,
-		QueryFmt:    `absent_over_time(node_cpu_seconds_total{%s}[5m] %s)`,
-		Label:       "Node-exporter metrics available",
-		Description: "Determine if metrics from node-exporter are available during last 5 minutes.",
-		DocLink:     fmt.Sprintf("%s#node-exporter-metrics-available", DocumentationBaseURL),
-	},
+	// Commented out because it is not critical to Opencost metrics and can cause confusion with false alarms about node-exporter
```
cliffcolvin: I'm curious why we comment out instead of removing? My opinion has always been remove as we have git commit history if we need to revert or look at something.
CC: @mbolt35 thoughts?

mbolt35: Yeah, let's just remove. Totally fine with that.

mittal-ishaan: sure removing it :+1:

pkg/prom/diagnostics.go
```diff
@@ -78,13 +78,6 @@ var diagnosticDefinitions map[string]*diagnosticDefinition = map[string]*diagnos
 		Label:       "Kubecost metrics available",
 		Description: "Determine if metrics from Kubecost are available during last 5 minutes.",
 	},
-	NodeExporterDiagnosticMetricID: {
-		ID:          NodeExporterDiagnosticMetricID,
```
thomasvn, line 82: Looks like you still need to delete references to this

https://github.com/opencost/opencost/blob/05f1603bc324d6ba5cd756ae123d54f10b5c56c1/pkg/prom/diagnostics.go#L29-L30

mittal-ishaan: there are references to this in kcm. I will remove have to remove from there Before I can remove it from here. Let me do that

mittal-ishaan: https://github.com/kubecost/kubecost-cost-model/pull/2998

thomasvn (approved):

cliffcolvin (approved):

PR #3008, opened 2024-11-28
Cluster Cache Optimization Effort - Additional Fields

mbolt35:
This changeset is meant to build off of @r2k1 s optimization efforts on the `ClusterCache` implementation. This adds a few fields to the flyweight types in order to support downstream dependents. 

Additionally merged in `develop` and updated it to use new flyweight types. 

Relevant Changeset: https://github.com/opencost/opencost/commit/f955eac61729ab1f0696a1e2474c7490a0954fbe


pkg/clustercache/clustercache.go
```diff
@@ -118,11 +133,52 @@ type PersistentVolume struct {
...
+	Namespace    string
 	SpecSelector *metav1.LabelSelector
+	Spec         appsv1.ReplicaSetSpec
+}
+
+type Volume struct {
+}
+
+// GetControllerOf returns a pointer to a copy of the controllerRef if controllee has a controller
+func GetControllerOf(pod *Pod) *metav1.OwnerReference {
```
r2k1, line 160: I don't really like copying things around unless it's necessary. It's also can be difficult to read. Maybe something like generic `deepCopy` do the job?

I think copying is the main reasons of high memory usage on querying opencost.

r2k1 (approved):
LGTM

pkg/clustercache/clustercache.go
```diff
@@ -77,30 +81,41 @@ type Deployment struct {
 	Name                    string
 	Namespace               string
 	Labels                  map[string]string
+	Annotations             map[string]string
 	MatchLabels             map[string]string
 	SpecSelector            *metav1.LabelSelector
 	SpecReplicas            *int32
+	SpecStrategy            appsv1.DeploymentStrategy
 	StatusAvailableReplicas int32
+	PodSpec                 PodSpec
```
r2k1, line 90: I believe spec will have the highest memory footprint. However, it still serves as a good starting point and offers potential for future optimization.

r2k1:
Sorry, I think I botched merge/conflict resolution. (It's been difficult to see the diff without it)

pkg/clustercache/clustercache.go
mbolt35: This function is actually straight from the k8s API, but leveraging the opencost types. See here: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/controller_ref.go

mbolt35:
> Sorry, I think I botched merge/conflict resolution. (It's been difficult to see the diff without it)

I think it was me who botched it. I should've just asked you to merge develop into your changes first before I branched. My fault :(

pkg/clustercache/clustercache.go
mbolt35: Let me merge this and open a new change set to reduce this further!

PR #3013, opened 2024-12-06
add support for adding local to the response objects

ameijer:
## What does this PR change?
* fixes an issue where we weren't encoding Local properly. Rather than have that be the sole exception, we add support for it here 

* No

* Makes the JSON representation more consistent with the golang one

## How was this PR tested?
* tested via downstream unit tests 

* 

*


cliffcolvin (approved):
LGTM

PR #3021, opened 2024-12-16
Fix Window Parsing in JSON

ameijer:
## What does this PR change?
* when marshaling from json into structs, we were reading start/end as the window start/end on the json objects. This was not correct and now we read from the window

* no

* will make windows in unmarshaled structs match what is in the json 

* Closes ...

## How was this PR tested?
* tested via unit tests 

* 

*

### DIFF
--- a/core/pkg/opencost/asset_json.go
+++ b/core/pkg/opencost/asset_json.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nik-kc (approved):

kubecost-pb (approved):

Sean-Holcomb (approved):
Looks good

PR #3023, opened 2024-12-17
[Cluster Cache] Block to Pre-Initialize Lists

mbolt35:
## What does this PR change?
* Splits the CreateStore and Watch behaviors into the New and Watch functions. 
* This now resembles the behavior of the legacy implementation
* Addresses issues where initialization of `Provider` was happening before specific resources were available. 
* Blocks on `Run()` until initialization is complete. 

## Notes
This is a pretty annoying change, but the k8s client API is very difficult to work with. The only specific behavior that seemed to signal an initial list load/init was `Store.Replace` on the _first_ call. We set the init func to nil after we execute it to prevent future replace calls from re-running the onInit.

### DIFF
--- a/pkg/clustercache/clustercache2.go
+++ b/pkg/clustercache/clustercache2.go
@@ -1,7 +1,7 @@
 package clustercache
 
 import (
-	"context"
+	"sync"
 
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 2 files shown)


pkg/clustercache/clustercache2.go
```diff
@@ -26,32 +27,60 @@ type KubernetesClusterCacheV2 struct {
...
+		jobStore:                   CreateStore(clientset.BatchV1().RESTClient(), "jobs", transformJob),
+		pdbStore:                   CreateStore(clientset.PolicyV1().RESTClient(), "poddisruptionbudgets", transformPodDisruptionBudget),
+		stopCh:                     make(chan struct{}),
 	}
 }
 
 func (kcc *KubernetesClusterCacheV2) Run() {
+	var wg sync.WaitGroup
+
+	if !env.IsETLReadOnlyMode() {
```
mbolt35, line 56: This env var is horribly named - the intended use, based on the previous cluster cache implementation, seems to be that if this env var is `true`, we don't access any k8s resources.

ameijer: brutal.

ameijer (approved):
LGTM

PR #3024, opened 2024-12-19
GPU Label Pricing

nik-kc:
## What does this PR change?
* Updates the CSV provider to allow for a new `gpulabel` asset class:
  * The user can now specify a label/value pair in their CSV that is used as a GPU cost filter. If a given node has the specified label, that label's value is equal to the value in the CSV, and that node has a GPU, the GPU's cost will be set to the `MarketPriceHourly` set for that asset in the CSV.

* Negative.

* Users that rely on the CSV provider will now be able to further control the costs of their GPUs.

* Closes ...

## How was this PR tested?
* Tested in InstaDeep's environment and validated expected GPU costs for nodes with the specified label/value pair.
* Tested the same in our own clusters:
  * A cluster without any GPUs was unaffected.
  * A cluster with a single GPU (that also matched the label/value) had its GPU cost set properly.
  * A cluster with two GPUs (where only one matched the label/value) had its GPU cost set properly, on the correct node.

* Yes indeed.


pkg/cloud/provider/csvprovider.go
```diff
@@ -178,18 +183,22 @@ func (c *CSVProvider) DownloadPricingData() error {
 		} else if p.AssetClass == "gpu" {
 			gpupricing[key] = &p
 			c.GPUMapFields = append(c.GPUMapFields, strings.ToLower(p.InstanceIDField))
+		} else if p.AssetClass == "gpulabel" {
+			labelKeyValue := p.InstanceIDField + "=" + p.InstanceID
```
nik-kc, line 187: Here, `InstanceIDField` represents the label name, and `InstanceID` represents the label value. So, any instance of `labelName=labelValue` on nodes with GPUs will match this filter.

pkg/cloud/provider/provider.go
```diff
@@ -149,12 +149,20 @@ func ShareTenancyCosts(p models.Provider) bool {
 // NewProvider looks at the nodespec or provider metadata server to decide which provider to instantiate.
 func NewProvider(cache clustercache.ClusterCache, apiKey string, config *config.ConfigFileManager) (models.Provider, error) {
 	nodes := cache.GetAllNodes()
-	if len(nodes) == 0 {
-		log.Infof("Could not locate any nodes for cluster.") // valid in ETL readonly mode
-		return &CustomProvider{
-			Clientset: cache,
-			Config:    NewProviderConfig(config, "default.json"),
-		}, nil
+	var numRetries int
```
nik-kc: This is kinda gross, and I'd love some thorough feedback on it. It worked for the customer in a pinch, but it doesn't seem _good_. Would love for there to be a guarantee that `cache.GetAllNodes()` will block until it has retrieved all available nodes. What's super interesting is that we only ever actually use the first node in the nodes list 🤔

pkg/cloud/provider/provider.go
```diff
@@ -291,41 +299,51 @@ func getClusterProperties(node *clustercache.Node) clusterProperties {
 	if env.IsUseCustomProvider() {
 		// Use CSV provider if set
 		if env.IsUseCSVProvider() {
+			log.Debug("using custom CSV provider")
```
nik-kc, line 311: It was nigh on impossible to determine what provider was in use by a given deployment, hence these new debug logs.

pkg/costmodel/cluster_helpers.go
```diff
@@ -191,12 +191,6 @@ func buildGPUCostMap(
 	gpuCostMap := make(map[NodeIdentifier]float64)
 	clusterAndNameToType := make(map[nodeIdentifierNoProviderID]string)
 
-	customPricingEnabled := provider.CustomPricesEnabled(cp)
```
nik-kc, line 194: Big thanks to @ameijer for figuring out that custom pricing was overriding the provider's GPU cost settings.

pkg/cloud/provider/provider.go
mbolt35: Do we still need this code?

ameijer: unfortunately, we did confirm that we needed this code, even with your PR from yesterday merged

ameijer: we had first tried with just your PR contents, but we still had the same error where nothing was coming back and we were defaulting to the default provider

mbolt35: There is some deeper bug here, because I did confirm the initial issue. Let me find where this provider is initialized.

pkg/cloud/provider/provider.go
```diff
@@ -149,12 +149,20 @@ func ShareTenancyCosts(p models.Provider) bool {
...
-		}, nil
+	var numRetries int
+	for len(nodes) == 0 {
+		log.Infof("No nodes found for cluster, will retry...")
+		time.Sleep(time.Second)
+		nodes = cache.GetAllNodes()
+
+		numRetries++
+		if numRetries == 10 {
+			log.Infof("Could not locate any nodes for cluster.") // valid in ETL readonly mode
```
mbolt35: Note, if I can't figure this out, and this code is required. You'll only want to implement the retry for the following condition: 

```
if nodes == 0 && !env.IsETLReadOnlyMode()
```

mbolt35: This is because k8s-less installs (or lack of k8s permissions) use a default provider and expect this scenario.

mbolt35: There is also an `github.com/opencost/opencost/core/pkg/util/retry` utility for this behavior if interested. https://github.com/opencost/opencost/blob/develop/core/pkg/util/retry/retry.go

pkg/costmodel/costmodel.go
```diff
@@ -1143,95 +1175,109 @@ func (cm *CostModel) GetNodeCost(cp costAnalyzerCloud.Provider) (map[string]*cos
...
+			if newCnode.GPUCost == "" {
+				defaultCPU, err := strconv.ParseFloat(cfg.CPU, 64)
+				if err != nil {
+					log.Errorf("Could not parse default cpu price")
+					defaultCPU = 0
+				}
+				if math.IsNaN(defaultCPU) {
+					log.Warnf("defaultCPU parsed as NaN. Setting to 0.")
+					defaultCPU = 0
+				}
```
mbolt35, line 1199: This block confuses me -- why do we check `newCnode.GPUCost == ""` before setting the `defaultCPU`?

mbolt35: Oh, it looks like we do this for the ratios? Ah man, this stuff hurts my head. 😆

nik-kc: The purpose of this block is to calculate `GPUCost` based on the ratios between the default GPU/RAM costs and the default GPU/RAM costs. This block has been around for a good while; the change I made is to only execute it if `GPUCost` wasn't already set by CSV pricing.

mbolt35:
@nik-kc @ameijer Overall, I think this change looks fine, but I'm not completely sure about why double resolving custom pricing breaks GPUs? Are we just saying that custom GPU pricing is unsupported? 

Also, are we going out of draft status? I think we're good to go - just the lingering questions :)

pkg/cloud/provider/provider.go
nik-kc: Updated! I appreciate the feedback.

pkg/costmodel/costmodel.go
nik-kc: Whoops, I posted my comment before your second one showed up. Agreed! This stuff's a headache.

pkg/cloud/provider/provider.go
```diff
@@ -148,9 +150,24 @@ func ShareTenancyCosts(p models.Provider) bool {
...
+	}
+
+	var nodes []*clustercache.Node
+	if !env.IsETLReadOnlyMode() {
+		// the error can be ignored because getAllNodesFunc only errors if nodes is empty, a case which we explicitly
+		// handle by checking the length of nodes below
+		nodes, _ = retry.Retry(context.Background(), getAllNodesFunc, 10, time.Second)
+	} else {
+		nodes, _ = getAllNodesFunc()
+	}
```
mbolt35, line 168: Looks great!

nik-kc:
Ah, I goofed a rebase. Will fix ASAP.

mbolt35 (approved):
Looks good to me!

PR #3041, opened 2025-01-17
use empty structs, pre-size array during dedupe

ameijer:
## What does this PR change?
* for the deduplication code, this pre-sizes the array and switches to empty structs from bools to reduce space requirements 

* no

* we are hoping this should reduce memory consumption for large users 

## How was this PR tested?
* unit test was added

* 

*


mbolt35 (approved):
Looks great! Thanks for the contribution :)

PR #3075, opened 2025-03-15
feat: Use custom PV price from annotations

### LINKED ISSUE #3074: Custom pricing options for individual PVs across different storage classes
My clusters have different PVs with custom Storage Classes. Custom prices don't help me; I want to set different prices for each PV.

I would like to have some options to set different prices for all my custom PV.

**Additional context**
My AWS clusters usually use default storage classes from provider (EBS) - but I need to extend these classes for workloads that need RWX (NFS) or high availability (CephFS). Prices for PVs of these custom classes are different.
```
NAME            PROVISIONER 
csi-cephfs-sc   cephfs.csi.ceph.com
gp3 (default)   ebs.csi.aws.com
gp2             kubernetes.io/aws-ebs
nfs             cluster.local/nfs-subdir-external-provisioner
```

maksim-paskal:
This change allows users to set custom prices for Persistent Volumes via annotation:
```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    # this will use hourly cost for this PV as 1234
    opencost.io/custom-price: "1234"
  name: pv-custom-data
spec:
...
```
Or Storage Class annotations:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    # this will use hourly cost for all PV from this SC as 5678
    opencost.io/custom-price: "5678"
  name: custom-class
...
```

By default, this change doesn't affect users. This change allows users to set custom prices for each Persistent Volume with annotation.

This change was tested this way:
1. Annotate some PV 
```bash
kubectl annotate pv pv-custom-data opencost.io/custom-price=1234
# You can also annotate all your PVs by globally annotating their Storage Class.
kubectl annotate sc custom-class opencost.io/custom-price=5678
```
2. Run exporter
```bash
export KUBECONFIG=~/.kube/test
# kubectl -n prometheus port-forward svc/prometheus-server 9090:80
export PROMETHEUS_SERVER_ENDPOINT="http://127.0.0.1:9090"
export CLUSTER_ID=test-cluster
export KUBERNETES_PORT=1234

go run ./cmd/costmodel/main.go
```
3. Check pv_hourly_cost metrics - must be equal to annotation value
```bash
> curl -s 127.0.0.1:9003/metrics | grep pv_hourly_cost | grep pv-custom-data
pv_hourly_cost{persistentvolume="pv-custom-data",provider_id="",volumename="pv-custom-data"} 1234
```

Need documentation change for users.

Closes #3074

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -33,6 +33,12 @@ const (
 	profileThreshold = 1000 * 1000 * 1000 // 1s (in ns)
 
 	unmountedPVsContainer = "unmounted-pvs"
+
+	annotationDomain = "opencost.io"
+
+	annotationStorageCost = annotationDomain + "/storage-hourly-cost"
+	annotationNodeCPUCost = annotationDomain + "/node-cpu-hourly-cost"
+	annotationNodeRAMCost = annotationDomain + "/node-ram-hourly-cost"
 )
 
 // isCron matches a CronJob name and captures the non-timestamp name
@@ -820,6 +826,11 @@ func (cm *CostModel) addPVData(pvClaimMapping map[string]*PersistentVolumeClaimD
 			storageClassMap["default"] = params
 			storageClassMap[""] = params
 		}
+
+		// Add custom cost annotation to storage class map
+		if key, found := storageClass.Annotations[annotationStorageCost]; found && params != nil {
+			params[annotationStorageCost] = key
+		}
 	}
 
 	pvs := cache.GetAllPersistentVolumes()
@@ -862,6 +873,24 @@ func (cm *CostModel) addPVData(pvClaimMapping map[string]*PersistentVolumeClaimD
 	return nil
 }
 
+// Checks if the provided cost string can be parsed into a finite, non-negative float64.
+// If the cost is invalid, it logs a warning with the cost value and the reason.
+func (cm *CostModel) costIsValid(cost string) bool {
+	parsedCost, err := strconv.ParseFloat(cost, 64)
+	if err != nil {
+		log.Warnf("Invalid cost value: %s. Error: %s", cost, err.Error())
+		return false
+	}
+
+	// Check if the parsed cost is a valid number (not NaN, not Inf, and non-negative)
+	if math.IsNaN(parsedCost) || math.IsInf(parsedCost, 0) || parsedCost < 0 {
+		log.Warnf("Invalid cost value: %s. Error: cost must be a finite, non-negative number", cost)
+		return false
+	}
+
+	return true
+}
+
 func (cm *CostModel) GetPVCost(pv *costAnalyzerCloud.PV, kpv *clustercache.PersistentVolume, defaultRegion string) error {
 	cp := cm.Provider
 	cfg, err := cp.GetConfig()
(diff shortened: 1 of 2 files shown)


AjayTripathy:
@nikovacevic any thoughts here? Think you were working through some custom pricing stuff?

pkg/costmodel/costmodel.go
```diff
@@ -44,6 +44,9 @@ const (
 	epCleanTombstones = apiPrefix + "/admin/tsdb/clean_tombstones"
 	epConfig          = apiPrefix + "/status/config"
 	epFlags           = apiPrefix + "/status/flags"
+
+	annotationDomain      = "opencost.io"
+	annotationCustomPrice = annotationDomain + "/custom-price"
```
ameijer: concerns
- how does this apply to other assets?
- naming of annotation
- tests?

ctrought:
Would it be possible to have this annotation on the storageclass level, that way all PVs created through a specific storage class inherit that cost, versus having to annotate every PV?

maksim-paskal:
@ctrought thanks for question, it makes sense.

I just noticed that all my `pv_hourly_cost metrics` from my Azure clusters are equal to default values. Maybe I made some mistake while installing OpenCost in Azure. But if I have some options to customize my Storage Class prices, it would help me. I'm trying to figure out how to implement this.

I think that annotations for PV (for scenarios when users manually create PVs) are preferable if there are no annotations in the Storage Class.

pkg/costmodel/costmodel.go
maksim-paskal: @ameijer thanks for review:
1) These annotations will apply only for storages.
2) What do you suggest we name it?
3) Yes, sure. I will write tests for this if there is value in these changes.

ameijer: we are prepared to accept this @maksim-paskal , can you write tests for this, and create issues to cover the other asset types? Also, can you include the unit in the annotation name? like custom-price-per-unit? when you say `opencost.io/custom-price: 1234`, what does that mean?

maksim-paskal:
@ameijer for custom storage pricing - I change name of the annotation, also add test for that feature:
```yaml
opencost.io/storage-hourly-cost: "123.12"
```
What do you think about other assets that needs to be covered with that kind of annotations? I also add two new annotations for nodes to rewrite cpu and ram cost:
```yaml
opencost.io/node-cpu-hourly-cost: "111"
opencost.io/node-ram-hourly-cost: "222"
```
I think we need create another issue to cover all other assets?

joel931767:
Hey! Any news on this? We would also need this feature

joel931767:
Ping @maksim-paskal @ameijer

maksim-paskal:
@joel931767 you can try by you own this chages - you need to clone this repo and build with latest docker cli
```bash
git clone https://github.com/maksim-paskal/opencost.git -b use-custom-price-from-pv-annotations
docker build --platform=linux/amd64,linux/arm64 --pull --push -t your-host/your-repo:your-tag . 
```
or you can use my image `docker.io/paskalmaksim/opencost:3a9c25c`

maksim-paskal:
@AjayTripathy do we need triangle for this PR?

ameijer (approved):
@maksim-paskal merging after group review! thank you!

KevinGimbel:
Hi, this MR has been approved for a while, is there any estimation when it will land in a release? Thanks! (And sorry if I missed something, I'm quite new to the OpenCost project!)

ameijer:
@maksim-paskal sorry for all the delays here in getting this in. would you be able to look at the copilot findings and see what you agree with? also, the build is broken looks like

ameijer (approved):

PR #3080, opened 2025-03-17
Network Insights to OpenCost Core

mbolt35:
## What does this PR change?
* Addition of Network Insights structs and filters to opencost/core

### DIFF
--- a/core/pkg/opencost/networkinsightgraph.go
+++ b/core/pkg/opencost/networkinsightgraph.go
@@ -1,0 +1,224 @@
+package opencost
+
+import (
+	"fmt"
+
+	"github.com/opencost/opencost/core/pkg/log"
+)
+
+const (
+	NetworkInsightsNodeUnknown  = "unknownNode"
+	NetworkInsightsCloudService = "cloudService"
+	NetworkInsightsOther        = "other"
+	NetworkInsightsInternet     = "internet"
+	NetworkInsightsSameZone     = "sameZone"
+	NetworkInsightsSameRegion   = "sameRegion"
+)
+
+type NetworkInsightsEdge struct {
+	Source string  `json:"source"`
+	Target string  `json:"target"`
+	Cost   float64 `json:"cost"`
+	Bytes  float64 `json:"BytesInGiB"`
+}
+
+func NewNetworkInsightsEdge(source string,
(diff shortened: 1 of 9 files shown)


core/pkg/opencost/networkinsightgraph.go
```diff
@@ -0,0 +1,224 @@
+package opencost
```
ameijer, line 1: would it be better to put these in their own packages? or do they need to be here b/c of bingen?

core/pkg/opencost/opencost_codecs.go
```diff
@@ -18,7 +18,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/opencost/opencost/core/pkg/util"
+	util "github.com/opencost/opencost/core/pkg/util"
```
ameijer, line 21: is this aliasing thing new?

core/pkg/opencost/networkinsightgraph.go
mbolt35: They need to be here due to bingen.

core/pkg/opencost/opencost_codecs.go
mbolt35: It's not new because you can point that package to anything, and we always want to alias to util. In this case, it is the same, so I think saving on most IDEs removes it. It appears VS Code has stopped doing that for some reason.

mbolt35: For reference, you can pass the package with the buffer reference: https://github.com/opencost/opencost/blob/3713a75fc917f12bf9bc6614cf7ddaea31751ddb/core/pkg/opencost/bingen.go#L82

To avoid weird issues where bingen has to detect the package name, we just always alias to `util`.

ameijer (approved):

rajith77:
@mbolt35 just curious about this patch.
What's the TLDR?

mbolt35:
> @mbolt35 just curious about this patch. What's the TLDR?

These structures were part of kubecost's network insights, and we've decided to open source the network data collection (as part of the larger refactor). This changeset just moves all of the datas structures from our proprietary codebase to here :)

This will just be another optional pipeline that can be enabled if desired.

PR #3085, opened 2025-03-26
fix duplicate labels error in emitted metrics

mittal-ishaan:
## What does this PR change?
* Creates a deep copy of the stored items so the callers do not modify them which was earlier causing duplicate labels to be present when creating and emitting pod annotation metrics. We did similar to what we had in `cachev1` to prevent things like these:
https://github.com/opencost/opencost/blob/66879963f306bd066f9ccbc667936d31553ca00a/pkg/clustercache/watchcontroller.go#L90-L104

* issue came in: https://github.com/opencost/opencost/pull/2736
* similar issue that was fixed in: https://github.com/opencost/opencost/pull/2998

* Prometheus targets that were in failing state are now healthy

* Closes ...

## How was this PR tested?
* Earlier with the latest 1.114.0 image we are getting
```
 13 error(s) occurred:
* collected metric "kube_pod_annotations" ...
 has two or more labels with the same name: annotation_kubectl_kubernetes_io_last_applied_configuration
 ```
 at `/metrics` endpoint and the prometheus target `opencost` was down

* created an opencost image and tested it in an openshift env. where we were facing the issue earlier with these changes the issue was not present. the `/metrics` endpoint loaded with no error and prometheus target opencost was now up and healthy

* No

* Next release and be included in previors 1.114.0 release

### DIFF
--- a/pkg/clustercache/store.go
+++ b/pkg/clustercache/store.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	rt "k8s.io/apimachinery/pkg/runtime"
 )
 
 // GenericStore is a generic store implementation. It converts objects to a different type using a transform function.
@@ -86,9 +87,14 @@ func (s *GenericStore[Input, Output]) Delete(obj any) error {
 func (s *GenericStore[Input, Output]) GetAll() []Output {
 	s.mutex.RLock()
 	defer s.mutex.RUnlock()
+
+    // Deep copy the stored items to ensure that callers do not modify
+    // the original objects in the store.
 	allItems := make([]Output, 0, len(s.items))
 	for _, item := range s.items {
-		allItems = append(allItems, item)
+		if deepCopyable, ok := any(item).(rt.Object); ok {
+			allItems = append(allItems, deepCopyable.DeepCopyObject().(Output))
+		}
 	}
 	return allItems
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/clustercache/store.go [hunk 1]
```diff
@@ -86,9 +87,16 @@ func (s *GenericStore[Input, Output]) Delete(obj any) error {
...
+    // Deep copy the stored items to ensure that callers do not modify
+    // the original objects in the store.
 	allItems := make([]Output, 0, len(s.items))
 	for _, item := range s.items {
-		allItems = append(allItems, item)
+		deepCopyable, ok := any(item).(rt.Object);
+		if !ok {
+			continue
+		}
+		allItems = append(allItems, deepCopyable.DeepCopyObject().(Output))
```
thomasvn: Minor style change? Matches the code in watchcontroller.go which you've linked to

```suggestion
		if deepCopyable, ok := any(item).(rt.Object); ok {
                    allItems = append(allItems, deepCopyable.DeepCopyObject().(Output))
                }
```

thomasvn: Don't directly commit this ^^. I think it was formatted using spaces instead of tabs.

thomasvn:
@mittal-ishaan Remember to sign your commits to pass DCO!

pkg/clustercache/store.go
```diff
@@ -86,9 +87,14 @@ func (s *GenericStore[Input, Output]) Delete(obj any) error {
 func (s *GenericStore[Input, Output]) GetAll() []Output {
 	s.mutex.RLock()
 	defer s.mutex.RUnlock()
+
+    // Deep copy the stored items to ensure that callers do not modify
+    // the original objects in the store.
```
thomasvn: Minor thing, fix indentation here?

thomasvn (approved):
This looks good to me. Let's make sure this doesn't interfere with any of the work @mbolt35 is doing.

mbolt35 (approved):
LGTM

PR #3086, opened 2025-03-27
Add metric collectors for all metrics

Sean-Holcomb:
## What does this PR change?
* This PR adds metric collectors for each required metric

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


modules/collector-source/pkg/collector/activeminutes.go
```diff
@@ -0,0 +1,44 @@
...
+	return m.name
+}
+
+func (m *ActiveMinutesAggregator) LabelValues() []string {
+	return m.labelValues
+}
+
+func (m *ActiveMinutesAggregator) Update(value float64) {
+	now := time.Now().UTC()
+	if m.start != nil {
```
mbolt35: should this be? 
```
m.start == nil
```

modules/collector-source/pkg/collector/info.go
```diff
@@ -0,0 +1,30 @@
...
+func (m *InfoAggregator) LabelValues() []string {
+	return m.labelValues
+}
+
+func (m *InfoAggregator) Update(value float64) {
+
+}
+
+func (m *InfoAggregator) Value() float64 {
+	return 0
```
mbolt35: I think most of these return `1` in prom. We can do the same for equivalence. There may be some underlying reason /shrug

PR #3094, opened 2025-04-03
Add offset when extracting start time from Prometheus query

thomasvn:
## What does this PR change?

**Resolved an issue where OpenCost reported incorrect resource start times when used with Prometheus v3.** This affected Assets (nodes, disks, loadbalancers) and Allocations (containers) data.

BEFORE (occurs on queries to Assets and Allocations):

```
      "window": {
        "start": "2025-04-03T00:00:00Z",     HOURLYWINDOW
        "end": "2025-04-03T01:00:00Z"
      },
      "start": "2025-04-03T00:05:00Z",     5MINUTESOFF (DEFAULT VALUE OF ETL_RESOLUTION_SECONDS)
      "end": "2025-04-03T01:00:00Z",
      "minutes": 55.000000,
      "nodeType": "t4g.small",
```

AFTER:

```
      "window": {
        "start": "2025-04-03T00:00:00Z",
        "end": "2025-04-03T01:00:00Z"
      },
      "start": "2025-04-03T00:00:00Z",
      "end": "2025-04-03T01:00:00Z",
      "minutes": 60.000000,
      "nodeType": "t4g.small",
```

The breaking change introduced by Prometheus is described [here](https://prometheus.io/docs/prometheus/3.0/migration/#promql):

> Lookback and range selectors are left open and right closed (previously left closed and right closed). This change affects queries when the evaluation time perfectly aligns with the sample timestamps. For example assume querying a timeseries with evenly spaced samples exactly 1 minute apart. Before Prometheus v3, a range query with `5m` would usually return 5 samples. But if the query evaluation aligns perfectly with a scrape, it would return 6 samples. In Prometheus v3 queries like this will always return 5 samples.

Reintroduces offsetting logic, which was removed by https://github.com/opencost/opencost/pull/2059

Users will need to upgrade to a patched version.

- https://apptio.atlassian.net/browse/KCM-3304

## How was this PR tested?

**1. Ran OpenCost locally and tested the following queries:**

* `http://localhost:9003/assets?window=2025-04-06T00:00:00Z,2025-04-07T00:00:00Z`. Correct start times for nodes, loadbalancers, and disks
* `http://localhost:9003/assets?window=2025-04-06T12:30:00Z,2025-04-06T14:00:00Z`. Correct start and end times for nodes, loadbalancers, and disks
* `http://localhost:9003/allocation?window=2025-04-06T00:00:00Z,2025-04-07T00:00:00Z`. All long-running containers (not cronjobs) had start times aligned with the window start time.
* `http://localhost:9003/allocation?window=2025-04-06T12:00:00Z,2025-04-06T14:30:00Z` All long-running containers (not cronjobs) had start and end times aligned with the window.

**2. Ran OpenCost locally against a Prometheus v2.** Tested all queries listed above, and saw the expected results.

**3. Built and deployed a Kubecost cost-model image.** One deployed with a Promv3, another with a Promv2. Tested all queries listed above, and saw the expected results.

**4. Integration tests written, to catch any future regressions.** https://github.com/kubecost/kubecost-integration-tests/pull/209

N/A

v1.115

### DIFF
--- a/pkg/costmodel/allocation_helpers.go
+++ b/pkg/costmodel/allocation_helpers.go
@@ -122,7 +122,8 @@ func (cm *CostModel) buildPodMap(window opencost.Window, resolution, maxBatchSiz
 			}
 		}
 
-		applyPodResults(window, resolution, podMap, clusterStart, clusterEnd, resPods, ingestPodUID, podUIDKeyMap)
+		resPromVersion, _ := ctx.GetMajorVersion().Await()
+		applyPodResults(window, resolution, podMap, clusterStart, clusterEnd, resPods, ingestPodUID, podUIDKeyMap, resPromVersion)
 
 		coverage = coverage.ExpandEnd(batchEnd)
 		numQuery++
@@ -131,7 +132,7 @@ func (cm *CostModel) buildPodMap(window opencost.Window, resolution, maxBatchSiz
 	return nil
 }
 
-func applyPodResults(window opencost.Window, resolution time.Duration, podMap map[podKey]*pod, clusterStart, clusterEnd map[string]time.Time, resPods []*prom.QueryResult, ingestPodUID bool, podUIDKeyMap map[podKey][]podKey) {
+func applyPodResults(window opencost.Window, resolution time.Duration, podMap map[podKey]*pod, clusterStart, clusterEnd map[string]time.Time, resPods []*prom.QueryResult, ingestPodUID bool, podUIDKeyMap map[podKey][]podKey, promVersion int) {
 	for _, res := range resPods {
 		if len(res.Values) == 0 {
 			log.Warnf("CostModel.ComputeAllocation: empty minutes result")
@@ -171,7 +172,7 @@ func applyPodResults(window opencost.Window, resolution time.Duration, podMap ma
 
 		}
 
-		allocStart, allocEnd := calculateStartAndEnd(res, resolution, window)
+		allocStart, allocEnd := calculateStartAndEnd(res, resolution, window, promVersion)
 		if allocStart.IsZero() || allocEnd.IsZero() {
 			continue
 		}
@@ -1451,14 +1452,14 @@ func applyServicesToPods(podMap map[podKey]*pod, podLabels map[podKey]map[string
 	}
 }
 
-func getLoadBalancerCosts(lbMap map[serviceKey]*lbCost, resLBCost, resLBActiveMins []*prom.QueryResult, resolution time.Duration, window opencost.Window) {
+func getLoadBalancerCosts(lbMap map[serviceKey]*lbCost, resLBCost, resLBActiveMins []*prom.QueryResult, resolution time.Duration, promVersion int, window opencost.Window) {
 	for _, res := range resLBActiveMins {
 		serviceKey, err := resultServiceKey(res, env.GetPromClusterLabel(), "namespace", "service_name")
 		if err != nil || len(res.Values) == 0 {
 			continue
 		}
 
-		lbStart, lbEnd := calculateStartAndEnd(res, resolution, window)
+		lbStart, lbEnd := calculateStartAndEnd(res, resolution, window, promVersion)
 		if lbStart.IsZero() || lbEnd.IsZero() {
 			log.Warnf("CostModel.ComputeAllocation: pvc %s has no running time", serviceKey)
 		}
@@ -1903,7 +1904,7 @@ func (cm *CostModel) getNodePricing(nodeMap map[nodeKey]*nodePricing, nodeKey no
 
 /* PV/PVC Helpers */
 
-func buildPVMap(resolution time.Duration, pvMap map[pvKey]*pv, resPVCostPerGiBHour, resPVActiveMins, resPVMeta []*prom.QueryResult, window opencost.Window) {
+func buildPVMap(resolution time.Duration, pvMap map[pvKey]*pv, resPVCostPerGiBHour, resPVActiveMins, resPVMeta []*prom.QueryResult, promVersion int, window opencost.Window) {
 	for _, result := range resPVActiveMins {
 		key, err := resultPVKey(result, env.GetPromClusterLabel(), "persistentvolume")
 		if err != nil {
@@ -1911,7 +1912,7 @@ func buildPVMap(resolution time.Duration, pvMap map[pvKey]*pv, resPVCostPerGiBHo
 			continue
 		}
 
-		pvStart, pvEnd := calculateStartAndEnd(result, resolution, window)
+		pvStart, pvEnd := calculateStartAndEnd(result, resolution, window, promVersion)
 		if pvStart.IsZero() || pvEnd.IsZero() {
 			log.Warnf("CostModel.ComputeAllocation: pv %s has no running time", key)
 		}
@@ -1984,7 +1985,7 @@ func applyPVBytes(pvMap map[pvKey]*pv, resPVBytes []*prom.QueryResult) {
 	}
 }
 
-func buildPVCMap(resolution time.Duration, pvcMap map[pvcKey]*pvc, pvMap map[pvKey]*pv, resPVCInfo []*prom.QueryResult, window opencost.Window) {
+func buildPVCMap(resolution time.Duration, pvcMap map[pvcKey]*pvc, pvMap map[pvKey]*pv, resPVCInfo []*prom.QueryResult, promVersion int, window opencost.Window) {
 	for _, res := range resPVCInfo {
 		cluster, err := res.GetString(env.GetPromClusterLabel())
 		if err != nil {
@@ -2005,7 +2006,7 @@ func buildPVCMap(resolution time.Duration, pvcMap map[pvcKey]*pvc, pvMap map[pvK
 		pvKey := newPVKey(cluster, volume)
 		pvcKey := newPVCKey(cluster, namespace, name)
 
-		pvcStart, pvcEnd := calculateStartAndEnd(res, resolution, window)
+		pvcStart, pvcEnd := calculateStartAndEnd(res, resolution, window, promVersion)
 		if pvcStart.IsZero() || pvcEnd.IsZero() {
 			log.Warnf("CostModel.ComputeAllocation: pvc %s has no running time", pvcKey)
 		}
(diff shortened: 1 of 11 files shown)


pkg/costmodel/allocation_helpers.go
```diff
@@ -2353,18 +2352,18 @@ func getUnmountedPodForNamespace(window opencost.Window, podMap map[podKey]*pod,
...
 	s := time.Unix(int64(result.Values[0].Timestamp), 0).UTC()
 	e := time.Unix(int64(result.Values[len(result.Values)-1].Timestamp), 0).UTC()
 
+	// As of Prometheus v3, we have had to reintroduce the "offsetting" of the
+	// start time.
+	//
+	// E.g. avg(node_total_hourly_cost{}) by (node, provider_id)[1h:5m] with
+	// time=01:00:00 will return, for a node running the entire time, 12
+	// timestamps where the first is 00:05:00 and the last is 01:00:00.
+	s = s.Add(-resolution)
```
ameijer: will this break v2 prometheus installs?

pkg/costmodel/allocation_helpers_test.go
```diff
@@ -455,7 +455,6 @@ func TestGetUnmountedPodForCluster(t *testing.T) {
 }
 
 func TestCalculateStartAndEnd(t *testing.T) {
-
```
ameijer, line 458: can you unit test your change? seems very worth it

pkg/costmodel/allocation_helpers.go
thomasvn: This is something Ishaan and I are testing now! If so, we'll need to modify this solution.

pkg/costmodel/allocation_helpers_test.go
thomasvn: Yes! Unit tests need to be updated, currently failing.

Because unit tests rely on a mocked Prometheus QueryResult, I'd like to eventually get some integration tests around this to catch regressions.

pkg/costmodel/allocation_helpers.go
thomasvn: I'm guessing that we should stil be safe when querying Prometheus v2 because of:

```go
	if s.Before(*window.Start()) {
		s = *window.Start()
	}
```

pkg/costmodel/allocation_helpers_test.go
thomasvn: Unit tests now passing.

avrodrigues5:
@thomasvn so, the prometheus v3 was adding an 5m offset to start time but why specifically 5m?

avrodrigues5:
> @thomasvn so, the prometheus v3 was adding an 5m offset to start time but why specifically 5m?

Ah never mind, i see the notes. It the set ETL_RESOLUTION_SECONDS.

thomasvn:
@avrodrigues5 Yes exactly! The start time was off by however much `ETL_RESOLUTION_SECONDS` was set to. Let me update PR notes for clarity.

pkg/costmodel/allocation_helpers_test.go
```diff
@@ -455,37 +441,35 @@ func TestGetUnmountedPodForCluster(t *testing.T) {
...
 		"1 hour resolution, 1 hour window": {
 			resolution:    time.Hour,
+			window:        opencost.NewClosedWindow(windowStart, windowStart.Add(time.Hour)),
 			expectedStart: windowStart,
 			expectedEnd:   windowStart.Add(time.Hour),
 			result: &prom.QueryResult{
 				Values: []*util.Vector{
-					{
-						Timestamp: startFloat,
-					},
```
thomasvn, line 473: All unit tests that extract start and end times from a mocked Prometheus query result have been updated. Note that the first vector in the slice needed to be removed to align with the behavior introduced in Prometheus v3.

pkg/costmodel/allocation_helpers_test.go
```diff
@@ -495,36 +479,83 @@ func TestCalculateStartAndEnd(t *testing.T) {
...
+						Timestamp: startFloat + (minute * 30),
+					},
+					{
+						Timestamp: startFloat + (minute * 45),
+					},
+				},
+			},
+		},
+		// Example: avg(node_total_hourly_cost{}) by (node, provider_id)[30m:5m]
+		"5 minute resolution, 30 minute window": {
```
thomasvn, line 509: Added two new unit tests at 5 minute resolution, since that is the default value for `ETL_RESOLUTION_SECONDS`.

pkg/costmodel/allocation_helpers.go
ameijer: what if I am running v2 and my node legitimately started at 10 mins past the hour? will it get walked back to 5 min past the hour?

thomasvn: Yes, unfortunately that's correct. If you have Prometheus v2 and your node legitimately starts at `time=01:15:00`, this function will return a start time of `time=01:10:00` (assuming default `ETL_RESOLUTION_SECONDS`).

This only occurs if a node started sometime within the query window. Which is less bad than the current bug where all asset/allocation start times are affected regardless of query window.

thomasvn: If needed, we could refactor to use the [`/api/v1/status/buildinfo`](https://prometheus.io/docs/prometheus/latest/querying/api/#build-information) Prometheus API to determine the Prometheus version the user is running?

ameijer: I feel like we need _something_, right? otherwise we just break this in a different way for other users. I see in the tests you are removing certain vectors from the mock responses. is there any way you can read the response for the presence/non presence of those and infer from that?

avrodrigues5: @thomasvn im just putting some thought here but i total agree with Alex, you need a more robust way to identify prometheus version 

perhaps you could leverage the /version end point of prometheus as https://prom.kubecost.xyz/version

Then identify major version to be 3.X.X and only then add offset of 5 minutes. @ameijer thoughts?

ameijer: any of these are good options. Whatever is easiest

pkg/prom/query.go
```diff
@@ -419,3 +420,64 @@ func warningsFrom(result interface{}) v1.Warnings {
...
+		defer errors.HandlePanic()
+		if version, err := ctx.getMajorVersion(); err != nil {
+			ctx.errorCollector.Report(apiPrefix+"/status/buildinfo", nil, err, nil)
+		} else {
+			resCh <- version
+		}
+	}()
+
+	return resCh
+}
```
thomasvn: I wrote this based off the `QueryAtTime()` function in this file, since it is heavily used across this codebase. Open to any and all feedback on this!

```
func (ctx *Context) QueryAtTime(query string, t time.Time) QueryResultsChan {
```

pkg/prom/result.go
```diff
@@ -60,6 +60,18 @@ func ValueFieldDoesNotExistErr(query string, promResponse interface{}) error {
 	return fmt.Errorf("Error parsing Prometheus response: 'value' field does not exist in data result vector. Query: '%s'. Response: '%+v'", query, promResponse)
 }
 
+// VersionResultsChan is a channel for Prometheus major version results
+type VersionResultsChan chan int
```
thomasvn: My API request to [`/api/v1/status/buildinfo`](https://prometheus.io/docs/prometheus/latest/querying/api/#build-information) returns very different response contents compared to API requests for [`/api/v1/query`](https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries).

I didn't want to force my response contents into a `QueryResult` type if it didn't make sense. So I started this new type which is just a Channel for an int representing the Prometheus major version number. I'm starting with a very narrow scope, but open to broadening it if you think it makes sense.

pkg/prom/query.go
thomasvn: No longer relevant. I reverted this commit.

pkg/prom/result.go
thomasvn: No longer relevant. I reverted this commit.

pkg/costmodel/router.go
```diff
@@ -1223,11 +1227,13 @@ func Initialize(router *httprouter.Router, additionalConfigWatchers ...*watcher.
 	}
 
 	api := prometheusAPI.NewAPI(promCli)
-	_, err = api.Buildinfo(context.Background())
+	result, err := api.Buildinfo(context.Background())
```
thomasvn, line 1226: I found this API call we're already making when initializing cost-model. The response from this call will have a Prometheus version. I've saved that to a package-wide variable.

pkg/costmodel/allocation_helpers.go
thomasvn: @ameijer @avrodrigues5 Requesting another review! I've included logic to query/save the Prometheus version. Then only apply the offset if we are on a Prometheus version 3 or greater.

pkg/costmodel/allocation_helpers.go
```diff
@@ -2353,18 +2352,24 @@ func getUnmountedPodForNamespace(window opencost.Window, podMap map[podKey]*pod,
...
+	// As of Prometheus v3, we have had to reintroduce the "offsetting" of the
+	// start time.
+	//
+	// E.g. avg(node_total_hourly_cost{}) by (node, provider_id)[1h:5m] with
+	// time=01:00:00 will return, for a node running the entire time, 12
+	// timestamps where the first is 00:05:00 and the last is 01:00:00.
+	//
+	// The below logic checks if the Prometheus major version is 3 or higher.
+	if v := strings.Split(prometheusVersion, "."); len(v) > 0 {
+		if major, err := strconv.Atoi(v[0]); err == nil && major >= 3 {
```
ameijer: Your comment helps, but you have a magic number in here. I'd refactor this small bit of logic to a function isPrometheusV3() that returns a bool, then a simple if isPrometheusV3 + your comment will be plenty of documentation

pkg/costmodel/router.go
ameijer: very nice

ameijer (changes requested):
one last nit for readability, otherwise lgtm and we can ship

pkg/costmodel/allocation_helpers.go
thomasvn: Will do!

ameijer (approved):

PR #3096, opened 2025-04-03
Fix Azure public pricing download

kaelanspatel:
## What does this PR change?
Fixes an issue where Azure would not download public pricing data for configurations where the rate card info failed to download.

Alters/fixes behavior introduced in https://github.com/opencost/opencost/pull/2978.

Fixes an issue where Azure would not download public pricing data.

https://apptio.atlassian.net/browse/KCM-3373

## How was this PR tested?
Tested on Azure clusters (standard KC install + agent) without the rate card configured. Observed repeated failures to get data. After fix, observed successful data download and no additional failures beyond that.

No.

### DIFF
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -810,8 +810,9 @@ func (az *Azure) DownloadPricingData() error {
 	config.AzureClientID = clientID
 	config.AzureClientSecret = clientSecret
 	config.AzureTenantID = tenantID
-
 	var authorizer autorest.Authorizer
+
+	log.Debugf("Fetching azure rate card with credentials: %s:%s:%s:%s", config.AzureSubscriptionID, config.AzureClientID, config.AzureClientSecret, config.AzureTenantID)
 
 	azureEnv := determineCloudByRegion(az.ClusterRegion)
 
@@ -1100,8 +1101,10 @@ func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetada
 
 	meta := models.PricingMetadata{}
 
+	azurePricingExists := true
 	if az.Pricing == nil {
-		return nil, meta, fmt.Errorf("Unable to download Azure pricing data")
+		azurePricingExists = false
+		log.DedupedWarningf(5, "Azure pricing data not found")
 	}
 
 	azKey, ok := key.(*azureKey)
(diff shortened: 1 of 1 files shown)


pkg/cloud/azure/provider.go
```diff
@@ -810,9 +810,10 @@ func (az *Azure) DownloadPricingData() error {
 	config.AzureClientID = clientID
 	config.AzureClientSecret = clientSecret
 	config.AzureTenantID = tenantID
-
 	var authorizer autorest.Authorizer
 
+	log.Debugf("Fetching azure rate card with credentials: %s:%s:%s:%s", config.AzureSubscriptionID, config.AzureClientID, config.AzureClientSecret, config.AzureTenantID)
+
```
Sean-Holcomb: just planting a flag that this should be removed

pkg/cloud/azure/provider.go
```diff
@@ -1122,12 +1125,16 @@ func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetada
 		featureString = azKey.Features()
 	}
 
-	if n, ok := az.Pricing[featureString]; ok {
-		log.Debugf("Returning pricing for node %s: %+v from key %s", azKey, n, azKey.Features())
-		if azKey.isValidGPUNode() {
-			n.Node.GPU = azKey.GetGPUCount()
+	if azurePricingExists {
```
Sean-Holcomb: can we just do a nil check

pkg/cloud/azure/provider.go
```diff
@@ -1099,8 +1100,10 @@ func (az *Azure) NodePricing(key models.Key) (*models.Node, models.PricingMetada
 
 	meta := models.PricingMetadata{}
 
+	azurePricingExists := true
 	if az.Pricing == nil {
-		return nil, meta, fmt.Errorf("Unable to download Azure pricing data")
+		azurePricingExists = false
+		log.DedupedWarningf(5, "Azure pricing data not found")
```
Sean-Holcomb: This log maybe belongs somewhere else

kaelanspatel: It should be reflected in a log when set initially anyhow, I've just removed it.

thomasvn:
The changes in this PR generally look good! It'd be helpful to show more test results in the PR notes.

thomasvn:
Out of scope for this PR, but it's worth investigating whether we should be removing `DownloadPricingData()` and all other code which calls the Rate Card API, since it seems it is deprecated. https://learn.microsoft.com/en-us/previous-versions/azure/reference/mt219004(v%3Dazure.100)

That would make the `NodePricing()` function much easier to read.

Sean-Holcomb (approved):

nik-kc (approved):

nikovacevic (approved):

jessegoodier:
CP to 2.7 please @nik-kc ?

PR #3097, opened 2025-04-04
Revert "fix duplicate labels error in emitted metrics"

mittal-ishaan:
Reverts opencost/opencost#3085


ameijer (approved):

jessegoodier:
FYI, the reason this change was reverted was because this caused many /metrics to not be returned- 
things like the hourly_costs were missing.

PR #3098, opened 2025-04-07
[CDP/KCM-3570] - Establish Image Building in Merge Queue

cpetersen5:
## What does this PR change?
* 

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/.github/workflows/build-test-image.yml
+++ b/.github/workflows/build-test-image.yml
@@ -1,0 +1,44 @@
+name: Build Test Image
+
+on:
+  workflow_run:
+    workflows: [Build/Test]
+    types: [completed]
+
+concurrency:
+  group: build-opencost-test
+  cancel-in-progress: false
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+
+jobs:
+  build-and-publish-opencost:
+    runs-on: ubuntu-latest
+    if: |
+      github.event.workflow_run.conclusion == 'success' && 
+      contains(github.event.workflow_run.head_branch, 'gh-readonly-queue')
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-image.yml
```diff
@@ -0,0 +1,44 @@
+name: Build Test Image
+
+on:
+  workflow_run:
+    workflows: [Build/Test]
+    types: [completed]
+
+concurrency:
+  group: build-opencost-test
+  cancel-in-progress: false
```
ameijer, line 10: not needed

.github/workflows/build-test-image.yml
```diff
@@ -0,0 +1,44 @@
+name: Build Test Image
+
+on:
+  workflow_run:
+    workflows: [Build/Test]
+    types: [completed]
```
ameijer, line 6: I don't think you need this. this needs to instead be replaced by a merge group trigger

.github/workflows/build-test-image.yml
```diff
@@ -0,0 +1,44 @@
...
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+
+jobs:
+  build-and-publish-opencost:
+    runs-on: ubuntu-latest
+    if: |
+      github.event.workflow_run.conclusion == 'success' && 
+      contains(github.event.workflow_run.head_branch, 'gh-readonly-queue')
```
ameijer, line 21: interesting, I would actually read the event type

.github/workflows/build-test-image.yml
```diff
@@ -0,0 +1,44 @@
...
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
```
ameijer, line 27: checkout the specific head sha, this will check out a merge commit by default which isn't what we do, we squash

PR #3101, opened 2025-04-11
Sth/kcm 3385

Sean-Holcomb:
## What does this PR change?
* Update Storage to work for labels and filters
* scrapper for kubernetes data
* scrapper for network data

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/modules/collector-source/pkg/collector/collectordriver.go
+++ b/modules/collector-source/pkg/collector/collectordriver.go
@@ -1,21 +1,20 @@
 package collector
 
 import (
-	"sync/atomic"
 	"time"
 
 	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/util/atomic"
 )
 
 type Config struct {
 	ScrapeInterval time.Duration
 }
 type CollectorDriver struct {
-	config     Config
-	isRunning  atomic.Bool
-	isStopping atomic.Bool
-	stop       chan struct{}
-	collector  MetricsCollector
+	config    Config
+	runState  atomic.AtomicRunState
+	stop      chan struct{}
+	collector MetricsCollector
 }
 
 func NewCollectorDriver(config Config) *CollectorDriver {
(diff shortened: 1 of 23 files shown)


modules/collector-source/pkg/collector/metric.go
```diff
@@ -18,14 +22,36 @@ type MetricResult struct {
...
+			timestamp = float64(value.Timestamp.Unix())
+		}
+		values[i] = &util.Vector{
+			Timestamp: timestamp,
+			Value:     value.Value,
+		}
+	}
+
+	return source.NewQueryResult(metrics, values, nil)
+}
```
mbolt35, line 44: brilliant

modules/collector-source/pkg/collector/metric.go
```diff
@@ -41,38 +67,48 @@ type MetricCollector struct {
...
 		metrics:           make(map[uint64]MetricAggregator),
+		filter:            fn,
 	}
 }
 
-func (mi *MetricCollector) Update(labelValues []string, value float64, timestamp *time.Time) {
+func (mi *MetricCollector) Update(labels map[string]string, value float64, timestamp *time.Time, additionalInfo map[string]string) {
+	if mi.filter != nil && !mi.filter(labels) {
+		return
+	}
```
mbolt35, line 89: This is a super clean way of doing this, awesome idea.

modules/collector-source/pkg/collector/collectordriver.go
```diff
@@ -0,0 +1,66 @@
...
+		for {
+			select {
+			case <-cd.stop:
+				cd.isRunning.Store(false)
+				return
+			default:
+
+			}
+			time.Sleep(cd.config.ScrapeInterval)
+		}
```
mbolt35: This blocks? You can drop the whole dual atomic bool and stop channel design for an `AtomicRunState` -- Example here: https://github.com/opencost/opencost/blob/cf4d510ad4dd1c27e37d7ce98a9b2c4296e2a402/core/pkg/exporter/controller.go#L63-L131

Sean-Holcomb: That is definitely a better option, I never actually ended up plugging this in. I think the real question in my mind is do we actually even need start and stop for something like this.

PR #3114, opened 2025-04-18
KCM-3828 - Prevent duplicate Prometheus metric labels

nik-kc:
## What does this PR change?
* Updates the annotation/label parser to remove duplicate labels. Duplicates can occur when we have two annotations or labels with names that become equivalent post-sanitization (e.g. `app-` and `app_`).
* Note, the following behavior changes when we have labels that become duplicates post-sanitization:
  * Previously, metric emission would fail, as Prometheus doesn't allow duplicate labels.
  * Now, exactly one of the duplicate labels will remain, and the rest will be dropped. Exactly which one gets dropped is non-deterministic.

* Nope.

* Users will now be able to enable `kubecostMetrics.emitNamespaceAnnotations` and `kubecostMetrics.emitPodAnnotations` together.

* Closes [KCM-3828](https://apptio.atlassian.net/browse/KCM-3828).

## How was this PR tested?
* Unit tests.
* Tested by Jesse, validated that the error presented in the ticket no longer occurs.

* Nope.

*

### DIFF
--- a/core/pkg/util/promutil/promutil.go
+++ b/core/pkg/util/promutil/promutil.go
@@ -75,16 +75,22 @@ func LabelNamesFrom(labels map[string]string) []string {
 
 // Prepends a qualifier string to the keys provided in the m map and returns the new keys and values.
 func KubePrependQualifierToLabels(m map[string]string, qualifier string) ([]string, []string) {
-	keys := make([]string, 0, len(m))
-	for k := range m {
+	// sanitize the keys in m to prevent duplicate output keys
+	sanitizedM := make(map[string]string)
+	for k, v := range m {
+		sanitizedM[SanitizeLabelName(k)] = v
+	}
+
+	keys := make([]string, 0, len(sanitizedM))
+	for k := range sanitizedM {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
 
-	values := make([]string, 0, len(m))
+	values := make([]string, 0, len(sanitizedM))
 	for i, k := range keys {
-		keys[i] = qualifier + SanitizeLabelName(k)
-		values = append(values, m[k])
+		keys[i] = qualifier + k
+		values = append(values, sanitizedM[k])
 	}
 
 	return keys, values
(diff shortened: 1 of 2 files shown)


core/pkg/util/promutil/promutil.go
```diff
@@ -75,16 +75,22 @@ func LabelNamesFrom(labels map[string]string) []string {
 
 // Prepends a qualifier string to the keys provided in the m map and returns the new keys and values.
 func KubePrependQualifierToLabels(m map[string]string, qualifier string) ([]string, []string) {
-	keys := make([]string, 0, len(m))
-	for k := range m {
+	// sanitize the keys in m to prevent duplicate output keys
+	sanitizedM := make(map[string]string)
+	for k, v := range m {
+		sanitizedM[SanitizeLabelName(k)] = v
+	}
```
thomasvn, line 82: Just checking my understanding. By creating a `sanitizedM` we ensure that we prevent duplicates? So for example if you have the following labels:

```
app-name: myapp
app.name: yourapp
```

Both would be sanitized to `app_name`, and the code would then only keep one of the two values?

nik-kc: Yup, that's exactly right!

thomasvn: Ok thanks! I think it's worth calling out in the PR notes that there is still a limitation here.

Previously having two or more labels that would sanitize to the exact same label name would result in an error serving the `/metrics` endpoint.

Now having two or more labels sanitizing to the exact same label name will result in a single sanitized label, and one of the values associated with that label. (the other values are effectively, dropped).

thomasvn (approved):

core/pkg/util/promutil/promutil.go
nik-kc: Great callout, I'll add that to the PR notes.

nikovacevic (approved):

core/pkg/util/promutil/promutil.go
```diff
@@ -75,16 +75,22 @@ func LabelNamesFrom(labels map[string]string) []string {
 
 // Prepends a qualifier string to the keys provided in the m map and returns the new keys and values.
 func KubePrependQualifierToLabels(m map[string]string, qualifier string) ([]string, []string) {
-	keys := make([]string, 0, len(m))
-	for k := range m {
+	// sanitize the keys in m to prevent duplicate output keys
+	sanitizedM := make(map[string]string)
```
mbolt35, line 79: Possible that we can size this map? `make(map[string]string, len(m))`

PR #3134, opened 2025-04-28
we don't need to test when building the container

ameijer:
## What does this PR change?
* stops running unit tests during container build. We already run unit tests in a separate, required CI step so this just adds build time onto the container

* no

* Faster PR merging 

* Closes ...

## How was this PR tested?
* testing in CI 

* 

*


mbolt35 (approved):
LGTM

PR #3158, opened 2025-05-22
Sth/kcm 3420

Sean-Holcomb:
## What does this PR change?
* This PR adds disaster recovery to the promless data collector

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


modules/collector-source/pkg/scrape/scraper.go
```diff
@@ -1,5 +1,29 @@
...
+			scrapeResults := scrapeFunc()
+			resultCh <- scrapeResults
+		}()
+	}
+
+	var scrapeResults []metric.Update
+	for range scrapeFuncs {
+		targetResults := <-resultCh
+		scrapeResults = append(scrapeResults, targetResults...)
+	}
```
mbolt35, line 27: Do you iterate over `scrapeFuncs` here as a trick to ensure a specific set of iterations? Clever. I have no idea whether or not it's _better_ in go to use a channel based approach like this, or a Mutex protected slice + WaitGroup.

modules/collector-source/pkg/util/interval.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/opencost/opencost/core/pkg/util/timeutil"
 )
 
-var intervalRegex = regexp.MustCompile(`^(\d+)(m|h|d|w)$`)
+var intervalRegex = regexp.MustCompile(`^(\d+)(s|m|h|d|w)$`)
```
mbolt35, line 12: There is a function in our timeutil `ParseDuration` that is basically a one for one with go's `time.ParseDuration` except that it adds support for `w` and `d` I think. 

Honestly, this seems different enough that it doesn't really matter whether we reuse that package, but thought I'd at least mention it.

modules/collector-source/pkg/scrape/scraper.go
Sean-Holcomb: The main goal was to get every target started and scraping as close as possible to the timestamp of the universal scrape clock. Mutex protected slice + WaitGroup probably would have worked just as well

modules/collector-source/pkg/collector/datasource.go
```diff
@@ -41,24 +41,33 @@ func NewCollectorDataSource(
...
+		bucketConfig, err := os.ReadFile(config.BucketConfigFile)
+		if err != nil {
+			log.Errorf("Failed to initialize bucket output storage, please check your configuration and bucket security settings: %s", err)
+		} else {
+			store, err = storage.NewBucketStorage(bucketConfig)
+			if err != nil {
+				log.Errorf("Failed to create bucket storage, please check your configuration and bucket security settings: %s", err)
+			}
+		}
+	}
```
mbolt35, line 57: Unless you think we'll end up storing this data in a bucket other than the main configured bucket, maybe we have this accept a `storage.Storage`? Ah, it's a separate module, so maybe not? Eh

I added a `storage.NewMemoryStorage()` options for testing that let's you pass in an in-memory file system that can be used.

modules/collector-source/pkg/metric/repository.go
```diff
@@ -2,40 +2,122 @@ package metric
...
-		resoluationCollectors[resConf.Interval] = resCollector
+		resoluationCollectors[resolution.Interval()] = resCollector
 	}
 
 	repo := &MetricRepository{
 		resolutionStores: resoluationCollectors,
 	}
 
+	if store != nil {
+		pathFormatter, err := pathing.NewEventStoragePathFormatter("", clusterID, ControllerEventName)
```
mbolt35, line 57: If you get the latest updates from my branch, I changed the path formatters a bit to not automatically include `"federated"` as part of the path. Instead, the path formatters now use `"federated"` as the root directory: 

So if you want to keep this consistently stored with the other files in federated/, you'd need to change this line to:
```
pathing.NewEventStoragePathFormatter("federated", clusterID, ControllerEventName)
```

modules/collector-source/pkg/metric/repository.go
```diff
@@ -2,40 +2,122 @@ package metric
...
+		}
+		encoder := exporter.NewJSONEncoder[UpdateSet]()
+		repo.exporter = exporter.NewEventStorageExporter(
+			pathFormatter,
+			encoder,
+			store,
+		)
+		// attempt to restore state from files
+		// get path of saved files
+		dirPath := path.Dir(pathFormatter.ToFullPath("", time.Time{}, ""))
```
mbolt35, line 70: Feel free to add a `Dir()` to the PathFormatter interface. so you don't have to do this awkward empty call.

modules/collector-source/pkg/collector/datasource.go
Sean-Holcomb: Yeah it was kind of hard to pipe through because it was being created in the emitter. I put it one struct above where it was actually being used so that I could pass in whatever for the test.

modules/collector-source/pkg/metric/repository.go
Sean-Holcomb: Haha, yeah

modules/collector-source/pkg/metric/repository.go
Sean-Holcomb: I kind of think federated shouldn't be included, mainly for opencosts sake

modules/collector-source/pkg/collector/datasource.go
mbolt35: Yeah, this makes sense -- we're going to need to refactor some of the plumbing in the agent to better support this type of stuff, but totally with you here.

modules/collector-source/pkg/metric/repository.go
mbolt35: Yeah, the tricky part is actually configuring it all properly to not use federated -- diagnostics, heartbeat, etc... are all using it, but it's just keep it consistent with the old exporter directory structure. Maybe we can figure out some type of global configuration for root directories... I'll look more into it.

mbolt35 (approved):
This all looks great to me man. I'm going to work on a few of the smaller one off changes for the path formatter stuff to maybe cleanup a bit, but everything here looks in order to me. Let it rip :)

PR #3162, opened 2025-05-27
[feat]: emit carbon cost as metrics

maxvishy:
* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/costmodel/handlers.go
+++ b/pkg/costmodel/handlers.go
@@ -3,7 +3,6 @@ package costmodel
 import (
 	"fmt"
 	"net/http"
-	"strings"
 
 	"github.com/julienschmidt/httprouter"
 	assetfilter "github.com/opencost/opencost/core/pkg/filter/asset"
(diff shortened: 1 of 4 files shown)


pkg/costmodel/metrics.go
```diff
@@ -273,6 +274,17 @@ func initCostModelMetrics(clusterCache clustercache.ClusterCache, provider model
...
+			prometheus.GaugeOpts{
+				Name: "opencost_carbon_cost",
+				Help: "Carbon cost associated with resource usage",
+			},
+			[]string{"namespace", "pod", "cluster"},
+		)
+		if _, disabled := disabledMetrics["opencost_carbon_cost"]; !disabled {
+			toRegisterGV = append(toRegisterGV, CarbonCostRecorder)
+		}
+
```
ameijer, line 288: unit tests on all this, please!

ameijer (changes requested):

pkg/costmodel/metrics.go
maxvishy: Added .

pkg/costmodel/handlers.go
```diff
@@ -62,6 +63,18 @@ func (a *Accesses) ComputeAssetsCarbonHandler(w http.ResponseWriter, r *http.Req
...
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Error computing carbon estimates: %s", err), http.StatusInternalServerError)
+		return
+	}
+
+	for key, carbonRow := range carbonEstimates {
+		parts := strings.Split(key, "/")
+		if len(parts) == 2 {
+			namespace, pod := parts[0], parts[1]
+			a.MetricsEmitter.EmitCarbonCost(namespace, pod, "", carbonRow.Co2e)
```
ameijer: won't this only update the metrics when the handler is called?

ameijer: also please update PR template and link to issue

maxvishy: > won't this only update the metrics when the handler is called?

Yeah, only emit when the handler is called.

ameijer: @maxvishy can we make this run on the regular emitter loop? it isn't useful if it only updates when the handler is called

maxvishy: Ok I will add

PR #3180, opened 2025-06-05
Sth/kcm 4204

Sean-Holcomb:
## What does this PR change?
This PR addresses this panic
```
panic: runtime error: index out of range [0] with length 0

goroutine 15747 [running]:
github.com/opencost/opencost/pkg/costmodel.applyRAMBytesUsedMax(0xc00909a7b8, {0xc01eba6c88, 0x18d, 0x1?}, 0xc00909a2d0)
	/app/opencost/pkg/costmodel/allocation_helpers.go:606 +0x519
```
This is seems to happen when a metric for a label set returns empty. This should never happen because metric are initialized when a value for their label set appears

* 

* 

* Closes ...

## How was this PR tested?
* Update unit tests

* 

*

### DIFF
--- a/modules/collector-source/pkg/collector/datasource.go
+++ b/modules/collector-source/pkg/collector/datasource.go
@@ -94,6 +94,7 @@ func NewCollectorDataSource(
 	clusterMap := newCollectorClusterMap(clusterInfo)
 
 	return &collectorDataSource{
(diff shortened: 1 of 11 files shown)


mbolt35 (approved):
LGTM!

PR #3186, opened 2025-06-06
KCM-4123 - InstallInfo Null

nik-kc:
## What does this PR change?
* Explicitly initializes a slice in a function called by `/installInfo` such that, in the event that no pods/containers are found, we don't send a `null` value in the response (see below).

* Nope.

* The settings page should no longer crash when no pods or containers are found.

* Closes [KCM-4123](https://apptio.atlassian.net/browse/KCM-4123).

## How was this PR tested?
* Local testing against `/installInfo`.

* Nope.

* 

## Before
```
{
    "containers": null,
    "clusterInfo": {},
    "version": "dev (HEAD)"
}
```

## Now
```
{
    "containers": [],
    "clusterInfo": {},
    "version": "dev (HEAD)"
}
```


nikovacevic (approved):

nikovacevic:
@cliffcolvin recommend we bypass SQ for this one

PR #3192, opened 2025-06-09
Remove Incomplete/Broken Thanos Support 

mbolt35:
## What does this PR change?
* Drops all incomplete implementation of Thanos


Sean-Holcomb (approved):

philipp1992:
so opencost doesnt work with thanos? Ive installed the latest helm chart version 2.4 and set all the thanos parameters, but its still trying to connect to the default prometheus service and ignoring the thanos query completly.

PR #3194, opened 2025-06-10
fix issue where node cost not getting reported properly

ameijer:
## What does this PR change?
* adds some more part numbers, including defaults, to fix OCP pricing 

* no . @anders-swanson , you had originally contributed this code and price list. The issue is, we don't consider nodes with 0 memory or CPU cost to be real nodes. Therefore, we have big negative idle costs on nodes where you didn't provide memory/CPU part numbers. In most cases, I was able to find direct matches. However, for some nodes, there is no part number for the ram? so I picked the X9 memory part number to serve as a default: `"Memory": "B94177",`. Do you think this approach is defensible? 

* provides more accurate prices for oracle instances 

* Closes ...

## How was this PR tested?
* 

* 

*


mbolt35 (approved):
LGTM!

ameijer:
@anders-swanson going to merge this to unblock integration testing, but eager to discuss this further and PR changes as needed

anders-swanson:
I can take a look at this, maybe re-compiling the price list data will be help. I think the issue is, OCI calculates cost for some node types by specific metrics, hence why "memory" might not have a part number. For example, at the time I wrote this, the cost of a GPUs instance was based on how many GPUs that instance had.

anders-swanson:
A node with 0 cpu/memory costs may be a GPU node, so i am wondering if that might need to change to accommodate this?

ameijer:
@anders-swanson thank you! interesting. So, I did a quick look here just to explore more about what you are saying. Let's take a concrete example:

A `VM.GPU.A10.1` VM has 1 GPU and has a GPU cost of 2.00 an hour according to the pricing calculator. that instance comes with 15 OCPU and 240 gb of ram. Using the same pricing calculator, I see that a VM.Standard3.Flex with 15 OCPU and 240 GB ram costs almost $1 an hour - for a total of $725 a month.

You're saying that those 15 OCPU and 240 gb of ram are included in the 2.00 an hour GPU price? so the GPU itself costs closter to $1, and the implied cost of the CPU and RAM on the GPU instance make up the other ~$1? I have asked the folks @ CNCF for billing data to look into this further, as we are running a  `VM.GPU.A10.1` for our demo @ demo.infra.opencost.io

PR #3199, opened 2025-06-17
fix: apply filter before aggregation in ComputeAllocationHandler

### LINKED ISSUE #3198: We should filter the result before doing aggregation
**Describe the bug**
Right now, we filter records only after aggregating them, which can lead to some missing records in the final result. Ideally, we should filter the result first and then do the aggregation on top of that.

We need to fix the method [ComputeAllocationHandler](https://github.com/opencost/opencost/blob/6af6acce7258abbcf795c538e207e616dd98945d/pkg/costmodel/aggregation.go#L170)
The correct approach can be found in [ComputeAllocationHandlerSummary](https://github.com/opencost/opencost/blob/6af6acce7258abbcf795c538e207e616dd98945d/pkg/costmodel/aggregation.go#L60)

**To Reproduce**
Steps to reproduce the behaviour:
1. Try `/model/allocation` endpoint with both filter and aggregation. You may see some missing records

**Which version of OpenCost are you using?** : `1.114.0`

abhinav-1305:
### What does this PR change?

Refactors `ComputeAllocationHandler` to apply filtering before aggregation, ensuring that all records are included in the final result.

### Does this PR relate to any other PRs?

Fixes #3198

### DIFF
--- a/pkg/costmodel/aggregation.go
+++ b/pkg/costmodel/aggregation.go
@@ -177,6 +177,7 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	window, err := opencost.ParseWindowWithOffset(qp.Get("window", ""), env.GetParsedUTCOffset())
 	if err != nil {
 		http.Error(w, fmt.Sprintf("Invalid 'window' parameter: %s", err), http.StatusBadRequest)
+		return
 	}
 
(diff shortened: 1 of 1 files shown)


abhinav-1305:
please review - @Sean-Holcomb

ameijer:
cclosing as dup of #3394

PR #3202, opened 2025-06-17
feat: allow Azure service principal configuration via environment variables

### LINKED ISSUE #3156: Move non secret values of service key out of secret or allow them to be overridden through environment variables.
When deploying opencost to multiple cluster the azure integration becomes a pain point.

https://opencost.io/docs/configuration/azure/#supply-azure-service-principal-details-to-opencost
Here you are instructed to provide a secret:
```
{
    "subscriptionId": "<Azure Subscription ID>",
    "serviceKey": {
        "appId": "<Azure AD App ID>",
        "displayName": "OpenCostAccess",
        "password": "<Azure AD Client Secret>",
        "tenant": "<Azure AD Tenant ID>"
    }
}
```

Problem with this secret is that only 1 value is actually a secret ```"<Azure AD Client Secret>"``` the rest are configuration values that could just as well be injected through environment variables, provided in a config map or something else. 

One value is especially troublesome ```"<Azure Subscription ID>"``` it's not spelled out explicitly but since app registrations aka Azure App ID's are not bound to a subscription in the tenant, I assume that this subscription id is used to query for cluster resources and can be different for each cluster depending on setup.

Non secret values should not be part of a secret it makes the rollout of opencost cumbersome, we now need to either maintain a keyvault with a secret pr cluster (not scalable), or template this secret in multiple steps (using external secrets and our own templating solution, also not fun), and common for both of these "solutions" is that the opencost service-key will be different depending on the cluster. (we like keeping cluster very similar as it makes them easier to manage and troubleshoot)

Configuration values like subscription id, app id / client id, tenant, can all be provided through the helm chart, the easiest thing would probably be by using a config map. The secret can then either be injected as an environment variable or mounted as it is today. 

While the problem is specific to azure a similar approach could be reused for GCP, AWS.

**Additional context**
An alternate solution in this case could be to allow subscriptionId to be emitted from service key if it is provided in an environment variable. 

Since everything under serviceKey is static 
```
    "serviceKey": {
        "appId": "<Azure AD App ID>",
        "displayName": "OpenCostAccess",
        "password": "<Azure AD Client Secret>",
        "tenant": "<Azure AD Tenant ID>"
    }
```

the inclusion of subscriptionId is the only part that really hurts when doing rollout to multiple clusters.

abhinav-1305:
resolves: #3156, PTAL - @ameijer / @Sean-Holcomb.

### DIFF
--- a/pkg/env/costmodelenv_test.go
+++ b/pkg/env/costmodelenv_test.go
@@ -3,6 +3,8 @@ package env
 import (
 	"os"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestGetAPIPort(t *testing.T) {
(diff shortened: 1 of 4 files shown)


pkg/env/costmodelenv_test.go
```diff
@@ -234,3 +236,59 @@ func TestEnvVarsWithBackup(t *testing.T) {
 	})
 
 }
+
+func TestAzureEnvironmentVariables(t *testing.T) {
```
Sean-Holcomb: This test feels like filler. We don't need a special test case for these alias functions. If you want to do a test with test cases that tests env.Get() that would be fine.

Sean-Holcomb (commented):
This PR doesn't actually seem to address https://github.com/opencost/opencost/issues/3156 as it still requires you to pick an auth approach and stick with it. 

I left a comment on the issue, but I have stated that I am against the mixing of configuration techniques, in addition to being having an unfavorable opinion ENV variable configuration for this type of configuration.

PR #3203, opened 2025-06-17
fix: standardize node identification to prevent duplicate metrics

### LINKED ISSUE #3113: OpenCost on AKS is treating a single node as two separate nodes
**Describe the bug**
OpenCost is treating a single node as two separate ones because it's seeing two different identifiers for the same node — one by IP:port (10.10.130.223:10250) and the other by hostname (aks-xxxx-97016666-vmss000005).

**To Reproduce**
1. Install OpenCost via Helm Chart https://opencost.io/docs/installation/helm#installing-opencost-with-helm
2. Setup Azure integration https://opencost.io/docs/configuration/azure
3. Go to prometheus and run query `avg_over_time(container_cpu_allocation{container="prometheus"}[1d]`
4. See duplicated metrics with different values for the same node

```
{container="prometheus", instance="10.10.130.223:10250", job="opencost", namespace="observability", node="10.10.130.223:10250", pod="prometheus-kube-prometheus-stack-prometheus-0"} | 0.21294522547845512
-- | --
{container="prometheus", instance="aks-xxxx-97016666-vmss000005", job="opencost", namespace="observability", node="aks-xxxx-97016666-vmss000005", pod="prometheus-kube-prometheus-stack-prometheus-0"} | 0.5
```

**Expected behavior**
In result should be only one metric

**Screenshots**
-

OpenCost version: 1.114.0 (dcbd21f)

**Additional context**
I am using existing Prometheus, but I don't think there is an issues with it. When I fetch the metrics directly from opencost container (/metrics endpoint) I am getting same result.

I also see mentioning of both nodes "types" in the logs

```
No pricing data found for node `10.10.128.100:10250` , using custom pricing
....
Node `aks-xxxx-97016666-vmss00002a` in prometheus but not k8s api
...
DBG Node Name: aks-xxxxx-10180546-vmss00052s
DBG Blended CPU Discount: 0.000000
DBG Blended RAM Discount: 0.000000
...
Node Name: 10.10.128.150:10250
Blended CPU Discount: 0.000000
DBG Blended RAM Discount: 0.000000
```

abhinav-1305:
resolves: #3113 

It seems like the issue stems from inconsistent use of `instance` and `node` labels across the codebase, which likely caused confusion during metric aggregation and Prometheus queries.

To fix this, I standardized node identification by:
- Removing the `instance` label from metric definitions
- Grouping Prometheus queries consistently by `node`
- Refactoring metric recording to rely solely on `node`

For example:

```go
// Before
cmme.NodeSpotRecorder.WithLabelValues(nodeName, nodeName, nodeType, nodeRegion, node.ProviderID, node.ArchType)

// After
cmme.NodeSpotRecorder.WithLabelValues(nodeName, nodeType, nodeRegion, node.ProviderID, node.ArchType)

### DIFF
--- a/modules/collector-source/pkg/collector/collector.go
+++ b/modules/collector-source/pkg/collector/collector.go
@@ -288,7 +288,7 @@ func NewLocalStorageBytesMetricCollector() *metric.MetricCollector {
 		metric.LocalStorageBytesID,
 		scrape.NodeFSCapacityBytes,
 		[]string{
-			source.InstanceLabel,
+			source.NodeLabel,
 			source.DeviceLabel,
 		},
 		aggregator.AverageOverTime,
@@ -492,7 +492,7 @@ func NewNodeRAMSystemUsageAverageMetricCollector() *metric.MetricCollector {
 		metric.NodeRAMSystemUsageAverageID,
 		scrape.ContainerMemoryWorkingSetBytes,
 		[]string{
-			source.InstanceLabel,
+			source.NodeLabel,
 		},
 		aggregator.AverageOverTime,
 		func(labels map[string]string) bool {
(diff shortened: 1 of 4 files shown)


abhinav-1305:
please review - @mbolt35 / @ameijer.

PR #3206, opened 2025-06-17
Deep Copy Labels

nik-kc:
## What does this PR change?
* Performs a deep clone of pod labels such that [this line](https://github.com/opencost/opencost/blob/68279ce27b84606059a32959c8fbc835cb1f03f1/pkg/costmodel/costmodel.go#L396) can modify the labels, while allowing other goroutines to read them.

* Nope.

* No more concurrent map read/write panics when using cluster cache V2.

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -3,6 +3,7 @@ package costmodel
 import (
 	"errors"
 	"fmt"
+	"maps"
 	"math"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mbolt35 (approved):
Long term, this isn't exactly the solution we want, but I'll accept it on a temporary basis. For V1 cache, this will double copy. For V2, we need a safer implementation on GetAll

PR #3210, opened 2025-06-19
Add nil check for clustercache labels in node cost

kaelanspatel:
## What does this PR change?
Fixes nil panic for nodes returned from clustercache with no labels.

No.

Nodes which somehow have nil labels will not cause a panic when attempting to get cost based on labels.

## How was this PR tested?
Tested locally with mock data.

No.

Likely will be patched back to 2.7.


pkg/costmodel/costmodel.go
```diff
@@ -867,6 +867,11 @@ func (cm *CostModel) GetNodeCost() (map[string]*costAnalyzerCloud.Node, error) {
 	for _, n := range nodeList {
 		name := n.Name
 		nodeLabels := n.Labels
+		if nodeLabels == nil {
+			log.Warnf("Got node '%s' with no labels", name)
```
thomasvn:
```suggestion
			log.Debugf("GetNodeCost: Found node '%s' with no labels", name)
```

Just to keep logs clean, proposing we keep this at Debug? We can surface to Warn/Error if we believe there are downstream consequences of no labels?

pkg/costmodel/costmodel.go
```diff
@@ -867,6 +867,11 @@ func (cm *CostModel) GetNodeCost() (map[string]*costAnalyzerCloud.Node, error) {
 	for _, n := range nodeList {
 		name := n.Name
 		nodeLabels := n.Labels
+		if nodeLabels == nil {
+			log.Debugf("GetNodeCost: Found node '%s' with no labels", name)
```
ameijer: I suggest you make this a warn level logs

kaelanspatel: Done!

ameijer (approved):

PR #3218, opened 2025-06-23
Cluster Storage

Sean-Holcomb:
## What does this PR change?
*  This PR adds a storage implementation that connects to cluster storage

* 

* 

* Closes ...

## How was this PR tested?
* Tested using the integration test found in bucketstorage_test.go against the cluster storage service implementation. Additional I have exported this testing suite so that it can be run in KCM as a "unit test"

* 

*

### DIFF
--- a/core/pkg/storage/clusterstorage.go
+++ b/core/pkg/storage/clusterstorage.go
@@ -1,0 +1,290 @@
+package storage
+
+import (
+	"bytes"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/util/json"
+	"gopkg.in/yaml.v2"
+)
+
+var defaultClusterConfig = ClusterConfig{}
+
+// ClusterStorage is a Storage implementation which connects to a remote file storage over http
+type ClusterStorage struct {
+	client *http.Client
+	host   string
+	port   int
+}
+
+type ClusterConfig struct {
+	Host string `yaml:"host"`
+	Port int    `yaml:"port"`
+}
+
+// parseConfig unmarshals a buffer into a Config with default HTTPConfig values.
+func parseClusterConfig(conf []byte) (ClusterConfig, error) {
+	config := defaultClusterConfig
+	if err := yaml.Unmarshal(conf, &config); err != nil {
+		return ClusterConfig{}, err
+	}
+
+	return config, nil
+}
+
+func NewClusterStorage(conf []byte) (*ClusterStorage, error) {
+	config, err := parseClusterConfig(conf)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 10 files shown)


core/pkg/storage/clusterstorage.go
```diff
@@ -0,0 +1,290 @@
...
+		host = config.Host
+	}
+	port := 9006
+	if config.Port != 0 {
+		port = config.Port
+	}
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
```
mbolt35: We'll likely want to add configuration for TLS certs, and probably a way to auth?

mbolt35 (approved):
LGTM!

core/pkg/storage/clusterstorage.go
```diff
@@ -0,0 +1,345 @@
...
+	if err != nil {
+		return fmt.Errorf("request failed: %w", err)
+	}
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
+	if !(resp.StatusCode >= 200 && resp.StatusCode <= 299) {
+		if resp.Body != nil {
+			var errResp Response[*struct{}]
```
mbolt35: Does
```go
var errResp Response[any]
``` 

work here?

core/pkg/storage/clusterstorage.go
```diff
@@ -0,0 +1,345 @@
...
+		err := json.NewDecoder(resp.Body).Decode(&jsonResp)
+		if err != nil {
+			return fmt.Errorf("failed to decode json: %w", err)
+		}
+		return nil
+	}
+
+	err := c.makeRequest(
+		http.MethodGet,
+		fmt.Sprintf("%s://%s:%d/clusterStorage/fullPath?path=%s", c.scheme(), c.host, c.port, path),
```
mbolt35: You may want to do one of two things here, but totally optional because it _will_ get caught at some point in the request process no matter what I'm pretty sure. 

You can build out the URL via the obnoxious way: 
```go
u := new(url.URL)
u.Scheme = c.scheme()
u.Host = net.JoinHostPort(c.host, fmt.Sprintf("%d", c.port))
u = u.JoinPath("clusterStorage", "fullPath")

q := make(url.Values)
q.Set("path", path)

rawQuery, _ := url.QueryUnescape(q.Encode())
u.RawQuery = rawQuery

u.String() // <-- full URL string 
```

Or, just parse the string before calling:
```go
u, err := url.Parse(...)
```

Pretty sure that `http.NewRequest()` just calls `Parse()` on the url anyways and forwards the result. Potentially could add this for debugging potential formatting weirdness on query params? Just a thought :)

mbolt35: Sorry, the main point in all this was to at least incorporate:
```go 
net.JoinHostPort(c.host, fmt.Sprintf("%d", c.port))
```

which is a newer API that supports IPV6 hosts

PR #3221, opened 2025-06-26
KCM-4229: Ingesting Collector Diagnostics to bucket

mittal-ishaan:
## What does this PR change?
* Adds diagnostics for Promless(Collector) source that will get ingested to the export bucket.
Right now we are adding few bare minimum diagnostics which could be improved later on.

* No

* Users will be able to see diagnostics data for collector source in their export bucket

* Closes ... https://apptio.atlassian.net/browse/KCM-4229

## How was this PR tested?
* Ran unified agent locally with these changes and confirmed the diagnostics data being ingested in my export bucket(in my case gcp cloud storage)
diagnostics json present at `<cluster>/diagnostics/<application>/<file>` path in the bucket:
```json
{
    "startTime": "2025-06-30T19:33:33Z",
    "results": [
        {
            "id": "0197c254-981e-7aae-bee6-707a56b7fe75",
            "name": "kubecostMetric",
            "description": "Determine if kubecost metrics for a node are being scraped",
            "category": "collector",
            "timestamp": "2025-06-30T19:33:33.342724Z",
            "details": {
                "docLink": "",
                "label": "Kubecost metrics for a node are being scraped",
                "passed": true,
                "query": "node_total_hourly_cost",
                "result": {
                    "2025-06-30 19:33:00 +0000 UTC": 0.17681545601272583
                }
            }
        },
        {
            "id": "0197c254-981e-7ab2-9b55-36f80ada3bd6",
            "name": "nodesCPUMetrics",
            "description": "Determine if the node CPU cores capacity is being scraped",
            "category": "collector",
            "timestamp": "2025-06-30T19:33:33.34274Z",
            "details": {
                "docLink": "",
                "label": "Node CPU cores capacity is being scraped",
                "passed": true,
                "query": "kube_node_status_capacity_cpu_cores",
                "result": {
                    "2025-06-30 19:33:00 +0000 UTC": 4
                }
            }
        }
    ]
}
```
The format is the same as what we receive in case of prometheus env.

* NA

*


modules/collector-source/pkg/collector/datasource.go
```diff
@@ -106,7 +107,32 @@ func (c *collectorDataSource) RegisterEndPoints(router *httprouter.Router) {
 }
 
 func (c *collectorDataSource) RegisterDiagnostics(diagService diagnostics.DiagnosticService) {
-	return
+	const CollectorDiagnosticCategory = "collector"
+
+	metricStore := c.metricsQuerier.collectorProvider.GetStore(time.Now().Add(-time.Minute * 1), time.Now())
```
mittal-ishaan: Diagnostics are sent every minute so trying to check if the required metrics are present in the last minute. But unsure if GetStore works this way. @Sean-Holcomb can you help confirm this?

avrodrigues5: do we need nil checks on metricQuerier then on collectorProvider ? 

overall this is a good start to continue working on MCD 3.0 ingestion and API. Good job @mittal-ishaan !

mittal-ishaan: When creating a new collector data source metricsQuerier and collectorProvider are defined where they are needed to be set. So did not do nil checks for them

Sean-Holcomb: This will return something but just know that it will be the current window of the smallest resolution that is running. Likely 10m

mittal-ishaan: Thanks 👍

modules/collector-source/pkg/collector/diagnostics.go
```diff
@@ -0,0 +1,78 @@
...
+		Label:       "Kubecost metrics for a node are being scraped",
+		Description: "Determine if kubecost metrics for a node are being scraped",
+		QueryFmt:    "avg(node_total_hourly_cost{}[1m]) by (node, cluster_id, provider_id)",
+	},
+}
+
+type diagnosticDefinition struct {
+	ID          string
+	MetricsId   metric.MetricCollectorID
+	QueryFmt    string
```
Sean-Holcomb: QueryFmt doesn't really make sense in this context. you could print out the type of the aggregator being used and the metric name that is aggregated.

modules/collector-source/pkg/collector/diagnostics.go
```diff
@@ -0,0 +1,78 @@
+package collector
+
+import (
+
+	"github.com/opencost/opencost/core/pkg/source"
+	"github.com/opencost/opencost/modules/collector-source/pkg/metric"
+)
+
+// Collector Metric Diagnostic IDs
+const (
```
Sean-Holcomb: Do we really need seperate diagnostic ids if there are going to go one-to-one with metrics which already have their own ids?

Sean-Holcomb (commented):
My thought here is that you are setting yourself up to redefine every metric a second time to have a diagnostic on it when you could really be adding a couple fields (really just the description) to the original registration in https://github.com/opencost/opencost/blob/e97b8290246d5e2a202cfa5fe93ea69bf64cdf46/modules/collector-source/pkg/collector/collector.go#L100 and run diagnostics off of that with a separate function on the store.

modules/collector-source/pkg/collector/diagnostics.go
mittal-ishaan: The diagnostics Id goes on to become the diagnostics name stored in the bucket. The metrics name differs from the diagnostics name I want it to have. Thats why thought of separate diagnostic ids

mittal-ishaan: Similar thought with adding description to the collectors is that the diagnostics description would be different from what description we would add to the collectors

modules/collector-source/pkg/metric/diagnostics.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// NodesDiagnosticMetricID is the identifier for the query used to determine if the node CPU cores capacity is being scraped
+	NodesDiagnosticMetricID = "nodesCPUMetrics"
+)
+
+// diagnostic definitions mapping holds all of the diagnostic definitions that can be used for collector metrics diagnostics
+var diagnosticDefinitions map[string]*diagnosticDefinition = map[string]*diagnosticDefinition{
+	NodesDiagnosticMetricID: {
+		ID:          NodesDiagnosticMetricID,
+		MetricName:  "kube_node_status_capacity_cpu_cores",
```
Sean-Holcomb: This is defined as a const elsewhere

modules/collector-source/pkg/metric/diagnostics.go
```diff
@@ -0,0 +1,130 @@
...
+		Label:       "Node CPU cores capacity is being scraped",
+		Description: "Determine if the node CPU cores capacity is being scraped",
+	},
+
+	KubecostDiagnosticMetricID: {
+		ID:          KubecostDiagnosticMetricID,
+		MetricName:  "node_total_hourly_cost",
```
Sean-Holcomb: this is also defined as a const elsewhere

modules/collector-source/pkg/metric/diagnostics.go
```diff
@@ -0,0 +1,130 @@
...
+	// Create a deep copy for the async update to avoid race condition
+	updateSetCopy := &UpdateSet{
+		Timestamp: updateSet.Timestamp,
+		Updates:   make([]Update, len(updateSet.Updates)),
+	}
+	copy(updateSetCopy.Updates, updateSet.Updates)
+
+	// Run update asynchronously
+	go func() {
+		d.updater.Update(updateSetCopy)
```
Sean-Holcomb: I don't think it will really matter, but instead the diagnostics module functionality should be run async rather than the lower level updater functionality. The reason for this is that it is required that updates to the repo are executed in order. If they are executed async this is no longer guaranteed.

modules/collector-source/pkg/metric/diagnostics.go
mittal-ishaan: these constants are present in scrape package. Importing scrape package is adding an import cycle in metrics

modules/collector-source/pkg/metric/diagnostics.go
mittal-ishaan: understood 👍 
Thank you

modules/collector-source/pkg/metric/diagnostics.go
Sean-Holcomb: I would say that would be a good reason to move them out of that package. They could be all defined in one file with comments on which scraper provides them

mittal-ishaan: sure sure

modules/collector-source/pkg/metric/walinator.go
```diff
@@ -32,14 +32,14 @@ type Walinator struct {
 	paths           pathing.StoragePathFormatter[time.Time]
 	exporter        exporter.EventExporter[UpdateSet]
 	limitResolution *util.Resolution
-	repo            *MetricRepository
```
Sean-Holcomb, line 35: since we are changing this to an Updater can you change the arg and property names from repo

modules/collector-source/pkg/constants/metrics.go
Sean-Holcomb, line 1: I would just put this one in the metric package since that will avoid dependency issues without having to add a whole new package

Sean-Holcomb (approved):
Looks good, just a couple more optional style comments

modules/collector-source/pkg/constants/metrics.go
mittal-ishaan: sure

PR #3222, opened 2025-06-30
KCM-4268: Add log level to heartbeat data

mittal-ishaan:
Signed-off-by: Ishaan Mittal <ishaanmittal123@gmail.com>

## What does this PR change?
* Adds logLevel information in the heartbeat data

* No

* Heartbeat data shown in diagnostics page would also be able to see the log level set for the agent

* Closes ...

## How was this PR tested?
* Ran the agent locally and confirmed the heartbeat data posted in export bucket(gcp cloud storage in my case)

* No

*


core/pkg/heartbeat/exporter/metadataprovider.go
```diff
@@ -0,0 +1,85 @@
...
+
+	for k, v := range m {
+		metadata[k] = v
+	}
+
+	return metadata
+}
+
+// LogLevelMetadataProvider is a `HeartbeatMetadataProvider` implementation that provides the log level.
+type LogLevelMetadataProvider struct {
```
avrodrigues5: Can you think of a name that's more general? Today its LogLevel but in future it could be something else. Perhaps AgentInfo. Not great with names @mbolt35 any recommendation here?

avrodrigues5: I think you got this thou!

avrodrigues5: ok i did see the [conversation](https://kubecost.slack.com/archives/C08ESKMCFA4/p1751041321071199.) after ishaan pinged me the PR, so idea is every new metadataprovider has its own struct My question is invalidated then!

mittal-ishaan: Thank you 🙌

mbolt35: Yeah, this is generally how I wanted to solve the problem! Naming looks fine.

core/pkg/heartbeat/exporter/metadataprovider.go
```diff
@@ -0,0 +1,85 @@
...
+	logLevel string
+}
+
+// NewLogLevelMetadataProvider creates a new `LogLevelMetadataProvider` instance.
+func NewLogLevelMetadataProvider(logLevel string) *LogLevelMetadataProvider {
```
mbolt35: I would drop the `logLevel` as a parameter:
* We have endpoints that allow changing the log level real-time. This value would become invalidated if that happened

Totally fine if the `LogLevelMetadataProvider` remains stateless and just returns `log.GetLogLevel()` in the `GetMetadata()` call.

core/pkg/heartbeat/exporter/metadataprovider.go
```diff
@@ -0,0 +1,85 @@
...
+	return &LogLevelMetadataProvider{
+		logLevel: logLevel,
+	}
+}
+
+// GetMetadata returns the metadata for new heartbeat instances. It uses the `LogLevelMetadataProvider` 
+// to get the log level and injects it into the metadata map.
+func (l *LogLevelMetadataProvider) GetMetadata() map[string]any {
+	return map[string]any{
+		"logLevel": l.logLevel,
```
mbolt35: Probably best to use `log.GetLogLevel()` instead of `l.logLevel`

core/pkg/heartbeat/exporter/metadataprovider.go
```diff
@@ -0,0 +1,85 @@
...
+type MultiMetadataProvider struct {
+	providers []HeartbeatMetadataProvider
+}
+
+// NewMultiMetadataProvider creates a new `MultiMetadataProvider` instance.
+func NewMultiMetadataProvider(providers ...HeartbeatMetadataProvider) *MultiMetadataProvider {
+	return &MultiMetadataProvider{
+		providers: providers,
+	}
+}
```
mbolt35, line 69: Love it!

mbolt35 (approved):
Let me know your thoughts about the `log.GetLogLevel()` changes. Once those changes are made, I think you're free to merge.

core/pkg/heartbeat/exporter/metadataprovider.go
mittal-ishaan: Ohh yes. Agree to now passing logLevel to it then. we should be getting it in real time. updating it

mittal-ishaan:
Thanks @mbolt35 updated the PR with `log.GetLogLevel()` change. Might also need help with merging the PR too, I do not have write access permissions.

Issue #3231, opened 2025-07-14
Implement Opencost provider for DigitalOcean


prateek1192:
**Is your feature request related to a problem? Please describe.**
We want to add the code for Digital Ocean Kubernetes Services to Opencost.

**Describe the solution you'd like**
I would implement the provider for it.

**Describe alternatives you've considered**
This is something that is missing in Opencost right now.

**Additional context**
DOKS is a popular Managed Kubernetes service and it would be great if Opencost supports it

dmouse:
Hi, I'm running opencost:1.117.6 and I’m seeing some issues fetching the `v2/billing/pricing` endpoint.

```
10/22/2025 6:23:23 PM | 2025-10-23T00:23:23.378894299Z INF Fetching DigitalOcean pricing from: https://api.digitalocean.com/v2/billing/pricing
10/22/2025 6:23:23 PM | 2025-10-23T00:23:23.493026709Z WRN Pricing API returned unexpected status: 404
10/22/2025 6:23:23 PM | 2025-10-23T00:23:23.493163Z WRN Failed to fetch PV pricing data: pricing API returned status: 404, using fallback
```

Unfortunately I couldn't find the right endpoint for the new version of the API, I'm going to open a ticket with the digitalocean support team.

File related: https://github.com/opencost/opencost/blob/503c0cdea1871dcd5c2812f17daafd3c570a71d8/pkg/env/costmodel.go#L369

prateek1192:
@dmouse  : This endpoint does not exist right now and would be created after the GA. If you install Opencost on DO, you can still use it with the fallback pricing. Please feel free to open a support ticket with DO

PR #3232, opened 2025-07-14
Add DigitalOcean Kubernetes Services provider

prateek1192:
This commit adds functionality to read prices for DigitalOcean Kubernetes Services(DOKS) from a pricing URL and if the URL is not available, fall back to static pricing and calculate costs using it.

## What does this PR change?

This PR adds code to fetch the prices from a pricing URL for DOKS and if the URL is not available, fall back to static pricing. 

No

This PR would allow users to use Opencost with DOKS and see the prices of their CPU, RAM, Persistent Volume Claims and Load balancers

No

## How was this PR tested?
This PR was tested by deploying Opencost on DOKS and also looking at allocation and assets APIs. Prometheus was deployed along with kube state metrics and node exporter and then we used port forwarding to reach the opencost svc. We saw the following metrics in the url http://localhost:9091/allocation?window=7d&agg=pod&acc=false

image

No 

I have created a new issue for the PR
https://github.com/opencost/opencost/issues/3231 
and would see how to add labels here

### DIFF
--- a/configs/digitalocean.json
+++ b/configs/digitalocean.json
@@ -1,5 +1,4 @@
 {
-  "doPricingURL": "https://api.digitalocean.com/v2/billing/pricing",
   "zoneNetworkEgress": "0.00",
   "regionNetworkEgress": "0.00",
   "internetNetworkEgress": "0.01"
--- a/configs/digitalocean.json
+++ b/configs/digitalocean.json
@@ -1,0 +1,6 @@
+{
+  "doPricingURL": "https://api.digitalocean.com/v2/billing/pricing",
+  "zoneNetworkEgress": "0.00",
+  "regionNetworkEgress": "0.00",
+  "internetNetworkEgress": "0.01"
+}
--- a/pkg/cloud/digitalocean/provider.go
+++ b/pkg/cloud/digitalocean/provider.go
@@ -233,13 +233,44 @@ type SlugBase struct {
 	BaseRAMGiB int
 }
 
-// Base slugs and known pricing
-var baseSlugs = map[string]SlugBase{
-	"c": {BaseSlug: "c-4-intel", BaseCost: 0.1622, BaseVCPU: 4, BaseRAMGiB: 8},
-	"s": {BaseSlug: "s-4vcpu-8gb", BaseCost: 0.07143, BaseVCPU: 4, BaseRAMGiB: 8},
-	"m": {BaseSlug: "m-8vcpu-64gb", BaseCost: 0.50, BaseVCPU: 8, BaseRAMGiB: 64},
-	"g": {BaseSlug: "g-4vcpu-16gb-intel", BaseCost: 0.2247, BaseVCPU: 4, BaseRAMGiB: 16},
+type slugSeeds struct {
+	BaseVCPU    int
+	BaseHourly  float64
+	RamPerVCPU  int
+	IntelHourly float64
 }
+
+var slugFamilySeed = map[string]slugSeeds{
+	"c":     {BaseVCPU: 4, BaseHourly: 0.12500, RamPerVCPU: 2, IntelHourly: 0.16220},
+	"c2":    {BaseVCPU: 4, BaseHourly: 0.13988, RamPerVCPU: 2, IntelHourly: 0.18155},
+	"g":     {BaseVCPU: 4, BaseHourly: 0.18750, RamPerVCPU: 4, IntelHourly: 0.22470},
+	"gd":    {BaseVCPU: 4, BaseHourly: 0.20238, RamPerVCPU: 4, IntelHourly: 0.23512},
+	"m":     {BaseVCPU: 8, BaseHourly: 0.50000, RamPerVCPU: 8, IntelHourly: 0.58929},
+	"m3":    {BaseVCPU: 8, BaseHourly: 0.61905, RamPerVCPU: 8, IntelHourly: 0.65476},
+	"m6":    {BaseVCPU: 8, BaseHourly: 0.77976, RamPerVCPU: 8, IntelHourly: 0},
+	"s":     {BaseVCPU: 4, BaseHourly: 0.07143, RamPerVCPU: 2, IntelHourly: 0.08333},
+	"so":    {BaseVCPU: 8, BaseHourly: 0.77976, RamPerVCPU: 8, IntelHourly: 0.77976},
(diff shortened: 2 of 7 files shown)


values.yaml
```diff
@@ -0,0 +1,62 @@
+opencost:
```
ameijer: can this be removed?

pkg/env/costmodelenv.go
```diff
@@ -551,3 +552,7 @@ func GetNodeStatsCertFile() string {
 func GetNodeStatsKeyFile() string {
 	return env.Get(NodeStatsKeyFileEnvVar, "")
 }
+
+func GetDOKSPricingURL() string {
+	return env.Get(DOPricingURL, "https://api.digitalocean.com/v2/billing/pricing")
+}
```
ameijer: any reason this is configurable? our other providers generally have stable billing API

pkg/cloud/models/models.go
```diff
@@ -195,6 +195,7 @@ type CustomPricing struct {
 	GoogleAnalyticsTag           string `json:"googleAnalyticsTag"`
 	ExcludeProviderID            string `json:"excludeProviderID"`
 	DefaultLBPrice               string `json:"defaultLBPrice"`
+	DOPricingURL                 string `json:"doPricingURL,omitempty"`
```
ameijer: likely this is not needed here

pkg/cloud/digitalocean/provider_test.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+	}
+
+	assertEqual("Cost", node.Cost, "0.03571")
+	assertEqual("VCPUCost", node.VCPUCost, "0.00595")
+	assertEqual("RAMCost", node.RAMCost, "0.00595")
+	assertEqual("VCPU", node.VCPU, "2")
+	assertEqual("RAM", node.RAM, "4GiB")
+	assertEqual("ArchType", node.ArchType, "amd64")
+	assertEqual("PricingType", string(node.PricingType), string(models.DefaultPrices))
```
ameijer, line 119: thanks for tests!

pkg/env/costmodelenv.go
prateek1192: We don't plan on using a public pricing URL and would an internal one. The internal URL is in a flux as of now and as such I have made it configurable till we complete that.

values.yaml
prateek1192: Thanks for pointing this out !
Appreciate the review

pkg/cloud/models/models.go
prateek1192: Removed, thanks for pointing this out

ameijer:
@prateek1192 looks like you have a merge conflict

pkg/env/costmodelenv.go
```diff
@@ -104,6 +104,7 @@ const (
 	PluginExecutableDirEnvVar = "PLUGIN_EXECUTABLE_DIR"
 
 	OCIPricingURL = "OCI_PRICING_URL"
+	DOPricingURL  = "DO_PRICING_URL"
```
Sean-Holcomb: Since provider are exclusive to install, rather than having per provider env vars for pricing url can we do just "PROVIDER_PRICING_URL".

configs/digitalocean.json
```diff
@@ -0,0 +1,6 @@
+{
+  "doPricingURL": "https://api.digitalocean.com/v2/billing/pricing",
```
Sean-Holcomb: This field got removed from the struct I believe

Sean-Holcomb (approved):
Looks good, just some small comments

pkg/env/costmodelenv.go
prateek1192: Thanks for the review Sean. I would fix this.

configs/digitalocean.json
prateek1192: Removed, would upload the PR in a couple of days, I am adding some more DO slugs

pkg/cloud/digitalocean/provider.go
```diff
@@ -233,14 +233,45 @@ type SlugBase struct {
...
+	"c2":    {BaseVCPU: 4, BaseHourly: 0.13988, RamPerVCPU: 2, IntelHourly: 0.18155},
+	"g":     {BaseVCPU: 4, BaseHourly: 0.18750, RamPerVCPU: 4, IntelHourly: 0.22470},
+	"gd":    {BaseVCPU: 4, BaseHourly: 0.20238, RamPerVCPU: 4, IntelHourly: 0.23512},
+	"m":     {BaseVCPU: 8, BaseHourly: 0.50000, RamPerVCPU: 8, IntelHourly: 0.58929},
+	"m3":    {BaseVCPU: 8, BaseHourly: 0.61905, RamPerVCPU: 8, IntelHourly: 0.65476},
+	"m6":    {BaseVCPU: 8, BaseHourly: 0.77976, RamPerVCPU: 8, IntelHourly: 0},
+	"s":     {BaseVCPU: 4, BaseHourly: 0.07143, RamPerVCPU: 2, IntelHourly: 0.08333},
+	"so":    {BaseVCPU: 8, BaseHourly: 0.77976, RamPerVCPU: 8, IntelHourly: 0.77976},
+	"so1_5": {BaseVCPU: 8, BaseHourly: 0.97024, RamPerVCPU: 8, IntelHourly: 0.82738},
+}
```
ameijer, line 254: @prateek1192 will these go out of date as DO changes its pricing?

prateek1192: Thanks for pointing this out Alex and internally in DO as well we have had some discussion about this. DO raises prices infrequently(only one since launch since 2011) and this is a fallback and we also majorly plan on relying on the endpoint that is being built for billing. So we decided to keep it like this since this was a trade off of updating pricing every 10 years to having all nodes accurate prices here.
I looked at other providers and saw that they can add CPU and RAM prices using a config file. The problem with that for DO is that we have different CPU and RAM prices for different slugs(node types) and don't have one size fits all model.

ameijer: thanks prateek. If you can resolve the last merge conflict I think we will be good to merge!

prateek1192: Thanks for taking a look again Alex. I have resolved the merge conflict

PR #3233, opened 2025-07-15
fix: handle processing of multi page athena query results

### LINKED ISSUE #3131: Savings Plan Data Lost When Querying Multiple Pages from Athena
**Describe the bug**
When querying Savings Plan data from Athena, if the results span multiple pages, the data from previous pages is lost because the map is being cleared for each page. This results in only the data from the last page being retained.

Current Behavior:
In the `GetSavingsPlanDataFromAthena` method, the map `SavingsPlanDataByInstanceID` is cleared for each page of results:
```go
aws.SavingsPlanDataByInstanceID = make(map[string]*SavingsPlanData) // Clean out the old data
```

This means that when processing multiple pages of results, only the data from the last page is kept, effectively losing all previous data

**Expected behavior**
The map should only be cleared once at the beginning of the query, and subsequent pages should accumulate their results into the same map. This ensures all Savings Plan data is properly collected regardless of how many pages the query returns.

**Screenshots**

1.114.0
AWS(eks)

**Additional context**
This bug affects the accuracy of Savings Plan cost calculations when the query results span multiple pages. Users may see incorrect or incomplete Savings Plan data in their cost reports.

nishanthreddydd:
## What does this PR change?
*  This PR fixes #3131 

* N/A

* This PR address a existing bug while handling multi-page athena query results.

* Closes #3131 

## How was this PR tested?
* Tested the PR running locally, my setup makes athena queries that results in multi-page results and made sure the map doesn't get reset until all the results are processed and stored in memory.

* N/A

* Since, this is a bug we should have this in next release

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -2061,6 +2061,7 @@ func (aws *AWS) GetSavingsPlanDataFromAthena() error {
 	line_item_usage_start_date DESC`
 
 	page := 0
+	mostRecentDate := ""
 	processResults := func(op *athena.GetQueryResultsOutput) bool {
 		if op == nil {
 			log.Errorf("GetSavingsPlanDataFromAthena: Athena page is nil")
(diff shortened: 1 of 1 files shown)


ameijer:
group review, approved, handles RIs too

ameijer (approved):
LGTM

PR #3234, opened 2025-07-15
Code Clean Up

Sean-Holcomb:
## What does this PR change?
* This PR cleans up reference to concepts that are outside the scope of opencost in addition to cleaning up dead code

* https://github.com/kubecost/kubecost-cost-model/pull/3517
* https://github.com/kubecost/cost-analyzer-helm-chart/pull/4137

* Users previously using `ETL_READ_ONLY=true` will need to now set `KUBERNETES_RESOURCE_ACCESS=false`

* Closes ...

## How was this PR tested?
* 

* 

*


core/pkg/diagnostics/exporter/exporter.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // NewDiagnosticExporter creates a new `StorageExporter[DiagnosticsRunReport]` instance for exporting diagnostic run events.
 func NewDiagnosticExporter(clusterId string, applicationName string, storage storage.Storage) exporter.EventExporter[diagnostics.DiagnosticsRunReport] {
-	pathing, err := pathing.NewEventStoragePathFormatter("federated", clusterId, diagnostics.DiagnosticsEventName, applicationName)
+	pathing, err := pathing.NewEventStoragePathFormatter("", clusterId, diagnostics.DiagnosticsEventName, applicationName)
```
ameijer: will this bork downstream stuff?

core/pkg/heartbeat/exporter/heartbeat_test.go
```diff
@@ -47,7 +47,7 @@ func TestHeartbeatExporter(t *testing.T) {
 	time.Sleep(10 * time.Second)
 	controller.Stop()
 
-	files, _ := store.List(path.Join("federated", MockClusterId, heartbeat.HeartbeatEventName, MockApplicationName))
+	files, _ := store.List(path.Join("", MockClusterId, heartbeat.HeartbeatEventName, MockApplicationName))
```
ameijer: will break downstream?

ameijer:
also looks like you broke the protobuf

core/pkg/diagnostics/exporter/exporter.go
Sean-Holcomb: Since the files being exported are new in 3.0 I want to avoid creating more things in the "federated" subpath. If the kubemodel work ever moves forward I would want to fully sunset "federated". Moving forward bingen files will be placed in a directory called federated regardless of whether it is actually federated. Right now that vision looks like clusterID dirs sitting in the root directory. I would be open to replacing "federated" here with something like "kubernetes" so that we had something like
```
root - -> kubernetes -> clusterID_dirs...
       -> billing_integrations -> integrationID_dirs...
```

Sean-Holcomb: @mbolt35 what is your opinion here?

ameijer: cc @mittal-ishaan who is building all this

Sean-Holcomb: or "finops_agent"

mbolt35: @Sean-Holcomb @ameijer 

If we're just talking about the name here, I agree generally to drop the name ("federated") completely and simply use a flat listing of clusterIDs, but thinking ahead, there will exist data pipelines that may not specifically "link" to a cluster (and we should generally try to account for this by at least providing one initial root directory. The "choice" of `"federated"` in the opencost pipeline/diagnostic/heartbeat context is just for backwards compatibility, and I can't say I'm a huge fan of it in terms of accurately capturing the meaning of the data inside it. 

However, I also think that IF we're going to update this, we should really consider the entire pathing schema, and go with something that we can all agree on. @cliffcolvin has mentioned organizing data in a way that prefixes based on datetime in order to allow bucket maintainers to write cold store/removal policies more easily. 

Lastly, we should have code that is capable of migrating paths -- conceptually this is fairly straight-forward, but I know that support and devs tend to mix KCM versions with agent versions, so we have to be careful to ensure there exists a path that calls out legacy agents (ie: for a case where 1 agent is emitting files to federated and 14 emitting new data, a once-migration will fail to continually capture new data from the legacy agent). I _really_ want to avoid complex "multi-path" logic on the ingestion side if we can. 

Just want to make sure that if we're going to introduce a breaking change, we get enough juice from the squeeze!

mbolt35: I did want to add here that we can completely alter this for `kubemodel` work, and let legacy die of old age. As much sleep as I lose about the `"federated"` path name, I'm also ok with allowing it to live a bit longer if we have a better model plan in the chamber!

core/pkg/heartbeat/exporter/heartbeat_test.go
mbolt35: @ameijer Yeah, it will break where we expect to find files in the bucket downstream. I think @Sean-Holcomb's point is valid here though -- if we're coupling new diagnostics and heartbeat work with the legacy model, it might complicate moving away from the current structure later.

core/pkg/diagnostics/exporter/exporter.go
Sean-Holcomb: On your last comment, the point I am trying to make is that heartbeat and diagnostics will likely live into the `kubemodel` era, which is why I want to prevent further perpetuation of "federated".

Another note here is that this is not a breaking change, on anything that has been released, and posses no migration headaches. 

I am okay with a different schema but the reality of our data is that resolution is relevant before datetime and it wont be so simple regardless with windowed data and timestamped data.

pkg/cloud/provider/provider.go
```diff
@@ -159,13 +159,11 @@ func NewProvider(cache clustercache.ClusterCache, apiKey string, config *config.
 	}
 
 	var nodes []*clustercache.Node
-	if !env.IsETLReadOnlyMode() {
-		// the error can be ignored because getAllNodesFunc only errors if nodes is empty, a case which we explicitly
-		// handle by checking the length of nodes below
-		nodes, _ = retry.Retry(context.Background(), getAllNodesFunc, 10, time.Second)
-	} else {
-		nodes, _ = getAllNodesFunc()
-	}
```
mbolt35: Ok, just wanted to specifically call this out -- There are some users of opencost who do not have permissions to watch kubernetes resources (or who do not want to). The name of this environment variable is incredibly misleading (`IsETLReadOnlyMode()`) -- When this is the case, we will never get back a list of nodes, so it's a waste of time to retry it.

pkg/clustercache/clustercache.go
```diff
@@ -80,23 +80,22 @@ func NewKubernetesClusterCacheV1(client kubernetes.Interface) cc.ClusterCache {
...
-		go initializeCache(kcc.deploymentsWatch, &wg, cancel)
-		go initializeCache(kcc.statefulsetWatch, &wg, cancel)
-		go initializeCache(kcc.replicasetWatch, &wg, cancel)
-		go initializeCache(kcc.pvWatch, &wg, cancel)
-		go initializeCache(kcc.pvcWatch, &wg, cancel)
-		go initializeCache(kcc.storageClassWatch, &wg, cancel)
-		go initializeCache(kcc.jobsWatch, &wg, cancel)
-		go initializeCache(kcc.pdbWatch, &wg, cancel)
-		go initializeCache(kcc.replicationControllerWatch, &wg, cancel)
-	}
```
mbolt35: This needs to either be added back or the name needs to be updated to appropriately reflect it's purpose. There are some users of opencost who not have permissions (or choose to disable) watchers in K8s.

pkg/clustercache/clustercache2.go
```diff
@@ -54,24 +53,22 @@ func NewKubernetesClusterCacheV2(clientset kubernetes.Interface) *KubernetesClus
 func (kcc *KubernetesClusterCacheV2) Run() {
 	var wg sync.WaitGroup
 
-	if !env.IsETLReadOnlyMode() {
```
mbolt35, line 57: Needs to be added back and possibly renamed.

pkg/env/costmodelenv.go
```diff
@@ -9,8 +9,7 @@ import (
 )
 
 const (
-	APIPortEnvVar          = "API_PORT"
-	NetworkCostsPortEnvVar = "NETWORK_COSTS_PORT"
```
mbolt35, line 13: Did this get moved?

pkg/cloud/provider/provider.go
Sean-Holcomb: Awesome, thanks for the detail

pkg/env/costmodelenv.go
```diff
@@ -120,16 +106,11 @@ const (
...
-	KubecostConfigBucketEnvVar = "KUBECOST_CONFIG_BUCKET"
+	KubecostNamespaceEnvVar = "KUBECOST_NAMESPACE"
 )
 
 const DefaultConfigMountPath = "/var/configs"
 
-func IsETLReadOnlyMode() bool {
-	return env.GetBool(ETLReadOnlyMode, false)
-}
-
```
mbolt35: Needs to be added back with better name IMO.

pkg/env/costmodelenv.go
Sean-Holcomb: They were unused

core/pkg/opencost/asset.go
```diff
@@ -3186,107 +3002,6 @@ func (as *AssetSet) FindMatch(query Asset, aggregateBy []string, labelConfig *La
 	return nil, fmt.Errorf("Asset not found to match %s on %v", query, aggregateBy)
 }
 
-// ReconciliationMatch attempts to find an exact match in the AssetSet on
-// (Category, ProviderID). If a match is found, it returns the Asset with the
-// intent to adjust it. If no match exists, it attempts to find one on only
-// (ProviderID). If that match is found, it returns the Asset with the intent
-// to insert the associated Cloud cost.
-func (as *AssetSet) ReconciliationMatch(query Asset) (Asset, bool, error) {
```
mbolt35, line 3194: We don't do any sort of reconciliation with cloud costs right?

modules/prometheus-source/pkg/env/promenv.go
```diff
@@ -20,10 +20,10 @@ const (
 	PrometheusTLSHandshakeTimeoutEnvVar = "PROMETHEUS_TLS_HANDSHAKE_TIMEOUT"
 	ScrapeIntervalEnvVar                = "KUBECOST_SCRAPE_INTERVAL"
 
-	ETLMaxPrometheusQueryDurationMinutes = "ETL_MAX_PROMETHEUS_QUERY_DURATION_MINUTES"
+	PrometheusMaxQueryDurationMinutes      = "PROMETHEUS_MAX_QUERY_DURATION_MINUTES"
+	PrometheusQueryResolutionSecondsEnvVar = "PROMETHEUS_QUERY_RESOLUTION_SECONDS"
```
mbolt35: Amazing

pkg/cloud/provider/provider.go
```diff
@@ -159,13 +72,15 @@ func NewProvider(cache clustercache.ClusterCache, apiKey string, config *config.
 	}
 
 	var nodes []*clustercache.Node
-	if !env.IsETLReadOnlyMode() {
+
+	if env.HasKubernetesResourceAccess() {
```
mbolt35, line 76: Much better! Thank you 👍

pkg/config/configmanager.go
```diff
@@ -15,27 +13,17 @@ import (
...
-
 // DefaultConfigFileManagerOpts returns the default configuration options for the
 // config file manager
 func DefaultConfigFileManagerOpts() *ConfigFileManagerOpts {
 	return &ConfigFileManagerOpts{
-		BucketStoreConfig: "",
-		LocalConfigPath:   "/",
+		LocalConfigPath: "/",
 	}
 }
```
mbolt35: I think we can remove all of this now that ConfigFileManager accepts the `storage.Storage`

core/pkg/opencost/asset.go
Sean-Holcomb: Yeah this is dead code now

pkg/env/costmodelenv.go
mbolt35: `NETWORK_COSTS_PORT` is used in kubecost? Probably just recreated there.

mbolt35 (approved):
Looks great man!

PR #3236, opened 2025-07-16
Add ability to configure cloud provider type from env vars

### LINKED ISSUE #3117: OpenCost is unable to determine the cloud provider when the Kubernetes cluster includes virtual nodes.
**Describe the bug**
We are running a workload on our Kubernetes clusters that dynamically creates virtual nodes. These nodes have the `spec.providerID` field set to a custom implementation-specific value. As a result, OpenCost is unable to detect the cloud provider during startup, which leads to it entering an inconsistent or invalid state.

This issue occurs because the cloud provider is determined based on a randomly selected node in the cluster:
[Code](https://github.com/opencost/opencost/blob/develop/pkg/cloud/provider/provider.go#L177)

**To Reproduce**
Any eks cluster running both normal and virtual k8s nodes

**Expected behavior**
OpenCost should correctly detect the cloud provider in Kubernetes clusters that include virtual nodes. If possible expose a way to filter out these virtual nodes by using some kind of label selectors [here](https://github.com/opencost/opencost/blob/develop/pkg/clustercache/store.go#L42)

`1.114.0`

**Additional context**
Kubernetes Version - 1.30
Cloud provide : AWS (eks)

nishanthreddydd:
## What does this PR change?
This PR address the #3117 
Currently the logic to determine the cloud provider type is based on randomly selected node `.spec.ProviderId`. If we have clusters running special node like fargate or virtual kubelet nodes then this logic would fail to correctly determine the provider type.  This fix add an option for user to specify the provider type as env var, so when the actual logic fails to determine the provider type it will fall back to env var.

N/A

Shouldn't impact anyone

* Closes #3117 

## How was this PR tested?
Testing locally on a cluster with both virtual and eks nodes.

* Yes,  we could capture this configuration option .

* It should be part of next release

### DIFF
--- a/pkg/cloud/provider/provider.go
+++ b/pkg/cloud/provider/provider.go
@@ -175,6 +175,32 @@ func NewProvider(cache clustercache.ClusterCache, apiKey string, config *config.
 	}
 
 	cp := getClusterProperties(nodes[0])
+
+	// If provider is DEFAULT, check for explicitly set cloud provider from environment variable
+	envProvider := env.GetCloudProvider()
+	if cp.provider == "DEFAULT" && envProvider != "" {
+		log.Infof("Using cloud provider from environment variable: %s", envProvider)
+		cp.provider = envProvider
+		switch envProvider {
+		case opencost.AWSProvider:
+			cp.configFileName = "aws.json"
+		case opencost.AzureProvider:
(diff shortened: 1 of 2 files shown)


ameijer (approved):
LGTM - group review

PR #3242, opened 2025-07-21
fix(prometheus): update validation logic to not require every series to have "job" label

### LINKED ISSUE #3224: Prometheus validation code causes issues when bringing your own prometheus server
**Describe the bug**
I'm trying to install OpenCost, and I want to bring my own data source instead of installing a dedicated Prometheus server for OpenCost.  There are docs for bringing your own Prometheus here: https://opencost.io/docs/installation/prometheus/#providing-your-own-prometheus.

However, I actually want to bring my own VictoriaMetrics server, which should theoretically be possible since [MetricsQL](https://docs.victoriametrics.com/victoriametrics/metricsql/) (the query language for VictoriaMetrics) is backwards-compatible with PromQL.  I want to try this even though it is not officially supported, and I'm happy to open a documentation PR with some instructions if I can get this working.

I have the underlying connection and credentials working, but the validation code that runs on startup is causing issues for me.  The issue is that the validation code [runs an `up` query](https://github.com/opencost/opencost/blob/272afc5f0c2585395398f9555825dfb4cd7c6d1c/modules/prometheus-source/pkg/prom/validate.go#L9) against the data source, and [immediately fails when it encounters any series that does not have a `job` label](https://github.com/opencost/opencost/blob/272afc5f0c2585395398f9555825dfb4cd7c6d1c/modules/prometheus-source/pkg/prom/validate.go#L51-L57).  This is a problem because not every series returned from an `up` query via VictoriaMetrics will have a `job` label, as there are a variety of ways that metrics can be ingested into VictoriaMetrics (a prometheus-style scrape config being one of them).  For example, series ingested into VictoriaMetrics via OTEL will not have a `job` label.

To be clear - the data that OpenCost will eventually read from VictoriaMetrics will be ingested by a prometheus-style scrape config, which will have its own `job` label, but not all series returned from `up` will have this label.  I don't think this should completely block OpenCost from starting.

**To Reproduce**
You can reproduce this by trying to connect OpenCost to a PromQL-compatible data source that responds to an `up` query with some series that do not contain a `job` label.

**Expected behavior**
It should not be a requirement that every series returned from `up` against the datasource must have a `job` label.

**Suggested Fixes**
1. Remove the [check](https://github.com/opencost/opencost/blob/272afc5f0c2585395398f9555825dfb4cd7c6d1c/modules/prometheus-source/pkg/prom/validate.go#L51-L57) that forces every returned series from `up` to have a `job` label.  It seems unnecessary to me, but perhaps there is a valid reason for why it's there.
2. Be more specific with the validation query, for example, only selecting series that have the expected job label.  Something like
```
upQuery := fmt.Sprintf(`up{job="%s"}`, config.JobName)
```

I'm happy to open a PR with either of these changes, or any other change suggested in the comments.

1.115.0

mrparkers:
## What does this PR change?
* Updates the Prometheus validation logic such that it is no longer required for every series returned from `up` to contain a `job` label.

* No

* Users will now be able to bring their own Prometheus server, even if that server contains some series that do not have a `job` label.

* Fixes #3224 

## How was this PR tested?
* I tested this locally by importing the `github.com/opencost/opencost/modules/prometheus-source/pkg/prom` package with my local changes, and ran the validation logic against my own Prometheus server, which contains series that do not have a `job` label.

* No

* No, because I do not have permission to add labels to the issue I opened.  I would like this functionality to be part of the next release.


modules/prometheus-source/pkg/prom/validate.go
```diff
@@ -47,15 +47,19 @@ func validate(cli prometheus.Client, q string, config *OpenCostPrometheusConfig)
...
+
 	for _, result := range resUp {
 		job, err := result.GetString("job")
 		if err != nil {
-			return &PrometheusMetadata{
-				Running:            false,
-				KubecostDataExists: false,
-			}, fmt.Errorf("up query does not have job names")
+			// Stop evaluating result if it does not have a job label (continue to next result)
+			// We don't error here because not every result from `up` will necessarily have a job label.
```
ameijer, line 56: I think I agree with the approach here and the comment is helpful. Do you think we should at least have a DEBUG log here to help folks troubleshoot their prom configs if the error continues to fire?

ameijer (commented):
would you be willing to write a unit test using the MockPromClient in OpenCost to make sure this doesn't regress?

modules/prometheus-source/pkg/prom/validate.go
mbolt35: I don't think it's possible to have an `up` result that omits a `job` label, as this is automatically assigned by prometheus on scrape. 

This PR does expose some major flaws in the validation: 
* There can be _many_ `up` query results -- we only check the first 
* References to "Kubecost" and an env var defaulting to "kubecost" (the actual job name is `"opencost"`
* `KubecostDataExists` is never used 

@mrparkers Can you elaborate a bit on the specific context in which the `up` query result doesn't have a `job`? Have you overridden the job label as something else? 

I'm ok with this PR going up, but I would like to better understand the context behind the "no job label." -- I think we can also follow up this PR with some general improvements to the Validation process as well.

mbolt35 (approved):
Looks good - @mrparkers <del> had just one question for you in the review comments if you don't mind answering before we merge.</del>

UPDATE: @mrparkers Apologies, I read the issue you posted -- It does seem that vanilla prometheus can omit job labels too from the comment on the issue. Will ask for follow up there

Thanks for the contribution!

modules/prometheus-source/pkg/prom/validate.go
mbolt35: @mrparkers Apologies, I read the issue you posted -- It does seem that vanilla prometheus can omit job labels too from the comment on the issue. Will ask for follow up there

PR #3243, opened 2025-07-22
Sth/kcm 4335

Sean-Holcomb:
## What does this PR change?
* This PR updates the increase and rate aggregators which are meant to track positive increases in a counter to be able to handle counter restarts. To do this negative deltas between sample timestamps are dropped and not included in running totals. Ideally when the counter is reset a new metric aggregator would be initialized as the label set would change but that is not currently the case when dealing with certain things like containers which are part of a statefulset and do not have unique identifiers in their metric label sets.

* 

* 

* Closes ...

## How was this PR tested?
* Added unit tests which simulate the failing scenario.

* 

*

### DIFF
--- a/modules/collector-source/pkg/metric/aggregator/increase.go
+++ b/modules/collector-source/pkg/metric/aggregator/increase.go
@@ -21,6 +21,7 @@ func Increase(labelValues []string) MetricAggregator {
 	}
 }
 
+// getIncrease returns the current increase without updating the state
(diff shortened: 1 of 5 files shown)


Sean-Holcomb:
Not sure what is up with the sonarcloud check but every line in this PR is covered

nik-kc (approved):

ameijer (approved):
LGTM

PR #3245, opened 2025-07-23
fix(aws): retain all Savings Plan data when paginating Athena results

### LINKED ISSUE #3131: Savings Plan Data Lost When Querying Multiple Pages from Athena

shiavm006:
### Problem
When querying Savings Plan data from AWS Athena, if the results span multiple pages, only the data from the last page was retained. This was due to the SavingsPlanDataByInstanceID map being cleared on each page of results, causing all previously collected data to be lost. As a result, users with large datasets or multi-page Athena queries would see incomplete or incorrect Savings Plan cost reporting in OpenCost.

### what does PR do:
Fixes data loss for multi-page Athena Savings Plan queries.
Ensures accurate and complete Savings Plan cost calculations in OpenCost reports for all AWS users.
No regressions observed in related AWS or Athena functionality.

### issue being Fixed #3131

Previously, when querying Savings Plan data from Athena, if the results spanned multiple pages, only the data from the last page was retained. This was because the SavingsPlanDataByInstanceID map was being cleared on every page of results, causing all previously collected data to be lost.

would like to know ur views on it.

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -2039,6 +2039,11 @@ func (aws *AWS) GetSavingsPlanDataFromAthena() error {
 	}
 	if aws.SavingsPlanDataByInstanceID == nil {
 		aws.SavingsPlanDataByInstanceID = make(map[string]*SavingsPlanData)
+	} else {
+		// Always clear the map at the start of the query
+		for k := range aws.SavingsPlanDataByInstanceID {
+			delete(aws.SavingsPlanDataByInstanceID, k)
+		}
 	}
 	tNow := time.Now()
 	tOneDayAgo := tNow.Add(time.Duration(-25) * time.Hour) // Also get files from one day ago to avoid boundary conditions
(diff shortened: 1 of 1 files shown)


ameijer:
thanks for your contribution @shiavm006 ! closing as a dup of https://github.com/opencost/opencost/pull/3233, which also handles this issue for RIs

shiavm006:
> thanks for your contribution @shiavm006 ! closing as a dup of #3233, which also handles this issue for RIs

oh ya it was just like thta was not merged so just cretaed this 
is there any specefic issue i can workopwn u would suggest ?

PR #3261, opened 2025-07-30
fix(azure): Correct inverted spot and retail price assignment

### LINKED ISSUE #3259: Bug: Azure pricing logic incorrectly assigns standard prices to spot prices and vice versa
## Bug Description

The Azure pricing logic in `pkg/cloud/azure/provider.go` has a critical bug where standard on-demand prices are incorrectly assigned to spot prices and vice versa.

## Location

**File**: `pkg/cloud/azure/provider.go`  
**Lines**: 290-298  
**Function**: `getRetailPrice`

## Current Buggy Code

```go
    if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
        if !strings.Contains(strings.ToLower(item.SkuName), " spot") {
            spotPrice = fmt.Sprintf("%f", item.RetailPrice) 
        } else {
            retailPrice = fmt.Sprintf("%f", item.RetailPrice)
        }
    }
}
```

## Problem Analysis

The logic is **inverted**:
- `!strings.Contains(..., " spot")` means "does NOT contain spot", but it assigns to `spotPrice`
- `else` (contains spot) assigns to `retailPrice`

This causes:
1. Standard on-demand prices to be used as spot prices
2. Spot prices to be used as standard prices
3. Incorrect cost calculations in OpenCost
4. Mismatch between displayed costs and actual Azure billing

## Expected Behavior

Standard on-demand instances should use standard pricing, and spot instances should use spot pricing.

## Impact

- **High**: This affects all Azure users of OpenCost
- **Cost Accuracy**: Users see incorrect pricing in OpenCost metrics
- **Decision Making**: Cost optimization decisions based on wrong data
- **Billing Mismatch**: OpenCost costs don't match actual Azure bills

## Steps to Reproduce

1. Deploy OpenCost in an Azure Kubernetes cluster
2. Check the `/metrics` endpoint for `node_total_hourly_cost`
3. Compare with actual Azure pricing from the Retail Pricing API
4. Notice the price mismatch

## Environment

- **OpenCost Version**: Latest 
- **Cloud Provider**: Azure
- **Instance Types**: Any Azure VM type (e.g., Standard_D2as_v5)

## Additional Context

This bug was discovered when investigating why OpenCost was showing different pricing than expected from Azure's Retail Pricing API. The issue affects the core pricing logic and should be prioritized for fixing.

adity1raut:
### Description

This pull request addresses a critical bug in the Azure cloud provider where standard on-demand prices were being incorrectly assigned as spot prices, and vice-versa. This led to inaccurate cost reporting for all Azure users of OpenCost.

The core of the issue was an inverted logic check within the `getRetailPrice` function in `pkg/cloud/azure/provider.go`.

- The condition for a **standard instance** (`!strings.Contains(..., " spot")`) was incorrectly assigning the price to the `spotPrice` variable.
- The `else` block, which handles **spot instances**, was incorrectly assigning the price to the `retailPrice` variable.

This PR corrects the logic by swapping the assignments, ensuring that prices are allocated to the correct variables based on the instance type. This change ensures that OpenCost metrics will align with actual Azure billing.

### How to Test

1.  Deploy the code from this branch to an Azure Kubernetes Service (AKS) cluster.
2.  Access the `/metrics` endpoint on the OpenCost pod.
3.  Check the `node_total_hourly_cost` metric for a standard on-demand node (e.g., `Standard_D2as_v5`).
4.  Verify that the reported cost now matches the official **Azure on-demand price** for that instance type and region.
5.  If a spot node is present, verify its cost matches the official **Azure spot price**.

### Related Issue

Fixes #3259


pkg/cloud/azure/provider.go
```diff
@@ -288,15 +288,18 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
...
 	spotPrice := ""
-	for _, item := range pricingPayload.Items {
-		if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
-			if !strings.Contains(strings.ToLower(item.SkuName), " spot") {
-				spotPrice = fmt.Sprintf("%f", item.RetailPrice)
-			} else {
-				retailPrice = fmt.Sprintf("%f", item.RetailPrice)
-			}
-		}
-	}
```
ameijer, line 299: @adity1raut would you be willing to write a unit test or 2 around this?

adity1raut: Yes, absolutely! I’d be happy to write one or two unit tests for this.

ameijer: @adity1raut would love to see some tests! so that way this does not regress in the future

PR #3283, opened 2025-08-11
[PrometheusMetricsQuerier] DEBUG Query Logs

mbolt35:
## What does this PR change?
* Adds `DEBUG` level logging to the `PrometheusMetricsQuerier` which includes the specific query being executed as well as the specific value passed to `time=`.
* Added tests to validate each query logged the expected format.


ameijer (approved):
LGTM

PR #3295, opened 2025-08-13
Refactor ComputeCostData and add unit tests

sneaxhuh:
## What does this PR change?
Refactors the ComputeCostData function in pkg/costmodel/costmodel.go by breaking it into smaller, private helper functions for clarity and maintainability.

Adds unit tests in costmodel_test.go to validate the refactored logic.
Improves code readability, maintainability, and reduces risk of bugs in future modifications.

## How was this PR tested?
Ran the existing test suite in pkg/costmodel.

Verified all unit tests pass after refactor.

No, internal refactor only; behavior remains the same.

### DIFF
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -686,7 +686,7 @@ func findDeletedNodeInfo(dataSource source.OpenCostDataSource, missingNodes map[
 		}
 
 		if len(cpuCosts) == 0 {
-			log.Infof("Kubecost metrics not currently available. Ingest this server's /metrics endpoint to get that data.")
+			log.Infof("Opencost metrics not currently available. Ingest this server's /metrics endpoint to get that data.")
 		}
 
 		for node, costv := range cpuCosts {
--- a/pkg/costmodel/costmodel.go
+++ b/pkg/costmodel/costmodel.go
@@ -139,61 +139,17 @@ func (cm *CostModel) ComputeCostData(start, end time.Time) (map[string]*CostData
 	ds := cm.DataSource
 	mq := ds.Metrics()
 
-	grp := source.NewQueryGroup()
+	// Get Kubernetes data
 
-	resChRAMUsage := source.WithGroup(grp, mq.QueryRAMUsageAvg(start, end))
-	resChCPUUsage := source.WithGroup(grp, mq.QueryCPUUsageAvg(start, end))
-	resChNetZoneRequests := source.WithGroup(grp, mq.QueryNetZoneGiB(start, end))
-	resChNetRegionRequests := source.WithGroup(grp, mq.QueryNetRegionGiB(start, end))
-	resChNetInternetRequests := source.WithGroup(grp, mq.QueryNetInternetGiB(start, end))
-
-	// Pull pod information from k8s API
-	podlist := cm.Cache.GetAllPods()
-
-	podDeploymentsMapping, err := getPodDeployments(cm.Cache, podlist, clusterID)
+	podlist, podDeploymentsMapping, podServicesMapping, namespaceLabelsMapping, namespaceAnnotationsMapping, err := getKubeCostData(cm.Cache, clusterID)
 	if err != nil {
 		return nil, err
 	}
 
-	podServicesMapping, err := getPodServices(cm.Cache, podlist, clusterID)
+	// Get Prometheus data
+	resRAMUsage, resCPUUsage, resNetZoneRequests, resNetRegionRequests, resNetInternetRequests, err := getPrometheusData(mq, start, end)
 	if err != nil {
-		return nil, err
-	}
-
-	namespaceLabelsMapping, err := getNamespaceLabels(cm.Cache, clusterID)
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/costmodel/costmodel.go
```diff
@@ -134,42 +134,118 @@ func (cd *CostData) GetController() (name string, kind string, hasController boo
...
+	}
+
+	err = findDeletedPodInfo(cm.DataSource, missingContainers, start, end)
+	if err != nil {
+		log.Errorf("Error fetching historical pod data: %s", err.Error())
+	}
+	return containerNameCost, err
+}
+
+func getKubeCostData(cache clustercache.ClusterCache, clusterID string) ([]*clustercache.Pod, map[string]map[string][]string, map[string]map[string][]string, map[string]map[string]string, map[string]map[string]string, error) {
```
mbolt35: This seems unnecessary. Returning 5 maps doesn't feel like it made the functionality more concise or improved clarity. Recommendation here is to move the contents back inline to the original callsite.

pkg/costmodel/costmodel.go
```diff
@@ -193,53 +269,26 @@ func (cm *CostModel) ComputeCostData(start, end time.Time) (map[string]*CostData
...
 
-	networkUsageMap, err := GetNetworkUsageData(resNetZoneRequests, resNetRegionRequests, resNetInternetRequests, clusterID)
-	if err != nil {
-		log.Warnf("Unable to get Network Cost Data: %s", err.Error())
-		networkUsageMap = make(map[string]*NetworkUsageData)
-	}
+	return resRAMUsage, resCPUUsage, resNetZoneRequests, resNetRegionRequests, resNetInternetRequests, nil
+}
 
+func getContainerCostData(podlist []*clustercache.Pod, resRAMUsage, resCPUUsage []*source.ContainerMetricResult, pvClaimMapping map[string]*PersistentVolumeClaimData, unmountedPVs map[string][]*PersistentVolumeClaimData, networkUsageMap map[string]*NetworkUsageData, podDeploymentsMapping, podServicesMapping map[string]map[string][]string, namespaceLabelsMapping, namespaceAnnotationsMapping map[string]map[string]string, nodes map[string]*costAnalyzerCloud.Node, clusterMap clusters.ClusterMap, cp costAnalyzerCloud.Provider, clusterID string) (map[string]*CostData, map[string]*costAnalyzerCloud.Node, map[string]*CostData, error) {
```
mbolt35: Feel the same way about this function - I do not think we've gained anything from adding a function which accepts 10+ parameters - we're just moving variables around.

pkg/costmodel/costmodel.go
```diff
@@ -134,42 +134,118 @@ func (cd *CostData) GetController() (name string, kind string, hasController boo
...
+
+	namespaceAnnotationsMapping, err := getNamespaceAnnotations(cache, clusterID)
+	if err != nil {
+		return nil, nil, nil, nil, nil, err
+	}
+
+	return podlist, podDeploymentsMapping, podServicesMapping, namespaceLabelsMapping, namespaceAnnotationsMapping, nil
+}
+
+func getPrometheusData(mq source.MetricsQuerier, start, end time.Time) ([]*source.ContainerMetricResult, []*source.ContainerMetricResult, []*source.NetZoneGiBResult, []*source.NetRegionGiBResult, []*source.NetInternetGiBResult, error) {
```
mbolt35: I'm more ok with this, as it keeps the prometheus queries together as a single testable unit. While n-element tuple return types are generally unwieldy, I think having some type of `struct` return type containing these results is overkill.

mbolt35: Also, while there are references to prometheus in this code, prometheus is no longer the only source of data available, so I would prefer this function be named `queryMetrics` or `getMetricsData` - whichever you feel is more appropriate. 

That goes for any direct mention of "prometheus" by name in the code. The preference here is to be abstract about the data source. Hope that makes sense.

mbolt35 (changes requested):
Love the unit tests, but there were two functions that felt unnecessary to me, at least with respect to testing and code clarity. The data source query grouping in a single function seemed like a good change to me, although please read my notes on avoiding "prometheus" specific naming, as there are other metric queriers available, and this should be phrased more abstractly.

ameijer (approved):
LGTM

pkg/costmodel/costmodel.go
```diff
@@ -673,14 +686,14 @@ func findDeletedNodeInfo(dataSource source.OpenCostDataSource, missingNodes map[
 		}
 
 		if len(cpuCosts) == 0 {
-			log.Infof("Kubecost prometheus metrics not currently available. Ingest this server's /metrics endpoint to get that data.")
+			log.Infof("Kubecost metrics not currently available. Ingest this server's /metrics endpoint to get that data.")
```
mbolt35:
```suggestion
			log.Infof("Opencost metrics not currently available. Ingest this server's /metrics endpoint to get that data.")
```

mbolt35 (approved):
If you don't mind making that suggested change before merge - This looks good, thank you!

PR #3299, opened 2025-08-14
feat: Implement Unicode support in filter lexer with improved error h…

sneaxhuh:
## What does this PR change?
This PR addresses all the TODOs in the `lexer.go` file by:

1. **Adding full Unicode support**
   - Modified `advance()`, `match()`, and `peek()` functions to work with **runes** instead of bytes.
   - Using `utf8.DecodeRuneInString` for proper UTF-8 handling.
   - Updated `isIdentifierChar()` to support Unicode letters and digits while maintaining backward compatibility for ASCII.

2. **Improved error messages**
   - Added specific error handling for invalid UTF-8 sequences.
   - Enhanced error messages for Unicode characters with **code point information**.
   - Better distinction between invalid UTF-8 and unexpected characters.

3. **Maintained backward compatibility**
   - All existing tests continue to pass.
   - ASCII-only input is handled exactly as before.
   - Kubernetes name compatibility is preserved for ASCII characters.

* Allows Unicode characters in identifiers where supported.
* Produces clearer, more specific error messages for invalid input.
* Maintains consistent behavior for existing ASCII-only identifiers.

## How was this PR tested?
* Added `unicode_test.go` with tests covering:
  - Valid ASCII identifiers
  - Valid Unicode letters/digits
  - Invalid UTF-8 sequences
  - Unexpected Unicode characters when ASCII-only mode is enforced
* Verified all existing tests pass.

* No — behavior is backward compatible for ASCII users.
* Optional documentation update for Unicode support in identifiers.


core/pkg/filter/ast/lexer.go
```diff
@@ -258,25 +272,33 @@ func (s *scanner) scanToken() {
...
+		b := byte(r)
+		return (b >= '0' && b <= '9') || // 0-9
+			(b >= 'A' && b <= 'Z') || // A-Z
+			(b >= 'a' && b <= 'z') || // a-z
+			b == '-' || // hyphens are allowed according to K8s spec
+			b == '_' // underscores are allowed because of Prometheus sanitization
+	}
+	
+	// For Unicode characters, allow letters and numbers
+	return unicode.IsLetter(r) || unicode.IsDigit(r)
```
mbolt35: I don't think we need to check ascii specifically (the `unicode` checks should also check for ascii as well). We will have to alter the unicode checks to include punctuation as well. 

In general, we're a bit confused as to why we hard restricted the filter language fields and values to an ascii subset in the first place. We probably don't need any special ascii handling. What are your thoughts?

mbolt35: Also worth noting that there are likely some tests that will fail (because the expectation was that we _only_ support a limited ascii character set). You can either repurpose those tests, or remove them if they're no longer required.

mbolt35 (changes requested):
PR looks great, just a quick request for change/comment.

core/pkg/filter/ast/lexer.go
sneaxhuh: the ascii restriction is just a naming convention going on from the k8s/prometheus naming rules i think, and i'd suggest to keep having it since it maintains consistency (not that huge impact but still), also the current tests still pass over the updated `isIdentifierChar` as well, so i don't think we need updating the tests as they are sufficient(although do let me know if you need any changes in the tests)

mbolt35 (approved):
LGTM

PR #3300, opened 2025-08-14
Fix: resolve Azure ConfigMap timing issue in OpenCost 1.116.0

### LINKED ISSUE #3297: Opencost v1.116.0 is no longer respecting pricing-configs configmap
**Describe the bug**
When I upgraded my opencost from 1.115.0 to 1.116.0, it is no longer reading the pricing-configs configmap for region and currency. I am using for azure cloud and configured `azureBillingRegion` and `currencyCode` Its using the default US region and USD currency. I could not find anything in the release notes for this breaking change. 

**To Reproduce**
1. Use `azureBillingRegion` and `currencyCode` in 1.115.0, these are being used for azure rate card api.
2. Upgrade to 1.116.0 and opencost would not read these configurations for pricing-configs configmap. 

**Expected behavior**
Opencost should use configurations added in the pricing-configs just like it was using in 1.115.0

**Screenshots**

I upgraded from 1.115.0 to 1.116.0
**Additional context**
Logs:
No pricing-configs configmap found at install time, using existing configs: configmaps "pricing-configs" not found
Using azureRateCard query OfferDurableId eq 'MS-AZR-0003p' and Currency eq 'USD' and Locale eq 'en-US' and RegionInfo eq 'US

Jetshree:
### Related Issues
Fixes #3297 

### Problem
After upgrading from OpenCost 1.115.0 to 1.116.0, the `pricing-configs` ConfigMap is no longer being properly read for Azure cloud configuration. Specifically, `azureBillingRegion` and `currencyCode` parameters default to "US" region and "USD" currency instead of using the configured values from the ConfigMap.

### Root Cause
A timing issue (race condition) where `DownloadPricingData()` is called immediately after the config watcher starts, but before the config watcher has fully processed and applied the initial ConfigMap data.

### Solution
1. **Timing Fix**: Added a 100ms delay after `configWatchers.Watch()` to ensure the config watcher has sufficient time to process the initial ConfigMap before `DownloadPricingData()` is called.

2. **Enhanced Logging**: Added comprehensive logging across multiple components to provide better visibility into:
   - ConfigMap processing status
   - Configuration field updates
   - Azure Rate Card API query parameters
   - Success/failure of config updates

### Files Modified
- `opencost/pkg/cmd/agent/agent.go`: Added timing delay
- `opencost/pkg/costmodel/router.go`: Added timing delay  
- `opencost/pkg/cloud/azure/provider.go`: Enhanced logging for config updates and API calls
- `opencost/pkg/util/watcher/configwatchers.go`: Enhanced logging for ConfigMap processing
- `opencost/pkg/cloud/provider/providerconfig.go`: Enhanced logging for field updates

### Testing
- Verified that ConfigMap values are properly applied
- Confirmed Azure Rate Card API uses configured region and currency
- Added logging to facilitate future debugging

### Breaking Changes
None. This is a bug fix that restores the expected behavior from version 1.115.0.

### DIFF
--- a/pkg/costmodel/router.go
+++ b/pkg/costmodel/router.go
@@ -550,6 +550,9 @@ func Initialize(router *httprouter.Router, additionalConfigWatchers ...*watcher.
 	configWatchers.AddWatcher(metrics.GetMetricsConfigWatcher())
 	configWatchers.Watch()
 
+	// Give the config watcher a moment to process the initial config
+	time.Sleep(100 * time.Millisecond)
+
 	clusterMap := dataSource.ClusterMap()
 	settingsCache := cache.New(cache.NoExpiration, cache.NoExpiration)
 
@@ -571,6 +574,7 @@ func Initialize(router *httprouter.Router, additionalConfigWatchers ...*watcher.
 
 	// Initialize mechanism for subscribing to settings changes
 	a.InitializeSettingsPubSub()
+
 	err = a.CloudProvider.DownloadPricingData()
 	if err != nil {
 		log.Infof("Failed to download pricing data: %s", err)
--- a/pkg/util/watcher/configwatchers.go
+++ b/pkg/util/watcher/configwatchers.go
@@ -98,6 +98,7 @@ func (cmw *ConfigMapWatchers) Watch() {
 			log.Infof("No %s configmap found at install time, using existing configs: %s", cw, err.Error())
 		} else {
 			log.Infof("Found configmap %s, watching...", configs.Name)
+			// Ensure the config is applied synchronously before proceeding
 			watchConfigFunc(configs)
 		}
 	}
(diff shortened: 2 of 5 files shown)


pkg/costmodel/router.go
```diff
@@ -550,6 +550,9 @@ func Initialize(router *httprouter.Router, additionalConfigWatchers ...*watcher.
 	configWatchers.AddWatcher(metrics.GetMetricsConfigWatcher())
 	configWatchers.Watch()
 
+	// Give the config watcher a moment to process the initial config
+	time.Sleep(100 * time.Millisecond)
```
ameijer, line 554: generally sleeps are not a good approach, is there a way you can listen on a signal or observe some state?

Jetshree: thanks for pointing this out, I’ll check if there’s a signal or ready state we can wait on instead.

pkg/util/watcher/configwatchers.go
```diff
@@ -124,10 +125,13 @@ func (cmw *ConfigMapWatchers) toWatchFunc() func(any) {
 		name := conf.GetName()
 		data := conf.Data
 		if watchers, ok := cmw.watchers[name]; ok {
+			log.Infof("Updating config from %s configmap with %d entries", name, len(data))
 			for _, cw := range watchers {
 				err := cw.WatchFunc(name, data)
 				if err != nil {
 					log.Infof("ERROR UPDATING %s CONFIG: %s", name, err.Error())
+				} else {
+					log.Infof("Successfully updated config from %s", name)
```
ameijer: debug log @mittal-ishaan ?

PR #3303, opened 2025-08-15
fix(oracle): implement cluster type detection for management pricing

sneaxhuh:
## What does this PR change?
* This commit implements the TODO in the Oracle provider's `ClusterManagementPricing` method to properly detect cluster types and apply appropriate pricing.

Changes:
- Added logic to detect `BASIC_CLUSTER` vs `ENHANCED_CLUSTER` types by checking node labels
- Implemented pricing logic where `BASIC_CLUSTER` types are free (0.0 cost)
- Default to `ENHANCED_CLUSTER` pricing for other cluster types
- Verified all tests pass and no regressions introduced

This implementation checks for common labels that indicate a `BASIC_CLUSTER` type and returns zero cost for those clusters, while maintaining existing pricing logic for `ENHANCED_CLUSTER` types. This ensures customers are correctly charged based on their actual cluster type.

* No direct relation, but complements previous work on Oracle provider pricing logic.

* Oracle OKE users will now be correctly billed for cluster management:
  - `BASIC_CLUSTER` types will incur no management cost.
  - `ENHANCED_CLUSTER` types will continue to be charged as before.
* Prevents overcharging for free cluster types.

## How was this PR tested?
* Added unit tests for `ClusterManagementPricing` covering both `BASIC_CLUSTER` and `ENHANCED_CLUSTER` detection.
* Manually tested in an environment with simulated Oracle nodes containing different label configurations.
* Verified no regressions in existing pricing logic.

### DIFF
--- a/pkg/cloud/oracle/provider.go
+++ b/pkg/cloud/oracle/provider.go
@@ -245,7 +245,28 @@ func (o *Oracle) ClusterManagementPricing() (string, float64, error) {
 	}
 	o.DownloadPricingDataLock.Lock()
 	defer o.DownloadPricingDataLock.Unlock()
-	// TODO: Support lookup of cluster type, as BASIC_CLUSTER types are free.
+	
+	// Determine cluster type by checking node labels
+	clusterType := "ENHANCED_CLUSTER" // default to enhanced cluster
+	nodes := o.Clientset.GetAllNodes()
+	for _, node := range nodes {
+		// Check for BASIC_CLUSTER indicator in labels
+		if _, ok := node.Labels["clusterType"]; ok && node.Labels["clusterType"] == "BASIC_CLUSTER" {
+			clusterType = "BASIC_CLUSTER"
+			break
+		}
+		// Alternative check for basic cluster labels
+		if _, ok := node.Labels["oke.oraclecloud.com/basic-cluster"]; ok {
+			clusterType = "BASIC_CLUSTER"
+			break
+		}
+	}
+	
+	// BASIC_CLUSTER types are free, so return 0.0 cost
+	if clusterType == "BASIC_CLUSTER" {
+		return managementPlatformOKE, 0.0, nil
+	}
+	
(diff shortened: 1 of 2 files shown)


ameijer:
@sneaxhuh needs UT file committed

ameijer:
@sneaxhuh wanted to follow up here

pkg/cloud/oracle/provider.go
```diff
@@ -245,7 +245,28 @@ func (o *Oracle) ClusterManagementPricing() (string, float64, error) {
...
+	clusterType := "ENHANCED_CLUSTER" // default to enhanced cluster
+	nodes := o.Clientset.GetAllNodes()
+	for _, node := range nodes {
+		// Check for BASIC_CLUSTER indicator in labels
+		if _, ok := node.Labels["clusterType"]; ok && node.Labels["clusterType"] == "BASIC_CLUSTER" {
+			clusterType = "BASIC_CLUSTER"
+			break
+		}
+		// Alternative check for basic cluster labels
+		if _, ok := node.Labels["oke.oraclecloud.com/basic-cluster"]; ok {
```
ameijer, line 259: hi @anders-swanson - wanted to see if you could help us with something. is this the recommended way to determine a cluster type? I checked the labels on our OKE cluster but cannot see any of these labels

anders-swanson: Last time I checked, labels didn't present an option. The way to get the cluster type was to get the cluster by OCID using OCI APIs - since I was implementing this quite lightweight at the time without any auth'd tenancy APIs, I omitted this.

ameijer:
thanks @anders-swanson - going to close this PR since this doesn't currently seem possible. Do you know if propagating this information down to k8s objects is on the roadmap? Most of the other major providers do tell us metadata about the kind of cluster we are on via labels or annotations

anders-swanson:
I’ll check on some Oracle channels.

PR #3310, opened 2025-08-15
Remove superfluous write header

Sean-Holcomb:
## What does this PR change?
* This PR removes a superflous call to write header which was causing logs such as these
image
When json encoding fails write header is called twice.
The description for WriteHeader in the interface definition states that 

> If WriteHeader is not called explicitly, the first call to Write
will trigger an implicit WriteHeader(http.StatusOK). 
thus explicit calls to WriteHeader are mainly used to
send error codes or 1xx informational responses.

There is a call to `Write` in the decode so writing the 200 code is unnecessary.

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


mbolt35 (approved):
LGTM :)

PR #3312, opened 2025-08-15
refactor(costmodel): Accumulate AllocationSets iteratively

sneaxhuh:
## What does this PR change?
* Refactored to accumulate objects iteratively, reducing memory usage for large time ranges.

The previous implementation collected all objects into a slice before accumulating them. This change processes each object within the loop, avoiding the need to store all sets in memory at once.

This optimization has also been applied to the same function in the build for consistency.

* Users will experience reduced memory usage when handling large time ranges.
* Improves performance and stability in scenarios with high data volume.
* No breaking changes to the API or output.

## How was this PR tested?
* Ran `go test ./...` to execute all tests recursively and confirm no regressions.

### DIFF
--- a/pkg/costmodel/allocation.go
+++ b/pkg/costmodel/allocation.go
@@ -38,15 +38,26 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time) (*opencost.Allocati
 	}
 
 	// If the duration exceeds the configured MaxPrometheusQueryDuration, then
-	// query for maximum-sized AllocationSets, collect them, and accumulate.
+	// query for maximum-sized AllocationSets, and accumulate as we go.
 
 	// s and e track the coverage of the entire given window over multiple
 	// internal queries.
 	s, e := start, start
 
-	// Collect AllocationSets in a range, then accumulate
-	// TODO optimize by collecting consecutive AllocationSets, accumulating as we go
-	asr := opencost.NewAllocationSetRange()
+	// result is the accumulating AllocationSet
+	var result *opencost.AllocationSet
+
+	// Populate annotations, labels, and services on each Allocation. This is
+	// necessary because AllocationSet.Merge does not propagate any values
+	// stored in maps or slices for performance reasons. In this case, however,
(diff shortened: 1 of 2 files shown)


pkg/costmodel/allocation.go
```diff
@@ -38,15 +38,26 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time) (*opencost.Allocati
 	}
 
 	// If the duration exceeds the configured MaxPrometheusQueryDuration, then
```
mbolt35:
```suggestion
	// If the duration exceeds the configured BatchDuration, then
```

mbolt35 (changes requested):
This code looks good in general, much more straight forward to read. However, I do want to request that you add a go Benchmark to compare before the change and after. Best case, this approach allocates the same as the old method (or _maybe_ a bit less). However, I have the suspicion that this is going to allocate, at worst, N additional times. The reason is that the AccumulateAll method allocates a single new AllocationSet to accumulate all data in, whereas each call to Accumulate() directly will allocate a new AllocationSet every call.

pkg/costmodel/allocation.go
```diff
@@ -66,26 +77,22 @@ func (cm *CostModel) ComputeAllocation(start, end time.Time) (*opencost.Allocati
...
+		}
 
-	// Also record errors and warnings, then append them to the results later.
-	errors := []string{}
-	warnings := []string{}
+		// On the first successful computation, set the result, otherwise accumulate
+		if result == nil {
+			result = as
+		} else {
+			acc, err := result.Accumulate(as)
```
nikovacevic, line 89: Really just repeating @mbolt35 -- this function is going to allocate a completely new `AllocationSet` each time it is called. If we wanted to really improve memory allocation we would need a new version of `Accumulate` that modified the existing pointer instead of cloning all the data.

I'd be supportive of that change. But, until we change that behvior, I wouldn't necessarily expect a big performance improvement from this code.

PR #3320, opened 2025-08-19
CVE-2025-30204 Update golang jwt v4 and v5

lmello:
## What does this PR change?
*  This PR updates a dependency to fix a CVE. 
golang-jwt is a Go implementation of JSON Web Tokens. Starting in version 3.2.0 and prior to versions 5.2.2 and 4.5.2, the function parse.ParseUnverified splits (via a call to strings.Split) its argument (which is untrusted data) on periods. As a result, in the face of a malicious request whose Authorization header consists of Bearer followed by many period characters, a call to that function incurs allocations to the tune of O(n) bytes (where n stands for the length of the function's argument), with a constant factor of about 16.

This issue is fixed in versions greater or equal to 5.2.2 and 4.5.2.

*  No

* No

## How was this PR tested?
* 

* No

*


ameijer (approved):
LGTM

PR #3308, opened 2025-08-15
add name method for storage

mittal-ishaan:
## What does this PR change?
* Few of the storage types like s3, gcs already have Name method. I have added it for all storage types. 
* We would be able to log the name of the storage present at the start 

## Does this PR relate to any other PRs?
* https://github.com/kubecost/kubecost-cost-model/pull/3604

## How will this PR impact users?
* We would be able to log storage name like the s3 or GCS bucket name.

## Does this PR address any GitHub or Zendesk issues?
* https://apptio.atlassian.net/browse/KCM-4388

## How was this PR tested?
* Ran go test ./... locally

## Does this PR require changes to documentation?
* No

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next OpenCost release? If not, why not?
*

PR #3321, opened 2025-08-19
use string method instead of adding Name method in storage interface

mittal-ishaan:
## What does this PR change?
* Add a String method in multiple storage types to return their storage names like s3 bucket name, gcp bucket name, etc that can be used for logging.

* https://github.com/opencost/opencost/pull/3308

* 

* Closes ...

## How was this PR tested?
* Ran go test ./... and ran locally with cost-model changes

* 

*

### DIFF
--- a/core/pkg/storage/azurestorage.go
+++ b/core/pkg/storage/azurestorage.go
@@ -191,8 +191,8 @@ func NewAzureStorageWith(conf AzureConfig) (*AzureStorage, error) {
 	}, nil
 }
 
-// Name returns the bucket name for azure storage.
-func (as *AzureStorage) Name() string {
+// String returns the bucket name for azure storage.
+func (as *AzureStorage) String() string {
 	return as.name
 }
 
--- a/core/pkg/storage/memorystorage.go
+++ b/core/pkg/storage/memorystorage.go
@@ -25,6 +25,11 @@ func NewMemoryStorage() *MemoryStorage {
 	}
 }
 
+// String returns the storage type as a string for logging purposes.
+func (ms *MemoryStorage) String() string {
+	return string(ms.StorageType())
+}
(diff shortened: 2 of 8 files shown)


core/pkg/storage/azurestorage.go
```diff
@@ -191,8 +191,8 @@ func NewAzureStorageWith(conf AzureConfig) (*AzureStorage, error) {
 	}, nil
 }
 
-// Name returns the bucket name for azure storage.
-func (as *AzureStorage) Name() string {
+// String returns the bucket name for azure storage.
+func (as *AzureStorage) String() string {
```
mittal-ishaan, line 195: Name method in azure, gcp and s3 were not added by me in https://github.com/opencost/opencost/pull/3308 they were present earlier itself. Checked for its references and they are not being used anywhere(also considering Name was not part of storage interface unlikely it was being used anywhere other than itself). Updated these methods to `String()`

core/pkg/storage/memorystorage.go
```diff
@@ -25,10 +25,6 @@ func NewMemoryStorage() *MemoryStorage {
 	}
 }
 
-func (ms *MemoryStorage) Name() string {
```
mittal-ishaan, line 28: Not adding a `String()` method here as for memory we do not have any info as part of its name. When logging we would need to default to `""` when this method is not found for a storage.

mbolt35: You can return the StorageType (string type) for the String() calls which don't have obvious return types!

mittal-ishaan: done

mbolt35 (approved):
Looks great! Thanks for making these improvements!

PR #3322, opened 2025-08-20
use app name for wal file paths

Sean-Holcomb:
## What does this PR change?
* This PR changes the location of the Wal files generated by the collector source to be placed in dir `root/app-name/cluster-id/collector/` rather than `root/cluster-id/collector/`. This follows patterns used by other exporters and allows for the application running the collector source to define the output dir of this component.

* 

* 

* Closes ...

## How was this PR tested?
* Tested with unit tests

* 

*


mittal-ishaan (approved):

modules/collector-source/pkg/collector/datasource.go
```diff
@@ -67,6 +67,7 @@ func NewCollectorDataSource(
 	if store != nil {
 		wal, err := metric.NewWalinator(
 			config.ClusterID,
+			config.ApplicationName,
```
ameijer, line 70: should application name be before cluster iD?

mittal-ishaan: https://github.com/opencost/opencost/pull/3322/files#diff-bda42a9867cf80fc6ea454a5afb577fac30d38f287ced1946cdd6979ee3519aaR38-R42
the function here defines it such that clusterId is before application id. No change in end result

ameijer (approved):
LGTM

mbolt35 (approved):
LGTM

modules/collector-source/pkg/metric/walinator.go
```diff
@@ -47,7 +48,7 @@ func NewWalinator(
 			limitResolution = resolution
 		}
 	}
-	pathFormatter, err := pathing.NewEventStoragePathFormatter("", clusterID, ControllerEventName)
+	pathFormatter, err := pathing.NewEventStoragePathFormatter(applicationName, clusterID, ControllerEventName)
```
mittal-ishaan, line 51: Should there be `collector` instead of `ControllerEventName` in the parameters? @Sean-Holcomb

Sean-Holcomb: Yes it should be

PR #3325, opened 2025-08-21
Update retention variables, introduce prefixs

Sean-Holcomb:
## What does this PR change?
* This PR changes the environment variables that manage data retention for various aspects of opencost and also introduces the idea of a prefixed env var which allows for a general default to be set for an env variable on top of the hard coded default but also allows specific components to have their own settings. In this case data retention can be set to a contain default using the unprefixed version of the env var, but any prefixed version will overwrite that for the calling component.

* https://github.com/opencost/opencost-helm-chart/pull/297

* 

* Closes ...

## How was this PR tested?
* 

* Some environment variables may need to be changed in the documentation

*


mbolt35 (approved):
LGTM

PR #3330, opened 2025-08-22
update validate func used by aggregator to not check for write and delete

mittal-ishaan:
## What does this PR change?
* Aggregator uses this validate func. to check for fed bucket permissions but as aggregator does not need write and delete permissions we are removing them from validate func
* Validate fun is not used anywhere except in cost-model

* https://github.com/kubecost/kubecost-cost-model/pull/3604

* 

* Closes ...

## How was this PR tested?
* 

* 

*

### DIFF
--- a/core/pkg/storage/storage.go
+++ b/core/pkg/storage/storage.go
@@ -78,12 +78,6 @@ func Validate(storage Storage) error {
 		return errors.Wrap(err, "Failed to list path")
 	}
 
-	// attempt to write a path
-	err = storage.Write(testPath, []byte(testContent))
-	if err != nil {
-		return errors.Wrap(err, "Failed to write data to storage")
-	}
-
 	// attempt to read the path
 	data, err := storage.Read(testPath)
 	if err != nil {
@@ -91,12 +85,6 @@ func Validate(storage Storage) error {
 	}
 	if string(data) != testContent {
 		return errors.New("Failed to read the expected data from storage")
-	}
-
-	// delete the path
-	err = storage.Remove(testPath)
-	if err != nil {
-		return errors.Wrap(err, "Failed to remove data from storage")
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


core/pkg/storage/storage.go
```diff
@@ -62,7 +62,6 @@ type Storage interface {
 // Validate uses the provided storage implementation to write a test file to the store, followed by a removal.
 func Validate(storage Storage) error {
```
ameijer: rather than removing this code, want to make a boolean get passed in that validates read vs write?

ameijer (approved):

core/pkg/storage/storage.go
```diff
@@ -87,8 +88,10 @@ func Validate(storage Storage, validateWriteDelete bool) error {
 	}
 
 	// attempt to read the path
+	// If we are not validating write and delete, the file won't exist since we never wrote it.
+	// We only want to check read permissions, so ignore errors with "exist" and "404" in the error message to bypass the file not exist error.
 	data, err := storage.Read(testPath)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "exist") && !strings.Contains(err.Error(), "404") {
```
ameijer, line 94: also, check for an os.notExistError

mittal-ishaan: sure, I was thinking that check for exists covers check for os.notExistError itself

ameijer (approved):
LGTM

core/pkg/storage/storage.go
mbolt35: ```if err != nil && !strings.Contains(err.Error(), "exist") && !strings.Contains(err.Error(), "404") {```

`os.IsNotExist(err)` or `os.IsExist(err)` please. This uses the standard `error.Is()` method for finding wrapped errors. 

I don't even begin to understand what the `404` error is -- I don't think the storage buckets return 404? Maybe I'm wrong.

ameijer: we have found that not all cloud providers return the os is not exist error

PR #3333, opened 2025-08-22
Add application name in diagnostics report file when exporting

mittal-ishaan:
## What does this PR change?
* Add application name in diagnostics report file when exporting
* This was mainly done to get the application Name while ingested diagnostics in clickhouse table. 
* We could have also asked the cost model ingestor to get the application name from file paths while checking for files to ingest. But that would have required us to separately handle application name in file paths in the main ingestor that ingest allocations, assets, etc too with diagnostics. But we do not have a concept of application name for allocation, etc. Doable but might become messy at the ingestor side. 
* Heartbeat export response already has application name in it so no change required there.

* https://github.com/kubecost/kubecost-cost-model/pull/3624

* No

* Closes ...

## How was this PR tested?
* Ran locally

* 

*


mbolt35 (approved):
LGTM

PR #3340, opened 2025-09-01
feat: Customizable cluster management cost

### LINKED ISSUE #3339: Customizable cluster management fee
Currently, cluster management fee is hard-coded (usually with 0.10) and cannot be changed via custom pricing model.
Also, since opencost sets cluster management cost only if it finds specific labels (`eks.amazonaws.com/nodegroup` for EKS and `cloud.google.com/gke-nodepool` for GKE), this is a bit fragile.
In my case, all my EKS nodes are managed by Karpenter and they don't have these labels, so OpenCost set the cluster management fee to 0, which is incorrect.
https://github.com/opencost/opencost/blob/14a3a79ea6cb7c844db5c783ac1778ec5e69f212/pkg/cloud/aws/provider.go#L858-L863

Allow cluster management cost (`clusterManagementPrice`) to be configurable from [the custom pricing model](https://opencost.io/docs/configuration/on-prem/#on-prem-configuration).

We could also fix the detection logic of EKS cluster management cost to include `karpenter.sh/nodepool` label, but that fix is only temporary and limits use-cases.

**Additional context**

motoki317:
## What does this PR change?
* Adds `clusterManagementCost` to the custom pricing model.

* Allows `clusterManagementCost` to be configurable, even if OpenCost fails to detect the cluster provisioner.

* Closes #3339

## How was this PR tested?
* I have added unit tests, but I haven't tested it on a real environment yet.

* Preferably, add `clusterManagementCost` to the list of fields in the [On-Premises](https://opencost.io/docs/configuration/on-prem/#on-prem-configuration) page.

* I don't have the permission to set labels

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -874,6 +874,10 @@ func (aws *AWS) DownloadPricingData() error {
 	aws.SpotDataPrefix = c.SpotDataPrefix
 	aws.ProjectID = c.ProjectID
 	aws.SpotDataRegion = c.SpotDataRegion
+	clusterManagementCost, hasCustomClusterManagementCost := c.GetClusterManagementCost()
+	if hasCustomClusterManagementCost {
+		aws.clusterManagementPrice = clusterManagementCost
+	}
 
 	aws.ConfigureAuthWith(c) // load aws authentication from configuration or secret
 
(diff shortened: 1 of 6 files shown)


ameijer:
@motoki317 can we beef up testing here? thanks!

PR #3344, opened 2025-09-03
DCGM Scrape Fix + Diagnostics Cleanup

mbolt35:
## What does this PR change?
* ### DCGM Fixes
  * Add dcgm-exporter changes that should fix dcgm for multiple exporters (multi-node dcgm). 
  * Added note about more efficient way of pulling pod ip mapping. 
* ### Remove Non-Scrape Diagnostics
  * Scrape Diagnostics included the two Updater checks, so we no longer need to support both types.


modules/collector-source/pkg/metric/diagnostics.go
```diff
@@ -427,72 +303,7 @@ func (d *DiagnosticsModule) Update(updateSet *UpdateSet) {
...
-		}
-
-		// More duck typing sadly -- there are some fundamental design incompatibilities
-		// with the way DiagnosticService was written and this cached diagnostic approach
-		// that make things like "cycle" resets a bit difficult
-		if metricDiag, ok := diag.(*metricDiagnostic); ok {
-			metricDiag.updateTimestamps = append(metricDiag.updateTimestamps, t)
-		}
-	}
+	d.updater.Update(updateSet)
```
mittal-ishaan: we can remove this `update` function and also remove `NewDiagnosticsModule` from the collector data source update pipeline? considering we would not be adding any metric specific diagnostics and if needed we can always add a scraper diagnostics for it.

mittal-ishaan (approved):

modules/collector-source/pkg/metric/diagnostics.go
mbolt35: @mittal-ishaan Done - Can you please verify this looks correct?

Sean-Holcomb (approved):
Looks good

PR #3351, opened 2025-09-07
chores: Optimize Dockerfile with cross-compilation

motoki317:
## What does this PR change?
* Optimizes Dockerfile for faster builds using buildkit.
Since Golang can cross-compile well, we can use `--platform=$BUILDPLATFORM` to let buildkit cross-compile without using CPU emulators which is quite slow.
* I have excluded test step when building image, since this is not strictly required when building a docker image. We should probably separate these in CI/CD pipeline and check them in parallel, resulting in faster image builds.

* No as far as I'm aware of

* No impact to users, only build process optimization.

* No

## How was this PR tested?
* I've built locally using builx commands (`docker buildx build --platform linux/amd64,linux/arm64 .`) and have confirmed that cross-compiled images (amd64->arm64) works.

* No

* Not sure yet

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,6 @@
-FROM golang:1.24-alpine3.20 as build-env
+FROM --platform=$BUILDPLATFORM golang:1.24-alpine3.20 AS build-env
 
-RUN mkdir /app
 WORKDIR /app
-COPY go.mod .
-COPY go.sum .
 
 # This ensures that CGO is disabled for go test running AND for the build
 # step. This prevents a build failure when building an ARM64 image with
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,30 +1,34 @@
...
+COPY modules/prometheus-source/go.* modules/prometheus-source/
+RUN go mod download
+
 ARG version=dev
 ARG	commit=HEAD
 
+ARG TARGETOS
+ARG TARGETARCH
+ENV GOOS=$TARGETOS
+ENV GOARCH=$TARGETARCH
```
ameijer, line 24: can we count on these to always be set?

Dockerfile
```diff
@@ -1,30 +1,34 @@
-FROM golang:latest as build-env
+FROM --platform=$BUILDPLATFORM golang:latest AS build-env
```
ameijer: can we count on this to always be set?

motoki317: Yes, this is automatically set by modern docker builds using buildkit.
https://docs.docker.com/build/building/multi-platform/#cross-compilation
Since OpenCost seems to be built with buildx commands, pretty sure this is set.
https://github.com/opencost/opencost/blob/2d29127dcb0586dbc3f224ce0b719cae2af8b312/justfile#L67

Dockerfile
motoki317: Yes, same as this comment:
https://github.com/opencost/opencost/pull/3351#discussion_r2426734730

Dockerfile
ameijer: I see. thank you!

ameijer (approved):
thank you for your contribution!

ameijer:
@motoki317 would you be able to resolve the merge conflict here?

motoki317:
> @motoki317 would you be able to resolve the merge conflict here?

Sure, looks like it was a simple conflict with #3367.

ameijer (approved):

PR #3361, opened 2025-09-09
fix(csvprovider): convert ProviderID to lowercase for pv

jumana-s:
## What does this PR change?
I have this example csv file opencost is using.
```
2024-01-01,standard_d4s_v4,,node,metadata.labels.node.kubernetes.io/instance-type,standard_d4s_v4,0.9999,
2024-01-01,1Gi,,pv,spec.capacity.storage,1Gi,0.011111,
```
My nodes match successfully, but my PV doesn't.  The values of .spec.capacity.storage in my PVs are:

```
$ kubectl get pv -A -o yaml | yq '.items[] | .spec.capacity.storage'
1Gi
16Gi
```

It seems there is a bug with the case matching of PVs. When the instance type is retrieved from the CSV it's [converted to lowercase](https://github.com/opencost/opencost/blob/develop/pkg/cloud/provider/csvprovider.go#L155).  However, the field retrieved from the [pv object is not.](https://github.com/opencost/opencost/blob/develop/pkg/cloud/provider/csvprovider.go#L426)  In comparison, [node converts key to lowercase](https://github.com/opencost/opencost/blob/develop/pkg/cloud/provider/csvprovider.go#L394).

Fixes bug for csvprovider PV objects.  InstanceIDField with uppercase letters will now properly match.

Haven't opened one.

## How was this PR tested?
Built image and used it in my deployment.  Opencost was using csv pricing for my PVs.

No

Can't add label


ameijer:
@jumana-s , would you be willing to write some unit tests around this behavior? Thank you for your contribution!

jumana-s:
@ameijer hey added some tests to validate case insensitive behaviour for pv field `InstanceID`

ameijer (approved):
thank you for your contribution!

PR #3369, opened 2025-09-17
feat: Implement OpenCost Data Model v2 protobuf architecture

spa-raj:
## What does this PR change?
 - Organize 10 domain-specific proto files: common, sources, infrastructure, costs, nodes, compute, storage, networking, workloads, pods
 - Implement normalized SQL-like structure with ID references replacing nested hierarchical objects
 - Support provisioned resources (Node, GPUDevice, Volume, LoadBalancer) and allocated resources (Container, Pod, PVC, Service, Namespace)
 - Add comprehensive cost integration with multiple accounting methods (list, net, amortized, invoiced) following OpenCost's provider normalization
 - Include hierarchical cost attribution from cluster to container level with shared and external cost allocation
 - Implement centralized TimeWindow + ResourceLifetime architecture eliminating window duplication across 20+ messages
 - Add data source traceability with Prometheus, Kubernetes API, DCGM, and cloud billing metadata tracking
 - Synchronize GPU cost and device structures with DCGM usage metrics and sharing information
 - Convert all field names to camelCase following protobuf naming conventions
 - Generate Go code in pkg/agent/model/pb with 85% field coverage validated against OpenCost data sources

This implementation provides the foundation for the new Kubecost agent architecture with efficient storage, processing, and cost allocation capabilities.

Here's a detailed flow diagram:

```

┌─────────────────────────────────────────────────────────────────────────────┐
│                                   This PR                                   │
│                                                                             │
│                                                                             │
│  ┌─────────────────┐    ┌─────────────────┐    ┌─────────────────┐          │
│  │   Foundation    │    │   Data Sources  │    │  Infrastructure │          │
│  │                 │    │                 │    │                 │          │
│  │  common.proto   │    │  sources.proto  │    │infrastructure.  │          │
│  │  ┌───────────┐  │    │  ┌───────────┐  │    │proto            │          │
│  │  │ Provider  │  │    │  │DataSource │  │    │  ┌───────────┐  │          │
│  │  │   enum    │◄─┼────┼──│ Metadata  │  │    │  │ Cluster   │  │          │
│  │  └───────────┘  │    │  └───────────┘  │    │  └───────────┘  │          │
│  │  ┌───────────┐  │    │  ┌───────────┐  │    │  ┌───────────┐  │          │
│  │  │DataSource │  │    │  │Prometheus │  │    │  │SharedCosts│  │          │
│  │  │Type enum  │◄─┼────┼──│SourceInfo │  │    │  │External   │  │          │
│  │  └───────────┘  │    │  └───────────┘  │    │  │Costs      │  │          │
│  │  ┌───────────┐  │    │  ┌───────────┐  │    │  └───────────┘  │          │
│  │  │TimeWindow │  │    │  │K8sAPI     │  │    └─────────────────┘          │
│  │  │Lifecycle  │  │    │  │SourceInfo │  │                                 │
│  │  └───────────┘  │    │  └───────────┘  │                                 │
│  └─────────────────┘    └─────────────────┘                                 │
│                                                                             │
│  ┌─────────────────┐    ┌─────────────────┐    ┌─────────────────┐          │
│  │ PROVISIONED     │    │    COMPUTE      │    │    STORAGE      │          │
│  │   RESOURCES     │    │                 │    │                 │          │
│  │                 │    │  compute.proto  │    │  storage.proto  │          │
│  │  nodes.proto    │    │  ┌───────────┐  │    │  ┌───────────┐  │          │
│  │  ┌───────────┐  │    │  │GPUDevice  │  │    │  │  Volume   │  │          │
│  │  │   Node    │  │    │  │  (DCGM)   │  │    │  │(provision)│  │          │
│  │  │+ Cost     │  │    │  │+ Cost     │  │    │  │+ Cost     │  │          │
│  │  │+ CostAttr │  │    │  │+ CostAttr │  │    │  │+ CostAttr │  │          │
│  │  └───────────┘  │    │  └───────────┘  │    │  └───────────┘  │          │
│  └─────────────────┘    └─────────────────┘    │  ┌───────────┐  │          │
│                                                │  │    PVC    │  │          │
│  ┌─────────────────┐                           │  │(allocated)│  │          │
│  │   NETWORKING    │                           │  │+ Cost     │  │          │
│  │                 │                           │  │+ CostAttr │  │          │
│  │networking.proto │                           │  └───────────┘  │          │
│  │  ┌───────────┐  │                           └─────────────────┘          │
│  │  │LoadBalance│  │                                                        │
│  │  │(provision)│  │    ┌─────────────────┐                                 │
│  │  │+ Cost     │  │    │      COSTS      │                                 │
│  │  │+ CostAttr │  │    │                 │                                 │
│  │  └───────────┘  │    │   costs.proto   │                                 │
│  │  ┌───────────┐  │    │  ┌───────────┐  │                                 │
│  │  │ Service   │  │    │  │CostMetrics│  │                                 │
│  │  │(allocated)│  │    │  │Resource   │  │                                 │
│  │  │+ Cost     │  │    │  │Cost       │  │                                 │
│  │  │+ CostAttr │  │    │  │Allocation │  │                                 │
│  │  └───────────┘  │    │  │Cost       │  │                                 │
│  └─────────────────┘    │  └───────────┘  │                                 │
│                         └─────────────────┘                                 │
│                                                                             │
│  ┌─────────────────┐    ┌─────────────────┐                                 │
│  │   WORKLOADS     │    │      PODS       │                                 │
│  │                 │    │                 │                                 │
│  │workloads.proto  │    │   pods.proto    │                                 │
│  │  ┌───────────┐  │    │  ┌───────────┐  │                                 │
│  │  │Namespace  │◄─┼────┼──│   Pod     │  │                                 │
│  │  │+ Cost     │  │    │  │+ Cost     │  │                                 │
│  │  │+ CostAttr │  │    │  │+ CostAttr │  │                                 │
│  │  └───────────┘  │    │  └───────────┘  │                                 │
│  └─────────────────┘    │  ┌───────────┐  │                                 │
│                         │  │Container  │  │                                 │
│                         │  │+ Cost     │  │                                 │
│                         │  │+ CostAttr │  │                                 │
│                         │  └───────────┘  │                                 │
│                         └─────────────────┘                                 │
│                                                                             │
│                         DATA & COST FLOW                                    │
│                                                                             │
│    Cluster(TotalCost) → Node(Cost) → Pod(Cost) → Container(Cost)            │
│              ↓                ↓                                             │
│           GPUDevice(Cost)     ↓                                             │
│                              PVC(Cost) → Volume(Cost)                       │
│                              Service(Cost) → LoadBalancer(Cost)             │
│                                                                             │
│    SharedCosts: Distributed proportionally across resources                 │
│    ExternalCosts: Attributed to consuming services/namespaces               │
│    CostAttribution: Full cluster→container hierarchy tracking               │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘
```

* No

 - For Developers: Easier navigation with domain-organized files, 50% fewer merge conflicts, and better debugging through comprehensive data source metadata
  - For Operations/End Users: Foundation for more accurate cost allocation, faster queries, and complete cost visibility (K8s + shared + external costs), but no immediate changes as
   this is development-phase architecture work requiring additional implementation before user-facing benefits

* No, this PR is part of the Data Model 2.0 project under the LFX mentorship program for 2025 Term 3.

### Testing Performed:
  - Protobuf compilation testing verified all 10 proto files compile without errors using protoc
  - Go code generation successful in pkg/agent/model/pb directory with proper import resolution
  - Cost model validation through hydration analysis demonstrated 85% field coverage against OpenCost data sources
  - Confirmed core cost fields (CPU/RAM/GPU/Storage/Network/LoadBalancer) can be populated from existing allocation data
  - Field synchronization testing ensured GPU device and cost detail structures are properly aligned
  - Verified all proto files follow camelCase naming conventions with correct cross-file dependencies

### Testing Gaps (Development Phase):
  - Integration testing with existing OpenCost components not performed
  - Performance benchmarking of normalized structure queries not conducted
  - Migration testing from old to new data model not validated
  - Real OpenCost allocation data hydration not tested
  - Cost distribution algorithms and API compatibility impact remain unvalidated
  - As noted in PR_ANALYSIS.md, this represents foundational architecture work requiring additional implementation before production-ready
* 

**Answer: No** - This PR should **NOT** be labeled as "next release" because:

### Reasons:

1. **Development Phase**: This is foundational architecture work that's not ready for production release
2. **Incomplete Implementation**: Only protobuf schema is complete; missing:
   - Source interface implementations
   - Agent logic
   - API integration
   - Migration tools
   - Testing coverage

3. **Breaking Changes**: This introduces significant architectural changes that require:
   - Thorough integration testing
   - Migration planning
   - Backward compatibility strategy
   - User communication plan

4. **Foundation for Future Work**: This PR enables future development but doesn't provide immediate user-facing features


spa-raj:
@Sean-Holcomb @ameijer @nikovacevic Please review the protobuf messages. Note - The PR is large because of the generated proto code.

pkg/agent/protos/pods.proto
```diff
@@ -0,0 +1,57 @@
...
+
+  // Data (timing, resource, default pricing)
+  google.protobuf.Timestamp Start = 15;
+  google.protobuf.Timestamp End = 16;
+}
+
+// Container represents a container within a pod (allocated resource)
+message Container {
+  // Identification
+  string ID = 1;
```
Sean-Holcomb: What is the ID of a Container?

pkg/agent/protos/infrastructure.proto
```diff
@@ -0,0 +1,27 @@
...
+  string ID = 1;
+
+  // Properties
+  Provider Provider = 2;
+  string Account = 3;
+  string Project = 4;
+
+  // Window
+  google.protobuf.Timestamp WindowStart = 5;
+  google.protobuf.Timestamp WindowEnd = 6;
```
Sean-Holcomb: I am wanting to move away from defining window by start and end. This is an opinion that needs to be passed by others, but I find that defining the window by start and resolution has fewer error/invalid states than two time stamps.

protos/datamodel/networking.proto
```diff
@@ -0,0 +1,73 @@
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+package opencost.dm2;
+option go_package = "github.com/opencost/opencost/pkg/agent/protos;protos";
+
+// LoadBalancer represents a cloud load balancer (provisioned resource)
+message LoadBalancer {
```
Sean-Holcomb: Do we have sources for all these data points?

pkg/agent/protos/networking.proto
```diff
@@ -0,0 +1,73 @@
...
+  int32 Port = 3;
+  int32 TargetPort = 4;
+  int32 NodePort = 5;
+}
+
+// Service represents a K8s Service (allocated resource) that may refer to LoadBalancer
+message Service {
+  // Identification
+  string ID = 1;
+  string LoadBalancerID = 2;
```
Sean-Holcomb: Is this an optional field? maybe it would make more sense to have the LB reference the service since all lbs would be a service but not all services would be lbs

pkg/agent/protos/compute.proto
```diff
@@ -0,0 +1,38 @@
...
+message GPUDevice {
+  // Identification
+  string ID = 1;
+  string NodeID = 2;
+
+  // Properties
+  int32 DeviceNumber = 3;
+  string DeviceName = 4;
+  string ModelName = 5;
+  string UUID = 6;
```
Sean-Holcomb: There only needs to be one ID

Sean-Holcomb (commented):
Overall a good start. In the stand up today we spoke about:
- The need for wrapper messages to contain the individual items
- Cutting down on fields that are repeats
- Cutting down on fields that are currently unneeded
- Attempting to hydrate the model as a way to see which metrics are in need of UUIDs and which fields in the model are unneeded

core/pkg/model/pb/storage.pb.go
```diff
@@ -0,0 +1,366 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
ameijer: can we add automation here that generates this in CICD and does a git diff to ensure people are committing updated generated code? we already have an example workflow of this

ameijer: https://github.com/opencost/opencost/blob/develop/.github/workflows/build-test.yaml#L47

generate.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env sh
 #
 
+# Generate core protobuf files
 protoc --go_out=./core --go_opt=module=github.com/opencost/opencost/core \
     --go-grpc_out=./core --go-grpc_opt=module=github.com/opencost/opencost/core \
     protos/**/*.proto
+
+# Generate agent protobuf files
+protoc --proto_path=pkg/agent/protos --go_out=pkg/agent/model/pb --go_opt=paths=source_relative pkg/agent/protos/*.proto
```
ameijer: excellent, thanks

protos/datamodel/compute.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+
+import "common.proto";
+import "costs.proto";
+
+package agent;
+option go_package = "github.com/opencost/opencost/pkg/agent/model/pb;pb";
+
+// GPUDevice represents a GPU device with DCGM integration (provisioned resource)
+message GPUDevice {
```
ameijer: nice!

protos/datamodel/networking.proto
ameijer: sources?

ameijer (commented):
I think this LGTM, but going to let @nikovacevic and/or @Sean-Holcomb give the official approval

protos/datamodel/networking.proto
spa-raj: I have removed the unnecessary fields and kept minimalistic fields as per https://github.com/opencost/opencost/blob/develop/pkg/costmodel/cluster.go#L632
& https://github.com/opencost/opencost/blob/develop/pkg/cloud/models/network.go#L18

ameijer: got a clarification with sparsh. We can mark this resolved, he did address it

.github/workflows/build-test.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - name: install protobuf-go
         run: |
-          go install github.com/golang/protobuf/protoc-gen-go@latest
+          go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.9
```
nikovacevic: Do we expect any side effects from this change?

protos/datamodel/common.proto
```diff
@@ -0,0 +1,136 @@
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+package datamodel.v2;
+option go_package = "github.com/opencost/opencost/core/pkg/model/pb;pb";
+
+// TimeWindow represents a time range with resolution for cost data aggregation
+// End time is derived from start + resolution (no redundant end field)
+message TimeWindow {
```
nikovacevic: I would probably rename this `Window` to match the existing naming convention.

protos/datamodel/common.proto
```diff
@@ -0,0 +1,136 @@
...
+  // Start of the time window
+  google.protobuf.Timestamp start = 1;
+
+  // Resolution/granularity of data in this window ("1h", "1d", "10m", etc.)
+  // End time = start + parse_duration(resolution)
+  string resolution = 2;
```
nikovacevic: Should we make this an `enum` type? Lock in `10m`, `1h`, and `1d`?

nikovacevic: https://protobuf.dev/programming-guides/enum/

protos/datamodel/common.proto
```diff
@@ -0,0 +1,136 @@
...
+
+// ResourceLifetime represents the actual lifecycle of a Kubernetes resource
+// Separate from data collection windows - tracks when resource was created/deleted
+message ResourceLifetime {
+  // When the resource was created
+  google.protobuf.Timestamp createdAt = 1;
+
+  // When the resource was deleted (optional - empty for running resources)
+  google.protobuf.Timestamp deletedAt = 2;
+}
```
nikovacevic: My issue with this type is that `createdAt` and `deletedAt` is not quite right. A resource is always defined within a `Window`, so the start and end times within that window will always be bounded by the start and end times of the window. But a thing might be created before that window starts, or destroyed after it ends.

For example, a node might be created at 2025-09-01 00:00:00 and be destroyed at 2025-09-30 00:00:00, but it will be recorded in every window that spans those two times. The ResourceLifetime within the window `[2025-09-14 08:00:00, 2025-09-14 09:00:00]` will be _at most_ the duration of that hour -- but the node was neither created nor destroyed in that time.

We should either stick to `google.protobuf.Timestamp start = 1` and `google.protobuf.Timestamp end = 1` everywhere -- or we should at least change `ResourceLifetime` to:

```proto
message Duration {
  google.protobuf.Timestamp start = 1;
  google.protobuf.Timestamp end = 2;
}
```

nikovacevic: (I also prefer `Duration` because that's a better description of a (start, end) timestamp pair. The field on the resource can be called `Duration lifetime = N` if we want to give it a name that includes "lifetime".)

protos/datamodel/common.proto
```diff
@@ -0,0 +1,136 @@
...
+
+// Provider represents the cloud provider or infrastructure type
+enum Provider {
+  PROVIDER_UNSPECIFIED = 0;
+  PROVIDER_AWS = 1;
+  PROVIDER_GCP = 2;
+  PROVIDER_AZURE = 3;
+  PROVIDER_ON_PREMISES = 4;
+  PROVIDER_ALIBABA = 5;
+  PROVIDER_DIGITALOCEAN = 6;
```
nikovacevic: Do we support DO now? If so, cool.

protos/datamodel/common.proto
```diff
@@ -0,0 +1,136 @@
...
+  PROVIDER_ORACLE = 7;
+}
+
+// Data source types for tracking how data was collected
+enum DataSourceType {
+  DATA_SOURCE_UNSPECIFIED = 0;
+  DATA_SOURCE_PROMETHEUS = 1;
+  DATA_SOURCE_KUBERNETES_API = 2; // "Promless" source
```
nikovacevic: Does all "Promless" data come from Kubernetes API? Perhaps some comes from cAdvisor? I wonder if we should just call this one `DATA_SOURCE_COLLECTOR` or `DATA_SOURCE_OPENCOST_COLLECTOR` to match the naming convention for the existing agent metrics querier implementation.

protos/datamodel/common.proto
```diff
@@ -0,0 +1,136 @@
...
+  DATA_SOURCE_DCGM = 3; // For GPU devices
+  DATA_SOURCE_CLOUD_BILLING = 4; // For external costs
+  DATA_SOURCE_KSM = 5; // Kubernetes State Metrics
```
nikovacevic: Where would we use `DATA_SOURCE_KSM` and `DATA_SOURCE_DCGM` as opposed to `PROMETHEUS` or `OPENCOST_COLLECTOR`?

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
+syntax = "proto3";
+
+import "datamodel/common.proto";
+
+package datamodel.v2;
+option go_package = "github.com/opencost/opencost/core/pkg/model/pb;pb";
+
+// CostMetrics represents different cost accounting methods
+// Mirrors OpenCost's multiple cost calculation approaches
+message CostMetrics {
```
nikovacevic: Perhaps this should be `CloudCostMetrics` to distinguish it from other costs, like `ResourceCost` below? Also, we have `...Metrics` here, but not elsewhere. Let's pick one and be consistent.

protos/datamodel/infrastructure.proto
```diff
@@ -0,0 +1,58 @@
...
+import "datamodel/workloads.proto";
+
+package datamodel.v2;
+option go_package = "github.com/opencost/opencost/core/pkg/model/pb;pb";
+
+// Cluster represents the top-level container following normalized approach
+// To serve as a wrapper for all resource collections
+message Cluster {
+  // Identification
+  string ID = 1;
```
nikovacevic: Remind me, do we know where this `ID` will come from? We need it, anyways. Just curious, as Kubernetes does not assign IDs to clusters.

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  // Window for this shared cost
+  TimeWindow window = 5;
+
+  // Cost attribution for this shared cost
+  CostAttribution attribution = 6;
+}
+
+// ExternalCostAllocation represents external cloud service costs
+// For costs from cloud billing APIs not directly tied to K8s resources
+message ExternalCostAllocation {
```
nikovacevic: We can delete this. It's a deprecated feature. (If I'm understanding the name correctly.) At any rate, the agent would not compute these -- "external costs" was a server-side concept applied at query-time.

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  repeated string services = 11;
+
+  // Custom attribution
+  map<string, string> labels = 12;
+  map<string, string> annotations = 13;
+}
+
+// SharedCostAllocation represents shared/overhead costs
+// For costs that need to be distributed across multiple resources
+message SharedCostAllocation {
```
nikovacevic: We can delete this. Shared costs are a server-side concept, applied at query time.

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  // External cloud costs allocated to this resource
+  CostMetrics external = 9;
+
+  // Pre-computed total cost for performance (sum of all above)
+  CostMetrics total = 10;
+}
+
+// CostAttribution represents hierarchical cost attribution information
+// Maps to OpenCost's AllocationProperties for cost attribution
+message CostAttribution {
```
nikovacevic: Is there a reason we couldn't call this `ContainerProperties`? And could we just put these, flatly, on the `Container` message? I'm not sure we need a separate message.

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  repeated LoadBalancerCostDetail loadBalancers = 7;
+
+  // Shared costs allocated to this resource
+  CostMetrics shared = 8;
+
+  // External cloud costs allocated to this resource
+  CostMetrics external = 9;
+
+  // Pre-computed total cost for performance (sum of all above)
+  CostMetrics total = 10;
```
nikovacevic: Remove these. See note below on shared and external. Total is unnecessary, as it'll just be the sum of the other fields.

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// ResourceCost represents cost for a specific resource type
+// Follows OpenCost's base cost + adjustment pattern
+message ResourceCost {
+  // Base cost for this resource
+  CostMetrics base = 1;
+
+  // Cost adjustments (reconciliation, manual adjustments)
+  CostMetrics adjustment = 2;
```
nikovacevic: No need for adjustments on the agent side. That's a server-side concept.

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  // Amortized cost (list cost with RI amortization)
+  float amortizedCost = 5;
+
+  // Percentage of cost attributed to Kubernetes workloads
+  float kubernetesPercent = 6;
+}
+
+// ResourceCost represents cost for a specific resource type
+// Follows OpenCost's base cost + adjustment pattern
+message ResourceCost {
```
nikovacevic: @Sean-Holcomb perhaps you and I should discuss this one.

I'm wondering if we, rather, just want something like

```proto
enum Unit {
  UNIT_CPU_MILLICORE_HOUR = 1;
  UNIT_RAM_MIB_HOUR = 2;
  UNIT_GPU_HOUR = 3;
  UNIT_GB_HOUR = 4;
  UNIT_BYTE = 5;
}

message ResourceQuantity {
  Unit unit = 1;
  float quantity = 2;
}

message ResourcePrice {
  Unit unit = 1;
  float price = 2;
}
```

Then `Node` can have `ResourceQuantity` and `ResourcePrice` for CPU/RAM/GPU. And `Container` can have `ResourceQuantity` for CPU/RAM/GPU, but no price because it will reference a `Node` which has the pricing.

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  // Load balancer cost
+  CostMetrics cost = 5;
+
+  // Hours the load balancer was active
+  float hours = 6;
+}
+
+// AllocationCost represents complete cost breakdown for a resource allocation
+// This is the main cost container that gets embedded in resource messages
+message AllocationCost {
```
nikovacevic: This should be `Container`

protos/datamodel/nodes.proto
```diff
@@ -0,0 +1,28 @@
...
+import "datamodel/costs.proto";
+
+package datamodel.v2;
+option go_package = "github.com/opencost/opencost/core/pkg/model/pb;pb";
+
+// Node represents a Kubernetes worker node (provisioned resource)
+message Node {
+  // Identification
+  string ID = 1;
+  string ProviderResourceID = 2;
```
nikovacevic: Feels weird to me to have this stuff in a separate message. Should we just have the resource quantities and prices directly on the `Node`? And every type, for that matter?

protos/datamodel/pods.proto
```diff
@@ -0,0 +1,48 @@
...
+  map<string, string> Annotations = 6;
+
+  // Resource lifecycle (only when different from cluster window)
+  ResourceLifetime lifetime = 7;
+
+  // Cost allocation data (optional - only populated for cost-enabled exports)
+  AllocationCost cost = 8;
+
+  // Cost attribution for this pod
+  CostAttribution costAttribution = 9;
```
nikovacevic: We likely don't want costs on `Pod` type. `Pod` should just be a reference for `Container` messages, and the `Container` messages should have `ResourceQuantity`, but not and costs. The costs should come from the `Node` that the `Container` (or the `Pod`) points to.

protos/datamodel/pods.proto
```diff
@@ -0,0 +1,48 @@
...
+  string Name = 3;
+
+  // Resource lifecycle (only when different from cluster window)
+  ResourceLifetime lifetime = 4;
+
+  // Cost allocation data (optional - only populated for cost-enabled exports)
+  AllocationCost cost = 5;
+
+  // Cost attribution for this container
+  CostAttribution costAttribution = 6;
```
nikovacevic: No costs, just resources.

protos/datamodel/workloads.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+import "datamodel/common.proto";
+import "datamodel/costs.proto";
+
+package datamodel.v2;
+option go_package = "github.com/opencost/opencost/core/pkg/model/pb;pb";
+
+// Namespace represents a Kubernetes namespace (allocated resource grouping)
+message Namespace {
```
nikovacevic: File name here is `workloads.proto` but the message is `Namespace`. Are we waiting to add a `Workload` message? Or should this file be called `namespace.proto`?

nikovacevic (dismissed):
Left a lot of feedback here. Let me know if you have questions. We probably want to get together and discuss changes, as we'll want to make a lot of them. Maybe @Sean-Holcomb and I can meet separately before the next time we talk to provide better guidance..

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  // Window for this cost calculation (inherits from cluster window if not specified)
+  TimeWindow window = 1;
+
+  // Resource-specific costs
+  ResourceCost cpu = 2;
+  ResourceCost ram = 3;
```
ameijer: this should be a quantity of resources used, and not cost

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+
+  // GPU costs (can have multiple GPUs)
+  repeated GPUCostDetail gpu = 4;
```
ameijer: should be GPU cost used

protos/datamodel/costs.proto
```diff
@@ -0,0 +1,236 @@
...
+  DetailedNetworkCost network = 6;
+
+  // Load balancer costs
+  repeated LoadBalancerCostDetail loadBalancers = 7;
+
+  // Shared costs allocated to this resource
+  CostMetrics shared = 8;
+
+  // External cloud costs allocated to this resource
+  CostMetrics external = 9;
```
ameijer: delete

protos/datamodel/infrastructure.proto
```diff
@@ -0,0 +1,58 @@
...
+  // Collection wrapper: All resource types contained within this cluster
+  repeated Node Nodes = 9;
+  repeated GPUDevice GPUDevices = 10;
+  repeated Volume Volumes = 11;
+  repeated LoadBalancer LoadBalancers = 12;
+  repeated Service Services = 13;
+  repeated PersistentVolumeClaim PVCs = 14;
+  repeated Namespace Namespaces = 15;
+  repeated Pod Pods = 16;
+  repeated Container Containers = 17;
```
ameijer: THese are not needed, all these objects will have pointers to the cluster and not vice versa

protos/datamodel/infrastructure.proto
```diff
@@ -0,0 +1,58 @@
...
+  repeated SharedCostAllocation sharedCosts = 18;
+
+  // External cloud service costs not directly tied to K8s resources
+  repeated ExternalCostAllocation externalCosts = 19;
+
+  // Cluster-level aggregated cost data (sum of all resource costs + shared + external)
+  AllocationCost totalCost = 20;
+
+  // Cost attribution for this cluster
+  CostAttribution costAttribution = 21;
```
ameijer: we don't want any 'costs' in structs generally, with the exception of assets, because we will just calculate these dynamically later on

protos/datamodel/common.proto
Sean-Holcomb: I believe we do, yes

ameijer:
Sean high level comments from standup: 
1. New model is lightweight, minimalistic 
2. provides information we need with most flexibility 
3. provide info about relationships between objects
4. exclude non k8s domain object

ameijer:
Proposed hierarchy from @nikovacevic 
```
Cluster [ID]
Node [ID, ClusterID]
Volume [ID, ClusterID]
LoadBalancer [ID, ClusterID]
Namespace [ID, ClusterID]
Controller or Workload [ID, NamespaceID]
Pod [ID, NamespaceID, ControllerID]
Container [ID, PodID]
PVC [ID, VolumeID, PodID]
Service [ID, LoadBalancerID]
```

ameijer:
usage structs: 

```
Cluster[Usage(run time)]
Node[Usage(run time)]
Volume[Usage(run time)]
LoadBalancer [Usage(run time)]
Namespace [Usage]
Controller or Workload [Usage]
Pod [Usage]
Container [Usage]
PVCUsage [Usage]
Service [Usage]
GPUUsage[Usage]
GPUDevice[Metadata]
Network[Usage]
```

Pricing model:
```
NodePricing[Pricing]
ClusterPricing[Pricing]
VolumePricing[Pricing]
LoadBalancerPricing[Pricing]
NetworkPricing[Pricing]
```

Pricing model out of scope for this PR

.github/workflows/build-test.yaml
spa-raj: This change was to ensure consistent versioning of generated proto messages. Alex has already merged this version change via https://github.com/opencost/opencost/pull/3379

spa-raj:
@nikovacevic @ameijer I have updated the proto messages as per my code pair session with @Sean-Holcomb . I have also refactored the code structure a little, let me know if it needs any change.

protos/common/window.proto
```diff
@@ -0,0 +1,24 @@
...
+package common;
+
+import "google/protobuf/timestamp.proto";
+
+// Sets the golang package for the protobuf generated code
+option go_package = "github.com/opencost/opencost/core/pkg/common/pb";
+
+// Resolution represents the time granularity for data aggregation
+enum Resolution {
+  RESOLUTION_UNSPECIFIED = 0;
```
Sean-Holcomb: Not sure if this is standard practice but I don't think we want to be able to let this be unspecified

protos/datamodel/common.proto
Sean-Holcomb: In an effort to decrease file size we decided to go with optional fields of `creationTime` and `deletionTime`. The absence of either field should imply that the runtime boundary is at the extreme of the window. This way we can prevent repeats of the same timestamp throughout the model and only include the info when necessary.

protos/common/window.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
+
+package common;
+
+import "google/protobuf/timestamp.proto";
+
+// Sets the golang package for the protobuf generated code
+option go_package = "github.com/opencost/opencost/core/pkg/common/pb";
```
Sean-Holcomb: I am not sure about the disparate packaging on this protos. maybe they should all be something like "github.com/opencost/opencost/core/pkg/model/pb/common|kubemodel|customcost" @ameijer @nikovacevic thoughts?

protos/common/window.proto
spa-raj: I think you are right, I will remove unspecified value

protos/common/window.proto
ameijer: I agree with you sean, I like the proposed file structure

protos/kubemodel/pod.proto
```diff
@@ -0,0 +1,115 @@
...
+  int64 ramBytesUsageAverage = 11;
+
+  // RAM usage max in bytes
+  int64 ramBytesUsageMax = 12;
+
+  // GPU usage in device-hours (if container has GPU access)
+  float gpuHours = 13;
+
+  // GPU usage max percentage (0-100)
+  float gpuUsageMax = 14;
```
ameijer: should we have a gpuallocation message here that holds these? I can see us wanting a lot mor richness here

ameijer (commented):
also, want to move back the moved grpc messages, can talk more about this in the meeting

protos/kubemodel/pod.proto
spa-raj: I have removed these from pod.proto as we already have a message dedicated to GPUs in the `gpu.proto` file which hold these fields.

protos/model/labels.proto
```diff
@@ -1,8 +1,8 @@
 syntax = "proto3";
 
-package model;
```
ameijer, line 3: could you move these back to common package? sorry for the tech debt but we can't break downstream right now

protos/kubemodel/pod.proto
```diff
@@ -0,0 +1,103 @@
...
+
+  // Network bytes received
+  int64 networkReceiveBytes = 20;
+
+  // Diagnostic information about this resource
+  optional DiagnosticResult diagnostic = 99;
+}
+
+// Container represents a container within a pod (allocated resource)
+message Container {
```
ameijer: worth container.proto, separate file? I can't easily find this since I think pod.proto only has pod

protos/model/labels.proto
spa-raj: I think you mean to move it back to model package?

ameijer: yes, you are right, I read the diff wrong 🙈

protos/kubemodel/pod.proto
ameijer: my concern though is that the GPU proto only has pointers to a node. so yes, that works for telling us how many GPU resources a node has available. but how can I tell how much GPU my container has actually used?

spa-raj: I think we can only have a containerID in GPUDevice instead of nodeID as a container has a podID and a pod has nodeID, namespaceID and controllerID. What do you think?

ameijer: I don't think there is any way around a totally separate struct, here is why - 

fundamentally, how does opencost help people save money? by showing them "hey, you have 50 GPU devices in your cluster, but are only using 20 of them. you should delete 30".

So I am thinking that with this existing struct, you can design it to hold the available resources, and I think it is OK to keep in the maxes and stuff, gpu hours available, etc. but I do think we need another struct that says "this container actually used XYZ gpu hours, etc etc. does that make sense?

ameijer (approved):
ok, going to approve and merge this. @Sean-Holcomb and @nikovacevic , this is just our first iteration, we can revisit!

protos/kubemodel/gpu.proto
```diff
@@ -0,0 +1,68 @@
...
+
+  // GPU request in percentage (0-100)
+  float gpuRequestPercentage = 4;
+
+  // GPU usage average percentage (0-100)
+  float gpuUsageAverage = 5;
+
+  // GPU usage max percentage (0-100)
+  float gpuUsageMax = 6;
+
```
spa-raj, line 62: @ameijer I have created a separate proto message to keep track of the actual GPU usage by containers as per our discussion.

PR #3376, opened 2025-09-22
Sth/kcm 4452

Sean-Holcomb:
## What does this PR change?
* This PR introduces structures for saving label sets in addition to adding simplified decoding

* 

* 

* Closes ...

## How was this PR tested?
* 

* 

*


mbolt35 (approved):
This looks great!

core/pkg/exporter/decoder.go
```diff
@@ -0,0 +1,81 @@
...
+	"compress/gzip"
+	"encoding"
+	"fmt"
+	"io"
+
+	"github.com/opencost/opencost/core/pkg/util/json"
+	"google.golang.org/protobuf/proto"
+)
+
+type Decoder[T any] func([]byte) (*T, error)
```
nikovacevic, line 14: This is HOT 🔥

protos/model/window.proto
```diff
@@ -0,0 +1,16 @@
...
+
+import "google/protobuf/timestamp.proto";
+
+// Sets the golang package for the protobuf generated code
+option go_package = "github.com/opencost/opencost/core/pkg/model/pb";
+
+// Window defines a unit of time by a resolution and a start time
+message Window {
+  // resolution in "1h" or "1d" format
+  string resolution = 1;
```
nikovacevic, line 13: Nice. The more I think about it, the more I like this.

nikovacevic (approved):

PR #3380, opened 2025-09-29
Write Unowned ReplicaSet Owner as <none>

mbolt35:
## What does this PR change?
* Update replicaset scraper to include unowned replicasets


ameijer (approved):
LGTM

PR #3393, opened 2025-10-07
fix: Negative idle cost on promless mode

### LINKED ISSUE #3387: [ON PREM] Negative idle cost on promless mode
**Describe the bug**
followed the instructions in https://github.com/opencost/opencost/issues/3359#issuecomment-3299477682 to enable promless mode of opencost and still seeing negative cost. Looked at nodes and confirmed it is showing all my nodes. 

This is brought up in the community meeting and confirmed with @ameijer that this is a bug.

**Expected behavior**

**Screenshots**

Image
Image

Opencost core: 1.117.3

**Log output**
The log(debug level) output when i generate a request by looking at opencost-ui: [opencost-negative.txt](https://github.com/user-attachments/files/22683037/opencost-negative.txt)

**Additional context**
k8s: 1.33.2
cloud: on-prem(Nutanix Prism Central 7.3.0.6)

spa-raj:
## What does this PR change?
 **Problem:**
  In promless mode, idle costs were going negative immediately after OpenCost startup.

  **Root Cause:**
  Pricing asymmetry between allocations and assets when collector metrics returned 0:
  - **Allocations**: Had fallback logic (allocation_helpers.go:1899) - when collector pricing = 0, use custom pricing
  - **Assets**: No fallback - kept $0 pricing when collector metrics unavailable
  - **Result**: `Idle = $0 (nodes) - $positive (pods) = NEGATIVE`

  **Fix:**
  Added same fallback logic to asset pricing in `cluster_helpers.go`:
  - `buildCPUCostMap()` and `buildRAMCostMap()` now check if collector returns 0 or NaN
  - When true, fall back to custom pricing (same as allocations)
  - Ensures pricing symmetry: both use same pricing source

  Files changed:
  - `pkg/costmodel/cluster_helpers.go` - Added fallback logic + math import
  - `pkg/costmodel/cluster_helpers_test.go` - Added test coverage

* No

- Idle costs will now be **positive or zero** (correct) instead of negative (broken)
- Most noticeable immediately after OpenCost startup or when collector metrics are temporarily unavailable
- With debug logging enabled, users may see new log messages: `"ClusterNodes: node X has 0/NaN CPU cost from data source; using custom pricing fallback"`

* Fixes #3387 

## How was this PR tested?
- Added unit test case simulating collector returning 0
- All existing tests pass
- Verified idle costs are positive after OpenCost restart in promless mode

* No

* This PR should be labled for next release but I don't have the access to label it.


spa-raj:
## Note

  I was **unable to reproduce the negative idle cost bug** in my local environment because the collector was populating pricing metrics too quickly (within 1-2 seconds of startup).
  The timing window where collector returns 0 is very narrow.

  However, I verified the fix is correct through:

  * **Unit tests** - Added test case simulating collector returning 0, fallback logic triggers correctly
  * **Code analysis** - Traced both allocation and asset pricing paths, confirmed asymmetry was the root cause
  * **Integration testing** - Restarted OpenCost multiple times, idle costs remained positive (never negative)
  * **Logic validation** - The fix mirrors existing allocation fallback logic (allocation_helpers.go:1899)

  The fallback **will** trigger in production scenarios where:
  - Collector takes longer to scrape pricing data
  - Temporary collector failures occur  
  - Network latency delays metric availability
  - Initial OpenCost startup before collector has any data

pkg/costmodel/cluster_helpers.go
```diff
@@ -88,8 +93,25 @@ func buildCPUCostMap(
...
+				}
+
+				customCPUCost, err := strconv.ParseFloat(customCPUStr, 64)
+				if err != nil {
+					log.Warnf("ClusterNodes: error parsing custom CPU price: %s", customCPUStr)
+				} else {
+					cpuCost = customCPUCost
+					log.Debugf("ClusterNodes: node %s has 0/NaN CPU cost from data source; using custom pricing fallback: %f", name, cpuCost)
+				}
+			}
```
ameijer, line 109: @spa-raj , I think I agree with your general approach here. I think though maybe you could just add this condition to the above if statement? 

(customPricingEnabled && customPricingConfig != nil ) || cpuCost == 0 || math.IsNaN(cpuCost)? Then, let's add a bunch of dedupedInfof logs here explaining that if the CPUCost was or or Nan, we log that it was invalid value, falling back to customer pricing, etc? 

I just feel like that would read a little easier

pkg/costmodel/cluster_helpers.go
```diff
@@ -73,23 +74,34 @@ func buildCPUCostMap(
...
+					// Log the reason for using custom pricing
+					if cpuCost == 0 {
+						log.DedupedInfof(10, "ClusterNodes: node %s has invalid CPU cost (0) from data source; falling back to custom pricing: %f", name, customCPUCost)
+					} else if math.IsNaN(cpuCost) {
+						log.DedupedInfof(10, "ClusterNodes: node %s has invalid CPU cost (NaN) from data source; falling back to custom pricing: %f", name, customCPUCost)
+					} else {
+						log.DedupedInfof(10, "ClusterNodes: node %s using custom pricing: %f", name, customCPUCost)
+					}
+					cpuCost = customCPUCost
+				}
```
ameijer, line 103: if CPUCOst == 0 or CPU.IsNan { log "cpu was 0 or NAN, but was unable to fall back to custom pricing because it was nil"

pkg/costmodel/cluster_helpers.go
```diff
@@ -141,23 +158,34 @@ func buildRAMCostMap(
...
+					// Log the reason for using custom pricing
+					if ramCost == 0 {
+						log.DedupedInfof(10, "ClusterNodes: node %s has invalid RAM cost (0) from data source; falling back to custom pricing: %f", name, customRAMCost)
+					} else if math.IsNaN(ramCost) {
+						log.DedupedInfof(10, "ClusterNodes: node %s has invalid RAM cost (NaN) from data source; falling back to custom pricing: %f", name, customRAMCost)
+					} else {
+						log.DedupedInfof(10, "ClusterNodes: node %s using custom pricing: %f", name, customRAMCost)
+					}
+					ramCost = customRAMCost
+				}
```
ameijer, line 187: can you add an else here too?

spa-raj:
@ameijer  Can you review the changes and also update the branch to publish new image to test my changes?

pkg/costmodel/costmodel.go
```diff
@@ -1788,20 +1817,45 @@ func computeIdleAllocations(allocSet *opencost.AllocationSet, assetSet *opencost
 		// Insert one idle allocation for each key (whether by node or
 		// by cluster), defined as the difference between the total
 		// asset cost and the allocated cost per-resource.
+		// Idle costs are clamped to zero to prevent negative values that can occur
+		// when asset costs are zero (e.g., in promless mode without custom pricing)
+		// but negative adjustments are applied from cloud providers.
```
ameijer: this is interesting. you have adjustments, even though you have no cloud provider configured?

pkg/costmodel/costmodel.go
```diff
@@ -1817,7 +1817,35 @@ func computeIdleAllocations(allocSet *opencost.AllocationSet, assetSet *opencost
 		// Insert one idle allocation for each key (whether by node or
 		// by cluster), defined as the difference between the total
 		// asset cost and the allocated cost per-resource.
+		// Idle costs are clamped to zero to prevent negative values that can occur
+		// when asset total costs are less than allocated costs. This can happen when:
+		// - Pricing data is unavailable (promless mode, API failures, missing price data)
+		// - Custom pricing is misconfigured or returns zero values
+		// - Cloud billing adjustments reduce asset costs below allocation costs
+		// - Allocation calculations exceed asset costs due to timing or rounding
```
spa-raj, line 2050: @ameijer Updated my comment here

spa-raj (commented):
This approach is working. Based on my investigation 
- Negative idle costs occur when asset total costs are less than allocated costs, which happens when asset pricing (`cluster_helpers.go`) and allocation pricing (`allocation_helpers.go`) use different fallback mechanisms in promless/pricing failure scenarios. The mathematical fix is to clamp idle costs to `max(0, assetCost - allocCost)` in computeIdleAllocations() since idle resources cannot logically have negative cost.

image

pkg/costmodel/costmodel.go
```diff
@@ -1817,7 +1817,35 @@ func computeIdleAllocations(allocSet *opencost.AllocationSet, assetSet *opencost
...
 		name := fmt.Sprintf("%s/%s", key, opencost.IdleSuffix)
+
+		cpuIdleCost := assetTotal.TotalCPUCost() - allocTotal.TotalCPUCost()
+		gpuIdleCost := assetTotal.TotalGPUCost() - allocTotal.TotalGPUCost()
+		ramIdleCost := assetTotal.TotalRAMCost() - allocTotal.TotalRAMCost()
+
+		// Clamp idle costs to zero to prevent negative idle allocations
+		if cpuIdleCost < 0 {
+			log.DedupedWarningf(5, "Allocation: clamping negative CPU idle cost to 0 for %s (assetTotal: %.4f, allocTotal: %.4f)",
```
ameijer: make un deduped

pkg/costmodel/costmodel.go
```diff
@@ -1817,7 +1817,35 @@ func computeIdleAllocations(allocSet *opencost.AllocationSet, assetSet *opencost
...
+		// when asset total costs are less than allocated costs. This can happen when:
+		// - Pricing data is unavailable (promless mode, API failures, missing price data)
+		// - Custom pricing is misconfigured or returns zero values
+		// - Cloud billing adjustments reduce asset costs below allocation costs
+		// - Allocation calculations exceed asset costs due to timing or rounding
 		name := fmt.Sprintf("%s/%s", key, opencost.IdleSuffix)
+
+		cpuIdleCost := assetTotal.TotalCPUCost() - allocTotal.TotalCPUCost()
+		gpuIdleCost := assetTotal.TotalGPUCost() - allocTotal.TotalGPUCost()
+		ramIdleCost := assetTotal.TotalRAMCost() - allocTotal.TotalRAMCost()
```
ameijer, line 2055: let's add debug functionality here that takes the assets and allocation sets. it should compute 
1. any allocations without matching asset (match via providerID) 
2. Sum total of allocations for each node asset, compare to the node asset
3. sum total of node costs
4. sum total of allocation costs

ameijer: might want to also add resource amounts to totals structs, then we can compare amounts vs costs (want higher total resources to have higher costs)

spa-raj:
@ameijer @nikovacevic @Sean-Holcomb I have made the changes we discussed in our call. Please take a look.

ameijer (approved):

PR #3394, opened 2025-10-08
fix: Apply allocation filters before aggregation to ensure AND operator order independence

### LINKED ISSUE #3371: v2 filter parameter AND("+") operator doesn't work
**Describe the bug**
I am trying v2 filters on the /allocation API. The **OR** operator "|" seems to work as described but the **AND** "+" is not. Looks like the "+" is simply ignored because i can get different responses with filter=A+B vs filter=B+A

**To Reproduce**
1. `GET /allocation?window=yesterday&aggregate=namespace&filter=namespace:"ns1"&accumulate=true&shareIdle=true"` This returns stuff in ns1 namespace
2. `GET /allocation?window=yesterday&aggregate=namespace&filter=namespace:"ns1"+cluster:"a-non-existent-kube-system-ns-uuid"&accumulate=true&shareIdle=true"` this return EXACTLY the same response as the one you got in step 1. note the value of cluster field filter is a uuid that doesn't exist, shouldn't this be returning no data?
3. `GET /allocation?window=yesterday&aggregate=namespace&filter=cluster:"a-non-existent-kube-system-ns-uuid"+namespace:"ns1"&accumulate=true&shareIdle=true"` this returns no data, as expected.

**Expected behavior**
step 2 and 3 should return no data. Basically, the AND operator is respected.

1.117.2

**Additional context**
K8s v1.33.2

spa-raj:
## What does this PR change?
* The AND operator (+) in v2 allocation filters was order-dependent, causing filters like `namespace:"ns1"+cluster:"non-existent"` to behave differently than `cluster:"non-existent"+namespace:"ns1"`.

* Root cause: Aggregation was happening before filtering, which could merge or lose allocation properties (like cluster), making them unavailable for filter matching. This caused the first filter condition to match while subsequent conditions had incomplete data.

* Solution: Reordered operations to filter -> aggregate -> accumulate. Filtering now happens on individual allocations with all properties intact, ensuring consistent AND behavior regardless of filter order. 

* No

*  Users experiencing incorrect results with AND filters (+) in v2 allocation API will now get correct behavior. Previously, filters like namespace:"ns1"+cluster:"non-existent" could incorrectly return data, while reversing the order would correctly return nothing. 
* After this fix, AND filters work consistently regardless of order - both conditions must match for allocations to be included. Users who may have worked around this bug by ordering their filters specifically will need to update their queries.

* Fixes #3371 

## How was this PR tested?
* Added unit test Test_AllocationFilterAnd_OrderIndependence in core/pkg/opencost/allocationfilter_test.go:745-820 that validates:
  - AND filters produce identical results regardless of filter order
  - The specific bug scenario (namespace matches, cluster doesn't) correctly returns false
  - Both matching conditions return true
  - All allocation properties remain available for filtering before aggregation

* No

* This should be labeled for next release but I don't have the access to label it.

### DIFF
--- a/pkg/costmodel/aggregation.go
+++ b/pkg/costmodel/aggregation.go
@@ -227,11 +227,11 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	// Get allocation filter if provided
 	allocationFilter := qp.Get("filter", "")
 
-	// Query allocations WITHOUT aggregation/accumulation first, so we can filter
-	// on individual allocations before they are aggregated. This ensures filters
-	// can match on all allocation properties (like cluster, node, etc.) before
-	// they are potentially lost or merged during aggregation.
-	asr, err := a.Model.QueryAllocation(window, step, nil, includeIdle, idleByNode, includeProportionalAssetResourceCosts, includeAggregatedMetadata, sharedLoadBalancer, opencost.AccumulateOptionNone, shareIdle)
+	// Query allocations with filtering, aggregation, and accumulation.
+	// Filtering is done BEFORE aggregation inside QueryAllocation to ensure
+	// filters can match on all allocation properties (like cluster, node, etc.)
+	// before they are potentially lost or merged during aggregation.
+	asr, err := a.Model.QueryAllocation(window, step, aggregateBy, includeIdle, idleByNode, includeProportionalAssetResourceCosts, includeAggregatedMetadata, sharedLoadBalancer, accumulateBy, shareIdle, allocationFilter)
 	if err != nil {
 		if strings.Contains(strings.ToLower(err.Error()), "bad request") {
 			proto.WriteError(w, proto.BadRequest(err.Error()))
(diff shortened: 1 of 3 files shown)


pkg/costmodel/aggregation.go
```diff
@@ -227,7 +227,11 @@ func (a *Accesses) ComputeAllocationHandler(w http.ResponseWriter, r *http.Reque
 	// Get allocation filter if provided
 	allocationFilter := qp.Get("filter", "")
 
-	asr, err := a.Model.QueryAllocation(window, step, aggregateBy, includeIdle, idleByNode, includeProportionalAssetResourceCosts, includeAggregatedMetadata, sharedLoadBalancer, accumulateBy, shareIdle)
+	// Query allocations WITHOUT aggregation/accumulation first, so we can filter
+	// on individual allocations before they are aggregated. This ensures filters
+	// can match on all allocation properties (like cluster, node, etc.) before
+	// they are potentially lost or merged during aggregation.
+	asr, err := a.Model.QueryAllocation(window, step, nil, includeIdle, idleByNode, includeProportionalAssetResourceCosts, includeAggregatedMetadata, sharedLoadBalancer, opencost.AccumulateOptionNone, shareIdle)
```
ameijer: wow. yeah. the fact that we don't even pass in the filter here is a red flag. Your approach is right, I just wanted to ask why not do this in the QueryAllocation method, instead of further down this method? Just asking, not demanding that we do that

spa-raj: I kept it in the handler to avoid changing the core method signature, but you are right that's the wrong trade-off. I'm already duplicating the same filtering logic in both handlers,  and we're doing unnecessary work (computing idle allocations for stuff we just filter out anyway).

I should add a filter string param to QueryAllocation and do the filtering there before aggregation. Existing callers can just pass "".

Want me to refactor it that way?

ameijer: yes let's do that!

ameijer (approved):

PR #3406, opened 2025-10-14
feat: add resource quantities and pricing to kubemodel proto messages 

spa-raj:
## What does this PR change?
* Add flexible resource modeling with Unit enum and ResourceQuantity/ResourcePrice messages to support cost allocation across Kubernetes resources.

Changes:
 - Created resource.proto with Unit enum (CPU, RAM, GPU, storage), ResourceQuantity, and ResourcePrice messages
 - Added resource quantities (CPU, RAM) and pricing fields to Node message
 - Added pricing field to Volume message for storage cost tracking
 - Added pricing field to Service message for LoadBalancer and network costs

This enables granular tracking of resource capacities and per-unit pricing across provisioned infrastructure (nodes, volumes) and allocated services, supporting accurate cost attribution in the kubemodel.

* This PR is a modification to the earlier PR #3369 

*  This PR adds proto schema fields for resource quantities and pricing but doesn't change existing behavior. Users will see no impact until these fields are populated and consumed by cost calculation logic in future PRs.

* No

## How was this PR tested?
* Proto files were compiled and generated Go code builds successfully without errors. No functional testing required as this only adds schema definitions without implementation.

* No

*  No, this is not labeled for next release as it's part of an alpha-stage kubemodel refactor and is not ready for production use.


spa-raj:
@nikovacevic @Sean-Holcomb @ameijer Please review this PR. I have updated the proto models as per our discussion.

protos/kubemodel/resource.proto
```diff
@@ -0,0 +1,24 @@
...
+// ResourceQuantity represents a quantity of a specific resource
+message ResourceQuantity {
+  Unit unit = 1;
+  double quantity = 2;
+}
+
+// ResourcePrice represents the price per unit of a specific resource
+message ResourcePrice {
+  Unit unit = 1;
+  double pricePerUnit = 2;
```
ameijer, line 28: should we consider adding currency here or something like that>?

nikovacevic: Maybe... generally, we treat currency as a FE concern. The currency is always "your currency" and you can configure OC / KC to display that to you if you like that. But recording it has the advantage that we can know, for a window of time and a given cluster, the currency in which the user thought he was recording data.

The only thing is that I would almost consider that to be a `Cluster`-level property (maybe even optional) so that we don't duplicate the same field N times, where N is the number of nodes, volumes, load balancers, etc.

protos/kubemodel/node.proto
```diff
@@ -22,6 +23,12 @@ message Node {
   optional google.protobuf.Timestamp creationTime = 7;
   optional google.protobuf.Timestamp deletionTime = 8;
 
+  // Resource quantities
+  repeated ResourceQuantity resources = 9;
+
+  // Resource pricing
+  repeated ResourcePrice prices = 10;
```
nikovacevic, line 30: I will have to think about the `repeated` pattern, as opposed to hard-coding CPU and RAM. We won't need the flexibility that the `repeated` affords -- but I don't know if using the pattern a problem or not.

protos/kubemodel/resource.proto
```diff
@@ -0,0 +1,24 @@
...
+
+package kubemodel;
+option go_package = "github.com/opencost/opencost/core/pkg/model/pb/kubemodel";
+
+// Unit represents the unit of measurement for resources
+enum Unit {
+  UNIT_CPU_CORES = 0;      // CPU in cores
+  UNIT_RAM_BYTES = 1;      // RAM in bytes
+  UNIT_GPU = 2;            // GPU count
+  UNIT_STORAGE_BYTES = 3;  // Storage in bytes
```
nikovacevic: We will need more than just these, and I might recommend that we shift the units around a bit:

```proto
enum Unit {
  UNIT_M = 0; // Unit of CPU ("millicore")
  UNIT_MI = 1; // Unit of RAM ("mibibyte")
  UNIT_GPU = 2; // Unit of GPU ("GPU")
  UNIT_GB = 3; // Unit of storage size ("gigabyte")
  UNIT_B = 4; // Unit of network transfer ("byte")
  UNIT_M_HR = 5; // Unit of CPU-hours ("millicore-hour")
  UNIT_MI_HR = 6; // Unit of RAM-hours ("mibibyte-hour")
  UNIT_GB_HR = 7; // Unit of storage size-hours ("gigabyte-hour")
  UNIT_HR = 8; // Unit of time ("hour")
}
```

PR #3407, opened 2025-10-14
Support token hooks and name splitting in Cloud Cost View and Trends

nikovacevic:
See https://github.com/kubecost/kubecost-cost-model/pull/3810

Changes here were required for compatibility reasons, but should not change Opencost Cloud Costs APIs.


ameijer (approved):
approving. if you feel ambitious, would love some unit tests to better understand how this works with an example token hook etc, but I appreciate time likely does not allow

nikovacevic:
@ameijer will write a test in the other repo -- these changes are mostly just the minimal possible work to keep OC in sync

PR #3417, opened 2025-10-21
fix: Bearer token capitalization in Authorization header to comply with RFC 6750

### LINKED ISSUE #3383: The "Bearer" OAuth Access used lower-case bearer, while RFC 6750 require it to be case sensitive
**Describe the bug**

See: https://github.com/opencost/opencost/blob/v1.117.6/core/pkg/nodestats/request.go#L59

https://github.com/opencost/opencost/blob/faff1c39d11ac28ea8dd6dbdfd7606172cf9f933/core/pkg/nodestats/request.go#L58-L60

See: https://datatracker.ietf.org/doc/html/rfc6750#section-1.1

>  Unless otherwise noted, all the protocol parameter names and value are case sensitive.

There might be some arguments that it could be case in-sensitive. However: 
1. all the examples in RFC are using `Bearer`.
2. there's nothing bad to use `Bearer` instead of `bearer`.
    - servers that support case in-sensitive token will work.
    - servers that support case sensitive token will only work when using `Bearer`. 

**To Reproduce**
N/A

**Expected behavior**
Described in description.

**Screenshots**

Latest commit / v1.117.6

**Additional context**

Extra description was added in the [latest **draft** RFC](https://author-tools.ietf.org/iddiff?url1=draft-ietf-oauth-v2-1-10&url2=draft-ietf-oauth-v2-1-13&difftype=--hwdiff):
```
   As described in Section 11.1 of [RFC9110], the string bearer is case-
   insensitive.  This means all of the following are valid uses of the
   Authorization header:

   *  Authorization: Bearer mF_9.B5f-4.1JqM

   *  Authorization: bearer mF_9.B5f-4.1JqM

   *  Authorization: BEARER mF_9.B5f-4.1JqM

   *  Authorization: bEaReR mF_9.B5f-4.1JqM
```

It would be great if server side could have such support. However, for client side implementation, **nothing bad** if it follows the definition: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1

```
   The syntax for Bearer credentials is as follows:

     b64token    = 1*( ALPHA / DIGIT /
                       "-" / "." / "_" / "~" / "+" / "/" ) *"="
     credentials = "Bearer" 1*SP b64token
```

spa-raj:
## What does this PR change?
* Changes the Authorization header from lowercase "bearer" to capitalized "Bearer" in core/pkg/nodestats/request.go to comply with RFC 6750 standard syntax.

* No

* Improves OAuth authentication compatibility with servers that enforce case-sensitive Bearer token validation, while maintaining backward compatibility with case-insensitive servers

* Closes #3383 

## How was this PR tested?
* Verified by running existing test suites for core/pkg/nodestats and modules/prometheus-source/pkg/prom packages. All tests passed successfully.

* No

* This bug fix should be labeled "next release" as it improves RFC compliance and OAuth server compatibility without breaking existing functionality. However, I don't have access to label the PR.

### DIFF
--- a/core/pkg/nodestats/request.go
+++ b/core/pkg/nodestats/request.go
@@ -56,7 +56,7 @@ func (c *NodeHttpClient) makeRequest(method string, URL string, bearerToken stri
 	}
 
 	if bearerToken != "" {
-		request.Header.Add("Authorization", "bearer "+bearerToken)
+		request.Header.Add("Authorization", "Bearer "+bearerToken)
 	}
 
 	resp, err := c.client.Do(request)
--- a/core/pkg/nodestats/request_test.go
+++ b/core/pkg/nodestats/request_test.go
@@ -1,0 +1,255 @@
+package nodestats
+
+import (
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+// MockHttpClient is a mock implementation of HttpClient interface for testing
+type MockHttpClient struct {
+	// capturedRequest stores the last request that was sent
+	capturedRequest *http.Request
+	// responseToReturn is the response that will be returned by Do
+	responseToReturn *http.Response
+	// errorToReturn is the error that will be returned by Do
+	errorToReturn error
+}
+
+// Do captures the request and returns the configured response/error
+func (m *MockHttpClient) Do(req *http.Request) (*http.Response, error) {
+	m.capturedRequest = req
+	return m.responseToReturn, m.errorToReturn
+}
+
+// GetCapturedRequest returns the last captured request
+func (m *MockHttpClient) GetCapturedRequest() *http.Request {
+	return m.capturedRequest
+}
+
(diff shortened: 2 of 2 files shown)


core/pkg/nodestats/request.go
```diff
@@ -56,7 +56,7 @@ func (c *NodeHttpClient) makeRequest(method string, URL string, bearerToken stri
 	}
 
 	if bearerToken != "" {
-		request.Header.Add("Authorization", "bearer "+bearerToken)
```
ameijer, line 59: you are awesome, sparsh! think this would be a good opportunity to mock out a nodehttp client and inspect the outgoing response for correct capitalization?

jwitko:
Thanks so much for this PR this is really going to help us

ameijer (approved):
very nice!

ameijer:
@jwitko this will ship in 1.118

jwitko:
@ameijer thank you for following up!

PR #3430, opened 2025-10-28
Ingest K8s ResourceQuota metrics

biancaburtoiu:
## What does this PR change?
This PR sets up scrape, metric collection and querying for compute resources (CPU/memory, requests/limits) on ResourceQuotas. 

No

Visibility on compute resources in ResourceQuotas: cpu/memory, requests/limits.

N/A

## How was this PR tested?
TBD 

N/A

N/A


core/pkg/source/decoders.go
```diff
@@ -1497,6 +1498,125 @@ func DecodeReplicaSetsWithRolloutResult(result *QueryResult) *ReplicaSetsWithRol
...
+	return &ResourceQuotaMetricResult{
+		Namespace:     namespace,
+		ResourceQuota: resourceQuota,
+		Resource:      resource,
+		Unit:          unit,
+		Data:          result.Values,
+	}
+}
+
+type ResourceQuotaSpecCPURequestAvgResult = ResourceQuotaMetricResult
```
biancaburtoiu, line 1527: There's a lot of duplication in these types - are there any arguments against using `ResourceQuotaMetricResult` for all purposes?

nikovacevic: Not objections! If it generalizes well, then reusing the type is perfect.

modules/collector-source/pkg/metric/metrics.go
```diff
@@ -23,6 +23,10 @@ const (
 	ServiceSelectorLabels                                 = "service_selector_labels"
 	StatefulSetMatchLabels                                = "statefulSet_match_labels"
 	KubeReplicasetOwner                                   = "kube_replicaset_owner"
+	KubeResourceQuotaSpecResourceRequests                 = "kube_resource_quota_spec_resource_requests"
+	KubeResourceQuotaSpecResourceLimits                   = "kube_resource_quota_spec_resource_limits"
+	KubeResourceQuotaStatusUsedResourceRequests           = "kube_resource_quota_status_used_resource_requests"
+	KubeResourceQuotaStatusUsedResourceLimits             = "kube_resource_quota_status_used_resource_limits"
```
nikovacevic: I don't think we need to follow this (correct @Sean-Holcomb ?) but here is what the KSM metrics look like: https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/policy/resourcequota-metrics.md

Just interesting because we're _sort of_ following their naming conventions, but we're breaking with some things too.

nikovacevic: I don't see anything in ksm for "status used" though -- just spec AFAICT. So we're going to have to break with their conventions one way or another.

modules/prometheus-source/pkg/prom/metricsquerier.go
```diff
@@ -1484,6 +1484,294 @@ func (pds *PrometheusMetricsQuerier) QueryReplicaSetsWithRollout(start, end time
 	return source.NewFuture(source.DecodeReplicaSetsWithRolloutResult, ctx.QueryAtTime(queryReplicaSetsWithRolloutOwner, end))
 }
 
+func (pds *PrometheusMetricsQuerier) QueryResourceQuotaSpecCPURequestAverage(start, end time.Time) *source.Future[source.ResourceQuotaSpecCPURequestAvgResult] {
```
nikovacevic, line 1489: I guess I'm still a little bit confused about the Prometheus side of things... do the other changes in this package imply that we are not just scraping ResourceQuota data directly from the Kubernetes API (i.e. Cluster Cache), but also EMITTING it to a `/metrics` endpoint that a Prometheus could then scrape and ingest for the PrometheusMetricsQuerier to query?

nikovacevic (approved):
This looks great. Just that one question about the Prometheus Source implementation. If this works, I just need to clarify (for myself) the role that OpenCost is playing in _emitting_ metrics to a `/metrics` endpoint for Prom scraping.

modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -538,6 +539,81 @@ func (ccs *ClusterCacheScraper) scrapeReplicaSets(replicaSets []*clustercache.Re
 	return scrapeResults
 }
 
+func (ccs *ClusterCacheScraper) ScrapeResourceQuotas() []metric.Update {
+	nodes := ccs.clusterCache.GetAllResourceQuotas()
```
Sean-Holcomb: nit: variable name

core/pkg/source/decoders.go
```diff
@@ -1497,6 +1498,125 @@ func DecodeReplicaSetsWithRolloutResult(result *QueryResult) *ReplicaSetsWithRol
 	}
 }
 
+type ResourceQuotaMetricResult struct {
+	Namespace     string
+	ResourceQuota string
```
Sean-Holcomb, line 1504: Are these fields UUIDs

modules/collector-source/pkg/metric/metrics.go
Sean-Holcomb: are you suggesting we exclude the `kube_` prefix. I think that might be a good idea to prevent confusion for users who are digging around in the metrics.

modules/prometheus-source/pkg/prom/metricsquerier.go
Sean-Holcomb: As far as this PR goes we aren't actually emitting these metrics anywhere. Not sure if that happened in a previous PR. But yes we expose metrics in our own `/metrics` endpoint, prometheus scrapes them and then we query them back out of prometheus. That is currently the case KSM metrics and it is also the case with all of the pricing metrics that we use.

Sean-Holcomb (approved):
Generally looks good. I want to make sure that if we are emitting these prom metrics that we should be adding in UUIDs as labels. It is not clear to me that we are actually emitting these metrics ourselves and what is being done in the collector module isn't just an emulation of some other metrics endpoint.

modules/collector-source/pkg/metric/metrics.go
biancaburtoiu: Renamed metrics to `resourcequota_spec...` and `resourcequota_status_used...`

core/pkg/source/decoders.go
biancaburtoiu: I may just leave these as they are. I was worried about the duplication but the same pattern applies to all the other metric types.

core/pkg/source/decoders.go
nikovacevic: I think we need UUIDs here still -- but correct me if I'm wrong @biancaburtoiu .

That would be the last change to make prior to merging.

nikovacevic (changes requested):

nikovacevic (approved):

PR #3432, opened 2025-10-29
Add Go structs for data model 2.0

spa-raj:
## What does this PR change?
* Create Go model structs in core/pkg/model/cluster.go based on proto message definitions from protos/kubemodel/*.proto for the data model 2.0 project. These structs will replace proto-generated code when transitioning to bingen or JSON serialization.

Implemented structs from proto definitions:
 - Provider and ControllerKind enums (without unspecified values)
 - Window, Resolution types for time-based aggregation
 - Container with full metrics and diagnostic support
 - Controller for Kubernetes workload controllers
 - DiagnosticResult matching core/pkg/diagnostics/diagnostics.go
 - GPUDevice with sharing, usage metrics, and memory tracking
 - GPUUsage for container-level GPU consumption
 - Namespace, Service, ServicePort with network metrics
 - Node with cost and usage tracking
 - Pod with full resource metrics
 - Volume and PersistentVolumeClaim with storage tracking
 - Cluster as top-level aggregator

Key design decisions:
 - Use core/pkg/diagnostics/diagnostics.go as source of truth for DiagnosticResult (not proto) since proto will be deprecated
 - Include per-resource diagnostic fields, exclude cluster-level diagnostics list (not in proto schema)
 - Native Go types (time.Time, map[string]any) instead of proto types
 - Clean field naming without proto legacy duplicates

These native Go structs form the foundation for data model 2.0 serialization via bingen or JSON, removing dependency on protobuf generated code.

* Based on the proto definitions from #3369 and #3406 

* No immediate impact to users. This PR is the foundation for the Data model 2.0 project under the Linux foundation mentorship program.

* No

## How was this PR tested?
1. **Compilation verification**: Built core/pkg/model package successfully to
     ensure all struct definitions are syntactically correct.

2. **Proto alignment validation**: Cross-referenced each Go struct against
     corresponding proto messages in protos/kubemodel/*.proto to verify all 
     fields are present, correctly typed, and no duplicates exist.

* No

* No, this PR contains breaking changes.


spa-raj:
@nikovacevic @Sean-Holcomb @ameijer Please review this PR.

Sean-Holcomb:
Thanks for pivoting this over, we will get some comments on this later

core/pkg/model/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	CreationTime           *time.Time        `json:"creationTime,omitempty"`
+	DeletionTime           *time.Time        `json:"deletionTime,omitempty"`
+	CpuCoreHours           float32           `json:"cpuCoreHours"`
+	CpuCoreRequestAverage  float32           `json:"cpuCoreRequestAverage"`
+	CpuCoreUsageAverage    float32           `json:"cpuCoreUsageAverage"`
+	CpuCoreUsageMax        float32           `json:"cpuCoreUsageMax"`
+	RamByteHours           int64             `json:"ramByteHours"`
+	RamBytesRequestAverage int64             `json:"ramBytesRequestAverage"`
+	RamBytesUsageAverage   int64             `json:"ramBytesUsageAverage"`
+	RamBytesUsageMax       int64             `json:"ramBytesUsageMax"`
```
ameijer: do we want to break this out to a container, then have a pod hold n many containers?

ameijer: oh, never mind, I found container. I don't think we would need these fields here then?

core/pkg/model/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	Start      time.Time  `json:"start"`
+}
+
+// Container represents a container within a pod (allocated resource)
+type Container struct {
+	PodID                  string            `json:"podId"`
+	Name                   string            `json:"name"`
+	CreationTime           *time.Time        `json:"creationTime,omitempty"`
+	DeletionTime           *time.Time        `json:"deletionTime,omitempty"`
+	CpuCoreHours           float32           `json:"cpuCoreHours"`
```
aaliomer-ibm: Lot of those float32 should probably be float64

core/pkg/model/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	CpuCoreRequestAverage  float32           `json:"cpuCoreRequestAverage"`
+	CpuCoreUsageAverage    float32           `json:"cpuCoreUsageAverage"`
+	CpuCoreUsageMax        float32           `json:"cpuCoreUsageMax"`
+	RamByteHours           int64             `json:"ramByteHours"`
```
aaliomer-ibm: use uint64

core/pkg/model/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	Annotations  map[string]string `json:"annotations,omitempty"`
+	CreationTime *time.Time        `json:"creationTime,omitempty"`
+	DeletionTime *time.Time        `json:"deletionTime,omitempty"`
+	Diagnostic   *DiagnosticResult `json:"diagnostic,omitempty"`
+}
+
+// ServicePort represents a service port
+type ServicePort struct {
+	Name       string `json:"name"`
+	Port       int32  `json:"port"`
```
aaliomer-ibm: uint16

core/pkg/model/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+)
+
+// Window defines a unit of time by a resolution and a start time
+type Window struct {
+	Resolution Resolution `json:"resolution"`
+	Start      time.Time  `json:"start"`
+}
+
+// Container represents a container within a pod (allocated resource)
+type Container struct {
```
aaliomer-ibm: is it possible to attribute the volume storage / file system usage here?

core/pkg/model/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+// Cluster represents the top-level Kubernetes cluster
+type Cluster struct {
+	ID       string   `json:"id"`
+	Provider Provider `json:"provider"`
+	Account  string   `json:"account"`
+	Name     string   `json:"name"`
+	Window   *Window  `json:"window,omitempty"`
+
+	// Resources within the cluster
+	Containers             []Container             `json:"containers,omitempty"`
```
aaliomer-ibm: the model is very flat where the object structure is hierarchal

core/pkg/model/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	GpuHours          float32           `json:"gpuHours"`
+	GpuRequestAverage float32           `json:"gpuRequestAverage"`
+	GpuUsageAverage   float32           `json:"gpuUsageAverage"`
+	GpuUsageMax       float32           `json:"gpuUsageMax"`
+	MemoryBytes       int64             `json:"memoryBytes"`
+	Diagnostic        *DiagnosticResult `json:"diagnostic,omitempty"`
+}
+
+// GPUUsage represents GPU usage metrics
+type GPUUsage struct {
```
aaliomer-ibm: shouldn't this be part of a pod or a container?

core/pkg/model/cluster.go
spa-raj: Pods can have multiple containers, and these pod-level fields provide pre-aggregated metrics across all containers, eliminating the need for clients to manually sum container metrics. We can remove them if we want the model to be lightweight.

core/pkg/model/cluster.go
spa-raj: Added in container

core/pkg/model/kubemodel/cluster.go
```diff
@@ -0,0 +1,9 @@
+package kubemodel
+
+// Cluster represents the top-level Kubernetes cluster
+type Cluster struct {
+	ID       string   `json:"id"`
+	Provider Provider `json:"provider"`
+	Account  string   `json:"account"`
+	Name     string   `json:"name"`
```
nikovacevic, line 8: Need to decide where the data comes from for `ID` and `Name`. Cloudy uses `kube-system` namespace's `UID` as the cluster `ID`.

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,27 @@
...
+	"time"
+)
+
+// Container represents a container within a pod (allocated resource)
+type Container struct {
+	PodID                      string            `json:"podId"`
+	Name                       string            `json:"name"`
+	CreationTime               *time.Time        `json:"creationTime,omitempty"`
+	DeletionTime               *time.Time        `json:"deletionTime,omitempty"`
+	CpuCoreHours               float64           `json:"cpuCoreHours"`
```
nikovacevic: What unit for CPU?

nikovacevic: mCPU * sec

nikovacevic: `uint64`

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,27 @@
...
+	CpuCoreUsageMax            float64           `json:"cpuCoreUsageMax"`
+	RamByteHours               uint64            `json:"ramByteHours"`
+	RamBytesRequestAverage     uint64            `json:"ramBytesRequestAverage"`
+	RamBytesUsageAverage       uint64            `json:"ramBytesUsageAverage"`
+	RamBytesUsageMax           uint64            `json:"ramBytesUsageMax"`
+	StorageByteHours           uint64            `json:"storageByteHours"`
+	StorageBytesRequestAverage uint64            `json:"storageBytesRequestAverage"`
+	StorageBytesUsageAverage   uint64            `json:"storageBytesUsageAverage"`
+	StorageBytesUsageMax       uint64            `json:"storageBytesUsageMax"`
+	GpuUsages                  []GPUUsage        `json:"gpuUsages,omitempty"`
```
nikovacevic: We need to deal with attached devices (GPUs, storage, etc.) -- something map-based.

Allow for things to evolve into the future.

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"time"
+)
+
+// Container represents a container within a pod (allocated resource)
+type Container struct {
+	PodID                      string            `json:"podId"`
+	Name                       string            `json:"name"`
+	CreationTime               *time.Time        `json:"creationTime,omitempty"`
+	DeletionTime               *time.Time        `json:"deletionTime,omitempty"`
```
nikovacevic: Prefer (Start, End) within the Window

core/pkg/model/kubemodel/diagnostic.go
```diff
@@ -0,0 +1,29 @@
+package kubemodel
+
+import (
+	"time"
+)
+
+// DiagnosticResult represents the result of a diagnostic run
+type DiagnosticResult struct {
```
nikovacevic, line 8: Should Diagnostics live at the Object level or the Set level?

nikovacevic (approved):
Partial review with commentary from 11/4 meeting

nikovacevic (changes requested):
Let's make changes in accord with previous comment

PR #3434, opened 2025-10-29
Implment new authorizer: Connection String for azure cloud integration

mittal-ishaan:
## What does this PR change?
* This PR implements a new authorizer Storage Connection String for azure cloud integration.

* NA

* This mainly allows azure users that have everything behind a firewall/cloudflare to setup cloud integration. For them current authorizer methods would use default blob URLs baed on `cloud` value. The new authorizor type `AzureStorageConnectionString` allows giving access using SAS URL and its corresponding Connection string which can take in and query a custom URL.

* Closes ...

## How was this PR tested?
* Created a SAS URL and connection string for my storage account
* Ran cloud cost locally using the following cloud-integration json:
```json
  {
    "azure": {
      "storage": [
        {
          "subscriptionID": "AZ_cloud_integration_subscriptionId",
          "account": "AZ_cloud_integration_azureStorageAccount",
          "container": "AZ_cloud_integration_azureStorageContainer",
          "path": "",
          "cloud": "public/gov",
          "authorizer": {
            "authorizerType": "AzureStorageConnectionString",
            "storageConnectionString": "xyz",
            "httpConfig": {
              "insecureSkipVerify": true
            }
          }
        }
      ]
    }
  }
  ```
Was able to see my federated bucket getting populated with clould cost

* Yes. 

*

### CODE REVIEW

File: pkg/cloud/azure/storageauthorizer.go [hunk 1]
```diff
@@ -127,3 +145,74 @@ func (ah *AuthorizerHolder) GetBlobClient(serviceURL string) (*azblob.Client, er
...
+	}
+
+	return true
+}
+
+func (s *StorageConnectionStringHolder) Sanitize() cloud.Config {
+	return &StorageConnectionStringHolder{StorageConnectionString: s.StorageConnectionString, HTTPConfig: s.HTTPConfig}
+}
+
+func (s *StorageConnectionStringHolder) GetBlobClient(serviceURL string) (*azblob.Client, error) {
```
mittal-ishaan: Reference taken from: https://github.com/opencost/opencost/blob/d13a362c72868db1e155bf2ec01b9724a0e03dc4/core/pkg/storage/azurestorage.go#L411-L437

pkg/cloud/azure/storageauthorizer.go
```diff
@@ -127,3 +145,77 @@ func (ah *AuthorizerHolder) GetBlobClient(serviceURL string) (*azblob.Client, er
...
+
+func (s *StorageConnectionStringCredential) MarshalJSON() ([]byte, error) {
+	fmap := make(map[string]any, 3)
+	fmap[cloud.AuthorizerTypeProperty] = StorageConnectionStringAuthorizerType
+	fmap["storageConnectionString"] = s.StorageConnectionString
+	fmap["httpConfig"] = s.HTTPConfig
+	return json.Marshal(fmap)
+}
+
+func (s *StorageConnectionStringCredential) UnmarshalJSON(b []byte) error {
```
Sean-Holcomb: I am wondering if this unmarshalJSON would be necessary if the httpConfig was a pointer which and the default was used when it is nil in `GetBlobClient()`

Sean-Holcomb (approved):
Connection Strings seem like they are probably an azure storage only connection method but just for the sake of questioning it can connection strings be used with other services. I am just wondering if there is something in `azidentity` that we could use here which could be used to make an `Authorizer` rather than a `StorageAuthorizer`, which could be reusable with other services.

pkg/cloud/azure/storageauthorizer.go
mittal-ishaan: We could do that. Issue with this approach is that, let us suppose the user only defines `InsecureSkipVerify` in HttpConfig, the defaultHttpConfig does not get applied and rest of the values remain unset to zero(which means no limit in most cases).

mittal-ishaan:
To what I found, connections strings are not specific to just azure storage but also used in places, in places like `azure cosmos` Ref: https://github.com/search?q=repo%3AAzure%2Fazure-sdk-for-go+NewClientFromConnectionString&type=code
`azidentity` does not have anything specific that would help here. I am thinking the main reason might be because connection string does not need any credentials for creating a new client.

pkg/cloud/azure/storageauthorizer.go
mittal-ishaan: removed the unmarshalJson method as I am now setting defaultHttpConfig when creating the `StorageConnectionStringCredential` itself

mbolt35 (approved):
LGTM!

pkg/cloud/azure/storageauthorizer.go
Sean-Holcomb: moot point now, but I also and unsure that the default values remain after you unmarshal into the struct

mittal-ishaan: I did test around it and it works.
I have also added a test specific to it: https://github.com/opencost/opencost/blob/ad829e725cd280702ac632ef29d8a85bcedf79ab/pkg/cloud/azure/storageconfiguration_test.go#L638-L669

PR #3435, opened 2025-10-30
ResourceQuotas support: KubeModel types, source, exporter

nikovacevic:
## What does this PR change?
* New metrics for container limits
* New metrics for resource quotas
* New `info` metrics for cluster, namespace
* Complete bingen-encoded version 1 of each of the following KubeModel types: `Window`, `Cluster`, `Namespace`, `ResourceQuota`, `Stats`, `Window`
* `ComputeKubeModel` for the above types
* Pipeline and exporter for KubeModel under new path formatting (i.e. `finops-agent/<cluster>/kubemodel/<res>/<YYYY>/<MM>/<DD>/<WindowStart>`)
* NamespaceLabel filters for containers
* ResourceQuota diagnostics

## Does this PR relate to any other PRs?
* https://github.com/kubecost/kubecost-cost-model/pull/3843
* https://github.com/kubecost/ibm-finops-agent/pull/110

## How will this PR impact users?
* Allows computing and exporting of KubeModel data with ResourceQuotas

## Does this PR address any GitHub or Zendesk issues?
* Closes https://apptio.atlassian.net/browse/KCM-4714

## How was this PR tested?
* Unit testing
* Manual testing on https://project-resource-quotas.qa-eks3.kubecost.xyz/

## Does this PR require changes to documentation?
* Yes, API specs are included and documentation is in progress

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next OpenCost release? If not, why not?
* Yes `v1.119`

---

> [!NOTE]
> Introduces KubeModel data model and exporter pipeline with ResourceQuota support, updates pathing and filters, and integrates KubeModel into costmodel, pipelines, diagnostics, and tests.
> 
> - **KubeModel (new)**:
>   - Add full KubeModel schema (`Cluster`, `Namespace`, `ResourceQuota`, `Stats`, `Unit`, `Window`) with generated codecs and helpers; tests and mock data.
>   - Implement `KubeModel` compute in `pkg/kubemodel` and expose `ComputeKubeModelSet`; integrate into `CostModel`.
> - **Export Pipeline**:
>   - Add KubeModel compute source (`core/pkg/opencost/exporter/kubemodel`) and wire into `PipelineExportControllers` with configurable resolutions.
>   - Extend pipeline name map with `kubemodel`.
>   - Update exporter tests to cover KubeModel.
> - **Pathing**:
>   - Introduce `KubeModelStoragePathFormatter` and route KubeModel exports to .
[...]

PR #3443, opened 2025-11-07
Implement flat kubemodel architecture with bingen serialization and metric hydration

spa-raj:
## What does this PR change?
*  Implements a new flat kubemodel architecture with bingen annotations for binary serialization and full metric hydration from costmodel datasource, including registration functions and compute methods for all Kubernetes resource types (nodes, pods, containers, controllers, services, PVCs, volumes)

* This PR is built upon the work from #3435 

*  No immediate user-facing impact. This PR lays the foundation for Data Model 2.0 by enabling efficient binary serialization and structured metric collection, which will support future features like improved storage and analytics capabilities.

* Closes ...

## How was this PR tested?
*  Verified through successful compilation of all kubemodel and costmodel packages after fixing result struct field mappings. Part of ongoing LFX mentorship project testing will include integration testing with the broader Data Model 2.0 pipeline

* No

* This PR contains breaking changes and I leave it to the maintainers to decide its release.

---

> [!NOTE]
> Adds a new flat kubemodel with bingen-annotated schemas, export pipeline and pathing, and costmodel computation/hydration, plus node public IP helpers and pathing interface updates.
> 
> - **KubeModel Data Model (new)**:
>   - Introduces flat `kubemodel` package with bingen-annotated types (`Cluster`, `Namespace`, `Node`, `Pod`, `Container`, `Service`, `Volume`, `PVC`, `Controller`, `ResourceQuota`, GPU structures, stats/units, `KubeModelSet`).
>   - Adds validation/clone utilities and tests.
> - **CostModel Integration**:
>   - Implements `ComputeKubeModel` and helpers to hydrate all resources from metrics; computes `PublicIPSeconds` per node via cluster cache.
> - **Exporter/Pipelines**:
>   - Adds KubeModel pipeline source (`opencost/exporter/kubemodel`) and wires it into `PipelineExportControllers` with start/stop and default resolutions.
>   - Extends mocks/tests to cover KubeModel exporting.
> - **Pathing**:
>   - Adds KubeModel-specific storage formatter (`kubemodelpath.go`) and routes KubeModel pipeline to FinOps-agent root pathing.
>   - Simplifies `StoragePathFormatter` (removes `RootDir()`); adjusts bingen/event pathing and constructors.
> - **Cluster Cache**:
>   - Adds `Node.GetPublicIPAddresses()` and `Node.GetPublicIPCount()` helpers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f0b8bec6981b4f62f88bfa96d7bccd5ddf397b99. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


spa-raj:
@Sean-Holcomb @nikovacevic @ameijer  I have reverted to a flat model and added more implementations of hydration of the metrics as per Niko's PR.

core/pkg/model/kubemodel/kubemodel.go
```diff
@@ -0,0 +1,274 @@
...
+	ResourceQuotas         map[string]*ResourceQuota         `json:"resourceQuotas"`       // @bingen:field[version=1]
+	Containers             map[string]*Container             `json:"containers,omitempty"` // @bingen:field[version=1]
+	Controllers            map[string]*Controller            `json:"controllers,omitempty"` // @bingen:field[version=1]
+	Devices                map[string]*Device                `json:"devices,omitempty"`    // @bingen:field[version=1]
+	Nodes                  map[string]*Node                  `json:"nodes,omitempty"`      // @bingen:field[version=1]
+	Pods                   map[string]*Pod                   `json:"pods,omitempty"`       // @bingen:field[version=1]
+	PersistentVolumeClaims map[string]*PersistentVolumeClaim `json:"pvcs,omitempty"`       // @bingen:field[version=1]
+	Services               map[string]*Service               `json:"services,omitempty"`   // @bingen:field[version=1]
+	Volumes                map[string]*Volume                `json:"volumes,omitempty"`    // @bingen:field[version=1]
+	idx                    *kubeModelSetIndexes              // @bingen:field[ignore]
```
ameijer, line 25: shall we consider removing the indexes?

ameijer: edit - not a concern, we can keep these

core/pkg/model/kubemodel/window.go
```diff
@@ -0,0 +1,9 @@
+package kubemodel
+
+import "time"
+
+// @bingen:generate:Window
+type Window struct {
+	Start time.Time `json:"start"` // @bingen:field[version=1]
+	End   time.Time `json:"end"`   // @bingen:field[version=1]
```
Sean-Holcomb, line 8: I prefer this to opencost.Window but are we moving to this rather than start and resolution?

Sean-Holcomb (commented):
Looks really good, great work.

core/pkg/model/kubemodel/window.go
spa-raj: During our previous discussion, Jay recommended using dynamic windows as an approach, and Niko expressed agreement with this direction. 

I suggest we revisit this topic in our next standup meeting to align on the implementation details.

core/pkg/model/kubemodel/service.go
```diff
@@ -0,0 +1,29 @@
...
+	Name       string `json:"name"`       // @bingen:field[version=1]
+	Port       uint16 `json:"port"`       // @bingen:field[version=1]
+	TargetPort uint16 `json:"targetPort"` // @bingen:field[version=1]
+	NodePort   uint16 `json:"nodePort"`   // @bingen:field[version=1]
+	Protocol   string `json:"protocol"`   // @bingen:field[version=1]
+}
+
+// @bingen:generate:Service
+type Service struct {
+	ID                   string            `json:"id"`                   // @bingen:field[version=1]
```
ameijer: Load balancer vs service
ID vs UID

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,25 @@
...
+
+// @bingen:generate:Container
+type Container struct {
+	PodID                      string                  `json:"podId"`                      // @bingen:field[version=1]
+	Name                       string                  `json:"name"`                       // @bingen:field[version=1]
+	Start                      time.Time               `json:"start"`                      // @bingen:field[version=1]
+	End                        time.Time               `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicoreSeconds        uint64                  `json:"cpuMillicoreSeconds"`        // @bingen:field[version=1]
+	CpuMillicoreRequestAverage uint64                  `json:"cpuMillicoreRequestAverage"` // @bingen:field[version=1]
+	CpuMillicoreUsageAverage   uint64                  `json:"cpuMillicoreUsageAverage"`   // @bingen:field[version=1]
```
aaliomer-ibm: nit: Nano-core is the smallest unit. dont think it is a big deal however

core/pkg/model/kubemodel/pod.go
```diff
@@ -0,0 +1,31 @@
...
+	CpuMillicoreUsageMax       uint64                            `json:"cpuMillicoreUsageMax"`             // @bingen:field[version=1]
+	RamByteSeconds             uint64                            `json:"ramByteSeconds"`                   // @bingen:field[version=1]
+	RamBytesRequestAverage     uint64                            `json:"ramBytesRequestAverage"`           // @bingen:field[version=1]
+	RamBytesUsageAverage       uint64                            `json:"ramBytesUsageAverage"`             // @bingen:field[version=1]
+	RamBytesUsageMax           uint64                            `json:"ramBytesUsageMax"`                 // @bingen:field[version=1]
+	StorageByteSeconds         uint64                            `json:"storageByteSeconds"`               // @bingen:field[version=1]
+	NetworkTransferBytes       uint64                            `json:"networkTransferBytes"`             // @bingen:field[version=1]
+	NetworkReceiveBytes        uint64                            `json:"networkReceiveBytes"`              // @bingen:field[version=1]
+	Containers                 map[string]*Container             `json:"containers"`                       // @bingen:field[version=1]
```
aaliomer-ibm: I dont think this should be here. i would not have references to pointers. instead, i'd have a set of device IDs / volume claims.
Wasnt the index object supposed to do this role? this is no longer flat. but now is a graph which i dont think it is the intention

core/pkg/model/kubemodel/service.go
aaliomer-ibm: unrelated but I'd add every k8s object has a UUID id type. everything else should use a string id (container id should be a struct with pod-<name/id>, container-name)

core/pkg/model/kubemodel/window.go
aaliomer-ibm: I think the discussion was about the codebase already using resolutions and the huge amount of work it would require to refactor if i remember correctly. Not too familiar with the codebase to give a level of effort estimate

core/pkg/model/kubemodel/cluster.go
```diff
@@ -0,0 +1,13 @@
...
+import "time"
+
+// @bingen:generate:Cluster
+type Cluster struct {
+	UID      string    `json:"uid"`      // @bingen:field[version=1]
+	Provider Provider  `json:"provider"` // @bingen:field[version=1]
+	Account  string    `json:"account"`  // @bingen:field[version=1]
+	Name     string    `json:"name"`     // @bingen:field[version=1]
+	Start    time.Time `json:"start"`    // @bingen:field[version=1]
+	End      time.Time `json:"end"`      // @bingen:field[version=1]
```
ameijer, line 12: were we going to start going with Start + resolution?

ameijer: looking back over prior reviews, where did we land with https://github.com/opencost/opencost/pull/3443#discussion_r2517618073?

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,24 @@
...
+	End                        time.Time                  `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicoreSeconds        uint64                     `json:"cpuMillicoreSeconds"`        // @bingen:field[version=1]
+	CpuMillicoreRequestAverage uint64                     `json:"cpuMillicoreRequestAverage"` // @bingen:field[version=1]
+	CpuMillicoreUsageAverage   uint64                     `json:"cpuMillicoreUsageAverage"`   // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64                     `json:"cpuMillicoreUsageMax"`       // @bingen:field[version=1]
+	RAMBytesSeconds            uint64                     `json:"ramBytesSeconds"`            // @bingen:field[version=1]
+	RAMBytesRequestAverage     uint64                     `json:"ramBytesRequestAverage"`     // @bingen:field[version=1]
+	RAMBytesUsageAverage       uint64                     `json:"ramBytesUsageAverage"`       // @bingen:field[version=1]
+	RAMBytesUsageMax           uint64                     `json:"ramBytesUsageMax"`           // @bingen:field[version=1]
+	StorageBytesSeconds        uint64            `json:"storageBytesSeconds"`        // @bingen:field[version=1]
```
ameijer: nit- fix spacing

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,24 @@
...
+type Container struct {
+	PodUID                     string                     `json:"podUid"`                     // @bingen:field[version=1]
+	Name                       string                     `json:"name"`                       // @bingen:field[version=1]
+	Start                      time.Time                  `json:"start"`                      // @bingen:field[version=1]
+	End                        time.Time                  `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicoreSeconds        uint64                     `json:"cpuMillicoreSeconds"`        // @bingen:field[version=1]
+	CpuMillicoreRequestAverage uint64                     `json:"cpuMillicoreRequestAverage"` // @bingen:field[version=1]
+	CpuMillicoreUsageAverage   uint64                     `json:"cpuMillicoreUsageAverage"`   // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64                     `json:"cpuMillicoreUsageMax"`       // @bingen:field[version=1]
+	RAMBytesSeconds            uint64                     `json:"ramBytesSeconds"`            // @bingen:field[version=1]
```
ameijer: nit - I think all these should be byte-seconds, not bytes-seconds

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,24 @@
...
+	RAMBytesRequestAverage     uint64                     `json:"ramBytesRequestAverage"`     // @bingen:field[version=1]
+	RAMBytesUsageAverage       uint64                     `json:"ramBytesUsageAverage"`       // @bingen:field[version=1]
+	RAMBytesUsageMax           uint64                     `json:"ramBytesUsageMax"`           // @bingen:field[version=1]
+	StorageBytesSeconds        uint64            `json:"storageBytesSeconds"`        // @bingen:field[version=1]
+	StorageBytesRequestAverage uint64            `json:"storageBytesRequestAverage"` // @bingen:field[version=1]
+	StorageBytesUsageAverage   uint64            `json:"storageBytesUsageAverage"`   // @bingen:field[version=1]
+	StorageBytesUsageMax       uint64            `json:"storageBytesUsageMax"`       // @bingen:field[version=1]
```
ameijer: I wonder if we need support for a few kinds of storage:

1. ephemeral
2. PV

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,24 @@
...
+	Diagnostic                 *DiagnosticResult `json:"diagnostic,omitempty"`       // @bingen:field[version=1]
```
ameijer: what was the consensus on where diagnosticResult was going?

core/pkg/model/kubemodel/gpudevice.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+// @bingen:generate:GPUDevice
+// GPUDevice represents a GPU device with DCGM integration (provisioned resource)
+// This tracks available GPU capacity on a node
+type GPUDevice struct {
+	UID                string            `json:"uid"`                      // @bingen:field[version=1] GPU UUID (hardware identifier)
```
ameijer: fix spacing

core/pkg/model/kubemodel/gpudevice.go
```diff
@@ -0,0 +1,77 @@
...
+	DeviceNumber       int32             `json:"deviceNumber"`             // @bingen:field[version=1]
+	ModelName                   string            `json:"modelName"`                   // @bingen:field[version=1]
+	IsShared                    bool              `json:"isShared"`                    // @bingen:field[version=1] GPU sharing information
+	SharePercentage             float64           `json:"sharePercentage"`             // @bingen:field[version=1]
+	GpuSeconds                  float64           `json:"gpuSeconds"`                  // @bingen:field[version=1] GPU seconds available
+	GpuRequestPercentageAverage float64           `json:"gpuRequestPercentageAverage"` // @bingen:field[version=1] GPU request average percentage (0-100)
+	GpuUsagePercentageAverage   float64           `json:"gpuUsagePercentageAverage"`   // @bingen:field[version=1] GPU usage average percentage (0-100)
+	GpuUsagePercentageMax       float64           `json:"gpuUsagePercentageMax"`       // @bingen:field[version=1] GPU usage max percentage (0-100)
+	MemoryBytes                 uint64            `json:"memoryBytes"`                 // @bingen:field[version=1] GPU memory capacity in bytes
```
ameijer: did we want the device tree to hold pricing information too?

ameijer: byte-seconds?

core/pkg/model/kubemodel/gpuusage.go
```diff
@@ -0,0 +1,66 @@
...
+// This tracks actual GPU usage by containers for cost analysis
+type GPUUsage struct {
+	ContainerUID                string            `json:"containerUid"`                // @bingen:field[version=1] Container consuming GPU resources
+	GpuDeviceUID                string            `json:"gpuDeviceUid"`                // @bingen:field[version=1] Reference to the GPU device being used
+	GpuSeconds                  float64           `json:"gpuSeconds"`                  // @bingen:field[version=1] GPU usage in device-seconds consumed
+	GpuRequestPercentageAverage float64           `json:"gpuRequestPercentageAverage"` // @bingen:field[version=1] GPU request in percentage (0-100)
+	GpuUsagePercentageAverage   float64           `json:"gpuUsagePercentageAverage"`   // @bingen:field[version=1] GPU usage average percentage (0-100)
+	GpuUsagePercentageMax       float64           `json:"gpuUsagePercentageMax"`       // @bingen:field[version=1] GPU usage max percentage (0-100)
+	MemoryBytesUsed             uint64            `json:"memoryBytesUsed"`             // @bingen:field[version=1] GPU memory usage in bytes
```
ameijer: byte-seconds?

core/pkg/model/kubemodel/kubemodel.go
```diff
@@ -0,0 +1,296 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// @bingen:generate[stringtable]:KubeModelSet
+type KubeModelSet struct {
+	Metadata               *Metadata                         `json:"meta"`                  // @bingen:field[version=1]
```
ameijer, line 11: so should we remove the diagnostic results from each struct then ?

core/pkg/model/kubemodel/namespace.go
```diff
@@ -0,0 +1,17 @@
...
+type Namespace struct {
+	UID             string                    `json:"uid"`                      // @bingen:field[version=1]
+	ClusterUID      string                    `json:"clusterUID"`               // @bingen:field[version=1]
+	Name            string                    `json:"name"`                     // @bingen:field[version=1]
+	Labels          map[string]string         `json:"labels"`                   // @bingen:field[version=1]
+	Annotations     map[string]string         `json:"annotations"`              // @bingen:field[version=1]
+	Start           time.Time                 `json:"start"`                    // @bingen:field[version=1]
+	End             time.Time                 `json:"end"`                      // @bingen:field[version=1]
+	Controllers     map[string]*Controller    `json:"controllers,omitempty"`    // @bingen:field[ignore]
+	ResourceQuotas  map[string]*ResourceQuota `json:"resourceQuotas,omitempty"` // @bingen:field[ignore]
```
ameijer: I think these can be removed - these objects have uid pointers back to the namespace

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,27 @@
...
+	Name                 string             `json:"name"`                       // @bingen:field[version=1]
+	Labels               map[string]string  `json:"labels,omitempty"`           // @bingen:field[version=1]
+	Annotations          map[string]string  `json:"annotations,omitempty"`      // @bingen:field[version=1]
+	Start                     time.Time          `json:"start"`                      // @bingen:field[version=1]
+	End                       time.Time          `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicores             uint64             `json:"cpuMillicores"`              // @bingen:field[version=1]
+	RAMBytes                  uint64             `json:"ramBytes"`                   // @bingen:field[version=1]
+	CpuCost                   float64            `json:"cpuCost"`                    // @bingen:field[version=1]
+	RAMCost                   float64            `json:"ramCost"`                    // @bingen:field[version=1]
+	GpuCost                   float64            `json:"gpuCost"`                    // @bingen:field[version=1]
```
ameijer: should these be price rather than cost?

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,27 @@
...
+type Node struct {
+	UID                  string             `json:"uid"`                        // @bingen:field[version=1]
+	ClusterUID           string             `json:"clusterUid"`                 // @bingen:field[version=1]
+	ProviderResourceUID  string             `json:"providerResourceUid"`        // @bingen:field[version=1]
+	Name                 string             `json:"name"`                       // @bingen:field[version=1]
+	Labels               map[string]string  `json:"labels,omitempty"`           // @bingen:field[version=1]
+	Annotations          map[string]string  `json:"annotations,omitempty"`      // @bingen:field[version=1]
+	Start                     time.Time          `json:"start"`                      // @bingen:field[version=1]
+	End                       time.Time          `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicores             uint64             `json:"cpuMillicores"`              // @bingen:field[version=1]
```
ameijer: millicore-seconds?

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,27 @@
...
+	RAMBytes                  uint64             `json:"ramBytes"`                   // @bingen:field[version=1]
```
ameijer: byte-seconds?

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,27 @@
...
+	CpuCost                   float64            `json:"cpuCost"`                    // @bingen:field[version=1]
+	RAMCost                   float64            `json:"ramCost"`                    // @bingen:field[version=1]
+	GpuCost                   float64            `json:"gpuCost"`                    // @bingen:field[version=1]
+	CpuMillicoreUsageAverage  uint64             `json:"cpuMillicoreUsageAverage"`   // @bingen:field[version=1]
+	CpuMillicoreUsageMax      uint64             `json:"cpuMillicoreUsageMax"`       // @bingen:field[version=1]
+	RAMBytesUsageAverage      uint64             `json:"ramBytesUsageAverage"`       // @bingen:field[version=1]
+	RAMBytesUsageMax          uint64             `json:"ramBytesUsageMax"`           // @bingen:field[version=1]
+	Pods                 map[string]*Pod    `json:"pods"`                       // @bingen:field[version=1]
+	EphemeralVolumes     map[string]*Volume `json:"ephemeralVolumes,omitempty"` // @bingen:field[version=1]
```
ameijer: won't these objects have pointers back to this node? so we can toss this mapping?

core/pkg/model/kubemodel/pod.go
```diff
@@ -0,0 +1,31 @@
...
+	CpuMillicoreSeconds        uint64                            `json:"cpuMillicoreSeconds"`              // @bingen:field[version=1]
+	CpuMillicoreRequestAverage uint64                            `json:"cpuMillicoreRequestAverage"`       // @bingen:field[version=1]
+	CpuMillicoreUsageAverage   uint64                            `json:"cpuMillicoreUsageAverage"`         // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64                            `json:"cpuMillicoreUsageMax"`             // @bingen:field[version=1]
+	RAMBytesSeconds            uint64                            `json:"ramBytesSeconds"`                  // @bingen:field[version=1]
+	RAMBytesRequestAverage     uint64                            `json:"ramBytesRequestAverage"`           // @bingen:field[version=1]
+	RAMBytesUsageAverage       uint64                            `json:"ramBytesUsageAverage"`             // @bingen:field[version=1]
+	RAMBytesUsageMax           uint64                            `json:"ramBytesUsageMax"`                 // @bingen:field[version=1]
+	StorageBytesSeconds        uint64                            `json:"storageBytesSeconds"`              // @bingen:field[version=1]
```
ameijer: break this up into two kinds?

ameijer: actually - do we even need this? can't this be dynamically calculated?

core/pkg/model/kubemodel/pod.go
```diff
@@ -0,0 +1,31 @@
...
+	End                        time.Time                         `json:"end"`                              // @bingen:field[version=1]
+	CpuMillicoreSeconds        uint64                            `json:"cpuMillicoreSeconds"`              // @bingen:field[version=1]
+	CpuMillicoreRequestAverage uint64                            `json:"cpuMillicoreRequestAverage"`       // @bingen:field[version=1]
+	CpuMillicoreUsageAverage   uint64                            `json:"cpuMillicoreUsageAverage"`         // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64                            `json:"cpuMillicoreUsageMax"`             // @bingen:field[version=1]
+	RAMBytesSeconds            uint64                            `json:"ramBytesSeconds"`                  // @bingen:field[version=1]
+	RAMBytesRequestAverage     uint64                            `json:"ramBytesRequestAverage"`           // @bingen:field[version=1]
+	RAMBytesUsageAverage       uint64                            `json:"ramBytesUsageAverage"`             // @bingen:field[version=1]
+	RAMBytesUsageMax           uint64                            `json:"ramBytesUsageMax"`                 // @bingen:field[version=1]
```
ameijer: do we need these usages on the pod? Can't these be dynamically calculated?

core/pkg/model/kubemodel/resource.go
```diff
@@ -0,0 +1,33 @@
+package kubemodel
+
+// @bingen:generate:Resource
+type Resource string
+
+const (
+	ResourceCPU     Resource = "cpu"
+	ResourceMemory  Resource = "memory"
+	ResourceGPU     Resource = "gpu"
+	ResourceStorage Resource = "storage"
```
ameijer, line 10: is this dead?>

core/pkg/opencost/mock.go
```diff
@@ -1013,3 +1015,10 @@ func GenerateMockCloudCostSet(start, end time.Time, provider, integration string
 
 	return ccs
 }
+
+// GenerateMockKubeModelSet creates generic allocation set without idle allocations
+func GenerateMockKubeModelSet(start time.Time) *kubemodel.KubeModelSet {
+	// TODO: fill out mocks
```
ameijer, line 1021: were we planning on populating this in scope of this PR?

core/pkg/model/kubemodel/volume.go
```diff
@@ -0,0 +1,19 @@
+package kubemodel
+
+import "time"
+
+// @bingen:generate:Volume
+type Volume struct {
```
ameijer, line 6: should we add an 'ephemeral volume' type as well? Jay what kinds of things do we want to track for ephemeral storage?

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,27 @@
+package kubemodel
+
+import "time"
+
+// @bingen:generate:Node
+type Node struct {
```
ameijer, line 6: can we also add public IPs as a cost incurring resource type to this node?

core/pkg/model/kubemodel/service.go
```diff
@@ -0,0 +1,29 @@
...
+type ServicePort struct {
+	Name       string `json:"name"`       // @bingen:field[version=1]
+	Port       uint16 `json:"port"`       // @bingen:field[version=1]
+	TargetPort uint16 `json:"targetPort"` // @bingen:field[version=1]
+	NodePort   uint16 `json:"nodePort"`   // @bingen:field[version=1]
+	Protocol   string `json:"protocol"`   // @bingen:field[version=1]
+}
+
+// @bingen:generate:Service
+type Service struct {
```
ameijer, line 25: I think we also need a corresponding loadBalancer type

core/pkg/model/kubemodel/cluster.go
```diff
@@ -0,0 +1,13 @@
+package kubemodel
+
+import "time"
+
+// @bingen:generate:Cluster
+type Cluster struct {
```
ameijer, line 6: should cluster have a price field as well to capture mgmt costs?

ameijer: no pricing in kube model

core/pkg/model/kubemodel/window.go
ameijer: benefit of Start + resolution is that we can have constraints around allowed resolutions

ameijer: we will keep it start and end

core/pkg/model/kubemodel/gpudevice.go
ameijer: no pricing info in kube model

core/pkg/model/kubemodel/volume.go
ameijer: at min, pointer to node

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,27 @@
...
+	ProviderResourceUID  string             `json:"providerResourceUid"`        // @bingen:field[version=1]
+	Name                 string             `json:"name"`                       // @bingen:field[version=1]
+	Labels               map[string]string  `json:"labels,omitempty"`           // @bingen:field[version=1]
+	Annotations          map[string]string  `json:"annotations,omitempty"`      // @bingen:field[version=1]
+	Start                     time.Time          `json:"start"`                      // @bingen:field[version=1]
+	End                       time.Time          `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicores             uint64             `json:"cpuMillicores"`              // @bingen:field[version=1]
+	RAMBytes                  uint64             `json:"ramBytes"`                   // @bingen:field[version=1]
+	CpuCost                   float64            `json:"cpuCost"`                    // @bingen:field[version=1]
+	RAMCost                   float64            `json:"ramCost"`                    // @bingen:field[version=1]
```
ameijer: remove all pricing

core/pkg/model/kubemodel/pod.go
```diff
@@ -0,0 +1,31 @@
...
+	StorageBytesSeconds        uint64                            `json:"storageBytesSeconds"`              // @bingen:field[version=1]
+	NetworkTransferBytes       uint64                            `json:"networkTransferBytes"`             // @bingen:field[version=1]
+	NetworkReceiveBytes        uint64                            `json:"networkReceiveBytes"`              // @bingen:field[version=1]
+	Containers                 map[string]*Container             `json:"containers"`                       // @bingen:field[version=1]
+	AttachedGPUDevices         map[string]*GPUDevice             `json:"attachedGpuDevices,omitempty"`     // @bingen:field[version=1]
+	PersistentVolumeClaims     map[string]*PersistentVolumeClaim `json:"persistentVolumeClaims,omitempty"` // @bingen:field[version=1]
```
ameijer: remove these since they will have pointers to this?

core/pkg/model/kubemodel/gpuusage.go
aaliomer-ibm: +1

core/pkg/model/kubemodel/gpuusage.go
```diff
@@ -0,0 +1,66 @@
...
+
+// @bingen:generate:GPUUsage
+// GPUUsage represents GPU resources consumed by a container (allocated resource)
+// This tracks actual GPU usage by containers for cost analysis
+type GPUUsage struct {
+	ContainerUID                string            `json:"containerUid"`                // @bingen:field[version=1] Container consuming GPU resources
+	GpuDeviceUID                string            `json:"gpuDeviceUid"`                // @bingen:field[version=1] Reference to the GPU device being used
+	GpuSeconds                  float64           `json:"gpuSeconds"`                  // @bingen:field[version=1] GPU usage in device-seconds consumed
+	GpuRequestPercentageAverage float64           `json:"gpuRequestPercentageAverage"` // @bingen:field[version=1] GPU request in percentage (0-100)
+	GpuUsagePercentageAverage   float64           `json:"gpuUsagePercentageAverage"`   // @bingen:field[version=1] GPU usage average percentage (0-100)
```
aaliomer-ibm: I understand that GPU is like CPU, it has memory and compute dimensions. Both need to be exposed. Power consumption is also a metric that we probably want to keep track of

core/pkg/model/kubemodel/namespace.go
```diff
@@ -0,0 +1,17 @@
...
+// @bingen:generate:Namespace
+type Namespace struct {
+	UID             string                    `json:"uid"`                      // @bingen:field[version=1]
+	ClusterUID      string                    `json:"clusterUID"`               // @bingen:field[version=1]
+	Name            string                    `json:"name"`                     // @bingen:field[version=1]
+	Labels          map[string]string         `json:"labels"`                   // @bingen:field[version=1]
+	Annotations     map[string]string         `json:"annotations"`              // @bingen:field[version=1]
+	Start           time.Time                 `json:"start"`                    // @bingen:field[version=1]
+	End             time.Time                 `json:"end"`                      // @bingen:field[version=1]
+	Controllers     map[string]*Controller    `json:"controllers,omitempty"`    // @bingen:field[ignore]
```
aaliomer-ibm: I thought controllers and ResourceQuotas are going to be flat

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,27 @@
...
+
+// @bingen:generate:Node
+type Node struct {
+	UID                  string             `json:"uid"`                        // @bingen:field[version=1]
+	ClusterUID           string             `json:"clusterUid"`                 // @bingen:field[version=1]
+	ProviderResourceUID  string             `json:"providerResourceUid"`        // @bingen:field[version=1]
+	Name                 string             `json:"name"`                       // @bingen:field[version=1]
+	Labels               map[string]string  `json:"labels,omitempty"`           // @bingen:field[version=1]
+	Annotations          map[string]string  `json:"annotations,omitempty"`      // @bingen:field[version=1]
+	Start                     time.Time          `json:"start"`                      // @bingen:field[version=1]
```
aaliomer-ibm: whats the start and end are for?

core/pkg/model/kubemodel/node.go
aaliomer-ibm: +1

core/pkg/model/kubemodel/node.go
aaliomer-ibm: +1 always use commutative units

core/pkg/model/kubemodel/node.go
aaliomer-ibm: can a volume be de-attached and attached to a different node? how will this impact this model? then you will have multiple nodes pointing to the same volume. For example, a stateful set will follow the volume not the node necessarily i think but not sure

core/pkg/model/kubemodel/pod.go
aaliomer-ibm: you can get an average from byteseconds but not max/min/p95 etc...

core/pkg/model/kubemodel/unit.go
```diff
@@ -0,0 +1,20 @@
...
+// @bingen:generate:Unit
+type Unit string
+
+const (
+	UnitMillicore     = "mCPU"
+	UnitByte          = "B"
+	UnitMi            = "Mi"
+	UnitGB            = "GB"
+	UnitGPU           = "GPU"
+	UnitSecond        = "s"
```
aaliomer-ibm: not needed anymore

core/pkg/model/kubemodel/volume.go
aaliomer-ibm: in cldy, we have two types of storage. Root storage (storage attached to the node lifecycle aka ephemeral) and attached storage.
Sometimes in AWS, Node storage is backed by an EBS volume. 
The attached root storage is billed separately by AWS for example and an EBS. We do keep the ID of it to later allocate the cost to it. 

Getting back at your point whether both should be separate, it appears so. Root volume should be a part of the node instead of the volume IMO

core/pkg/model/kubemodel/node.go
spa-raj: Specifically for Nodes - they can join/leave clusters dynamically (auto-scaling, failures, maintenance).
Also, cumulative metrics are meaningless without a time range.

core/pkg/model/kubemodel/gpuusage.go
spa-raj: I have added fields for power consumption but I think opencost is not currently scraping DCGM metrics for the same. We need to update our scraper as well to see power metrics in effect.

core/pkg/model/kubemodel/service.go
spa-raj: @Sean-Holcomb @nikovacevic Do we want ID for containers? We don't have it in our current model. Shall I update the data type for all the K8s object UIDs to UUID?

core/pkg/model/kubemodel/volume.go
spa-raj: Perhaps we can rename "volume" to "PersistantVolume" and add these to node?

`EphemeralStorageBytes       uint64 'json:"ephemeralStorageBytes"'`
`EphemeralStorageByteSeconds uint64 'json:"ephemeralStorageByteSeconds"'`
`RootVolumeProviderID        string 'json:"rootVolumeProviderId,omitempty"'`

core/pkg/clustercache/clustercache.go
```diff
@@ -171,6 +171,28 @@ type ResourceQuota struct {
 type Volume struct {
 }
 
+// GetPublicIPAddresses returns all external IP addresses associated with the node
+func (n *Node) GetPublicIPAddresses() []string {
+	var publicIPs []string
+	for _, addr := range n.Status.Addresses {
+		if addr.Type == v1.NodeExternalIP {
+			publicIPs = append(publicIPs, addr.Address)
```
ameijer, line 179: wow that is nice that we have this so accessible.

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,24 @@
...
+	CpuMillicoreUsageMax       uint64            `json:"cpuMillicoreUsageMax"`       // @bingen:field[version=1]
+	RAMByteSeconds             uint64            `json:"ramByteSeconds"`             // @bingen:field[version=1]
+	RAMByteRequestAverage      uint64            `json:"ramByteRequestAverage"`      // @bingen:field[version=1]
+	RAMByteUsageAverage        uint64            `json:"ramByteUsageAverage"`        // @bingen:field[version=1]
+	RAMByteUsageMax            uint64            `json:"ramByteUsageMax"`            // @bingen:field[version=1]
+	StorageByteSeconds         uint64            `json:"storageByteSeconds"`         // @bingen:field[version=1]
+	StorageByteRequestAverage  uint64            `json:"storageByteRequestAverage"`  // @bingen:field[version=1]
+	StorageByteUsageAverage    uint64            `json:"storageByteUsageAverage"`    // @bingen:field[version=1]
+	StorageByteUsageMax        uint64            `json:"storageByteUsageMax"`        // @bingen:field[version=1]
+	Diagnostic                 *DiagnosticResult `json:"diagnostic,omitempty"`       // @bingen:field[version=1]
```
ameijer: following up here - I think we only want Diagnostics in the root kubemodel object

core/pkg/model/kubemodel/gpuusage.go
```diff
@@ -0,0 +1,92 @@
...
+	GpuRequestPercentageAverage float64           `json:"gpuRequestPercentageAverage"` // @bingen:field[version=1] GPU compute request in percentage (0-100)
+	GpuUsagePercentageAverage   float64           `json:"gpuUsagePercentageAverage"`   // @bingen:field[version=1] GPU compute usage average percentage (0-100)
+	GpuUsagePercentageMax       float64           `json:"gpuUsagePercentageMax"`       // @bingen:field[version=1] GPU compute usage max percentage (0-100)
+	MemoryByteSecondsUsed       uint64            `json:"memoryByteSecondsUsed"`       // @bingen:field[version=1] GPU memory usage in byte-seconds
+	PowerWattSeconds            float64           `json:"powerWattSeconds"`            // @bingen:field[version=1] GPU power consumption in watt-seconds (Joules)
+	PowerWattAverage            float64           `json:"powerWattAverage"`            // @bingen:field[version=1] GPU average power consumption in watts
+	PowerWattMax                float64           `json:"powerWattMax"`                // @bingen:field[version=1] GPU max power consumption in watts
```
ameijer: should the watts get moved to the device? and then device loses the Usage percentage/max etc?

core/pkg/model/kubemodel/gpuusage.go
```diff
@@ -0,0 +1,92 @@
...
+	Diagnostic                  *DiagnosticResult `json:"diagnostic,omitempty"`        // @bingen:field[version=1]
```
ameijer: toss diagnostics since we have that in kubemodel root?

core/pkg/model/kubemodel/node.go
```diff
@@ -0,0 +1,26 @@
...
+	CpuMillicoreSeconds      uint64            `json:"cpuMillicoreSeconds"`      // @bingen:field[version=1]
+	RAMByteSeconds           uint64            `json:"ramByteSeconds"`           // @bingen:field[version=1]
+	// PublicIPSeconds represents the cumulative public IP allocation (count × seconds) for this node.
+	// Calculated as: number of ExternalIP addresses from Kubernetes node Status.Addresses × window duration in seconds.
+	// Used for cost attribution of public IP addresses associated with the node.
+	PublicIPSeconds          uint64            `json:"publicIpSeconds"`          // @bingen:field[version=1]
+	CpuMillicoreUsageAverage uint64            `json:"cpuMillicoreUsageAverage"` // @bingen:field[version=1]
+	CpuMillicoreUsageMax     uint64            `json:"cpuMillicoreUsageMax"`     // @bingen:field[version=1]
+	RAMByteUsageAverage      uint64            `json:"ramByteUsageAverage"`      // @bingen:field[version=1]
+	RAMByteUsageMax          uint64            `json:"ramByteUsageMax"`          // @bingen:field[version=1]
```
ameijer: I think all these usage fields can be removed. We can get the usage of the node by summing the usage of all the containers on the node. What we care about from the node is how many CPU cores are allocatable, which we get from the CPUMillicoreseconds and etc above

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,24 @@
...
+	CpuMillicoreUsageAverage   uint64            `json:"cpuMillicoreUsageAverage"`   // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64            `json:"cpuMillicoreUsageMax"`       // @bingen:field[version=1]
+	RAMByteSeconds             uint64            `json:"ramByteSeconds"`             // @bingen:field[version=1]
+	RAMByteRequestAverage      uint64            `json:"ramByteRequestAverage"`      // @bingen:field[version=1]
+	RAMByteUsageAverage        uint64            `json:"ramByteUsageAverage"`        // @bingen:field[version=1]
+	RAMByteUsageMax            uint64            `json:"ramByteUsageMax"`            // @bingen:field[version=1]
+	StorageByteSeconds         uint64            `json:"storageByteSeconds"`         // @bingen:field[version=1]
+	StorageByteRequestAverage  uint64            `json:"storageByteRequestAverage"`  // @bingen:field[version=1]
+	StorageByteUsageAverage    uint64            `json:"storageByteUsageAverage"`    // @bingen:field[version=1]
+	StorageByteUsageMax        uint64            `json:"storageByteUsageMax"`        // @bingen:field[version=1]
```
ameijer: I think we need to think about putting this in a map of different storage types

core/pkg/model/kubemodel/pod.go
```diff
@@ -0,0 +1,28 @@
...
+	End                        time.Time         `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicoreSeconds        uint64            `json:"cpuMillicoreSeconds"`        // @bingen:field[version=1]
+	CpuMillicoreRequestAverage uint64            `json:"cpuMillicoreRequestAverage"` // @bingen:field[version=1]
+	CpuMillicoreUsageAverage   uint64            `json:"cpuMillicoreUsageAverage"`   // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64            `json:"cpuMillicoreUsageMax"`       // @bingen:field[version=1]
+	RAMByteSeconds             uint64            `json:"ramByteSeconds"`             // @bingen:field[version=1]
+	RAMByteRequestAverage      uint64            `json:"ramByteRequestAverage"`      // @bingen:field[version=1]
+	RAMByteUsageAverage        uint64            `json:"ramByteUsageAverage"`        // @bingen:field[version=1]
+	RAMByteUsageMax            uint64            `json:"ramByteUsageMax"`            // @bingen:field[version=1]
+	StorageByteSeconds         uint64            `json:"storageByteSeconds"`         // @bingen:field[version=1]
```
ameijer: I don't think pod needs to duplicate any of the fields in the containers - with the exception of network (which I do think needs to stay since we only have network stats at pod level), all of these CPU*, RAM*, Storage fields can be calculated from the containers under pod

core/pkg/model/kubemodel/pod.go
```diff
@@ -0,0 +1,28 @@
...
+	NetworkTransferBytes       uint64            `json:"networkTransferBytes"`       // @bingen:field[version=1]
+	NetworkReceiveBytes        uint64            `json:"networkReceiveBytes"`        // @bingen:field[version=1]
+	Diagnostic                 *DiagnosticResult `json:"diagnostic,omitempty"`       // @bingen:field[version=1]
```
ameijer: toss?

core/pkg/model/kubemodel/pvc.go
```diff
@@ -0,0 +1,22 @@
+package kubemodel
+
+import "time"
+
+// @bingen:generate:PersistentVolumeClaim
+type PersistentVolumeClaim struct {
+	UID                 string            `json:"uid"`                   // @bingen:field[version=1]
+	NamespaceUID        string            `json:"namespaceUid"`          // @bingen:field[version=1]
+	VolumeUID           *string           `json:"volumeUid,omitempty"`   // @bingen:field[version=1]
+	PodUID              *string           `json:"podUid,omitempty"`      // @bingen:field[version=1]
```
ameijer: should this be an array? A PVC will frequently be attached to multiple pods in a window if a pod for example moves. I almost wonder if we need like a map here that maps a pod UID to a start,end pair so we can tell when the PVC was attached to which pod in a window

core/pkg/model/kubemodel/volume.go
ameijer: I think for root volume, we do need to record its total size in the Node object sparsh. For the consumers, I think at container level we enhance storage to be a map which can have different types, PVC vs ephemeral

pkg/costmodel/kubemodel.go
```diff
@@ -0,0 +1,872 @@
...
+	if err != nil {
+		diagnostic := &kubemodel.DiagnosticResult{
+			UID:         fmt.Sprintf("pod-compute-%d", time.Now().Unix()),
+			Name:        "PodCompute",
+			Description: "Failed to compute pod data",
+			Category:    "pod",
+			Timestamp:   time.Now().UTC(),
+			Error:       err.Error(),
+		}
+		kms.Metadata.Diagnostics = append(kms.Metadata.Diagnostics, diagnostic)
```
ameijer, line 93: looks like you are indeed loading the diagnostics only in the kube model set. let's toss diagnostics from the other objects

core/pkg/model/kubemodel/service.go
Sean-Holcomb: The map in the flat struct should be keyed on a concatenated string of the pod UID and the container name. It should be a string because keying on a struct limits serialization options. I don't think this key id needs to be part of the container struct since it is already available in other fields of the container struct.

core/pkg/model/kubemodel/node.go
Sean-Holcomb: I don't think that we can calculate maxes or even averages that way. If these are being pulled from metrics that we have access to then I dont think we should remove these fields

core/pkg/model/kubemodel/diagnostic.go
```diff
@@ -0,0 +1,14 @@
+package kubemodel
+
+import "time"
+
+// @bingen:generate:DiagnosticResult
+type DiagnosticResult struct {
+	UID         string            `json:"uid"`               // @bingen:field[version=1]
```
Sean-Holcomb, line 7: What is the UID of a DiagnosticResult?

core/pkg/model/kubemodel/gpudevice.go
```diff
@@ -0,0 +1,75 @@
...
+// GPUDevice represents a GPU device with DCGM integration (provisioned resource)
+// This tracks available GPU capacity on a node
+type GPUDevice struct {
+	UID                         string  `json:"uid"`                         // @bingen:field[version=1] GPU UUID (hardware identifier)
+	NodeUID                     string  `json:"nodeUid"`                     // @bingen:field[version=1] Node hosting this GPU device
+	DeviceNumber                int32   `json:"deviceNumber"`                // @bingen:field[version=1]
+	ModelName                   string  `json:"modelName"`                   // @bingen:field[version=1]
+	IsShared                    bool    `json:"isShared"`                    // @bingen:field[version=1] GPU sharing information
+	SharePercentage             float64 `json:"sharePercentage"`             // @bingen:field[version=1]
+	GpuSeconds                  float64 `json:"gpuSeconds"`                  // @bingen:field[version=1] GPU seconds available
```
Sean-Holcomb: flagging this to investigate what the metric populating looks like here. Could we use "MilliGPU"?

core/pkg/model/kubemodel/kubemodel.go
Sean-Holcomb: That seems right to me

core/pkg/model/kubemodel/resource.go
Sean-Holcomb: I don't believe it is, but we also need to pick if we are going to use this or not and use it everywhere consistently

core/pkg/model/kubemodel/node.go
ameijer: assuming that we can get this information separately from container level maxes, then we should keep these. if not, then we should remove

core/pkg/model/kubemodel/gpudevice.go
```diff
@@ -0,0 +1,75 @@
...
+// @bingen:generate:GPUDevice
+// GPUDevice represents a GPU device with DCGM integration (provisioned resource)
+// This tracks available GPU capacity on a node
+type GPUDevice struct {
+	UID                         string  `json:"uid"`                         // @bingen:field[version=1] GPU UUID (hardware identifier)
+	NodeUID                     string  `json:"nodeUid"`                     // @bingen:field[version=1] Node hosting this GPU device
+	DeviceNumber                int32   `json:"deviceNumber"`                // @bingen:field[version=1]
+	ModelName                   string  `json:"modelName"`                   // @bingen:field[version=1]
+	IsShared                    bool    `json:"isShared"`                    // @bingen:field[version=1] GPU sharing information
+	SharePercentage             float64 `json:"sharePercentage"`             // @bingen:field[version=1]
```
ameijer: GPU can only be mounted to one node at a time

core/pkg/model/kubemodel/service.go
ameijer: tabling this, just service will work

core/pkg/model/kubemodel/service.go
```diff
@@ -0,0 +1,28 @@
...
+	UID                  string            `json:"uid"`                   // @bingen:field[version=1]
+	ClusterUID           string            `json:"clusterUid"`            // @bingen:field[version=1]
+	NamespaceUID         string            `json:"namespaceUid"`          // @bingen:field[version=1]
+	Name                 string            `json:"name"`                  // @bingen:field[version=1]
+	Type                 string            `json:"type"`                  // @bingen:field[version=1]
+	Labels               map[string]string `json:"labels,omitempty"`      // @bingen:field[version=1]
```
ameijer, line 32: add optional field for the hostname

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,24 @@
...
+	Start                      time.Time         `json:"start"`                      // @bingen:field[version=1]
+	End                        time.Time         `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicoreSeconds        uint64            `json:"cpuMillicoreSeconds"`        // @bingen:field[version=1]
+	CpuMillicoreRequestAverage uint64            `json:"cpuMillicoreRequestAverage"` // @bingen:field[version=1]
+	CpuMillicoreUsageAverage   uint64            `json:"cpuMillicoreUsageAverage"`   // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64            `json:"cpuMillicoreUsageMax"`       // @bingen:field[version=1]
+	RAMByteSeconds             uint64            `json:"ramByteSeconds"`             // @bingen:field[version=1]
+	RAMByteRequestAverage      uint64            `json:"ramByteRequestAverage"`      // @bingen:field[version=1]
+	RAMByteUsageAverage        uint64            `json:"ramByteUsageAverage"`        // @bingen:field[version=1]
+	RAMByteUsageMax            uint64            `json:"ramByteUsageMax"`            // @bingen:field[version=1]
```
aaliomer-ibm: average is not worth storing. cumulative  to average is a simple calculation (byte-hours to average bytes for example). it makes sense to store max however just not averages

@ameijer @spa-raj  what do you think? I never seen these averages used anywhere

ameijer: I think I agree with you

aaliomer-ibm: to clarify, i consider average a computed value in this scenario. We should never store computed values. 
I'm sure im wrong and there is some context but at least wanted to share my idea

core/pkg/model/kubemodel/gpudevice.go
spa-raj: SharePercentage - it's required for cost attribution even though DCGM metrics already scope to virtual GPU slices.

  ---
  How DCGM Metrics Handle Virtual GPUs

  | Aspect                | Behavior                                                        |
  |-----------------------|-----------------------------------------------------------------|
  | UUID labels           | MIG slices get distinct UUIDs (MIG-GPU-<uuid>)                  |
  | Utilization metrics   | DCGM_FI_PROF_GR_ENGINE_ACTIVE reports 0-1 relative to the slice |
  | Per-container scoping | Metrics grouped by (container, pod, namespace, UUID)            |

  A MIG slice at 100% utilization reports 1.0, not 0.143 (1/7th of physical).

  ---
  Why SharePercentage Is Still Needed

  Metrics tell you: "How much of my allocated slice am I using?"
  SharePercentage tells you: "What fraction of the physical GPU is my slice?"

  | Scenario                   | Without SharePercentage | With SharePercentage          |
  |----------------------------|-------------------------|-------------------------------|
  | MIG 1/7 slice @ 100% usage | Billed as "1 GPU" ❌     | Billed as 14.3% of GPU cost ✅ |
  | Time-sliced (4 pods)       | Each sees 100% capacity | Each pays 25% of cost         |

  ---
  Current Implementation Gap

  IsGPUShared:     ✅ Populated (from nvidia_com_gpu_shared resource type)
  SharePercentage: ❌ NOT populated by Prometheus queries

  To populate SharePercentage, we'd need:
  - Node-level GPU inventory data
  - MIG partition configuration
  - Or explicit annotation/label on the workload

  ---
  Recommendation

  Keep both fields in GPUDevice:
  - IsShared - boolean flag for quick filtering
  - SharePercentage - required for accurate cost attribution

  Future work: Add a data source for MIG/vGPU partition configuration to populate SharePercentage.

core/pkg/model/kubemodel/diagnostic.go
spa-raj: I was thinking of UUIDv7 like in https://github.com/opencost/opencost/blob/develop/core/pkg/diagnostics/service.go#L148

core/pkg/model/kubemodel/gpudevice.go
spa-raj: I looked into this and MilliGPU doesn't really match how the industry works.

  Here's what I found:

  - Kubernetes: GPUs are allocated as whole integers (nvidia.com/gpu: 1). You can't request fractional GPUs natively.
  - Cloud providers (AWS/GCP/Azure): They bill in whole GPU-hours or GPU-seconds (like "$0.74/GPU/hour"), not milliGPUs.
  - GPU sharing (MIG, time-slicing): When GPUs are shared, they use decimal fractions like 0.25 or 0.5 GPU, not millicores.
  - DCGM metrics: Reports utilization as percentages (0-100%), not millicores.
  - Intel exception: Intel GPUs do use millicores, but that's Intel-specific, not an industry standard.

  I'd suggest keeping GpuSeconds as float64 representing GPU device-seconds. This naturally handles fractional GPU usage:
  - 3600.0 = 1 whole GPU for 1 hour
  - 1800.0 = 1 whole GPU for 30 minutes (or 0.5 GPU for 1 hour)
  - 900.0 = 0.25 GPU for 1 hour

  This matches how cloud providers bill and how time-slicing works. CPUs use millicores because they're built for fine-grained time-sharing, but GPUs are different—they're
  monolithic compute devices where the base unit is still a whole GPU (or GPU instance for MIG).

core/pkg/model/kubemodel/container.go
spa-raj: I agree with you. I think we can calculate averages like this
`RAMByteUsageAverage = RAMByteSeconds / (End - Start).Seconds()`

Let's discuss this in our standup.

core/pkg/model/kubemodel/node.go
spa-raj: I dug into how the existing OpenCost code handles node usage and you're both right.

The current cluster.go Node struct doesn't store simple avg/max values - it uses breakdowns (idle/system/user percentages) from `QueryNodeCPUModeTotal` and similar queries. These give way more insight than just averages.

@Sean-Holcomb - you're right that node max can't be computed from container maxes. But we don't have node-level usage queries in kubemodel yet anyway.

 @ameijer  - your condition applies: we can't get this separately from containers right now, so we should remove these fields.

**Proposing to remove:**
  - `CpuMillicoreUsageAverage`
  - `CpuMillicoreUsageMax`
  - `RAMByteUsageAverage`
  - `RAMByteUsageMax`

If we need node-level usage metrics later, we should implement them properly with breakdowns or node_exporter queries, not half-populated fields. For now, nodes track capacity, containers track usage.

What do you think?

core/pkg/model/kubemodel/gpuusage.go
spa-raj: Moved Power fields to GPUDevice and removed from Usage.

core/pkg/model/kubemodel/pod.go
spa-raj: yes, I have removed these duplicate fields from pod.

core/pkg/model/kubemodel/pod.go
spa-raj: I have removed averages and kept only max

core/pkg/model/kubemodel/node.go
ameijer: yes I think we should go ahead and remove these then since we never collected them in the first place. @Sean-Holcomb should we track a TODO to implement getting these for nodes from somewhere?

core/pkg/model/kubemodel/container.go
```diff
@@ -0,0 +1,23 @@
...
+
+import "time"
+
+// @bingen:generate:Container
+type Container struct {
+	PodUID                     string    `json:"podUid"`                     // @bingen:field[version=1]
+	Name                       string    `json:"name"`                       // @bingen:field[version=1]
+	Start                      time.Time `json:"start"`                      // @bingen:field[version=1]
+	End                        time.Time `json:"end"`                        // @bingen:field[version=1]
+	CpuMillicoreSeconds        uint64    `json:"cpuMillicoreSeconds"`        // @bingen:field[version=1]
```
ameijer:
```suggestion
	CpuMillicoreSecondsAllocated        uint64    `json:"cpuMillicoreSeconds"`        // @bingen:field[version=1]
```
Same for the others

ameijer:
goald before thanksgiving:
 merge everything EXCEPT 
1. GPUs (removed from kubemodel set etc)
2. Volumes (removed from kubemodel set etc)

core/pkg/model/kubemodel/controller.go
```diff
@@ -0,0 +1,27 @@
...
+	ControllerKindDaemonSet   ControllerKind = "daemonset"
+	ControllerKindJob         ControllerKind = "job"
+	ControllerKindCronJob     ControllerKind = "cronjob"
+	ControllerKindReplicaSet  ControllerKind = "replicaset"
+)
+
+// @bingen:generate:Controller
+type Controller struct {
+	UID          string            `json:"uid"`                   // @bingen:field[version=1]
+	NamespaceUID string            `json:"namespaceUid"`          // @bingen:field[version=1]
```
ameijer: OwnerID as well

core/pkg/model/kubemodel/controller.go
```diff
@@ -0,0 +1,27 @@
...
+	ControllerKindDeployment  ControllerKind = "deployment"
+	ControllerKindStatefulSet ControllerKind = "statefulset"
+	ControllerKindDaemonSet   ControllerKind = "daemonset"
+	ControllerKindJob         ControllerKind = "job"
+	ControllerKindCronJob     ControllerKind = "cronjob"
+	ControllerKindReplicaSet  ControllerKind = "replicaset"
+)
+
+// @bingen:generate:Controller
+type Controller struct {
```
ameijer: change to Owner

Sean-Holcomb:
closing in favor these changes to KubeModel https://github.com/opencost/opencost/pull/3875

PR #3448, opened 2025-11-11
feat: Add cluster label filtering to Prometheus queries via OPENCOST_CLUSTER_FILTER

badal773:
## What does this PR change?

This PR introduces a mechanism to **filter all Prometheus queries in OpenCost by a specific cluster label**, enabling proper multi-cluster separation when using backends such as **Mimir** that store metrics from multiple clusters under a single `OrgID`.

In our environment, we operate **hundreds of clusters**, all reporting metrics into a **single Mimir OrgID**. Creating and maintaining a separate OrgID for each cluster is not practical at our scale.
This change injects a **cluster label filter** into every Prometheus query so that OpenCost only retrieves metrics for the selected cluster.

#### Why `PROMETHEUS_HEADER_X_SCOPE_ORGID` isn’t sufficient

`PROMETHEUS_HEADER_X_SCOPE_ORGID` allows setting a custom HTTP header (`X-Scope-OrgID`) to select the Mimir tenant. However, in our setup, **all clusters share the same OrgID**, so this header only selects the tenant, not the cluster.
As a result, OpenCost receives metrics for all clusters under that OrgID, making it impossible to isolate costs per cluster.

#### What this PR does instead

This PR enables a new environment variable, `OPENCOST_CLUSTER_FILTER`, which injects a label matcher (e.g., `{cluster="stage-cluster"}`) into all Prometheus queries.
This ensures OpenCost retrieves metrics only for the specified cluster, even when multiple clusters share a single OrgID.

---

No.

---

In environments where multiple clusters share one Mimir OrgID, OpenCost previously queried and displayed data for all clusters combined.
With this change, users can now isolate metrics and costs for a specific cluster, improving accuracy and usability in **multi-cluster, single-tenant setups**.

---

## Code changes

* Added logic in `query.go` to check for the `OPENCOST_CLUSTER_FILTER` environment variable.
* If set, all Prometheus queries (instant and range) are automatically modified to include `{cluster="<value>"}` as a label matcher.
* Introduced a helper function `injectClusterLabelFilter` to correctly apply the label filter, whether or not existing matchers are present.

---
* Closes ...

---

## How was this PR tested?
Testing was performed in a local OpenCost setup using **Mimir** as the Prometheus backend.

### Steps:

1. Set the environment variable:

   ```bash
   export OPENCOST_CLUSTER_FILTER=stage-cluster
   ```
2. Queried the following OpenCost endpoints:

   * `http://localhost:9003/allocation?window=today&aggregate=namespace`
   * `http://localhost:9003/allocation?window=today`

### Results:

* **Before the change:**

  * The `/allocation` API returned namespaces and metrics from **all clusters**, even though OpenCost was installed on only one.
  * The total core count reflected **aggregate usage across all clusters**.

* **After the change:**

  * The same APIs now correctly return data **only for the specified cluster (`stage-cluster`)**.
  * The core count and namespace-level data now reflect **single-cluster metrics**, confirming that the cluster label filter works as expected.

---

---

If merged, this feature will require a small documentation update describing the new `OPENCOST_CLUSTER_FILTER` environment variable.

---

Marked for next release if accepted.

### DIFF
--- a/modules/prometheus-source/pkg/prom/query.go
+++ b/modules/prometheus-source/pkg/prom/query.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"time"
 
@@ -223,9 +224,17 @@ func runQuery(query string, ctx *Context, resCh source.QueryResultsChan, t time.
 
 // RawQuery is a direct query to the prometheus client and returns the body of the response
 func (ctx *Context) RawQuery(query string, t time.Time) ([]byte, error) {
-	u := ctx.Client.URL(epQuery, nil)
-	q := u.Query()
-	q.Set("query", query)
+
+       // Inject cluster label filter if OPENCOST_CLUSTER_FILTER is set
+       clusterFilter := ""
+       if val := os.Getenv("OPENCOST_CLUSTER_FILTER"); val != "" {
+	       clusterFilter = val
+       }
+       filteredQuery := injectClusterLabelFilter(query, clusterFilter)
+
+       u := ctx.Client.URL(epQuery, nil)
+       q := u.Query()
+       q.Set("query", filteredQuery)
 
 	if t.IsZero() {
 		t = time.Now()
@@ -244,7 +253,7 @@ func (ctx *Context) RawQuery(query string, t time.Time) ([]byte, error) {
 	if ctx.name != "" {
 		req = httputil.SetName(req, ctx.name)
 	}
-	req = httputil.SetQuery(req, query)
+	req = httputil.SetQuery(req, filteredQuery)
 
 	// Note that the warnings return value from client.Do() is always nil using this
 	// version of the prometheus client library. We parse the warnings out of the response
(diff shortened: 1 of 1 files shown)


mbolt35:
Thanks for the contribution, and taking the time to reach out. I _think_ that we already have support for the functionality you're looking for, but let me know:

We have three env vars that you'll want to set in order to add the cluster filters:
* `PROM_CLUSTER_ID_LABEL` - In your case, you'll want to set this to "cluster" -- I believe it defaults to "cluster_id". 
* `CURRENT_CLUSTER_ID_FILTER_ENABLED` - Set this to "true"
* `CLUSTER_ID` - This is already a requirement, but you'll want to make sure that this is set to the local cluster (which is also what you presumably want to filter).

Note the following env checks: https://github.com/opencost/opencost/blob/9bf926c15c1b29fdd3beab86354fc323a6d01b61/modules/prometheus-source/pkg/env/promenv.go#L145-L157

So, for example: Let's say your `CLUSTER_ID="foo"`, `PROM_CLUSTER_ID_LABEL="cluster"` and `CURRENT_CLUSTER_ID_FILTER_ENABLED="true"` then all outbound prometheus queries will include `cluster="foo"` in the metric filters. 

Does this sound like it covers your use cases? 

Note that if you set `LOG_LEVEL=debug` you should get prometheus query logging to show the exact queries we're making!

badal773:
Thanks a lot for the help! 🙌

I tested again with the following environment variables:

```bash
PROM_CLUSTER_ID_LABEL="cluster"
CURRENT_CLUSTER_ID_FILTER_ENABLED="true"
CLUSTER_ID="stage-cluster"
```

Everything works perfectly now! 
Looks like my earlier issue was due to `CURRENT_CLUSTER_ID_FILTER_ENABLED` not being set to `true` during previous tests.

Really appreciate your guidance , we can go ahead and close this PR.

PR #3451, opened 2025-11-12
feat: Introduce configurable metrics emitter query window

noahnut:
Added a new environment variable to set the metrics emitter query window in minutes, allowing for dynamic adjustment of the time range used in ComputeCostData queries. Default is set to 2 minutes.

## What does this PR change?

* Adds a new environment variable `METRICS_EMITTER_QUERY_WINDOW` to configure the time window for CostModelMetricsEmitter queries
* Replaces hardcoded 2-minute query window in `CostModelMetricsEmitter.Start()` with a configurable value from environment variable
* Adds `GetMetricsEmitterQueryWindow()` helper function in `pkg/env/costmodel.go` with a default value of 2 minutes

* No

* **Positive Impact**: Users with non-standard Prometheus scrape intervals can now adjust the metrics emitter query window to match their setup, improving data accuracy and reducing query errors
* **No Breaking Changes**: Default behavior remains unchanged (2 minutes), existing deployments will continue to work without modification
* **Use Case**: Particularly beneficial for users with scrape intervals > 1 minute (e.g., 3 minutes), who can now set the query window to 6-8 minutes to ensure sufficient data points for reliable metrics

* Closes ... *(if applicable)*

## How was this PR tested?

* Code builds successfully without linter errors
* Manual testing recommended: Deploy with different `METRICS_EMITTER_QUERY_WINDOW` values (e.g., 2, 3, 5) and verify:
  - Metrics are emitted correctly
  - Prometheus queries complete without errors
  - No increase in query failures for longer time windows

* **Yes** - The new environment variable `METRICS_EMITTER_QUERY_WINDOW` should be documented in:
  - Environment variables reference documentation
  - Configuration guide (with recommendations on how to set it based on scrape interval)
  - Suggested guidance: Set to at least 2× the Prometheus scrape interval

* Yes, this should be labeled as "next release" as it's a new configuration feature that improves flexibility for different deployment scenarios

---

### Additional Notes

Recommended values based on Prometheus scrape interval:
- 30s scrape → 2 minutes (default)
- 1m scrape → 3 minutes  
- 3m scrape → 6-8 minutes

### DIFF
--- a/pkg/costmodel/metrics.go
+++ b/pkg/costmodel/metrics.go
@@ -477,7 +477,8 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 			}
 
 			end := time.Now()
-			start := end.Add(-time.Minute * 2)
+			queryWindow := env.GetMetricsEmitterQueryWindow()
+			start := end.Add(-queryWindow)
 
 			data, err := cmme.Model.ComputeCostData(start, end)
 			if err != nil {
--- a/pkg/env/costmodel.go
+++ b/pkg/env/costmodel.go
@@ -1,6 +1,8 @@
 package env
 
 import (
+	"time"
+
 	"github.com/opencost/opencost/core/pkg/env"
 )
 
@@ -91,6 +93,9 @@ const (
 	// MCP Server
 	MCPServerEnabledEnvVar = "MCP_SERVER_ENABLED"
 	MCPHTTPPortEnvVar      = "MCP_HTTP_PORT"
+
+	// Metrics Emitter
+	MetricsEmitterQueryWindowEnvVar = "METRICS_EMITTER_QUERY_WINDOW"
 )
 
 func GetGCPAuthSecretFilePath() string {
@@ -382,4 +387,11 @@ func IsMCPServerEnabled() bool {
 // the HTTP port for the MCP server.
 func GetMCPHTTPPort() int {
 	return env.GetInt(MCPHTTPPortEnvVar, 8081)
+}
+
+// GetMetricsEmitterQueryWindow returns the time window for the metrics emitter
+// to query historical data. This controls the time range used in ComputeCostData queries.
+// Default is 2m.
+func GetMetricsEmitterQueryWindow() time.Duration {
+	return env.GetDuration(MetricsEmitterQueryWindowEnvVar, 2*time.Minute)
 }


pkg/costmodel/metrics.go
```diff
@@ -477,7 +477,8 @@ func (cmme *CostModelMetricsEmitter) Start() bool {
 			}
 
 			end := time.Now()
-			start := end.Add(-time.Minute * 2)
+			queryWindowMinutes := env.GetMetricsEmitterQueryWindowMinutes()
+			start := end.Add(-time.Minute * time.Duration(queryWindowMinutes))
```
ameijer: do we need time.duration here?

noahnut: I think it's important to be flexible, because prometheus scraper interval is set by user.
If the prometheus scraper interval bigger than opencost interval, some data point will missing.

mbolt35:
@Noahnut Thanks for the PR here - I agree that this is important, and I think your point is valid. What I'm concerned about is exposing 2 configuration values that allow a disagreement between the scrape interval and this value. 

Prometheus Scrape Interval is extracted from the prometheus config: https://github.com/opencost/opencost/blob/59fdcb5260e086da91cb4fb61db7ea40ca5aa346/modules/prometheus-source/pkg/prom/datasource.go#L141-L150

https://github.com/opencost/opencost/blob/59fdcb5260e086da91cb4fb61db7ea40ca5aa346/modules/prometheus-source/pkg/prom/helpers.go#L35-L62

The scrape interval for the collector-source is a bit different, so we'll want to find a way to expose it from there as well. 

My instinct is that we could expose the scrape interval on the `DataSource` and pull that duration from the CostModel. However, I'm also not sure that creating this dependency is worth it - might just be simpler to go with the separate environment variable. Let me review some things further and let you know. This is a valid solution, I just need to investigate whether or not there is a path that would automatically account for different scrape intervals.

noahnut:
Thanks for you reply. That is important point which I'm missing.
Agree with you. This may be easy for user.
I think If I use the [VictoriaMetrics](https://docs.victoriametrics.com/victoriametrics/) as my metrics component and `Vmselect` is my external prometheus data source. Is it still have correct scrape interval?
because Vmselect doesn't have job name or scrape interval in it, it belong Vmagent component.

noahnut:
Another want to ask, `err == nil` is the expected error handling? 

https://github.com/opencost/opencost/blob/59fdcb5260e086da91cb4fb61db7ea40ca5aa346/modules/prometheus-source/pkg/prom/datasource.go#L146-L149

mbolt35:
> Another want to ask, `err == nil` is the expected error handling?
> 
> https://github.com/opencost/opencost/blob/59fdcb5260e086da91cb4fb61db7ea40ca5aa346/modules/prometheus-source/pkg/prom/datasource.go#L146-L149

Yeah, so I think you're right that Victoria Metrics doesn't return this value. That's mainly what I wanted to look more into. Give me a few hours this afternoon. Thanks for the patience here.

noahnut:
Hi @mbolt35, I just wanted to follow up on the status of this PR. Have anything I can help or provide any information?
Thanks for take time read this PR.

mbolt35:
> Hi @mbolt35, I just wanted to follow up on the status of this PR. Have anything I can help or provide any information? Thanks for take time read this PR.

@Noahnut Apologies for going missing in action this past month. I had some unexpected life events followed by the holidays, but I am taking a look at this right now. Should have an update by EOD. Thanks for your patience :)

pkg/env/costmodel.go
```diff
@@ -90,6 +90,9 @@ const (
 	// MCP Server
 	MCPServerEnabledEnvVar = "MCP_SERVER_ENABLED"
 	MCPHTTPPortEnvVar      = "MCP_HTTP_PORT"
+
+	// Metrics Emitter
+	MetricsEmitterQueryWindowMinutesEnvVar = "METRICS_EMITTER_QUERY_WINDOW_MINUTES"
```
mbolt35: I'd prefer this be:

```go
MetricsEmitterQueryWindowEnvVar = "METRICS_EMITTER_QUERY_WINDOW"
```

(in addition to the changes below)

pkg/env/costmodel.go
```diff
@@ -384,3 +387,10 @@ func IsMCPServerEnabled() bool {
 func GetMCPHTTPPort() int {
 	return env.GetInt(MCPHTTPPortEnvVar, 8081)
 }
+
+// GetMetricsEmitterQueryWindowMinutes returns the time window in minutes for the metrics emitter
+// to query historical data. This controls the time range used in ComputeCostData queries.
+// Default is 2 minutes.
+func GetMetricsEmitterQueryWindowMinutes() int {
+	return env.GetInt(MetricsEmitterQueryWindowMinutesEnvVar, 2)
+}
```
mbolt35: I'd prefer you use `env.GetDuration(...)` here instead of doing the `time.Duration` conversion at the call site. 

for example (note the naming):

```go
func GetMetricsEmitterQueryWindow() time.Duration {
	return env.GetDuration(MetricsEmitterQueryWindowEnvVar, 2 * time.Minute)
}
```

mbolt35: Note that the env var would need to be something like: "5m", "2m" etc...

pkg/costmodel/metrics.go
mbolt35: Provided the other review comments, you would want to change this to the following: 

```go
queryWindow:= env.GetMetricsEmitterQueryWindow()
start := end.Add(-queryWindow)
```

mbolt35 (changes requested):
Requested changes provided.

pkg/costmodel/metrics.go
mbolt35: Also, so that we're not parsing a duration every few minutes, if you want to want to set a variable outside of the metrics loop and reference it each loop instead, that would be preferable.

noahnut:
Hi @mbolt35 
Happy new year ! I'm appreciate you for take your time to review this Pull Request.
I already fix all, if have any further question please tell me.
Thank you

mbolt35 (approved):
Looks good to me!

noahnut:
Hi @mbolt35 @ameijer,

Sorry to disturb you again.

I noticed that the GitHub Actions tests failed. However, the failures seem unrelated to my PR.

Do I need to fix anything on my end? If so, could you provide some guidance? I couldn't find any documentation regarding testing in this repository.

Thank you!

mbolt35:
> Hi @mbolt35 @ameijer,
> 
> Sorry to disturb you again.
> 
> I noticed that the GitHub Actions tests failed. However, the failures seem unrelated to my PR.
> 
> Do I need to fix anything on my end? If so, could you provide some guidance? I couldn't find any documentation regarding testing in this repository.
> 
> Thank you!

@noahnut Nothing on your end needed, we're working on fixing some things. We'll get this merged as soon as we can!

PR #3459, opened 2025-11-17
Enable the use of mTLS with the prometheus backend

mrak:
## What does this PR change?

Adds three new environment variables describing file locations for the ca certificate, client certificate, and client key, respectively. When used together, they instruct the prometheus client to connect using mTLS.

* `DB_MTLS_AUTH_CA_FILE`
* `DB_MTLS_AUTH_CRT_FILE`
* `DB_MTLS_AUTH_KEY_FILE`

* No

* This is an opt-in feature and should not impact existing users. The three environment variables will need to be specified together and their respective files mounted for the feature to be enabled.

* N/A

## How was this PR tested?
* Build and deployed to a private kubernetes cluster using Prometheus Mimir with mTLS as the prometheus endpoint.

* Yes, to the helm chart documentation describing how to enable the feature.

* TBD

### DIFF
--- a/modules/prometheus-source/pkg/env/promenv.go
+++ b/modules/prometheus-source/pkg/env/promenv.go
@@ -34,9 +34,9 @@ const (
 	DBBasicAuthPassword = "DB_BASIC_AUTH_PW"
 	DBBearerToken       = "DB_BEARER_TOKEN"
 
-	DBmTLSAuthCAFile  = "DB_MTLS_AUTH_CA_FILE"
-	DBmTLSAuthCrtFile = "DB_MTLS_AUTH_CRT_FILE"
-	DBmTLSAuthKeyFile = "DB_MTLS_AUTH_KEY_FILE"
+	PromMtlsAuthCAFile  = "PROM_MTLS_AUTH_CA_FILE"
+	PromMtlsAuthCrtFile = "PROM_MTLS_AUTH_CRT_FILE"
+	PromMtlsAuthKeyFile = "PROM_MTLS_AUTH_KEY_FILE"
 
 	CurrentClusterIdFilterEnabledVar = "CURRENT_CLUSTER_ID_FILTER_ENABLED"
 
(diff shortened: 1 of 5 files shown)


modules/prometheus-source/pkg/env/promenv.go
```diff
@@ -34,6 +34,10 @@ const (
 	DBBasicAuthPassword = "DB_BASIC_AUTH_PW"
 	DBBearerToken       = "DB_BEARER_TOKEN"
 
+	DBmTLSAuthCAFile  = "DB_MTLS_AUTH_CA_FILE"
+	DBmTLSAuthCrtFile = "DB_MTLS_AUTH_CRT_FILE"
+	DBmTLSAuthKeyFile = "DB_MTLS_AUTH_KEY_FILE"
```
ameijer: How about Prom rather than DB?

mrak: Sure, I don't mind. I was copying the naming convention of the `DB_BASIC_AUTH_PW` environment variables. I see both `PROMETHEUS_` and `PROM_` prefixes are currently in use, too. Which is preferred?

ameijer: maybe PROM? less typing lol

modules/prometheus-source/pkg/prom/config.go
```diff
@@ -88,6 +91,27 @@ func NewOpenCostPrometheusConfigFromEnv() (*OpenCostPrometheusConfig, error) {
...
+		// The /etc/ssl/cert.pem location is correct for Alpine Linux, the container base used here
+		systemCa, err := os.ReadFile("/etc/ssl/cert.pem")
+		if err != nil {
+			log.Errorf("mTLS options were set but failed to load system CAs: %s", err)
+		} else {
+			tlsCaCert.AppendCertsFromPEM(systemCa)
+		}
+		mTlsCa, err := os.ReadFile(env.GetPromMtlsAuthCAFile())
+		if err != nil {
+			log.Errorf("mTLS options were set but failed to load DB_MTLS_AUTH_CA_FILE: %s", err)
```
ameijer:
```suggestion
			log.Errorf("mTLS options were set but failed to load PROM_MTLS_AUTH_CA_FILE: %s", err)
```

modules/prometheus-source/pkg/prom/config.go
```diff
@@ -88,6 +91,27 @@ func NewOpenCostPrometheusConfigFromEnv() (*OpenCostPrometheusConfig, error) {
...
+		} else {
+			tlsCaCert.AppendCertsFromPEM(mTlsCa)
+		}
+		mTlsKeyPair, err := tls.LoadX509KeyPair(env.GetPromMtlsAuthCrtFile(), env.GetPromMtlsAuthKeyFile())
+		if err != nil {
+			log.Errorf("mTLS options were set but failed to load DB_MTLS_AUTH_CRT_FILE or DB_MTLS_AUTH_KEY_FILE: %s", err)
```
ameijer:
```suggestion
			log.Errorf("mTLS options were set but failed to load PROM_MTLS_AUTH_CRT_FILE or PROM_MTLS_AUTH_KEY_FILE: %s", err)
```

honestly, bonus points if you can use %s here and just use the var name const for these so we don't lose track of the naming

modules/prometheus-source/pkg/prom/config.go
```diff
@@ -88,6 +91,27 @@ func NewOpenCostPrometheusConfigFromEnv() (*OpenCostPrometheusConfig, error) {
 		if err != nil {
 			log.Errorf("%s was set to true but failed to load service-ca.crt: %s", env.KubeRbacProxyEnabledEnvVar, err)
 		}
+	} else if env.IsPromMtlsAuthEnabled() {
+		tlsCaCert = x509.NewCertPool()
+		// The /etc/ssl/cert.pem location is correct for Alpine Linux, the container base used here
+		systemCa, err := os.ReadFile("/etc/ssl/cert.pem")
+		if err != nil {
+			log.Errorf("mTLS options were set but failed to load system CAs: %s", err)
```
ameijer, line 99: what do you think about returning an error here, which would result in the boot blowing up? My thinking is that if someone wanted MTLS, we should fail safe if there are any issues at all?

mrak: For consistency I was following the current convention we have for, say, kube-rbac-proxy being enabled. If kube-rbac-proxy is enabled but we fail to set it up, we just log errors instead of failing.

ameijer (approved):
thanks for the contribution @mrak

PR #3470, opened 2025-11-21
Remove unused configs

thomasvn:
## What does this PR change?
* Deletes unused HTTP handlers to GET/UPDATE configs
* Updates PR template
* Removes configs from `CustomPricing` struct which are no longer used downstream. These configs were originally added to pass values through from a ConfigMap -> Backend -> Frontend. In retrospect the `CustomPricing` struct was probably not the right avenue to do so.
* Removes `ShareTenancyCost`. A default value for it was being set in multiple places, but was never used anywhere in OpenCost or downstream repos.

* N/A

* N/A

* N/A

### 1. Default install

A default install of OpenCost via the Helm chart will not have Custom Pricing enabled. Therefore the impact of these changes is negligible on a default install.

```
$ helm upgrade -i opencost --repo https://opencost.github.io/opencost-helm-chart opencost
```

```log
INF Could not find Custom Pricing file at path '/var/configs/aws.json'
INF Could not write Custom Pricing file to path '/var/configs/aws.json'
INF No pricing-configs configmap found at install time, using existing configs: configmaps "pricing-configs" not found
```

### 2. No build/test errors

Confirmed that none of the removed configs/vars are used elsewhere in OpenCost. Therefore if configuring your own custom pricing JSON there should be no effect.

Example config ([docs ref](https://opencost.io/docs/configuration/on-prem/#custom-pricing-using-the-opencost-helm-chart)):

```yaml
opencost:
  customPricing:
    enabled: true
    costModel:
      description: Modified prices based on your internal pricing
      CPU: 1.25
      RAM: 0.50
      storage: 0.25
```

* N/A

* N/A

### DIFF
--- a/pkg/cloud/models/models.go
+++ b/pkg/cloud/models/models.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/microcosm-cc/bluemonday"
 	"github.com/opencost/opencost/core/pkg/clustercache"
-	"github.com/opencost/opencost/core/pkg/log"
 	"github.com/opencost/opencost/pkg/config"
 )
 
@@ -166,7 +165,6 @@ type CustomPricing struct {
 	AthenaCURVersion             string `json:"athenaCURVersion,omitempty"` // "1.0" or "2.0", defaults to "2.0"
 	BillingDataDataset           string `json:"billingDataDataset,omitempty"`
 	CustomPricesEnabled          string `json:"customPricesEnabled"`
-	DefaultIdle                  string `json:"defaultIdle"`
 	AzureSubscriptionID          string `json:"azureSubscriptionID"`
 	AzureClientID                string `json:"azureClientID"`
 	AzureClientSecret            string `json:"azureClientSecret"`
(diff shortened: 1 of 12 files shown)


pkg/costmodel/settings.go
```diff
@@ -86,8 +86,8 @@ func (a *Accesses) customPricingHasChanged() bool {
 	// describe parameters by which we determine whether or not custom
 	// pricing settings have changed
 	encodeCustomPricing := func(cp *models.CustomPricing) string {
-		return fmt.Sprintf("%s:%s:%s:%s:%s:%s:%s:%s:%s", cp.CustomPricesEnabled, cp.CPU, cp.SpotCPU,
-			cp.RAM, cp.SpotRAM, cp.GPU, cp.Storage, cp.CurrencyCode, cp.SharedOverhead)
+		return fmt.Sprintf("%s:%s:%s:%s:%s:%s:%s:%s", cp.CustomPricesEnabled, cp.CPU, cp.SpotCPU,
```
nik-kc, line 89: Would `customPricingHasChanged()` return true for every customer who already has custom pricing enabled? Given that the parameters are guaranteed to have changed? Will this cause any problems?

pkg/cloud/models/models.go
```diff
@@ -166,7 +164,6 @@ type CustomPricing struct {
 	AthenaCURVersion             string `json:"athenaCURVersion,omitempty"` // "1.0" or "2.0", defaults to "2.0"
```
nik-kc, line 164: How did you determine which attributes are no longer used downstream?

nik-kc (commented):
Just a couple questions, but overall looks great!

pkg/costmodel/settings.go
thomasvn: Good question, I hadn't thought of that!

I don't foresee any issues. `customPricingHasChanged()` would only return true if the CustomPricing config has changed while OpenCost was continuously running. Upon restart of OpenCost, this function is entirely reset.

```go
    // cpStr is the encoded string I have changed in this PR
    // cpStrCached is the encoded CustomPricing string in the Settings Cache. That cache is flushed upon restart of the app.
	if cpStr == cpStrCached {
		return false
	}
```

pkg/cloud/models/models.go
thomasvn: For the lines I deleted, there are no other references to them in OpenCost! They were primarily used in the `kubecost-cost-model` codebase, but are no longer. I only considered Kubecost and the IBM FinOps Agent as downstream users of this section of CustomPricing.

ameijer (approved):

druanoor:
Hi. This configuration seems to removed config used for AWS Spot cost allocation like:

```
    "awsSpotDataRegion":"us-east-2",
    "awsSpotDataBucket": "my-spot-bucket",
    "awsSpotDataPrefix": "spot_pricing_prefix",
```

This is referenced here: https://opencost.io/docs/configuration/aws#aws-spot-instance-data-feed

After upgrading my helm chart from 2.4.1 to 2.5.8 and the opencost version from 1.118 to 1.119.2, im facing this error:

```
2026-02-17T15:03:36.272000427Z INF ERROR UPDATING custom-pricing-model CONFIG: error updating provider config: error setting custom pricing field: no such field: AwsSpotDataBucket in obj
```

PR #3480, opened 2025-12-02
Sth/kcm 4927

Sean-Holcomb:
## What does this PR change?
This PR provides a ClusterUID for the KubeModel (`kube-system` namespace spec UID) upon initialization because it is required for the export pathing of the KubeModel data. This is required because there is no guarantee that this information will be in any particular implementation of ClusterCache when the KubeModel exporter starts running. To accomplish this a I have added a function that retrieves this string from a Kubernetes client and saved the UID in each implementation of ClusterCache will also adding a getter for it to the ClusterCache interface.

A missing or empty ClusterUID is also grounds for a fatal error. Since continuity of this value is imperative and as far as we know it will always be present this choice has been made to avoid any undesirable behavior from silently taking place and call immediate attention to any issues

Another approach that I tried was to retrieve the the ClusterUID on the initialization of the Kubernetes client (see commit 085fa7e). The issue with this approach is that the required signature changes make it so that there would need to be changes in each command that uses the `costmodel` across multiple repos. While the modification of the ClusterCache interface is not desirable it is less invasive than this alternative. Finally as my next task I expect to include this ClusterUID in the ClusterInfo metric which I will be adding for my next task. Including the ClusterUID in ClusterCache prevents further signature changes there.

* 

* 

* Closes ...

## How was this PR tested?
* Tested running locally and ensuring that ClusterUID was properly retieved.

* 

*


pkg/kubemodel/kubemodel.go
```diff
@@ -18,16 +18,14 @@ type KubeModel struct {
 	clusterUID string
 }
 
-func NewKubeModel(dataSource source.OpenCostDataSource) (*KubeModel, error) {
+func NewKubeModel(clusterUID string, dataSource source.OpenCostDataSource) (*KubeModel, error) {
```
nikovacevic, line 21: Excellent

nikovacevic (approved):

PR #3482, opened 2025-12-04
Cloud Integration Status Refresh

nik-kc:
## Description
* The intent behind this change is to add "status refresh" functionality to cloud integrations. Currently, when a cloud integration is created, the "check if credentials are valid, and if data exists" logic occurs asynchronously. Additionally, the "does data exist?" check requires that data be pulled down from the provider.
* This PR adds `RefreshStatus()` to each cloud provider, which synchronously performs an "are the credentials valid, and does data exist?" check, without pulling down any data. It does this by attempting to list a very small subset of data from the provider, which will either fail due to bad credentials, or will succeed with "missing data" or "connection successful".
* This functionality allows downstream apps to quickly determine whether their cloud integration credentials are valid.

## Related Issues
N/A.

## User Impact
Users will have fast-feedback regarding the connection status of their cloud configs.

## Testing
Updated `kubecost-cost-model` to call `RefreshStatus()` via `/cloud/config/test`, and validated that requests for AWS, Azure, and GCP cloud integrations pass and fail as expected.
Also validated that existing cloud integration functionality continues to function as before.

### DIFF
--- a/pkg/cloud/aws/athenaintegration.go
+++ b/pkg/cloud/aws/athenaintegration.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/athena/types"
 	"github.com/opencost/opencost/core/pkg/log"
 	"github.com/opencost/opencost/core/pkg/opencost"
+	"github.com/opencost/opencost/core/pkg/util/timeutil"
 	"github.com/opencost/opencost/pkg/cloud"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/cloud/oracle/usageapiintegration.go
```diff
@@ -158,6 +158,11 @@ func (uai *UsageApiIntegration) GetStatus() cloud.ConnectionStatus {
 	return uai.ConnectionStatus
 }
 
+func (uai *UsageApiIntegration) RefreshStatus() cloud.ConnectionStatus {
+	// this function is not implemented for the oracle provider
+	return uai.ConnectionStatus
```
ameijer, line 164: logs this at warn level

pkg/cloud/aws/athenaintegration.go
```diff
@@ -68,6 +69,22 @@ type AthenaIntegration struct {
...
+}
+
+func (ai *AthenaIntegration) RefreshStatus() cloud.ConnectionStatus {
+	end := time.Now().UTC().Truncate(timeutil.Day)
+	start := end.Add(-7 * timeutil.Day)
+
+	_, err := ai.getCloudCost(start, end, 1)
+	if err != nil {
+		ai.ConnectionStatus = cloud.FailedConnection
+	}
```
ameijer, line 84: if the error is nil, does that mean we should say goot status? maybe error needs to be nil AND you get your one result in the result set?

nik-kc: `getCloudCost()` already handles that, I've added a comment to clarify this.

ameijer (approved):

PR #3485, opened 2025-12-07
initial kubemodel proposal

aaliomer-ibm:
Initial model for kubemodel.

The model uses metered fields that allows for merging kubemodelsets.
The model is improved adding standardized fields and units for type safety and self documentation

### DIFF
--- a/core/pkg/model/kubemodel/container.go
+++ b/core/pkg/model/kubemodel/container.go
@@ -6,22 +6,89 @@ import (
 )
 
 type Container struct {
-	PodUID                              string    `json:"podUid"`
-	Name                                string    `json:"name"`
-	Start                               time.Time `json:"start"`
-	End                                 time.Time `json:"end"`
-	CpuMillicoreSecondsAllocated        uint64    `json:"cpuMillicoreSecondsAllocated"`
-	CpuMillicoreRequestAverageAllocated uint64    `json:"cpuMillicoreRequestAverageAllocated"`
-	CpuMillicoreUsageAverage            uint64    `json:"cpuMillicoreUsageAverage"`
-	CpuMillicoreUsageMax                uint64    `json:"cpuMillicoreUsageMax"`
-	RAMByteSecondsAllocated             uint64    `json:"ramByteSecondsAllocated"`
-	RAMByteRequestAverageAllocated      uint64    `json:"ramByteRequestAverageAllocated"`
-	RAMByteUsageAverage                 uint64    `json:"ramByteUsageAverage"`
-	RAMByteUsageMax                     uint64    `json:"ramByteUsageMax"`
-	StorageByteSecondsAllocated         uint64    `json:"storageByteSecondsAllocated"`
-	StorageByteRequestAverageAllocated  uint64    `json:"storageByteRequestAverageAllocated"`
-	StorageByteUsageAverage             uint64    `json:"storageByteUsageAverage"`
-	StorageByteUsageMax                 uint64    `json:"storageByteUsageMax"`
+	PodUID                     string                 `json:"podUid"`
+	Name                       string                 `json:"name"`
+	DurationSeconds            Measurement            `json:"durationSeconds"`
+	CpuMillicoreSeconds        Measurement            `json:"cpuMillicoreSeconds"`
+	CpuMillicoreUsageMax       Measurement            `json:"cpuMillicoreUsageMax"`
+	CpuMillicoreRequestSeconds Measurement            `json:"cpuMillicoreRequestSeconds"`
+	RAMByteSeconds             Measurement            `json:"ramByteSeconds"`
+	RAMByteUsageMax            Measurement            `json:"ramByteUsageMax"`
+	RAMByteSecondRequest       Measurement            `json:"ramByteSecondRequest"`
+	VolumeStorageByteSeconds   map[string]Measurement `json:"volumeStorageByteSeconds,omitempty"`
+	VolumeStorageByteUsageMax  map[string]Measurement `json:"volumeStorageByteUsageMax,omitempty"`
+	CpuMillicoreLimitSeconds   Measurement            `json:"cpuMillicoreLimitSeconds,omitempty"`
+	RAMByteSecondsLimit        Measurement            `json:"ramByteSecondsLimit,omitempty"`
+	Start                      time.Time              `json:"start"`
+	End                        time.Time              `json:"end"`
+}
+
+func (c *Container) CpuMillicoreUsageAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	return c.CpuMillicoreSeconds / c.DurationSeconds
+}
+
+func (c *Container) RAMByteUsageAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	return c.RAMByteSeconds / c.DurationSeconds
+}
+
+func (c *Container) TotalStorageByteSeconds() Measurement {
+	var total Measurement
+	for _, ByteSeconds := range c.VolumeStorageByteSeconds {
+		total += ByteSeconds
+	}
+	return total
+}
+
+func (c *Container) TotalStorageByteUsageMax() Measurement {
+	var max Measurement
+	for _, usage := range c.VolumeStorageByteUsageMax {
+		if usage > max {
+			max = usage
+		}
+	}
+	return max
+}
+
+func (c *Container) StorageByteUsageAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	totalByteSeconds := c.TotalStorageByteSeconds()
+	return totalByteSeconds / c.DurationSeconds
+}
+
+func (c *Container) CpuMillicoreRequestAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	return c.CpuMillicoreRequestSeconds / c.DurationSeconds
+}
+
+func (c *Container) RAMByteRequestAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	return c.RAMByteSecondRequest / c.DurationSeconds
+}
+
+func (c *Container) CpuMillicoreLimitAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	return c.CpuMillicoreLimitSeconds / c.DurationSeconds
+}
+
+func (c *Container) RAMByteLimitAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	return c.RAMByteSecondsLimit / c.DurationSeconds
 }
 
 func (kms *KubeModelSet) RegisterContainer(uid, name, podUID string) error {
--- a/core/pkg/model/kubemodel/device.go
+++ b/core/pkg/model/kubemodel/device.go
@@ -1,0 +1,92 @@
+package kubemodel
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// @bingen:generate:Device
+type Device struct {
+	UID               string      `json:"uid"`            // Device UUID (hardware identifier)
+	Type              string      `json:"type,omitempty"` // Device type (e.g., "device", "tpu")
+	NodeUID           string      `json:"nodeUid"`        // Node hosting this device
+	DeviceNumber      int32       `json:"deviceNumber"`
+	ModelName         string      `json:"modelName"`
+	IsShared          bool        `json:"isShared"` // Device sharing information
+	SharePercentage   float64     `json:"sharePercentage"`
+	UsageSeconds      float64     `json:"usageSeconds"`      // Device seconds available
+	MemoryByteSeconds Measurement `json:"memoryByteSeconds"` // Device memory capacity in Byte-seconds
+	PowerWattSeconds  float64     `json:"powerWattSeconds"`  // Device power consumption in watt-seconds (Joules)
+	PowerWattMax      float64     `json:"powerWattMax"`      // Device max power consumption in watts
+	// Version 2 fields - Lifecycle tracking
+	Start           time.Time   `json:"start,omitempty"` // Device availability start
+	End             time.Time   `json:"end,omitempty"`   // Device availability end
+	DurationSeconds Measurement `json:"durationSeconds"` // Duration device was available
+}
+
+// Validate validates the Device fields
+func (d *Device) Validate() error {
+	if d.UID == "" {
+		return errors.New("UID is required")
+	}
+	if d.NodeUID == "" {
+		return errors.New("NodeUID is required")
+	}
+	if d.SharePercentage < 0 || d.SharePercentage > 100 {
+		return fmt.Errorf("SharePercentage must be 0-100, got %.2f", d.SharePercentage)
+	}
+	if d.PowerWattSeconds < 0 {
+		return fmt.Errorf("PowerWattSeconds cannot be negative, got %.2f", d.PowerWattSeconds)
+	}
+	if d.PowerWattMax < 0 {
+		return fmt.Errorf("PowerWattMax cannot be negative, got %.2f", d.PowerWattMax)
+	}
+	return nil
+}
+
+// Clone creates a deep copy of the Device
+func (d *Device) Clone() *Device {
+	if d == nil {
+		return nil
+	}
+
+	cloned := &Device{
+		UID:               d.UID,
+		Type:              d.Type,
+		NodeUID:           d.NodeUID,
+		DeviceNumber:      d.DeviceNumber,
+		ModelName:         d.ModelName,
+		IsShared:          d.IsShared,
+		SharePercentage:   d.SharePercentage,
+		UsageSeconds:      d.UsageSeconds,
+		MemoryByteSeconds: d.MemoryByteSeconds,
+		PowerWattSeconds:  d.PowerWattSeconds,
+		PowerWattMax:      d.PowerWattMax,
+		DurationSeconds:   d.DurationSeconds,
+	}
+
+	cloned.Start = d.Start
+	cloned.End = d.End
+
+	return cloned
+}
+
+func (kms *KubeModelSet) RegisterDevice(uid, nodeUID string) error {
+	if uid == "" {
+		err := fmt.Errorf("UID is nil for Device")
+		kms.Error(err)
+		return err
+	}
+
+	if _, ok := kms.Devices[uid]; !ok {
+		kms.Devices[uid] = &Device{
+			UID:     uid,
+			NodeUID: nodeUID,
+		}
+
+		kms.Metadata.ObjectCount++
+	}
+
+	return nil
+}
--- a/core/pkg/model/kubemodel/namespace.go
+++ b/core/pkg/model/kubemodel/namespace.go
@@ -7,13 +7,13 @@ import (
 
 // @bingen:generate:Namespace
 type Namespace struct {
-	UID         string            `json:"uid"`         // @bingen:field[version=1]
-	ClusterUID  string            `json:"clusterUID"`  // @bingen:field[version=1]
-	Name        string            `json:"name"`        // @bingen:field[version=1]
-	Labels      map[string]string `json:"labels"`      // @bingen:field[version=1]
-	Annotations map[string]string `json:"annotations"` // @bingen:field[version=1]
-	Start       time.Time         `json:"start"`       // @bingen:field[version=1]
-	End         time.Time         `json:"end"`         // @bingen:field[version=1]
+	UID         string            `json:"uid"`             // @bingen:field[version=1]
+	ClusterUID  string            `json:"clusterUID"`      // @bingen:field[version=1]
+	Name        string            `json:"name"`            // @bingen:field[version=1]
+	Labels      map[string]string `json:"labels"`          // @bingen:field[version=1]
+	Annotations map[string]string `json:"annotations"`     // @bingen:field[version=1]
+	Start       time.Time         `json:"start,omitempty"` // @bingen:field[version=1]
+	End         time.Time         `json:"end,omitempty"`   // @bingen:field[version=1]
 }
 
 func (kms *KubeModelSet) RegisterNamespace(uid, name string) error {
--- a/core/pkg/model/kubemodel/node.go
+++ b/core/pkg/model/kubemodel/node.go
@@ -5,25 +5,81 @@ import (
 	"time"
 )
 
+// @bingen:generate:Node
+// Node represents a Kubernetes node with capacity-based resource tracking.
+// All resource measures (CPU, RAM) represent node capacity, not requests or limits.
+// This aligns with the principle that cost allocation should be based on provisioned capacity.
 type Node struct {
-	UID                          string            `json:"uid"`
-	ClusterUID                   string            `json:"clusterUid"`
-	ProviderResourceUID          string            `json:"providerResourceUid"`
-	Name                         string            `json:"name"`
-	Labels                       map[string]string `json:"labels,omitempty"`
-	Annotations                  map[string]string `json:"annotations,omitempty"`
-	Start                        time.Time         `json:"start"`
-	End                          time.Time         `json:"end"`
-	CpuMillicoreSecondsAllocated uint64            `json:"cpuMillicoreSecondsAllocated"`
-	RAMByteSecondsAllocated      uint64            `json:"ramByteSecondsAllocated"`
-	// PublicIPSeconds represents the cumulative public IP allocation (count × seconds) for this node.
-	// Calculated as: number of ExternalIP addresses from Kubernetes node Status.Addresses × window duration in seconds.
-	// Used for cost attribution of public IP addresses associated with the node.
-	PublicIPSecondsAllocated uint64 `json:"publicIpSecondsAllocated"`
-	CpuMillicoreUsageAverage uint64 `json:"cpuMillicoreUsageAverage"`
-	CpuMillicoreUsageMax     uint64 `json:"cpuMillicoreUsageMax"`
-	RAMByteUsageAverage      uint64 `json:"ramByteUsageAverage"`
-	RAMByteUsageMax          uint64 `json:"ramByteUsageMax"`
+	UID                  string                      `json:"uid"`
+	ProviderResourceUID  string                      `json:"providerResourceUid"`
+	Name                 string                      `json:"name"`
+	Labels               map[string]string           `json:"labels,omitempty"`
+	Annotations          map[string]string           `json:"annotations,omitempty"`
+	DurationSeconds      Measurement                 `json:"durationSeconds"`
+	CpuMillicoreSeconds  Measurement                 `json:"cpuMillicoreSeconds"` // Node CPU capacity in millicore-seconds
+	RAMByteSeconds       Measurement                 `json:"ramByteSeconds"`      // Node RAM capacity in Byte-seconds
+	AttachedVolumes      map[string]*NodeVolumeUsage `json:"attachedVolumes,omitempty"`
+	CpuMillicoreUsageMax Measurement                 `json:"cpuMillicoreUsageMax"` // Peak CPU usage observed
+	RAMByteUsageMax      Measurement                 `json:"ramByteUsageMax"`      // Peak RAM usage observed
+	Start                time.Time                   `json:"start,omitempty"`      // Node creation/start timestamp
+	End                  time.Time                   `json:"end,omitempty"`        // Node deletion/end timestamp (nil if still running)
+}
+
+// NodeVolumeUsage tracks storage usage for a disk volume attached to a node.
+// Used for cost allocation of cloud storage resources (e.g., AWS EBS volumes).
+type NodeVolumeUsage struct {
+	VolumeUID        string      `json:"volumeUid"`        // "root" for primary disk, or actual volume UID for additional volumes
+	CapacityBytes    Measurement `json:"capacityBytes"`    // Total capacity of the volume in bytes
+	UsageByteSeconds Measurement `json:"usageByteSeconds"` // Cumulative usage (Byte × seconds) over measurement window
+	VolumeType       string      `json:"volumeType"`       // "root" for primary disk, "persistent" for additional PVs
+	ProviderID       string      `json:"providerId"`       // Cloud provider volume ID (e.g., "vol-xxxxx" for AWS EBS)
+	DurationSeconds  Measurement `json:"durationSeconds"`  // Duration the volume was attached during measurement window in seconds
+}
+
+// CpuMillicoreUsageAverage calculates the average CPU usage in millicores over the uptime period.
+// Returns 0 if uptime is 0 to avoid division by zero.
+func (n *Node) CpuMillicoreUsageAverage() Measurement {
+	if n.DurationSeconds == 0 {
+		return 0
+	}
+	return n.CpuMillicoreSeconds / n.DurationSeconds
+}
+
+// RAMByteUsageAverage calculates the average RAM usage in bytes over the uptime period.
+// Returns 0 if uptime is 0 to avoid division by zero.
+func (n *Node) RAMByteUsageAverage() Measurement {
+	if n.DurationSeconds == 0 {
+		return 0
+	}
+	return n.RAMByteSeconds / n.DurationSeconds
+}
+
+// TotalVolumeUsageByteSeconds returns the sum of all volume usage Byte-seconds across all attached volumes.
+func (n *Node) TotalVolumeUsageByteSeconds() Measurement {
+	var total Measurement
+	for _, volume := range n.AttachedVolumes {
+		total += volume.UsageByteSeconds
+	}
+	return total
+}
+
+// TotalVolumeCapacityBytes returns the sum of all volume capacities across all attached volumes.
+func (n *Node) TotalVolumeCapacityBytes() Measurement {
+	var total Measurement
+	for _, volume := range n.AttachedVolumes {
+		total += volume.CapacityBytes
+	}
+	return total
+}
+
+// GetVolumeUsageAverage calculates the average storage usage in bytes for a specific volume over the uptime period.
+// Returns 0 if uptime is 0 or volume doesn't exist.
+func (n *Node) GetVolumeUsageAverage(volumeUID string) Measurement {
+	volume, exists := n.AttachedVolumes[volumeUID]
+	if !exists || n.DurationSeconds == 0 {
+		return 0
+	}
+	return volume.UsageByteSeconds / n.DurationSeconds
 }
 
 func (kms *KubeModelSet) RegisterNode(uid, name string) error {
@@ -34,18 +90,14 @@ func (kms *KubeModelSet) RegisterNode(uid, name string) error {
 	}
 
 	if _, ok := kms.Nodes[uid]; !ok {
-		clusterUID := ""
-
 		if kms.Cluster == nil {
 			kms.Warnf("RegisterNode(%s, %s): Cluster is nil", uid, name)
-		} else {
-			clusterUID = kms.Cluster.UID
 		}
 
 		kms.Nodes[uid] = &Node{
-			UID:        uid,
-			ClusterUID: clusterUID,
-			Name:       name,
+			UID:             uid,
+			Name:            name,
+			AttachedVolumes: make(map[string]*NodeVolumeUsage),
 		}
 
 		kms.Metadata.ObjectCount++
--- a/core/pkg/model/kubemodel/pod.go
+++ b/core/pkg/model/kubemodel/pod.go
@@ -8,17 +8,16 @@ import (
 type Pod struct {
 	UID                  string            `json:"uid"`
 	NamespaceUID         string            `json:"namespaceUid"`
-	OwnerUID             string            `json:"ownerUid"`
+	OwnerUID             string            `json:"ownerUid"` // Reference to Owner (Deployment, StatefulSet, etc.)
 	NodeUID              string            `json:"nodeUid"`
 	Name                 string            `json:"name"`
 	Labels               map[string]string `json:"labels,omitempty"`
 	Annotations          map[string]string `json:"annotations,omitempty"`
-	Start                time.Time         `json:"start"`
-	End                  time.Time         `json:"end"`
-	CpuMillicoreUsageMax uint64            `json:"cpuMillicoreUsageMax"`
-	RAMByteUsageMax      uint64            `json:"ramByteUsageMax"`
-	NetworkTransferBytes uint64            `json:"networkTransferBytes"`
-	NetworkReceiveBytes  uint64            `json:"networkReceiveBytes"`
+	DurationSeconds      Measurement       `json:"durationSeconds"`
+	NetworkTransferBytes Measurement       `json:"networkTransferBytes"`
+	NetworkReceiveBytes  Measurement       `json:"networkReceiveBytes"`
+	Start                time.Time         `json:"start,omitempty"` // Pod creation/start timestamp
+	End                  time.Time         `json:"end,omitempty"`   // Pod deletion/end timestamp (nil if still running)
 }
 
 func (kms *KubeModelSet) RegisterPod(uid, name, namespace string) error {
--- a/core/pkg/model/kubemodel/pv.go
+++ b/core/pkg/model/kubemodel/pv.go
@@ -1,0 +1,72 @@
+package kubemodel
+
+import (
+	"fmt"
+	"time"
+)
+
+// @bingen:generate:PersistentVolume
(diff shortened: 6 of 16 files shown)


core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,17 @@
+//nolint:stylecheck // generated code and package conventions
```
ameijer: why are you disabling the style linter on all the files?

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,17 @@
...
-
 // @bingen:generate:Cluster
 type Cluster struct {
-	UID      string    `json:"uid"`      // @bingen:field[version=1]
-	Provider Provider  `json:"provider"` // @bingen:field[version=1]
-	Account  string    `json:"account"`  // @bingen:field[version=1]
-	Name     string    `json:"name"`     // @bingen:field[version=1]
-	Start    time.Time `json:"start"`    // @bingen:field[version=1]
-	End      time.Time `json:"end"`      // @bingen:field[version=1]
+	UID          string   `json:"uid"`          // @bingen:field[version=1]
```
ameijer: were we going to turn this into a uuid.UUID type?

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,17 @@
...
-import "time"
-
 // @bingen:generate:Cluster
 type Cluster struct {
-	UID      string    `json:"uid"`      // @bingen:field[version=1]
-	Provider Provider  `json:"provider"` // @bingen:field[version=1]
-	Account  string    `json:"account"`  // @bingen:field[version=1]
-	Name     string    `json:"name"`     // @bingen:field[version=1]
-	Start    time.Time `json:"start"`    // @bingen:field[version=1]
-	End      time.Time `json:"end"`      // @bingen:field[version=1]
```
ameijer, line 15: did we want to keep start/end here?

core/pkg/model/kubemodel/container.go
```diff
@@ -1,23 +1,105 @@
...
 package kubemodel
 
-import "time"
-
 // @bingen:generate:Container
 type Container struct {
-	PodUID                              string    `json:"podUid"`                              // @bingen:field[version=1]
-	Name                                string    `json:"name"`                                // @bingen:field[version=1]
-	Start                               time.Time `json:"start"`                               // @bingen:field[version=1]
-	End                                 time.Time `json:"end"`                                 // @bingen:field[version=1]
```
ameijer: I think we do still want start/end here

core/pkg/model/kubemodel/node.go
```diff
@@ -1,25 +1,93 @@
...
+}
+
+// GetVolumeUsageAverage calculates the average storage usage in bytes for a specific volume over the uptime period.
+// Returns 0 if uptime is 0 or volume doesn't exist.
+func (n *Node) GetVolumeUsageAverage(volumeUID string) uint64 {
+	volume, exists := n.AttachedVolumes[volumeUID]
+	if !exists || n.DurationSeconds == 0 {
+		return 0
+	}
+	return KiBToBytes(volume.UsageKiBSeconds) / n.DurationSeconds
```
Sean-Holcomb: If we are worried about uint overflow, maybe we should divide by seconds first than convert to bytes

core/pkg/model/kubemodel/node.go
```diff
@@ -1,25 +1,93 @@
...
+	RAMKiBSeconds       uint64            `json:"ramKiBSeconds"`         // @bingen:field[version=1] - Node RAM capacity in KiB-seconds
 	// PublicIPSeconds represents the cumulative public IP allocation (count × seconds) for this node.
 	// Calculated as: number of ExternalIP addresses from Kubernetes node Status.Addresses × window duration in seconds.
 	// Used for cost attribution of public IP addresses associated with the node.
-	PublicIPSecondsAllocated uint64 `json:"publicIpSecondsAllocated"` // @bingen:field[version=1]
-	CpuMillicoreUsageAverage uint64 `json:"cpuMillicoreUsageAverage"` // @bingen:field[version=1]
-	CpuMillicoreUsageMax     uint64 `json:"cpuMillicoreUsageMax"`     // @bingen:field[version=1]
-	RAMByteUsageAverage      uint64 `json:"ramByteUsageAverage"`      // @bingen:field[version=1]
-	RAMByteUsageMax          uint64 `json:"ramByteUsageMax"`          // @bingen:field[version=1]
+	PublicIPSeconds uint64 `json:"publicIpSeconds"` // @bingen:field[version=1]
```
Sean-Holcomb: not too sure about this one. It implies that we have a means of tracking the number of seconds that each public IP was attached to the node

core/pkg/model/kubemodel/namespace.go
```diff
@@ -1,14 +1,10 @@
...
-import "time"
-
 // @bingen:generate:Namespace
 type Namespace struct {
 	UID         string            `json:"uid"`         // @bingen:field[version=1]
-	ClusterUID  string            `json:"clusterUID"`  // @bingen:field[version=1]
 	Name        string            `json:"name"`        // @bingen:field[version=1]
 	Labels      map[string]string `json:"labels"`      // @bingen:field[version=1]
 	Annotations map[string]string `json:"annotations"` // @bingen:field[version=1]
-	Start       time.Time         `json:"start"`       // @bingen:field[version=1]
```
Sean-Holcomb, line 15: In general we are using stat and end on all types. start and end represent the time at which we started receiving metrics for a particular object

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,17 @@
...
+	UID          string   `json:"uid"`          // @bingen:field[version=1]
+	Provider     Provider `json:"provider"`     // @bingen:field[version=1]
+	Account      string   `json:"account"`      // @bingen:field[version=1]
+	Name         string   `json:"name"`         // @bingen:field[version=1]
+	Agent        string   `json:"agent"`        // @bingen:field[version=1]
+	AgentVersion string   `json:"agentVersion"` // @bingen:field[version=1]
+	K8sVersion   string   `json:"version"`      // @bingen:field[version=1]
+	ClusterType  string   `json:"type"`         // @bingen:field[version=1]
```
Sean-Holcomb: What is cluster type?

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,17 @@
...
-	Provider Provider  `json:"provider"` // @bingen:field[version=1]
-	Account  string    `json:"account"`  // @bingen:field[version=1]
-	Name     string    `json:"name"`     // @bingen:field[version=1]
-	Start    time.Time `json:"start"`    // @bingen:field[version=1]
-	End      time.Time `json:"end"`      // @bingen:field[version=1]
+	UID          string   `json:"uid"`          // @bingen:field[version=1]
+	Provider     Provider `json:"provider"`     // @bingen:field[version=1]
+	Account      string   `json:"account"`      // @bingen:field[version=1]
+	Name         string   `json:"name"`         // @bingen:field[version=1]
+	Agent        string   `json:"agent"`        // @bingen:field[version=1]
```
Sean-Holcomb: What is Agent?

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,17 @@
...
+	AgentVersion string   `json:"agentVersion"` // @bingen:field[version=1]
```
Sean-Holcomb: This code is in Opencost, it doesn't not have a notion of any parent application that is calling it. The bingen codex do have a version, maybe we could include that here.

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,17 @@
...
+	K8sVersion   string   `json:"version"`      // @bingen:field[version=1]
+	ClusterType  string   `json:"type"`         // @bingen:field[version=1]
+	Region       string   `json:"region"`       // @bingen:field[version=1]
```
Sean-Holcomb: I know that generally speaking a cluster will always be in a single region, but it is possible to create a cluster that spans multiple regions.

core/pkg/model/kubemodel/device.go
```diff
@@ -0,0 +1,118 @@
...
+	DeviceNumber     int32   `json:"deviceNumber"`     // @bingen:field[version=1]
+	ModelName        string  `json:"modelName"`        // @bingen:field[version=1]
+	IsShared         bool    `json:"isShared"`         // @bingen:field[version=1] GPU sharing information
+	SharePercentage  float64 `json:"sharePercentage"`  // @bingen:field[version=1]
+	UsageSeconds     float64 `json:"usageSeconds"`     // @bingen:field[version=1] GPU seconds available
+	MemoryKiBSeconds uint64  `json:"memoryKiBSeconds"` // @bingen:field[version=1] GPU memory capacity in KiB-seconds
+	PowerWattSeconds float64 `json:"powerWattSeconds"` // @bingen:field[version=1] GPU device power consumption in watt-seconds (Joules)
+	PowerWattMax     float64 `json:"powerWattMax"`     // @bingen:field[version=1] GPU device max power consumption in watts
+	// Version 2 fields - Lifecycle tracking
+	Start           *time.Time `json:"start,omitempty"` // @bingen:field[version=1] - Device availability start
```
Sean-Holcomb: consistence on start and end type

core/pkg/model/kubemodel/pvc.go
```diff
@@ -0,0 +1,38 @@
...
+	// Version 2 fields - FinOps enhancements
+	// ReadWriteOnce, ReadWriteMany, ReadOnlyMany
+	AccessModes []string `json:"accessModes,omitempty"` // @bingen:field[version=1]
+	// Real usage KiB-seconds from kubelet stats
+	ActualUsedKiBSeconds uint64 `json:"actualUsedKiBSeconds,omitempty"` // @bingen:field[version=1]
+	// Generic attributes (IOPS, throughput, etc.) from CSI or provider
+	VolumeAttributes map[string]string `json:"volumeAttributes,omitempty"` // @bingen:field[version=1]
+	// PVC lifecycle timestamps
+	Start time.Time  `json:"start"`         // @bingen:field[version=1] - PVC creation timestamp
+	End   *time.Time `json:"end,omitempty"` // @bingen:field[version=1] - PVC deletion timestamp (nil if still active)
```
Sean-Holcomb: end consistence

core/pkg/model/kubemodel/window.go
```diff
@@ -6,4 +7,6 @@ import "time"
 type Window struct {
 	Start time.Time `json:"start"` // @bingen:field[version=1]
 	End   time.Time `json:"end"`   // @bingen:field[version=1]
+	// Sometimes the duration is smaller than the time between start and end (missing data for example)
+	DurationSeconds uint64 `json:"durationSeconds"` // @bingen:field[version=1]
```
Sean-Holcomb: Can you elaborate a bit more?

core/pkg/model/kubemodel/namespace.go
```diff
@@ -1,14 +1,10 @@
+//nolint:stylecheck // generated code and package conventions
 package kubemodel
 
-import "time"
-
 // @bingen:generate:Namespace
 type Namespace struct {
 	UID         string            `json:"uid"`         // @bingen:field[version=1]
-	ClusterUID  string            `json:"clusterUID"`  // @bingen:field[version=1]
```
Sean-Holcomb, line 11: Good call

Sean-Holcomb (changes requested):
Just looking for elaboration with most of the comments, and some standardization with some of the fields

core/pkg/model/kubemodel/device.go
```diff
@@ -0,0 +1,118 @@
...
+package kubemodel
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// @bingen:generate:Device
+// Device represents a physical GPU device with DCGM integration (provisioned resource)
```
ameijer: I'd remove all gpu references in this comment if we are trying to generalize

core/pkg/model/kubemodel/device.go
```diff
@@ -0,0 +1,118 @@
...
+	IsShared         bool    `json:"isShared"`         // @bingen:field[version=1] GPU sharing information
+	SharePercentage  float64 `json:"sharePercentage"`  // @bingen:field[version=1]
+	UsageSeconds     float64 `json:"usageSeconds"`     // @bingen:field[version=1] GPU seconds available
+	MemoryKiBSeconds uint64  `json:"memoryKiBSeconds"` // @bingen:field[version=1] GPU memory capacity in KiB-seconds
+	PowerWattSeconds float64 `json:"powerWattSeconds"` // @bingen:field[version=1] GPU device power consumption in watt-seconds (Joules)
+	PowerWattMax     float64 `json:"powerWattMax"`     // @bingen:field[version=1] GPU device max power consumption in watts
+	// Version 2 fields - Lifecycle tracking
+	Start           *time.Time `json:"start,omitempty"` // @bingen:field[version=1] - Device availability start
+	End             *time.Time `json:"end,omitempty"`   // @bingen:field[version=1] - Device availability end
+	DurationSeconds uint64     `json:"durationSeconds"` // @bingen:field[version=1] - Duration device was available
```
ameijer: I don't think we need duration w/ start and end

core/pkg/model/kubemodel/namespace.go
ameijer: why would having a namespace point to a cluster be a bad thing?

core/pkg/model/kubemodel/namespace.go
ameijer: agreed, we should keep these

core/pkg/model/kubemodel/node.go
ameijer: I think we do have such a means?

core/pkg/model/kubemodel/pod.go
```diff
@@ -1,20 +1,26 @@
...
+	NetworkTransferBytes uint64            `json:"networkTransferBytes"`  // @bingen:field[version=1]
+	NetworkReceiveBytes  uint64            `json:"networkReceiveBytes"`   // @bingen:field[version=1]
+	// Version 2 fields - Lifecycle tracking
+	Start *time.Time `json:"start,omitempty"` // @bingen:field[version=1] - Pod creation/start timestamp
+	End   *time.Time `json:"end,omitempty"`   // @bingen:field[version=1] - Pod deletion/end timestamp (nil if still running)
+	// Version 2 fields - Network breakdown by destination type
+	NetworkInternetEgressBytes uint64 `json:"networkInternetEgressBytes,omitempty"` // @bingen:field[version=1]
+	NetworkCrossRegionBytes    uint64 `json:"networkCrossRegionBytes,omitempty"`    // @bingen:field[version=1]
+	NetworkSameRegionBytes     uint64 `json:"networkSameRegionBytes,omitempty"`     // @bingen:field[version=1]
+	NetworkIntraAZBytes        uint64 `json:"networkIntraAzBytes,omitempty"`        // @bingen:field[version=1]
```
ameijer: should we break this out to a new type? check out the networkInsights objects in this repo

core/pkg/model/kubemodel/pv.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// Version 2 fields - FinOps enhancements
+	// awsElasticBlockStore, azureDisk, gcePersistentDisk, csi, nfs, local, etc.
+	Type string `json:"type,omitempty"` // @bingen:field[version=1]
+	// ebs.csi.aws.com, disk.csi.azure.com, etc.
+	CSIDriver string `json:"csiDriver,omitempty"` // @bingen:field[version=1]
+	// Cloud provider's volume identifier
+	ProviderVolumeID string `json:"providerVolumeId,omitempty"` // @bingen:field[version=1]
+	// ReadWriteOnce, ReadWriteMany, ReadOnlyMany
+	AccessModes []string `json:"accessModes,omitempty"` // @bingen:field[version=1]
```
ameijer: do we need this?

core/pkg/model/kubemodel/service.go
```diff
@@ -36,4 +58,15 @@ type Service struct {
...
+	// Maps label keys to values (e.g., {"app": "nginx", "tier": "frontend"})
+	// Pods with matching labels will receive traffic from this service
+	Selector map[string]string `json:"selector,omitempty"` // @bingen:field[version=1]
+	// Lifecycle tracking
+	DurationSeconds uint64 `json:"durationSeconds"` // @bingen:field[version=1] - Duration service existed within measurement window
+	// Version 2 fields - Network breakdown by destination type
+	NetworkInternetEgressBytes uint64 `json:"networkInternetEgressBytes,omitempty"` // @bingen:field[version=1]
+	NetworkCrossRegionBytes    uint64 `json:"networkCrossRegionBytes,omitempty"`    // @bingen:field[version=1]
+	NetworkSameRegionBytes     uint64 `json:"networkSameRegionBytes,omitempty"`     // @bingen:field[version=1]
+	NetworkIntraAZBytes        uint64 `json:"networkIntraAzBytes,omitempty"`        // @bingen:field[version=1]
```
ameijer: services don't have network traffic, they are just an abstraction

core/pkg/model/kubemodel/node.go
aaliomer-ibm: I'm honestly not confident about the networking metrics. I'm thinking about just taking it out completely until we talk about it and consult matt bolt at another time on a different PR. @ameijer 
Generally these are the fields we want but no idea how to get them
```
	CrossZoneBytes uint64
	CrossRegionBytes uint64
	InternetBytes uint64
	NatGatewayBytes uint64
}
```

core/pkg/model/kubemodel/cluster.go
aaliomer-ibm: will fix it. was giving issues with cldy before moving it

core/pkg/model/kubemodel/cluster.go
aaliomer-ibm: Yes! forgot about that but will get on it

core/pkg/model/kubemodel/cluster.go
aaliomer-ibm: agent type. I'm actually reverting this file. I like your suggestion of the heartbeat mechanism better

core/pkg/model/kubemodel/cluster.go
aaliomer-ibm: will put it back just in case

core/pkg/model/kubemodel/cluster.go
aaliomer-ibm: openshift, k8s, etc

aaliomer-ibm: gonna go away

core/pkg/model/kubemodel/device.go
aaliomer-ibm: you're right. downtime is not relevant in this case probably

core/pkg/model/kubemodel/pv.go
aaliomer-ibm: Brock think it is useful. better have it and not need it

core/pkg/model/kubemodel/window.go
aaliomer-ibm: i can add an example to the comment :
`if the agent is down for 20 minutes in the middle of an hour span, then start and end will be an hour span but the duration will be only 40 minutes to account for the missing 20 minute metric window`

core/pkg/model/kubemodel/cluster.go
aaliomer-ibm: actually probably not belong here.

core/pkg/model/kubemodel/device.go
aaliomer-ibm: actually we do. we have MemoryKiBSeconds and other cumulative units. anywhere we have those, we need duration

core/pkg/model/kubemodel/namespace.go
aaliomer-ibm: its one cluster for the entire kubemodelset

core/pkg/model/kubemodel/pod.go
aaliomer-ibm: removed for now till we think it properly

core/pkg/model/kubemodel/service.go
aaliomer-ibm: should i remove 59, 60?

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,23 @@
+//nolint:stylecheck
```
ameijer: can we remove the linter disables?

core/pkg/model/kubemodel/node.go
ameijer: I agree with sean here

core/pkg/model/kubemodel/node.go
ameijer: can't we use this method to obtain this? https://github.com/opencost/opencost/blob/76a4bffbe455539bcfac26e567e84fa47d80de94/core/pkg/clustercache/clustercache.go#L186

core/pkg/model/kubemodel/container.go
```diff
@@ -1,23 +1,95 @@
...
-	RAMByteUsageAverage                 uint64    `json:"ramByteUsageAverage"`                 // @bingen:field[version=1]
-	RAMByteUsageMax                     uint64    `json:"ramByteUsageMax"`                     // @bingen:field[version=1]
-	StorageByteSecondsAllocated         uint64    `json:"storageByteSecondsAllocated"`         // @bingen:field[version=1]
-	StorageByteRequestAverageAllocated  uint64    `json:"storageByteRequestAverageAllocated"`  // @bingen:field[version=1]
-	StorageByteUsageAverage             uint64    `json:"storageByteUsageAverage"`             // @bingen:field[version=1]
-	StorageByteUsageMax                 uint64    `json:"storageByteUsageMax"`                 // @bingen:field[version=1]
+	PodUID                     uuid.UUID            `json:"podUid"`                              // @bingen:field[version=1]
+	Name                       string               `json:"name"`                                // @bingen:field[version=1]
+	DurationSeconds            uint64               `json:"durationSeconds"`                     // @bingen:field[version=1]
```
ameijer: couldn't duration be computed as end - start time?

core/pkg/model/kubemodel/cluster.go
ameijer: looks like it is still present. @Sean-Holcomb are you good with this getting tossed?

core/pkg/model/kubemodel/cluster.go
ameijer: looks to be still present

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,13 +1,23 @@
...
+	Provider     Provider  `json:"provider"`        // @bingen:field[version=1]
+	Account      string    `json:"account"`         // @bingen:field[version=1]
+	Name         string    `json:"name"`            // @bingen:field[version=1]
+	Agent        string    `json:"agent"`           // @bingen:field[version=1]
+	AgentVersion string    `json:"agentVersion"`    // @bingen:field[version=1]
+	K8sVersion   string    `json:"version"`         // @bingen:field[version=1]
+	ClusterType  string    `json:"type"`            // @bingen:field[version=1]
+	Region       string    `json:"region"`          // @bingen:field[version=1]
+	Start        time.Time `json:"start,omitempty"` // @bingen:field[version=1]
+	End          time.Time `json:"end,omitempty"`   // @bingen:field[version=1]
```
ameijer: set all bingen fields to 'ignore' until we finish hydrating them

core/pkg/model/kubemodel/cluster.go
ameijer: looks to be still present

core/pkg/model/kubemodel/namespace.go
ameijer: oh! of course

core/pkg/model/kubemodel/device.go
```diff
@@ -0,0 +1,79 @@
...
+// @bingen:generate:Device
+type Device struct {
+	UID              uuid.UUID `json:"uid"`              // @bingen:field[version=1] GPU UUID (hardware identifier)
+	Type             string    `json:"type,omitempty"`   // @bingen:field[version=1] Device type (e.g., "gpu", "tpu")
+	NodeUID          uuid.UUID `json:"nodeUid"`          // @bingen:field[version=1] Node hosting this GPU device
+	DeviceNumber     int32     `json:"deviceNumber"`     // @bingen:field[version=1]
+	ModelName        string    `json:"modelName"`        // @bingen:field[version=1]
+	IsShared         bool      `json:"isShared"`         // @bingen:field[version=1] GPU sharing information
+	SharePercentage  float64   `json:"sharePercentage"`  // @bingen:field[version=1]
+	UsageSeconds     float64   `json:"usageSeconds"`     // @bingen:field[version=1] GPU seconds available
```
ameijer: GPU references in comments

core/pkg/model/kubemodel/service.go
ameijer: probably, I think those would then be computed from other network data

core/pkg/model/kubemodel/container.go
aaliomer-ibm: no because if there is the agent goes down for any reason then it wont be accurate. remember that the model needs to be mergable

core/pkg/model/kubemodel/pod.go
```diff
@@ -1,19 +1,22 @@
...
-	NetworkReceiveBytes  uint64            `json:"networkReceiveBytes"`
+	UID                  uuid.UUID         `json:"uid"`                   // @bingen:field[version=1]
+	NamespaceUID         uuid.UUID         `json:"namespaceUid"`          // @bingen:field[version=1]
+	OwnerUID             uuid.UUID         `json:"ownerUid"`              // @bingen:field[version=1] - Reference to Owner (Deployment, StatefulSet, etc.)
+	NodeUID              uuid.UUID         `json:"nodeUid"`               // @bingen:field[version=1]
+	Name                 string            `json:"name"`                  // @bingen:field[version=1]
+	Labels               map[string]string `json:"labels,omitempty"`      // @bingen:field[version=1]
+	Annotations          map[string]string `json:"annotations,omitempty"` // @bingen:field[version=1]
+	DurationSeconds      uint64            `json:"durationSeconds"`       // @bingen:field[version=1]
+	NetworkTransferBytes uint64            `json:"networkTransferBytes"`  // @bingen:field[version=1]
```
ameijer:
```suggestion
	NetworkTransferBytes []NetworkUsage          `json:"networkTransferBytes"`  // @bingen:field[version=1]
```
```
type NetworkUsage struct {
    Service          string
	Bytes            float64                 `json:"bytes"`
	EndPoint         string                  `json:"endPoint"`
	TrafficDirection NetworkTrafficDirection `json:"trafficDirection"`
	TrafficType      NetworkTrafficType      `json:"trafficType"`
}
```

ameijer:
```
type NetworkService struct {
    ServiceName string 
    TotalBytes float64/uint64
}

type NetworkTraffic struct {
  CrossZoneBytes float64/uint64
  CrossRegionBytes float64/uint64
  InternetBytes float64/uint64
  NatGatewayBytes float64/uint64
   Services []*NetworkService
}

On Pod/Container:
{ 
    ...
    NetworkEgress *NetworkTraffic 
    NetworkIngress *NetworkTraffic
    ...
}
```

ameijer:
@aaliomer-ibm can you resolve the merge issues? then I will take a review pass

core/pkg/model/kubemodel/container.go
```diff
@@ -1,23 +1,94 @@
...
+	CpuMillicoreSeconds        uint64               `json:"cpuMillicoreSeconds"`                 // @bingen:field[version=1]
+	CpuMillicoreUsageMax       uint64               `json:"cpuMillicoreUsageMax"`                // @bingen:field[version=1]
+	CpuMillicoreRequestSeconds uint64               `json:"cpuMillicoreRequestSeconds"`          // @bingen:field[version=1]
+	RAMKiBSeconds              uint64               `json:"ramKiBSeconds"`                       // @bingen:field[version=1]
+	RAMByteUsageMax            uint64               `json:"ramByteUsageMax"`                     // @bingen:field[version=1]
+	RAMKiBRequestSeconds       uint64               `json:"ramKiBRequestSeconds"`                // @bingen:field[version=1]
+	VolumeStorageKiBSeconds    map[uuid.UUID]uint64 `json:"volumeStorageKiBSeconds,omitempty"`   // @bingen:field[version=1]
+	VolumeStorageByteUsageMax  map[uuid.UUID]uint64 `json:"volumeStorageByteUsageMax,omitempty"` // @bingen:field[version=1]
+	CpuMillicoreLimitSeconds   uint64               `json:"cpuMillicoreLimitSeconds,omitempty"`  // @bingen:field[version=1]
```
ameijer: change all to be ignore except for the ones that were not marked ignored

core/pkg/model/kubemodel/device.go
```diff
@@ -0,0 +1,75 @@
...
+	IsShared         bool      `json:"isShared"`         // @bingen:field[version=1] Device sharing information
+	SharePercentage  float64   `json:"sharePercentage"`  // @bingen:field[version=1]
+	UsageSeconds     float64   `json:"usageSeconds"`     // @bingen:field[version=1] Device seconds available
+	MemoryKiBSeconds uint64    `json:"memoryKiBSeconds"` // @bingen:field[version=1] Device memory capacity in KiB-seconds
+	PowerWattSeconds float64   `json:"powerWattSeconds"` // @bingen:field[version=1] Device power consumption in watt-seconds (Joules)
+	PowerWattMax     float64   `json:"powerWattMax"`     // @bingen:field[version=1] Device max power consumption in watts
+	// Version 2 fields - Lifecycle tracking
+	Start           time.Time `json:"start,omitempty"` // @bingen:field[version=1] - Device availability start
+	End             time.Time `json:"end,omitempty"`   // @bingen:field[version=1] - Device availability end
+	DurationSeconds uint64    `json:"durationSeconds"` // @bingen:field[version=1] - Duration device was available
```
ameijer: bingen ignore

core/pkg/model/kubemodel/device.go
```diff
@@ -0,0 +1,75 @@
...
+type Device struct {
+	UID              uuid.UUID `json:"uid"`              // @bingen:field[version=1] Device UUID (hardware identifier)
+	Type             string    `json:"type,omitempty"`   // @bingen:field[version=1] Device type (e.g., "device", "tpu")
+	NodeUID          uuid.UUID `json:"nodeUid"`          // @bingen:field[version=1] Node hosting this device
+	DeviceNumber     int32     `json:"deviceNumber"`     // @bingen:field[version=1]
+	ModelName        string    `json:"modelName"`        // @bingen:field[version=1]
+	IsShared         bool      `json:"isShared"`         // @bingen:field[version=1] Device sharing information
+	SharePercentage  float64   `json:"sharePercentage"`  // @bingen:field[version=1]
+	UsageSeconds     float64   `json:"usageSeconds"`     // @bingen:field[version=1] Device seconds available
+	MemoryKiBSeconds uint64    `json:"memoryKiBSeconds"` // @bingen:field[version=1] Device memory capacity in KiB-seconds
```
ameijer: floats, not uints

ameijer (commented):
can we also add the networking fields?

core/pkg/model/kubemodel/cluster.go
```diff
@@ -1,10 +1,14 @@
...
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
 
 // @bingen:generate:Cluster
 type Cluster struct {
-	UID      string    `json:"uid"`      // @bingen:field[version=1]
+	UID      uuid.UUID `json:"uid"`      // @bingen:field[version=1]
```
Sean-Holcomb: While I agree with the `uuid.UUID` type in principle. The ship has kind of sailed on this field being a `string` in version 1 and probably in general given that it currently exists in the wild for a paying costumer. The fact that it is true here and with names space and resourcequota, makes me feel that the whole idea should probably be abondon.

core/pkg/model/kubemodel/container.go
Sean-Holcomb: In general I think we should have a policy here that fields should be marked ignored until they have code which populates them. I am fine with asperational fields to enable functionality we want, but keeping them ignored protects us from bloat if it turns out that we cannot populate those fields. Once a version goes live we are stuck with those fields.

core/pkg/model/kubemodel/namespace.go
```diff
@@ -1,14 +1,17 @@
...
 // @bingen:generate:Namespace
 type Namespace struct {
-	UID         string            `json:"uid"`         // @bingen:field[version=1]
-	ClusterUID  string            `json:"clusterUID"`  // @bingen:field[version=1]
-	Name        string            `json:"name"`        // @bingen:field[version=1]
-	Labels      map[string]string `json:"labels"`      // @bingen:field[version=1]
-	Annotations map[string]string `json:"annotations"` // @bingen:field[version=1]
-	Start       time.Time         `json:"start"`       // @bingen:field[version=1]
-	End         time.Time         `json:"end"`         // @bingen:field[version=1]
+	UID         uuid.UUID         `json:"uid"`             // @bingen:field[version=1]
```
Sean-Holcomb: see other UUID comment

core/pkg/model/kubemodel/resourcequota.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// @bingen:generate:ResourceQuota
+type ResourceQuota struct {
+	UID          uuid.UUID            `json:"uid"`             // @bingen:field[version=1]
```
Sean-Holcomb: see previous uuid comment

core/pkg/model/kubemodel/kubemodel_test.go
```diff
@@ -1,237 +0,0 @@
-package kubemodel
```
Sean-Holcomb, line 1: Why are these tests removed?

core/pkg/model/kubemodel/container.go
Sean-Holcomb: It is still unclear to me why Kubemodel needs to be mergable.

core/pkg/model/kubemodel/window.go
Sean-Holcomb: Window describes a period of time delimited by a start and end in which the runtimes of various entites can occur. The uptime of the agent is inconsequential to the window.

core/pkg/model/kubemodel/container.go
```diff
@@ -1,23 +1,94 @@
...
+	RAMKiBLimitSeconds         uint64               `json:"ramKiBLimitSeconds,omitempty"`        // @bingen:field[version=1]
+	Start                      time.Time            `json:"start"`                               // @bingen:field[version=1]
+	End                        time.Time            `json:"end"`                                 // @bingen:field[version=1]
+}
+
+func (c *Container) CpuMillicoreUsageAverage() uint64 {
```
Sean-Holcomb: I think this function is emblematic of a core misunderstand of Kubemodel. It is populated via existing datasources which are used to aggregate values such as these. We do not have access to a counter like CpuMillicoreSeconds rather the value is created in a sample based approach that is not susceptible to the issue that duration seconds is trying to solve.

nikovacevic:
@aaliomer-ibm -- I'd like to propose we use `float64` everywhere we've previously decided to use `unit64`. Here's a full analysis of the difference between the types, and their fitness for your purposes: https://apptio.atlassian.net/wiki/spaces/Kubecost/pages/934248701/KubeModel+numeric+precision

Let me know what you think.

core/pkg/model/kubemodel/cluster.go
aaliomer-ibm: thats a bummer. sounds good

core/pkg/model/kubemodel/container.go
aaliomer-ibm: 1. In Cldy, we had a lot of business logic that is static window aware (daily, hourly, etc...). That led to a very ugly code
2. Assuming requirements wont change or assuming we'll just do a rewrite is not a good idea
3-A. in Cldy, storage was an issue in multi-tenant environment. Kubecost doesn't support multi-tenant yet and have not faced this issue yet. Cldy has. We need to be able to combine smaller windows into a larger one without losing information. hence my push for units like seconds, byteSeconds, etc. it is something cldy didn't do and kubecost did which is brilliant on your side. 
3-B. Most containers workloads dont change drastically. If you collect data every 10 minutes will be ~5-6 times bigger than once an hour
5.  You have to assume interruptions with the agent. Also you have to assume non-contiguous use cases. For example getting daytime cost consumption 
6. Saying that mergable units (like seconds, byteseconds, bytes, etc) is a downstream concern is just not true. Those are what we call contagious units. It will impact what use-cases you're able to support downstream
7. You need flexibility. we don't know the future nor the use cases. and i don't think there is a cost for it to do it this way


Those are some of the reasons off the top of my head. I reduced those points into two requirements:
1. Dynamic windows
2. Mergable units

core/pkg/model/kubemodel/container.go
```diff
@@ -6,22 +6,89 @@ import (
...
+	}
+	return max
+}
+
+func (c *Container) StorageByteUsageAverage() Measurement {
+	if c.DurationSeconds == 0 {
+		return 0
+	}
+	totalByteSeconds := c.TotalStorageByteSeconds()
+	return KiBToBytes(totalByteSeconds) / c.DurationSeconds
```
Sean-Holcomb: Do we still need this KiB conversion func

core/pkg/model/kubemodel/device.go
```diff
@@ -0,0 +1,92 @@
+package kubemodel
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// @bingen:generate:Device
+type Device struct {
```
Sean-Holcomb, line 10: Are devices relevant on the CLDY side of things?

Sean-Holcomb, line 10: Part of the data here is coming from the DCGM exporter which specifically pertains to Nvidea GPUs I wonder I am starting to think we may want a gpu specific type that has these additional fields on it and references a specific device

core/pkg/model/kubemodel/window.go
```diff
@@ -6,4 +6,8 @@ import "time"
 type Window struct {
 	Start time.Time `json:"start"` // @bingen:field[version=1]
 	End   time.Time `json:"end"`   // @bingen:field[version=1]
+	// Sometimes the duration is smaller than the time between start and end (missing data for example)
+	// If the agent is down for 20 minutes in the middle of an hour span, then start and end will be
+	// an hour span but the duration will be only 40 minutes to account for the missing 20 minute metric window
+	DurationSeconds Measurement `json:"durationSeconds"` // @bingen:field[version=1]
```
Sean-Holcomb: This still doesn't seem to fit to me. It definitely cannot be bingen version 1

Sean-Holcomb (commented):
The main concern I have with any changes here is that they need to be hydrated and we don't currently have the metric querier functions to do so. As long as we leave things on bingen ignore until we have that capability I think it is fine to move forward with this version of the model.

Besides that there are still a lot of references to KiB which should probably go.

core/pkg/model/kubemodel/window.go
aaliomer-ibm: Ah should've been `ignore`
The idea is that is the duration for how long the agent was running between the start and end times. If the agent was down for 10 minutes but start/end is an hour, then it would be 50 minutes. 
I think it needs a different name maybe a different place as well so i'll remove it. 
Not sure where it should be though

core/pkg/model/kubemodel/device.go
aaliomer-ibm: So if the node enables MIG mode (virtual GPU), then we'll need to support the concept of a virtual device. The device concept itself is a first class k8s concept.
Maybe f `virtual` flag with `physicalDeviceUID` field? 
But to answer your original question, i think device is fine

aaliomer-ibm: GPU, TPU thats it for now. Looked into special networking devices that some companies use but we don't support it yet. they are usually CAPEX spend so no customer pushed for it yet

core/pkg/model/kubemodel/container.go
aaliomer-ibm: Nope! should take it off. Good catch!

Sean-Holcomb (approved):

PR #3499, opened 2025-12-12
Fix listing directories in root path(empty string path)

mittal-ishaan:
## Description
Fix listing directories in root path(empty string path).
Right now when we list directories in GCS for root path (`path=""`) it gives an empty result due to the if statement filtering out all root path prefixes(directories) as `attrs.Name` is an empty string for all prfixes(directories) in gcs and we were checking `if attrs.Name==""` when `path=""`

## Testing

Tested locally by running the aggregator with a GCP bucket.
Ran the added tests with a gcp bucket in config path and they pass

### DIFF
--- a/core/pkg/storage/gcsstorage.go
+++ b/core/pkg/storage/gcsstorage.go
@@ -264,11 +264,6 @@ func (gs *GCSStorage) ListDirectories(path string) ([]*StorageInfo, error) {
 			return nil, errors.Wrap(err, "list gcs objects")
 		}
 
-		// ignore the root path directory
-		if attrs.Name == path {
-			continue
-		}
-
 		// We filter directories using DirDelim, so a nameless entry is a dir
 		// See gcs.ObjectAttrs Prefix property
 		if attrs.Name == "" {
--- a/core/pkg/storage/gcsstorage.go
+++ b/core/pkg/storage/gcsstorage.go
@@ -239,7 +239,7 @@ func (gs *GCSStorage) List(path string) ([]*StorageInfo, error) {
 func (gs *GCSStorage) ListDirectories(path string) ([]*StorageInfo, error) {
 	path = trimLeading(path)
 
-	log.Debugf("GCSStorage::List(%s)", path)
+	log.Debugf("GCSStorage::ListDirectories(%s)", path)
 	ctx := context.Background()
 
 	// Ensure the object name actually ends with a dir suffix. Otherwise we'll just iterate the
@@ -261,12 +261,12 @@ func (gs *GCSStorage) ListDirectories(path string) ([]*StorageInfo, error) {
 			break
 		}
 		if err != nil {
-			return nil, errors.Wrap(err, "list gcs objects")
+			return nil, errors.Wrap(err, "list gcs prefixes")
 		}
 
-		// We filter directories using DirDelim, so a nameless entry is a dir
+		// We filter directories using DirDelim, so a non empty prefix entry is a prefix(directory)
 		// See gcs.ObjectAttrs Prefix property
-		if attrs.Name == "" {
+		if attrs.Prefix != "" {
 			stats = append(stats, &StorageInfo{
 				Name:    attrs.Prefix,
 				Size:    attrs.Size,
--- a/core/pkg/storage/s3storage.go
+++ b/core/pkg/storage/s3storage.go
@@ -419,7 +419,7 @@ func (s3 *S3Storage) List(path string) ([]*StorageInfo, error) {
 func (s3 *S3Storage) ListDirectories(path string) ([]*StorageInfo, error) {
 	path = trimLeading(path)
 
-	log.Tracef("S3Storage::List(%s)", path)
+	log.Tracef("S3Storage::ListDirectories(%s)", path)
 	ctx := context.Background()
 
 	if path != "" {
--- a/core/pkg/storage/test.go
+++ b/core/pkg/storage/test.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"fmt"
 	"path"
+	"strings"
 	"testing"
 
 	"github.com/opencost/opencost/core/pkg/util/json"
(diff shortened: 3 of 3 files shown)


core/pkg/storage/gcsstorage.go
```diff
@@ -261,17 +261,12 @@ func (gs *GCSStorage) ListDirectories(path string) ([]*StorageInfo, error) {
...
-		if attrs.Name == path {
-			continue
+			return nil, errors.Wrap(err, "list gcs prefixes")
 		}
 
-		// We filter directories using DirDelim, so a nameless entry is a dir
+		// We filter directories using DirDelim, so a non empty prefix entry is a prefix(directory)
 		// See gcs.ObjectAttrs Prefix property
-		if attrs.Name == "" {
+		if attrs.Prefix != "" {
```
thomasvn, line 269: This seems like the right approach to me according to Google Storage's pkg comments.

```go
    // Name is the name of the object within the bucket.
    // This field is read-only.
    Name string

    // Prefix is set only for ObjectAttrs which represent synthetic "directory
    // entries" when iterating over buckets using Query.Delimiter. See
    // ObjectIterator.Next. When set, no other fields in ObjectAttrs will be
    // populated.
    Prefix string
```

This is a pretty critical function in OpenCost we are modifying though, so we should be careful here. Could you share some details in the PR notes on how you tested this?

mittal-ishaan: Sure adding

kaelanspatel (commented):
As far as I remember, this was added to service deprecated Fed ETL. This looks fine; is there any context as to why we need to list content in the root dir?

mittal-ishaan:
The current federated bucket structure can have multiple application directories in the root path in which the diagnostics, heartbeat and walinator files are present
```
finops-agent/cluster/diagnostics
opencost/cluster/diagnositcs
```
so the diagnostics ingestor has to List directories in the root dir to first get all the applications.
reference: https://github.com/kubecost/kubecost-cost-model/blob/fa77690ba4437bcd31fb77789647334db6196f80/pkg/db/ingest/filechecker/diagnosticschecker.go#L100-L107

core/pkg/storage/gcsstorage.go
```diff
@@ -261,17 +261,12 @@ func (gs *GCSStorage) ListDirectories(path string) ([]*StorageInfo, error) {
...
 		}
 		if err != nil {
-			return nil, errors.Wrap(err, "list gcs objects")
-		}
-
-		// ignore the root path directory
-		if attrs.Name == path {
-			continue
+			return nil, errors.Wrap(err, "list gcs prefixes")
 		}
```
kaelanspatel, line 265: Will removing this not result in counting the path itself as an item? Or is the `prefix` attribute relative to the path?

mbolt35: Yeah, this probably still needs to exist. This line removes the path itself from the returned listing. Not sure why we removed it.

mbolt35: The comment in the description of this change says exactly why it's problematic. When path==""

Why are we changing it for all cases and not the one stated in the description?

mittal-ishaan: we would never be getting back the same path itself, as we add a [delimiter](https://github.com/opencost/opencost/blob/dc0660f80b39ae345d314ea900e29cf15beb4e0c/core/pkg/storage/gcsstorage.go#L245-L249) at the end of the path before listing, and iterator would give back the path upto next `delimiter`.

For non empty path the above case will apply and we never get the same path back. for empty path(root path) too, we would never get an empty path back, as we are using delimiter to find the next path and there is no such path as `"/"`

we want to remove this check as the directories have `attrs.Name` set to `""`(empty string) and when we are listing directories in the root path(`path=""`), we would be missing out all the directories present there.

mittal-ishaan: when `path != ""`, this if check was doing nothing, as `attrs.Name` is an empty string for all the directories so it was redundant check which never becomes truw for non root paths.

kaelanspatel: That seems consistent with the documentation for GCS `Object` API, though I am also seeing some historical issues on the general internet where `path: "foo/bar/"` and `delimiter: "/"` have still returned the directory itself. Have we tested this explicitly?

mittal-ishaan: no, I havent explicitly tested for it, as I original code was itself skipping that case. If this case needs to be taken care of we can replace the current if statement with:
```go
if attrs.Prefix == path {
```
I will test it right away

mittal-ishaan: Though, there is already a test for it. Ref: https://github.com/mittal-ishaan/opencost/blob/dc0660f80b39ae345d314ea900e29cf15beb4e0c/core/pkg/storage/test.go#L177-L183
which passed when I ran it. But I will locally run the aggregator too, to test it out

mittal-ishaan: tested and does not look like we are getting the same path back

kaelanspatel: Ok, searching a little more I can't find any documentation that says it would list the directory itself with the combination of delimiter/path here. If we're confident in the testing I'll approve.

kaelanspatel (approved):

core/pkg/storage/gcsstorage.go
mbolt35: As long as we are sure there are no circumstances where this is true (I'm certain there existed cases previously where this happened). I really hope we're certain here because being wrong and returning the current dir could create some potentially really awful infinite recursion issues. What happens if you omit the "/"? 

I don't think we have any room to be guessing here.

mittal-ishaan: agree, we do not want to get into that where we are are getting the same path back. 
The current `if` check that I removed, was not helping to check and filter out the same path if received. Reason: `attrs.Name` is an empty string for all the directories, and the if statement would be false all the time for a non empty path. 
If we omit appending "/" to the path, in that case we would be getting the same path back, `ListDirectories("foo/bar")` would give back "foo/bar/". (Though this is not a case that we will encounter as we append "/" to the path)

If we want to be very cautious here, We can have: 
```go
if attrs.Prefix == path {
```
to filter the same path directory if received. 
To note again: The original `if` statement checked for directory being a root path directory and did not check if the returned directory is same as the path

mbolt35 (approved):
As long as we feel good about the `.Name` checks, I think this is fine to merge.

PR #3503, opened 2025-12-15
Fix nil pointer dereference in transformCloudCostSetRange

### LINKED ISSUE #3502: fix(mcp): Add nil checks to prevent panic in cloud cost transformation
**Describe the bug**
The `transformCloudCostSetRange` function in the MCP server dereferences `Window.Start()` and `Window.End()` pointers without nil checks, which can panic and crash the entire MCP server when cloud providers return malformed window data.

**To Reproduce**
1. Enable MCP server with cloud cost integration (`ENABLE_MCP_SERVER=true`)
2. Configure a cloud provider (AWS, GCP, Azure, etc.)
3. Query cloud costs via MCP when provider returns incomplete time window data
4. Observe MCP server panic with nil pointer dereference

**Location**
- **File:** `pkg/mcp/server.go`

**Current Code:**

```
Window: &TimeWindow{
    Start: *ccSet.Window.Start(),  // Panic if nil!
    End:   *ccSet.Window.End(),    // Panic if nil!
} 
```

**Expected behavior**
Check for nil windows/timestamps, log warning, and skip malformed data instead of crashing. Proposed Fix:

```
if ccSet.Window == nil || ccSet.Window.Start() == nil || ccSet.Window.End() == nil {
    log.Warnf("Skipping cloud cost set with invalid window")
    continue
}
```

Current develop branch

peatey:
…Range

The transformCloudCostSetRange function dereferences Window.Start() and Window.End() pointers without nil checks, which can panic and crash the MCP server when cloud providers return malformed window data.

This fix adds nil checks before dereferencing:
- CloudCostSet.Window.Start() and End() - skips sets with invalid windows
- CloudCost item Window.Start() and End() - skips items with invalid windows
- CloudCostSet in Kubernetes percent calculation loop - prevents nil access

When invalid data is encountered, a warning is logged and the malformed data is skipped instead of crashing the server.

Fixes #3502

### DIFF
--- a/pkg/mcp/server.go
+++ b/pkg/mcp/server.go
@@ -871,6 +871,12 @@ func transformCloudCostSetRange(ccsr *opencost.CloudCostSetRange) *CloudCostResp
 			continue
 		}
 
+		// Skip cloud cost sets with invalid windows to prevent nil pointer dereference
+		if ccSet.Window.Start() == nil || ccSet.Window.End() == nil {
+			log.Warnf("Skipping cloud cost set %d with invalid window", i)
+			continue
+		}
+
 		setName := fmt.Sprintf("cloudcosts_%d", i)
 		mcpSet := &CloudCostSet{
 			Name:                  setName,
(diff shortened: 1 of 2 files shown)


vtushar06:
@ameijer could you please advise on how you’d like to proceed here? 
as the linked PR addresses part of the issue, but some tests are currently failing and a few edge cases may still be uncovered.
I had been preparing a fix after opening the issue and am happy to help refine or extend the current PR if that’s useful or should I proceed with my fix.

peatey:
I haven't addressed the failures yet - if you want to go ahead and build a fix please feel free to do so don't let this work be sunk cost - it's claude cycles + some minor revisions from me.

I also think these are our standard failures we're working on fixing atm - @vtushar06 could I confirm what functionality, not testing issues, this isn't covering off?

vtushar06:
hey @ameijer,I’m planning a small follow-up PR to handle remaining edge cases (test failures + partial window validation). Will link it here once ready.

ameijer:
closing as dup of https://github.com/opencost/opencost/pull/3511

PR #3504, opened 2025-12-17
fix image building and testing for non maintainers in MQ

ameijer:
## Description
fixes issue where tests not getting run in MQ for non maintainers


mbolt35 (approved):
LGTM

PR #3531, opened 2026-01-09
KCM-4049: Fixes Ondemand cost In Azure Cluster

avrodrigues5:
## Description

Fixes Azure ondemand cost for linux based computes.

## Related Issues

https://apptio.atlassian.net/browse/KCM-4079

## User Impact

* KCM drilldown will have accurate Ondemand cost.
* Opencost is enhanced with accurate ondemand cost for linux based computes.

## Testing

Azure ondemand cost were incorrect, because of missing filters.

Lets take Opencost with develop branch for example
The ondemand cost for **Standard_D8ds_v5** was shown as 0.712. This is due to the filters not having the condition to only go fetch compute related cost from retail price URL. This is reflected in node_total_hourly_cost in Opencost local run

Screenshot 2026-01-09 at 2 42 13 PM

This is the URL that is called in OC for node type  **Standard_D8ds_v5** https://prices.azure.com/api/retail/prices?$skip=0&currencyCode=%27USD%27&$filter=armRegionName+eq+%27eastus2%27+and+armSkuName+eq+%27Standard_D8ds_v5%27

The last satisfying json with existing condition is

```
{
  "currencyCode": "USD",
  "tierMinimumUnits": 0,
  "retailPrice": 0.712,
  "unitPrice": 0.712,
  "armRegionName": "eastus2",
  "location": "US East 2",
  "effectiveStartDate": "2023-03-01T00:00:00Z",
  "meterId": "4a00cbca-ecfa-5b8b-8407-7a68219bd58d",
  "meterName": "vCore",
  "productId": "DZH318Z0KDZJ",
  "skuId": "DZH318Z0KDZJ/00BC",
  "productName": "Azure Database for PostgreSQL Flexible Server General Purpose - Ddsv5 Series Compute",
  "skuName": "8 vCore",
  "serviceName": "Azure Database for PostgreSQL",
  "serviceId": "DZH3199QPQTD",
  "serviceFamily": "Databases",
  "unitOfMeasure": "1 Hour",
  "type": "Consumption",
  "isPrimaryMeterRegion": true,
  "armSkuName": "Standard_D8ds_v5"
},
```

Which clearly shows that the serviceFamily is Databases. 

The code is further enhanced to exclude Low priority instances. Basically Opencost logic for Azure Ondemand price match Our logic in [Instance builder script](https://github.com/kubecost/provider-instances-builder/blob/main/pkg/instances/azure.go#L311-L320).

The data after fix clearly shows accurate cost which align with azure cost on their website.

Screenshot 2026-01-09 at 2 42 34 PM

### DIFF
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -273,9 +273,14 @@ func buildAzureRetailPricesURL(region string, skuName string, currencyCode strin
 	return pricingURL
 }
 
-func extractAzureVMRetailAndSpotPrices(resp []byte) (retailPrice string, spotPrice string, err error) {
+func extractAzureVMRetailAndSpotPrices(resp *http.Response) (retailPrice string, spotPrice string, err error) {
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", "", fmt.Errorf("Error getting response: %v", err)
+	}
+
 	pricingPayload := AzureRetailPricing{}
-	jsonErr := json.Unmarshal(resp, &pricingPayload)
+	jsonErr := json.Unmarshal(body, &pricingPayload)
 	if jsonErr != nil {
 		return "", "", fmt.Errorf("error unmarshalling data: %v", jsonErr)
 	}
@@ -306,12 +311,7 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 		return "", fmt.Errorf("retail price responded with error status code %d", resp.StatusCode)
 	}
 
-	body, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return "", fmt.Errorf("Error getting response: %v", err)
-	}
-
-	retailPrice, spotPrice, err := extractAzureVMRetailAndSpotPrices(body)
+	retailPrice, spotPrice, err := extractAzureVMRetailAndSpotPrices(resp)
 	if err != nil {
 		return "", fmt.Errorf("failed to extract azure prices: %v", err)
 	}
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -239,7 +239,7 @@ func getRegions(service string, subscriptionsClient subscriptions.Client, provid
 	}
 }
 
-func getRetailPrice(region string, skuName string, currencyCode string, spot bool) (string, error) {
+func buildAzureRetailPricesURL(region string, skuName string, currencyCode string) string {
 	pricingURL := "https://prices.azure.com/api/retail/prices?$skip=0"
 
 	if currencyCode != "" {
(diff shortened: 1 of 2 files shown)


pkg/cloud/azure/provider.go
```diff
@@ -258,16 +258,23 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 		filterParams = append(filterParams, skuNameParam)
 	}
 
-	if len(filterParams) > 0 {
-		filterParamsEscaped := url.QueryEscape(strings.Join(filterParams[:], " and "))
-		pricingURL += fmt.Sprintf("&$filter=%s", filterParamsEscaped)
-	}
+	// Make sure only service name with Virtual Machines are parsed with skuName
```
avrodrigues5, line 261: I'm happy to write a unit test case but two challenges
1. Calls azure retails API everytime test is run.
2. The ondemand prices are subjected to change. So we might create flaky test if we assert ondemand cost. 

Any opinions here @ameijer

pkg/cloud/azure/provider.go
```diff
@@ -289,7 +296,9 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 	retailPrice := ""
 	spotPrice := ""
 	for _, item := range pricingPayload.Items {
-		if item.Type == "Consumption" && !strings.Contains(item.ProductName, "Windows") {
+		// note: Windows OS ondemand price will be equal to Linux, Adoption of Windows based
```
avrodrigues5, line 288: Something to add to roadmap.. We only test Linux based Computes, windows based are on rise. I have some ideas on how to do it.

nik-kc (commented):
Very nice! How'd you determine the filters that you need to make these changes?

avrodrigues5:
> Very nice! How'd you determine the filters that you need to make these changes?

Honestly, this was based on a combination of the Azure Retail Prices API documentation and practical validation using the API.
The key filter is serviceFamily eq 'Compute', which is explicitly documented in the sample queries here:
https://learn.microsoft.com/en-us/rest/api/cost-management/retail-prices/azure-retail-prices#api-sample-calls

Azure publishes on-demand, spot, and low-priority prices for most VM SKUs under the same skuName, so filtering correctly is important. In this flow we include Spot pricing but exclude Low Priority, which is why we do not filter out Spot here (though another [script](https://github.com/kubecost/provider-instances-builder/blob/main/pkg/instances/azure.go#L311-L320) does exclude it).

You’ll also see in the develop branch that we filter on type = 'Consumption' using string contains. The retails API already has that as input,  ref API documentation and significantly reduces the result set, making queries faster and easier to process.

I’m happy to walk through any of the filters in more detail if helpful. I spent a fair amount of time validating this during the node-group right-sizing work over the summer.

pkg/cloud/azure/provider.go
ameijer: what others have done is they call the API, and stick the JSON into test assets. then the "API call" just reads in the response JSON. what do you think about that approach?

pkg/cloud/azure/provider.go
ameijer: can you file an issue on OC alan?

ameijer (commented):
overall looks good, see what you think about the unit testing idea?

pkg/cloud/azure/provider.go
avrodrigues5: Filed the issue and proposed solution https://github.com/opencost/opencost/issues/3532.

pkg/cloud/azure/provider.go
avrodrigues5: Split the big function into 3 functions with 2 of them being testable and was able to beat SQ code coverage.

pkg/cloud/azure/provider.go
```diff
@@ -258,45 +258,63 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
...
-	log.Infof("starting download retail price payload from \"%s\"", pricingURL)
-	resp, err := http.Get(pricingURL)
+	// Add type eq 'Consumption' to the filter to avoid reservation cost
+	filterParams = append(filterParams, "type eq 'Consumption'")
 
-	if err != nil {
-		return "", fmt.Errorf("bogus fetch of \"%s\": %v", pricingURL, err)
-	}
+	// Exclude Low Priority instances[Azure has special computes that let you run workloads on spare capacity at a deeply discounted price in exchange for no SLA and the possibility of being evicted]
+	filterParams = append(filterParams, "contains(meterName,'Low Priority') eq false")
```
ameijer, line 268: is this us basically removing support for spot nodes?

avrodrigues5: So i have never used Low Priority VMs but there is a difference between Low priority VMs and spot VMs explained in https://jaychapel.medium.com/azure-low-priority-vms-and-spot-vms-for-cost-savings-b08f5ac691bc.

This condition added to the URL does not discard spot prices and only low priority VMs which in fact are discontinued by Azure but retail prices still show in the URl. 

When the Low Priority VM price appears after the regular on-demand price in the Retail Prices API response, the low-priority price can overwrite the real on-demand price if this condition does not exist.

For example, look at this URL: https://prices.azure.com/api/retail/prices?$skip=0&currencyCode=%27USD%27&$filter=armRegionName+eq+%27eastus2%27+and+armSkuName+eq+%27Standard_NC64as_T4_v3%27

The Low Priority entry appears before the actual on-demand price of $4.35/hour. If our logic does not filter it out, the lower Low Priority price would overwrite the correct on-demand price if it had appeared below the $4.35/hour response.

Let me know if you want further explanation, happy to hop on a call!

ameijer (approved):
LGTM, thanks Alan!

Issue #3549, opened 2026-01-21
[Bug]: Node pricing fallback lacks diagnostic logging when node labels indicate a match should exist


peatey:
## Goal

Enhance debug logging such that when a node from the provider contains a label that indicates a match should be available, we should log what caused the match to fail.

## Describe the bug

When running OpenCost/Kubecost on AWS EKS clusters, the logs are flooded with two types of messages that create significant noise and may indicate underlying issues with node pricing resolution and CPU metrics collection:

1. **INFO level** - `Could not get node pricing for node [node-ip].us-east-2.compute.internal. Falling back to default pricing`
2. **WARNING level** - `ClusterNodes: unable to read CPU mode data`

These messages appear continuously (every few seconds) for multiple nodes in the cluster, making it difficult to identify genuine issues in the logs and potentially indicating that cost calculations are relying on fallback/default pricing rather than accurate AWS pricing data.

## Environment

- **OpenCost/Kubecost Version**: [INSERT VERSION - check with `kubectl logs` or `/version` endpoint]
- **Cloud Provider**: AWS
- **Region**: us-east-2
- **Kubernetes Distribution**: EKS (assumed based on `.us-east-2.compute.internal` node naming)
- **Node Count**: 15+ nodes affected
- **Prometheus**: [INSERT VERSION]

## Expected behavior

1. Node pricing should be successfully retrieved from AWS pricing APIs without falling back to defaults for standard EC2 instance types
2. CPU mode data should be readable from Prometheus/node-exporter metrics without warnings
3. Log verbosity should be reasonable and not flood with repeated messages for known/handled conditions

### Log Sample (node pricing)
```
2026-01-21T14:56:54.401017347Z INF Could not get node pricing for node ip-x.us-east-2.compute.internal. Falling back to default pricing
2026-01-21T14:56:54.401055597Z INF Could not get node pricing for node ip-x.us-east-2.compute.internal. Falling back to default pricing
2026-01-21T14:56:54.401084282Z INF Could not get node pricing for node ip-x.us-east-2.compute.internal. Falling back to default pricing
[... repeated for 15+ nodes ...]
```

### Log Sample (CPU mode data)
```
2026-01-21T14:58:55.782847821Z WRN ClusterNodes: unable to read CPU mode data.
2026-01-21T14:58:55.782857191Z WRN ClusterNodes: unable to read CPU mode data.
2026-01-21T14:58:55.782863982Z WRN ClusterNodes: unable to read CPU mode data.
[...]

PR #3552, opened 2026-01-21
Add Prometheus metrics for node pricing lookups

peatey:
Add opencost_node_pricing_lookup_total counter metric to track the success/failure rate of node pricing lookups. This helps operators identify when pricing data is falling back to defaults, which could indicate configuration issues or cloud provider API problems.

The metric includes a "status" label with values:
- "success": Node pricing was retrieved successfully from the provider
- "fallback": Node pricing lookup failed, using fallback/default pricing

Addresses #3549 (Part 4 - Prometheus metrics for node pricing lookups)


jessegoodier:
This is so helpful- thank you!

I can't comment on the code. 

But we really do need this to inspire confidence in the accuracy of the costs.

PR #3566, opened 2026-01-26
Downgrade k8s/api to revert a breakng change due to certain protobuf package depriciation

mittal-ishaan:


ameijer:
can we just upgrade the downstream?

ameijer (changes requested):
holding until we can learn about all options available

mittal-ishaan:
There will be some work need to be done on the cldy side. 
k8s completely depricated it: https://github.com/kubernetes/kubernetes/pull/134256
Not sure if simple upgrade would work or not there. so downgrading it as a temporary solution so that we are atleast able to compile ibm finops agent.

we can try to find other ways or a permanent solution, but upgrading them in ibm finops agent was not working. cldy team is checking, if they can fix it. that is the root cause

PR #3570, opened 2026-01-29
feat: Apply configured discounts to Prometheus node cost metrics 

badal773:
## Description
Apply configured discounts to Prometheus node cost metrics. Previously discounts only affected allocation calculations, not the exported metrics (node_total_hourly_cost, node_cpu_hourly_cost, etc.). Also adds separate spotDiscount and spotNegotiatedDiscount config options for different discounts on spot vs ondemand nodes.

## User Impact

- Non-breaking: Existing behavior unchanged if no discount configured
- New config options: spotDiscount, spotNegotiatedDiscount in CustomPricing
- Metrics now reflect configured discounts, matching allocation API behavior

## Testing
Verified discount multiplier applied correctly via debug logs

```
2026-01-29T07:21:43.821860748Z DBG GetNodeCost: node=ip-100-xx-xx-xx.ec2.internal type=ondemand applying discount multiplier 0.960000
 2026-01-29T07:21:43.82194443Z DBG GetNodeCost: node=ip-100-xx-xx-xx.ec2.internal type=spot applying discount multiplier 0.980000

```
 
Tested spot vs ondemand node detection

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM golang:1.25.3-alpine3.21 AS build-env
+FROM --platform=$BUILDPLATFORM golang:1.25.5-alpine3.21 AS build-env
 
 WORKDIR /app
 
--- a/pkg/cloud/aws/athenaconfiguration.go
+++ b/pkg/cloud/aws/athenaconfiguration.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 
+	"github.com/opencost/opencost/core/pkg/log"
 	"github.com/opencost/opencost/core/pkg/opencost"
 	"github.com/opencost/opencost/core/pkg/util/json"
 	"github.com/opencost/opencost/pkg/cloud"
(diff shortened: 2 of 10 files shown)


badal773:
@ameijer @mbolt35 

Hi maintainers!!

I noticed that the configured discount and negotiatedDiscount values were only being applied to allocation calculations, but not to the Prometheus metrics (node_cpu_hourly_cost, node_ram_hourly_cost, etc.). This PR adds discount application to the metrics as well.

I also added support for separate spotDiscount and spotNegotiatedDiscount config options, since spot instances often have different discount arrangements than on-demand.

A few questions for the maintainers:

- Is this the intended behavior, or was the discount intentionally excluded from metrics for a reason I'm not aware of?
- I couldn't find existing documentation on whether discounts should apply to metrics - please let me know if this approach is incorrect.
- Should I add any additional tests or documentation?


Happy to make changes based on your feedback. Thanks for your time!

mbolt35:
@badal773 Pretty sure allocations uses the cost metrics directly( this would apply a double discount. This is a major change that is going to need a lot of iteration and review. 

@nikovacevic do you remember why we apply discounts later?

badal773:
Hi! Do we have any updates on this?
It's only applying once for me.

PR #3573, opened 2026-02-01
Update swagger.json

eliran-ops:
Key Improvements:
Version Update: Bumped version to 1.0.0 to reflect modern OpenCost compatibility.

Dynamic Keys: In the old spec, the keys "opencost" and "kube-system" were hard-coded. I've replaced these with additionalProperties, allowing the spec to work whether you aggregate by namespace, pod, or label.

Refined Components: Separated the response wrappers (AllocationResponse, AssetsResponse) from the data models (Allocation, Asset) to make the YAML more readable and reusable.

Schema Accuracy: Added the minutes and externalCost fields which are standard in current OpenCost output but were missing or grouped differently in older versions.

### DIFF
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -7,173 +7,130 @@
       "name": "Apache 2.0",
       "url": "https://www.apache.org/licenses/LICENSE-2.0.html"
     },
-    "version": "0.1"
+    "version": "1.0.0"
   },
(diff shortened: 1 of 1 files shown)


ameijer (approved):
LGTM, thanks for the contribution!

PR #3578, opened 2026-02-04
Move failed RI lookup log to warn

kaelanspatel:
## Description
Small log cleanup to move RI lookup failure to `warn` instead of `error`.

## Related Issues
N/A

## User Impact
More accurate depiction of RI lookup failure as not an explicit error in some cases (no RIs, no Athena bucket).


ameijer (approved):
LGTM

PR #3599, opened 2026-02-17
fix: prevent negative costs from rolling window queries with data gaps

### LINKED ISSUE #3574: Negative totalCost values returned for namespace with cronjobs
## Describe the bug
OpenCost is returning negative totalCost values for some namespaces via the /allocation API. This results in negative percentages when calculating cost allocation across namespaces.
For example, the namespace `service-argo-metrics` shows `-0.37%` OpenCost percentage
our `service-argo-metrics` namespace contains cronjobs only 

## To Reproduce
I'm unable to reliably reproduce this issue, but it appears in our production environment. The negative costs are visible in the UI (and through the API).

Image

## Expected behavior
All namespace costs should be zero or positive.

## Additional context
- OpenCost version: `1.119.1 (0a1cceb)`
- Cloud provider:  Azure AKS
- Kubernetes version: 
  ```
  Client Version: v1.35.0
  Kustomize Version: v5.7.1
  Server Version: v1.34.1
  ```
- Helm Chart version: 2.5.5

Helm chart values:
``` yaml
serviceAccount:
  # --  Annotations to add to the service account
  annotations:
    eks.amazonaws.com/role-arn:  <??>

opencost:
  
  exporter:
    defaultClusterId: <??>
    
    # Collector DataSource configuration (promless mode)
    # Allows OpenCost to run without Prometheus dependency
    collectorDataSource:
      enabled: true
      # How often to collect data (10s, 15s, 30s, 1m, etc.)
      scrapeInterval: 30s
      # Port for network pod egress
      networkPort: 3001
    
    # Enable persistence to preserve state between pod restarts
    # This is strongly recommended when using Collector DataSource
    persistence:
      enabled: true
      size: 10Gi
      accessMode: ReadWriteOnce
      storageClass: disk-csi-azure-standardssd-zrs
      mountPath: /var/configs
    
  prometheus:
    internal:
      # Disable internal Prometheus when using promless mode
      enabled: false
      namespaceName: opencost
      serviceName: opencost-prometheus-server
```

Logs from OpenCost:
```
2026-01-30T11:15:11.096949268Z WRN failed to retrieve retail pricing: Couldn't find price for product "Standard_d2_v5" in "westeurope" region
2026-01-30T11:15:11.097051857Z WRN No pricing data found for node &{map[agentpool:default beta.kubernetes.io/arch:amd64 beta.kubernetes.io/instance-type:Standard_d2_v5 beta.kubernetes.io/os:linux failure-domain.beta.kubernetes.io/region:westeurope failure-domain.beta.kubernetes.io/zone:westeurope-2 kubernetes.azure.com/agentpool:default kubernetes.azure.com/cluster:<REDACTED> kubernetes.azure.com/consolidated-additional-properties:<REDACTED> kubernetes.azure.com/kubelet-identity-client-id:<REDACTED> kubernetes.azure.com/kubelet-serving-ca:cluster kubernetes.azure.com/localdns-state:disabled kubernetes.azure.com/mode:system kubernetes.azure.com/network-policy:none kubernetes.azure.com/node-image-version:AKSAzureLinux-V3gen2-202601.07.0 kubernetes.azure.com/nodepool-type:VirtualMachineScaleSets kubernetes.azure.com/os-sku:AzureLinux kubernetes.azure.com/os-sku-effective:AzureLinux3 kubernetes.azure.com/os-sku-requested:AzureLinux kubernetes.azure.com/role:agent kubernetes.azure.com/sku-cpu:2 kubernetes.azure.com/sku-memory:8192 kubernetes.azure.com/storageprofile:managed kubernetes.azure.com/storagetier:Standard_LRS kubernetes.io/arch:amd64 kubernetes.io/hostname:<REDACTED> kubernetes.io/os:linux node.kubernetes.io/instance-type:Standard_d2_v5 providerID:azure:///subscriptions/<REDACTED>/resourceGroups/<REDACTED>/providers/Microsoft.Compute/virtualMachineScaleSets/<REDACTED>/virtualMachines/0 storageprofile:managed storagetier:Standard_LRS topology.disk.csi.azure.com/zone:westeurope-2 topology.kubernetes.io/region:westeurope topology.kubernetes.io/zone:westeurope-2] accelerator nvidia} from key westeurope,Standard_d2_v5,ondemand
```

I found related issues that mention negative costs being caused by pricing asymmetry between allocations and assets:
- #3359 - "Negative idle cost despite correct scraping"
- #3387 - "ON PREM Negative idle cost on promless mode"
- #3393 - Fix merged for promless mode

dag-andersen:
## What does this PR change?

**Problem:**
When querying rolling windows (e.g. `24h`), `calculateStartAndEnd()` can produce `start > end` after clamping, resulting in negative durations that propagate through `Minutes()` → `CPUCoreHours` → `CPUCost`, producing negative costs in API responses.

**Root Cause:**
In promless/collector mode, the store uses daily buckets. When a rolling window like `24h` clips into a daily bucket that contains data from *before* the window start (e.g. from before a pod restart gap), the clamping logic:
1. Clamps `start` forward to `window.Start()` (e.g. 16:00)
2. Leaves `end` at its original timestamp (e.g. 12:30)
3. Result: `start=16:00, end=12:30` → negative duration → negative costs

Calendar-aligned windows (e.g. `1d`, `today`) don't trigger this because their boundaries align with bucket boundaries.

**Fix:**
Added a guard after all clamping in `calculateStartAndEnd()`: when `start > end`, return a zero-duration interval at the window start (`*window.Start(), *window.Start()`). This keeps the returned timestamps within the query window while ensuring callers naturally compute zero minutes and zero cost.

**Files changed:**
- `pkg/costmodel/allocation_helpers.go` — Added `s.After(e)` guard returning `*window.Start(), *window.Start()`
- `pkg/costmodel/allocation_helpers_test.go` — Added 2 regression test cases

## How was this PR tested?
- Added unit tests covering both multi-point and single-point data before window start
- All existing `pkg/costmodel` tests continue to pass
- `go test ./pkg/costmodel/... -v` passes cleanly

## Does this PR address any GitHub issues?
Fixes #3574

- Rolling window queries (`24h`, `6h`, `12h`) will no longer produce negative costs when there are data gaps (e.g. after pod restarts)
- Instead of reporting negative costs for pods with stale data, those pods will correctly report zero cost for that time window
- No behavior change for calendar-aligned windows (`1d`, `today`, `yesterday`)

No


pkg/costmodel/allocation_helpers.go
```diff
@@ -2532,6 +2532,15 @@ func calculateStartAndEnd(result []*util.Vector, resolution time.Duration, windo
 		e = now
 	}
 
+	// After all clamping, if start is after end, the data does not
+	// overlap with the query window. Return zero times so callers can
+	// detect this and skip the result. This can happen when a rolling
+	// window (e.g. "24h") clips into a daily store bucket that contains
+	// data from before the window start but none after it.
+	if s.After(e) {
+		return time.Time{}, time.Time{}
```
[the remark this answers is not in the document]
dag-andersen: Good point bot 🎉

pkg/costmodel/allocation_helpers.go
```diff
@@ -2532,6 +2532,17 @@ func calculateStartAndEnd(result []*util.Vector, resolution time.Duration, windo
...
+	// After all clamping, if start is after end, the data does not
+	// overlap with the query window. Return a zero-duration interval
+	// at the clamped end so that callers naturally compute zero
+	// minutes and zero cost, without leaking zero-value timestamps
+	// into API responses. This can happen when a rolling window
+	// (e.g. "24h") clips into a daily store bucket that contains
+	// data from before the window start but none after it.
+	if s.After(e) {
+		return e, e
+	}
```
ameijer, line 2544: @nikovacevic bolt asked for you to look at this line, we were fixing related things

nikovacevic:
@dag-andersen thank you for this contribution. In general I think this PR is a good idea, but I have a few questions.

First, let me summarize the context, the problem, and the root cause. Let me know if you would dispute any of this:

1. `calculateStartAndEnd` receives a vector and the window used to query that vector, then returns the start and end times for that vector, within the window.

2. The function serves two main purposes:
  a. to clamp the vector to the window (in case either the start or end of the vector exceeds the window); and
  b. to provide a minimal non-zero duration (make end > start) in the case that only a single datum is provided in the vector.

3. If all data given in the vector fall within the window, there is no problem, i.e., (2.a) is a non-issue, and (2.b) might take effect, but there is no concern with its implementation.

4. However, if the vector's data and the window are _very badly_ misaligned, then _both_ start and end might fall before the window. In that case, start gets clamped, but end doesn't. (In principle, it seems like the same might be said of a clamped end time paired with start time that falls after the window. But perhaps this does not occur in practice.) This problem results in negative durations, which generate negative resource quantities and negative costs.

5. Your proposed fix detects the problem described above (4) and returns `end, end` to create a duration of zero (and, thus, resource quantities and costs of zero).

If the above description is accurate, then I have a few questions:

1. Returning `end, end` would zero out negative values, but would it not also of necessity return time values outside the window? If so, it seems like we'd be replacing a big problem with a smaller problem -- but not fixing the real root cause of the problem at hand. **Should we instead fix the true root cause of (4), above, which is a radical misalignment between query windows and returned vector data?**

2. You claim that querying `24h` (as opposed to `today` or `1d`) will access daily buckets, even though the window will be aligned by hours, not days. For example, a `24h` query submitted on 2026-01-02T16:00:00Z will produce a window of `start: 2026-01-01T16:00:00Z, end: 2026-01-02T16:00:00Z` -- but it will draw from daily buckets that have boundaries `2026-01-01T00:00:00Z`, `2026-01-02T00:00:00Z`, and `2026-01-03T00:00:00Z`. **Did you look into why that is occurring?** The Opencost team can (and should) investigate why this is happening, because it seems to me that once vector data is drawn from a resolution `1d` source for a resolution `1h` query, there is nothing we can do to adequately fix the problem downstream (e.g. here, in this function).

All of that said, I'm not opposed to accepting this PR and merging it as a stopgap measure to prevent negative costs -- at least until the real root cause is fixed.

What are your thoughts? Does this all make sense, or am I misunderstanding part of the problem/proposal?

pkg/costmodel/allocation_helpers.go
nikovacevic: As mentioned in my [comment on the PR](https://github.com/opencost/opencost/pull/3599#issuecomment-3998757856), this change will suffice to set negative costs to zero -- but those times would necessarily _also_ fall outside of the window. (If start is after end, that means that end is before window start -- which means this `return e, e` will return a time outside the window.

It might be nice to simply `return window.Start(), window.Start()` here instead.

What do you think of that as an alternative stopgap measure, until the real root cause is addressed, @dag-andersen ?

nikovacevic:
Just following up, here, to confirm that there is a large problem.

In `modules/collector-source/pkg/collector/collectorprovider.go` there is a function `getStoreKeys` ([link](https://github.com/opencost/opencost/blob/develop/modules/collector-source/pkg/collector/collectorprovider.go#L51-L74)):

```go
// getStoreKeys compares the given start and end against each resolution by truncating the start time and
// add one interval to the truncated value. The duration between start and end is compared with the duration
// between the interval generated times, with the lowest
func (r *repoStoreProvider) getStoreKeys(start, end time.Time) (string, time.Time) {
	windowDuration := int64(end.Sub(start))
	var minDiff *int64
	var minKey string
	var minStart time.Time
	for key, interval := range r.intervals {
		intStart := interval.Truncate(start)
		intEnd := interval.Add(start, 1)
		intDuration := int64(intEnd.Sub(intStart))
		diffDuration := windowDuration - intDuration
		if diffDuration < 0 {
			diffDuration = -diffDuration
		}
		if minDiff == nil || diffDuration < *minDiff {
			minDiff = &diffDuration
			minKey = key
			minStart = intStart
		}
	}
	return minKey, minStart
}
```

That function has a unit test, `func Test_repoStoreProvider_getStoreKeys(t *testing.T)`, in `modules/collector-source/pkg/collector/collectorprovider_test.go` ([link](https://github.com/opencost/opencost/blob/develop/modules/collector-source/pkg/collector/collectorprovider_test.go#L11)).

If you add a unit test case for `"24h"` it fails, picking the `1d` interval instead of the `1h` interval.

```go
// Parse the window string
win24h, _ := opencost.ParseWindowUTC("24h")

// Add a "24h" test case to the tests map
tests := map[string]struct {
		configs    []util.ResolutionConfiguration
		start      time.Time
		end        time.Time
		intevalKey string
		startKey   time.Time
	}{
		"24h": {
			configs:    defaultResConfigs,
			start:      *win24h.Start(),
			end:        *win24h.End(),
			intevalKey: "1h",
			startKey:   *win24h.Start(),
		},
	}
}
```
```
--- FAIL: Test_repoStoreProvider_getStoreKeys (0.00s)
    --- FAIL: Test_repoStoreProvider_getStoreKeys/24h (0.00s)
        /Users/niko/code/kubecost/model/opencost/modules/collector-source/pkg/collector/collectorprovider_test.go:82: getStoreKeys() got = 1d, want 1h
        /Users/niko/code/kubecost/model/opencost/modules/collector-source/pkg/collector/collectorprovider_test.go:85: getStoreKeys() got = 2026-03-03 00:00:00 +0000 UTC, want 2026-03-03 19:00:00 +0000 UTC
```

If we fix this underlying bug, my guess is that it will obviate the need for the fix in this PR.

Again -- I'm not opposed to merging this PR, too, just to be extra safe. But I don't think it will suffice for a real fix to the real root cause.

nikovacevic:
Opened an issue to track the underlying root cause here: https://github.com/opencost/opencost/issues/3630

dag-andersen:
Thanks for the suggestion, @nikovacevic — that makes a lot of sense. Returning `*window.Start(), *window.Start()` keeps the timestamps within the window, which is cleaner than leaking out-of-window values.

I've updated the code and tests accordingly in the latest push. All existing tests plus the two regression tests pass.

Happy to adjust further if needed. Hopefully this can serve as a reasonable stopgap until the root cause in `getStoreKeys()` (#3630) is addressed.

pkg/costmodel/allocation_helpers.go
dag-andersen: Applied — updated to return `*window.Start(), *window.Start()` in the latest commit. Thanks!

PR #3603, opened 2026-02-22
[Fix]:err != err typo and stream handling issues in S3 Select querier

### LINKED ISSUE #3600: err != err typo and stream handling issues in S3 Select querier
**Describe the bug**
While going through pkg/cloud/aws/s3selectquerier.go, I noticed a few things that don't look right:
**1) Typo on line 69** —` err != err `**instead of** `err != nil`

In _GetQueryKeys()_, the error check after _getMonthStrings()_ reads:

```
monthStrings, err := getMonthStrings(start, end)
if err != err {
    return nil, err
}
```
This compares err to itself so the check never triggers. Should be err != nil. If getMonthStrings fails, the error gets ignored and you end up with "no CUR files for given time range" instead of the real error.

Should be `if err != nil` {.

**2. In fetchCSVReader (lines 112-127), the stream handling has a couple of issues.** There's already a /_/ todo: this needs work_ comment there:

- **resStream.Events()** is called on line 117 and the return value is thrown away. Then it's called again on line 118 in the **for range**. The first call might consume events that the second one never sees.

- **break** on line 123 (End event case) only breaks out of the **switch**, not the **for** loop. **return** would actually stop the goroutine.

**To Reproduce**

For the first bug — call GetQueryKeys with start after end. getMonthStrings returns an error but it gets swallowed. You'll see "no CUR files for given time range" instead of the time range error.
For the stream bugs — visible in the code. The double Events() call and the break vs return can be verified by reading the function.

**Expected behavior**

- Error check should use **err != nil**
- **resStream.Events()** should be called once
- End event should use **return** to exit the goroutine

**Screenshots**
N/A

v1.119.2

vtushar06:
## Description

Fix three bugs in S3SelectQuerier: a self-comparison typo ([err != err](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)), a redundant [resStream.Events()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) call that discarded stream events, and a break that only exited the switch instead of the goroutine on End event.

## Related Issues

Fixes #3600

## User Impact

No breaking changes. Error messages from [getMonthStrings](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) are now correctly surfaced instead of being silently swallowed. Stream event handling is now correct, preventing potential missed records or goroutine leaks.

## Testing

Verified by code inspection:

- err != nil now correctly propagates errors from getMonthStrings
- Single [resStream.Events()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) call ensures no events are consumed before the loop
- return in the End event case correctly terminates the goroutine


vtushar06:
Hii @ameijer, need your guidance over this PR, whenever you have time.

ameijer (approved):
lgtm, thanks Tushar!

PR #3612, opened 2026-02-26
KCM-5276: Update asset pipeline to add account id

Hudong-hub:
## Description

add account id for assetpipeline and autocomplete

## Related Issues

KCM-5276


ameijer (commented):
would you be willing to write some tests here so we can confirm this does filter on the account ID field?

Hudong-hub:
> would you be willing to write some tests here so we can confirm this does filter on the account ID field?

added some tests

PR #3631, opened 2026-03-05
KCM-5177 Update allocations autocomplete pipeline

Hudong-hub:
## Description

add account to allocation

## Related Issues

KCM-5277


core/pkg/opencost/allocationprops.go
```diff
@@ -184,6 +187,7 @@ type AllocationProperties struct {
 	Pod                  string                `json:"pod,omitempty"`
 	Services             []string              `json:"services,omitempty"`
 	ProviderID           string                `json:"providerID,omitempty"`
+	Account              string                `json:"account,omitempty"` // @bingen:field[version=18]
```
ameijer: I wonder if this PR is somewhat overkill. I don;t think we need to store accountID in these objects, it can just exist in the database, right? the user doesn't need the account ID field back in the response, we are just using it to filter. 

I wonder then if we trim this PR down to filter changes, then just introduce the account ID concept as a database-only abstraction?

PR #3632, opened 2026-03-05
KCM-5277 add account to allocation field

Hudong-hub:
## Related Issues

KCM-5277


core/pkg/opencost/allocationprops.go
```diff
@@ -175,6 +178,7 @@ func ParseProperty(text string) (AllocationProperty, error) {
 
 // AllocationProperties describes a set of Kubernetes objects.
 type AllocationProperties struct {
+	Account              string                `json:"account,omitempty"`              // @bingen:field[version=18]
```
ameijer: similar to my earlier request - I don't think this needs to exist, right? we don't want to modify the API response or the bingen object

core/pkg/opencost/allocationmatcher.go
```diff
@@ -72,6 +72,8 @@ func allocationFieldMap(a *Allocation, identifier ast.Identifier) (string, error
 		return a.Properties.Container, nil
 	case afilter.FieldProvider:
 		return a.Properties.ProviderID, nil
+	case afilter.FieldAccount:
+		return "", nil
```
ameijer: actually, I'd return a 'not implemented' error

ameijer (approved):

PR #3650, opened 2026-03-11
KubeModel

Sean-Holcomb:
## Description
 Feature PR for KubeModel hydration

This branch implements the full KubeModel hydration pipeline.  new data model and ingestion path that builds a extensible, UID-keyed graph of Kubernetes resources with associated pricing data. The work spans the core model, both data sources (prometheus-source and collector-source), cloud provider pricing integrations, and Prometheus
metric emission.

  KubeModel core (core/pkg/model/kubemodel/):
  - Expanded domain types for all Kubernetes resource kinds: Node, Pod, Container, Namespace, PVC, PV, Service, Deployment, StatefulSet, DaemonSet, ReplicaSet, Job, CronJob,
  ResourceQuota
  - New DCGM type replacing the previous generic Device/DeviceUsage types for GPU tracking
  - Full binary codec (kubemodel_codecs.go) for serialization/deserialization of all types
  - NetworkTrafficDetail type for per-pod network ingress/egress breakdown

  PricingModel (core/pkg/model/pricingmodel/):
  - New pricingmodel package with PricingModel, NodePricing, and PricingSource types
  - Binary codecs and a Window-keyed storage model
  - PricingSource interface implemented for AWS (price list API), Azure (retail pricing API), and GCP (billing API)

  Shared model (core/pkg/model/shared/):
  - Extracted Provider and UsageType enums into a shared package used by both kubemodel and pricingmodel

  Data sources:
  - prometheus-source: Implemented all previously-stubbed MetricsQuerier methods covering nodes, pods, containers, PVCs, PVs, services, namespaces, deployments, statefulsets,
   DCGM devices, and network traffic
  - collector-source: Corresponding collector-side query implementations and scraper updates

  KubeModel pipeline (pkg/kubemodel/kubemodel.go):
  - Hydration logic combining kubemodel + pricing data into a unified cluster model
  - Pricing pipeline service (pkg/pricingmodel/) with storage, runner, and status tracking

  Prometheus metric emission (pkg/metrics/kubemodel.go):
  - New KubeModelCollector emitting node_info, cluster_info, pod_info, namespace_info, deployment_info, statefulset_info, daemonset_info, job_info, cronjob_info,
  replicaset_info, resourcequota_info, and associated _labels/_annotations variants sourced from the cluster cache

  Exporter (core/pkg/exporter/):
  - Static path routing for kubemodel/pricingmodel pipeline exports

## User Impact
This is a foundational internal change introducing the KubeModel hydration pipeline. No existing public API endpoints are changed. The new Prometheus metrics emitted by
  KubeModelCollector are additive and individually disableable via the existing metrics config. No breaking changes to current cost allocation or asset APIs.

## Testing
- Unit tests pass (just test) including new tests for pricingmodel, kubemodel codecs, AWS price list API, Azure retail pricing, and GCP billing pricing source
  - Codec round-trip tests cover all new kubemodel types
  - Manual integration testing recommended against a live cluster with Prometheus to validate end-to-end hydration across all resource types and cloud providers
  - Integration tests: https://github.com/opencost/opencost-integration-tests


abhyudayareddy:
Thanks for this — the KubeModel hydration pipeline is exactly the right foundation for extensible cost attribution.

I have a companion PR open at #3704 (`feat(kubemodel): add GPUDevice and GPUUsage structs for GPU cost attribution`) that builds on this same domain. It adds:
- `GPUDevice` struct (device UUID, model, vendor, memory capacity, CUDA compute capability)
- `GPUUsage` struct (used/requested memory, utilization pct, GPU hours)
- Integration with the existing `Node` and `Pod` cost models

Given the scale of this PR (+16k lines, 88 files), I wanted to flag the overlap early so we can coordinate. A few questions:

1. **Struct placement**: Does the KubeModel hydration pipeline have a preferred location for device-level hardware structs (e.g., alongside `PVCAllocation` or as a separate `devices/` package)?
2. **Hydration source**: Are you sourcing GPU metrics from the existing Prometheus datasource, or planning a separate DCGM/NVML scrape path?
3. **UID keying**: The PR description mentions a UID-keyed graph — does `GPUDevice` fit naturally as a node keyed by `device-uuid` label from DCGM metrics?

Happy to rebase #3704 onto this branch once there's a merge path, or contribute the GPU structs directly into this PR if that's cleaner. Let me know how you'd like to coordinate.

abhyudayareddy:
Hey @Sean-Holcomb — thanks for the detailed question thread. Here are our answers to the three coordination questions, reflecting the design we've landed on in our GPU work:

---

**1. Struct placement**

`GPUDevice` and `GPUUsage` will live as top-level types in `pkg/kubemodel`, alongside `KubeNode`, `KubePod`, etc. They'll be stored in two new flat maps on `KubeModelSet`:

```go
GPUDevices map[string]*GPUDevice  // key: "{nodeUID}/{deviceIndex}"
GPUUsages  map[string]*GPUUsage   // key: "{containerUID}/{deviceUID}"
```

This mirrors the exact pattern you've established for nodes, pods, and containers — no nested structs, no separate sub-package.

---

**2. Hydration source**

We've defined a `GPUHydrator` interface:

```go
type GPUHydrator interface {
    HydrateGPU(ctx context.Context, kms *KubeModelSet, node *KubeNode, devices []*GPUDevice) error
}
```

`DefaultGPUHydrator` implements this using NVIDIA DCGM Exporter metrics via Prometheus (`DCGM_FI_DEV_GPU_UTIL`, `DCGM_FI_DEV_FB_USED`, etc.). AMD/Intel support is a separate `GPUHydrator` implementation behind the same interface.

We'd love to hook into whatever pipeline extension point you're planning for pluggable hydrators in #3650 — happy to align on the exact call site once that structure is visible.

---

**3. UID keying**

We're using `"{nodeUID}/{deviceIndex}"` as the GPU device key — for example `"abc-123-def/0"`.

- `nodeUID` is the Kubernetes `node.metadata.uid`, durable across cluster restarts and node-name reuse.
- `deviceIndex` matches `CUDA_VISIBLE_DEVICES` / ROCm ordinals and is stable on the same physical node.
- We intentionally do **not** use the DCGM UUID as the primary key because it resets on driver reload. It's stored in `GPUDevice.UUID` for diagnostics.

This is consistent with how #3650 keys other objects by Kubernetes UID.

---

We're holding our GPU PR as a draft until #3650 merges so we can depend on the finalised `KubeModelSet`, `KubeNode`, and `Window` types directly. Let us know if there's anything in the pipeline wiring we should pre-wire differently to make the integration smoother.

core/pkg/exporter/encoder.go
```diff
@@ -11,6 +11,14 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	BingenExt           = "bin"
```
mbolt35, line 15: Request that this be changed to `bingen`, as `bin` is associated with other file types. This works with the bingen file type I changed for the WAL exports.

core/pkg/exporter/encoder.go
```diff
@@ -11,6 +11,14 @@ import (
...
+	BingenVersionExtFMT = "v%d.bin"
+	JSONExt             = "json"
+	GZipExt             = ".gz"
```
mbolt35, line 18: `.gz` or `gz`? Seems like the others don't have a `.` but this one does.

core/pkg/exporter/encoder.go
```diff
@@ -11,6 +11,14 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	BingenExt           = "bin"
+	BingenVersionExtFMT = "v%d.bin"
```
mbolt35, line 16: I would love to understand the reasoning here a bit more. Are we thinking of not loading older versions at all? Is it just for visual representation on S3? I think this has some reasonable potential uses, but it does seem that there are some potential hazards as well: 
* no guarantees about version in file name and actual model version synchronization
* creating a hard dependency on bingen version representation, locking both file naming and bingen to the same scheme. 

Again, just want to understand the reasoning if we're going to take on the responsibility of the above two points.

core/pkg/model/kubemodel/kubemodel.go
mbolt35, line 8: No. For a variety of reasons but here are some key points: 
* A single lock on `KubeModelSet` doesn't enforce the use of that lock, and there are no clear indictations on what level of the structure the lock should be used. 
* A single lock will not protect any of the resources within the maps on the `KubeModelSet`. 
* A single lock will not protect any of the pointers on the `KubeModelSet` 

Data models should really never require thread safety, it's a really bad code smell. 

What type of "modifications" are we talking here? Because if this is just about building the `KubeModelSet` in parallel, then you can build all of the components in parallel, then build the set on completion. 

There is no problem at all reading the model in parallel, but I can't fathom a [good] reason to ever modify in-place over multiple threads/goroutines.

mbolt35, line 10:
```
// @bingen:generate[stringtable,streamable]:KubeModelSet
```
to allow streaming reads.

mbolt35: Note this will not change any of the existing Unmarshal methods, it just generates the code to stream it. Some light docs here: https://github.com/opencost/bingen#streamable-types

core/pkg/exporter/encoder.go
Sean-Holcomb: happy to change. It is how it is to avoid any extra logical changes. Gzip is distinct from the other types in that it is a wrapper that cannot exist on its own.

core/pkg/model/kubemodel/kubemodel_codecs.go
```diff
@@ -1,23 +1,29 @@
...
+//	              ┻━┻ ︵ヽ(`Д´)ﾉ︵ ┻━┻
 //
 //
-//            This source file was automatically generated by bingen.
+//	This source file was automatically generated by bingen.
 //
-////////////////////////////////////////////////////////////////////////////////
-
+// //////////////////////////////////////////////////////////////////////////////
+// TODO This file was infact modified due to an issue with including aliased string types from 'shared'
```
mbolt35, line 11: What did you have to specifically modify?

core/pkg/exporter/encoder.go
Sean-Holcomb: sure I can do that.

core/pkg/model/kubemodel/kubemodel.go
Sean-Holcomb: I would really like to get Streaming in kubemodel before this ships

core/pkg/model/kubemodel/kubemodel_codecs.go
mbolt35: Just want to be really clear that bingen is designed for a single package because it operates under the assumption that it has access to all package scoped data. If it is an external reference, it will always be treated as requiring a `MarshalBinary()` and `UnmarshalBinary()` implementation, as we just don't have any other details about the external type. 

I believe there is a way to allow the definition of an external aliased type as long as it's an alias to a primitive or built-in, but the nuance of this requirement has me a little uneasy.

core/pkg/model/kubemodel/namespace.go
```diff
@@ -7,38 +7,48 @@ import (
...
-	ClusterUID  string            `json:"clusterUID"`      // @bingen:field[version=1]
-	Name        string            `json:"name"`            // @bingen:field[version=1]
-	Labels      map[string]string `json:"labels"`          // @bingen:field[version=1]
-	Annotations map[string]string `json:"annotations"`     // @bingen:field[version=1]
-	Start       time.Time         `json:"start,omitempty"` // @bingen:field[version=1]
-	End         time.Time         `json:"end,omitempty"`   // @bingen:field[version=1]
+	UID string `json:"uid"` // @bingen:field[version=1]
+	// This field was included in the initial bigen codec but was never populated. If you need to add a new string
+	// field to this structure rename this field and delete this comment
+	bingenPlaceHolder string            // @bingen:field[version=1]
```
mbolt35, line 13: Smart!

core/pkg/model/kubemodel/kubemodel_codecs.go
Sean-Holcomb: I am not completely married to the shared idea if it is truly untenable here. It is certainly not my intention to leave this TODO past the merge.

The main issue here that requires manual intervention here is 
```
d := &shared.Provider{}
```
is not valid for a string alias type. With that being the case, The edit here just make it so this type is treated as the string alias that it is.

core/pkg/exporter/pathing/staticpath.go
```diff
@@ -0,0 +1,53 @@
...
+//
+// If prefix is empty the leading dot is omitted. If fileExt is empty the trailing dot is omitted.
+type StaticFileStoragePathFormatter struct {
+	rootDir  string
+	pipeline string
+}
+
+// NewStaticFileStoragePathFormatter creates a StaticFileStoragePathFormatter with the given
+// root directory and pipeline name.
+func NewStaticFileStoragePathFormatter(rootDir, pipeline string) (*StaticFileStoragePathFormatter, error) {
```
mbolt35, line 21: Just wanted to callout that for the http storage option we have, this could be a potential issue. The total number of files allowed in a single directory on `ext3` file systems is like 10-20k a directory or something before performance starts to really tank. However, I think most everything we're running on is either `XFS` or `ext4` which should scale fine for single directory files up to the millions. However, we may want to investigate which file system better supports our single dir usage for this data. 

We could also implement a storage routing system at the application layer to route files to various directories to avoid stacking on a single dir.

core/pkg/model/shared/bingen.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// type Foo struct {
+//     ExistingField1 string
+//     ExistingField2 int
+//     NewField       float64 // @bingen: <- annotation ref: bingen README
+// }
+//
+////////////////////////////////////////////////////////////////////////////////
+
+//go:generate bingen -package=shared -version=1 -buffer=github.com/opencost/opencost/core/pkg/util
```
mbolt35, line 23: You likely don't need this if all that exists in here are string alias types.

core/pkg/source/decoders.go
```diff
@@ -1046,14 +1231,55 @@ func DecodePVInfoResult(result *QueryResult) *PVInfoResult {
...
+	}
+
+	return &PodNetworkBytesResult{
+		UID:        uid,
+		Cluster:    cluster,
+		Service:    service,
+		Internet:   internet == "true",
+		SameRegion: sameRegion == "true",
+		SameZone:   sameZone == "true",
+		NatGateway: natGateway == "true",
```
mbolt35, line 1281: We should be using some type of helper here like:

```
func parseBool(value string) bool {
    b, err := strconv.ParseBool(value)
    if err != nil {
        // log or just return false 
        return false 
    }
    return b 
}
```

and then
```
Internet: parseBool(internet),
SameRegion: parseBool(sameRegion),
// etc....
```

`strconv` just handles a lot more cases than all lcase "true"

modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -29,25 +60,71 @@ func newClusterCacheScraper(clusterCache clustercache.ClusterCache) Scraper {
...
+	for _, pvc := range pvcs {
+		pvcNameToUID.Set(pvcKey{
+			name:      pvc.Name,
+			namespace: pvc.Namespace,
+		}, pvc.UID)
+	}
+	pvNameToUID := newSyncMap[string, types.UID](len(pvs))
+	for _, pv := range pvs {
+		pvNameToUID.Set(pv.Name, pv.UID)
+	}
```
mbolt35, line 104: These don't need to use sync maps, or use any mutexes at all. You're prepopulating them up front, and then reading in parallel, which is allowed with a normal `map`

core/pkg/exporter/encoder.go
Sean-Holcomb: Yes this definitely deserves some scrutiny. The main reason for it is that we need a way to identify the bingen version of Kubemodel files on the ingestor side of things to determine what should be ingested and to determine the cutoff of when we no longer ingest asset and allocation data. As you noted one of the big side benefits of this is a having a quick way of determining update times when looking at bingen data in buckets.

Realistically this information is also available by streaming the bytes and targeting the version. Or even just adding the bingen suffix and targetting files that have it versus ones that don't.
 
 The flip side of this is that it requires all consumers to identify `xyz.v2.bin` and `xyz.v3.bin` as the same file. I wasn't too concerned about the synchronization issue because the fmt is populated by the version constant in the codec file. Your second point can become more tricky when you think about codecs which might have different subversion within them, but I would say it should just be the version of the primary object being encoded.

core/pkg/model/kubemodel/kubemodel_codecs.go
mbolt35: Ohhhh! Yeah, ok, that makes sense. I forgot that we instantiated types like that. I can definitely fix this.

core/pkg/model/kubemodel/kubemodel.go
mbolt35: All you gotta do is add that tag, and it will generate for you without changing any of the existing code.

core/pkg/model/shared/provider.go
```diff
@@ -0,0 +1,37 @@
+package shared
+
+import "strings"
+
+// @bingen:generate:Provider
```
mbolt35: Don't need this unless there are types in shared that use this and require generating.

core/pkg/model/shared/usagetype.go
```diff
@@ -0,0 +1,10 @@
+package shared
+
+// @bingen:generate:UsageType
```
mbolt35: Same as Provider alias. Don't need any bingen generation unless there are generated types in the shared package that have fields of these types.

core/pkg/source/queryresult.go
```diff
@@ -173,6 +174,25 @@ func (qr *QueryResult) GetString(field string) (string, error) {
 	return strField, nil
 }
 
+func (qr *QueryResult) GetBool(field string) (bool, error) {
```
mbolt35, line 177: 🙌

core/pkg/model/kubemodel/dcgm.go
```diff
@@ -0,0 +1,66 @@
...
+	Start     time.Time
+	End       time.Time
+	Device    string
+	ModelName string
+	PodUsage  map[string]DCGMPod
+}
+
+// @bingen:generate:DCGMPod
+type DCGMPod struct {
+	ContainerUsage map[string]DCGMContainer
```
ameijer: add mem avg/max

core/pkg/model/kubemodel/dcgm.go
```diff
@@ -0,0 +1,66 @@
...
+// from the k8s API. It is left to the end user to interpret the relationships to the rest of the cluster based on
+// container unique identifiers
+// @bingen:generate:DCGMDevice
+type DCGMDevice struct {
+	UUID      string
+	Start     time.Time
+	End       time.Time
+	Device    string
+	ModelName string
+	PodUsage  map[string]DCGMPod
```
ameijer:
```suggestion
	Pod UUID
	UsageCPUAvg
	UsageCPUMax
	UsageRamAvg
	UsageRamMAx
	Power
```

core/pkg/model/kubemodel/owner.go
```diff
@@ -16,44 +18,29 @@ const (
...
+	default:
+		log.Warnf("failed to find owner kind for '%s'", kind)
+		return OwnerKind(strings.ToLower(kind))
 	}
+}
 
-	return nil
+// @bingen:generate:Owner
+type Owner struct {
+	UID        string
```
bbylovas: json tags

core/pkg/model/kubemodel/owner.go
```diff
@@ -16,44 +18,29 @@ const (
...
+		return OwnerKindReplicaSet
+	default:
+		log.Warnf("failed to find owner kind for '%s'", kind)
+		return OwnerKind(strings.ToLower(kind))
 	}
+}
 
-	return nil
+// @bingen:generate:Owner
+type Owner struct {
```
bbylovas, line 43: only see this as part of ReplicaSet and Pod, any reason it isn't used elsewhere?

core/pkg/model/kubemodel/node.go
```diff
@@ -10,95 +10,57 @@ import (
...
-	RAMByteUsageMax      Measurement                 `json:"ramByteUsageMax"`      // Peak RAM usage observed
-	Start                time.Time                   `json:"start,omitempty"`      // Node creation/start timestamp
-	End                  time.Time                   `json:"end,omitempty"`        // Node deletion/end timestamp (nil if still running)
+	UID                  string             `json:"uid"`
+	ProviderID           string             `json:"providerId"`
+	Name                 string             `json:"name"`
+	Labels               map[string]string  `json:"labels"`
+	InstanceType         string             `json:"instanceType"`
+	Preemptible          bool               `json:"preemptible"` // TODO unpopulated
+	ResourceCapacities   ResourceQuantities `json:"resourceCapacities"`
```
bbylovas, line 17: nit - just flagging that ResourceCapacities and Resource**s**Allocatable use different singular/plural forms

core/pkg/model/kubemodel/pod.go
```diff
@@ -5,43 +5,61 @@ import (
 	"time"
 )
 
+// @bingen:generate:PodPVCVolumes
+type PodPVCVolumes struct {
```
bbylovas: nit - would change to `PodPVCVolume`

core/pkg/model/kubemodel/pod.go
```diff
@@ -5,43 +5,61 @@ import (
...
+	Name                     string `json:"name"`
+	PersistentVolumeClaimUID string `json:"persistentVolumeClaimUID"`
```
bbylovas: nit - flagging that `UID` breaks the `Uid` convention used in other json tags

core/pkg/model/kubemodel/service.go
```diff
@@ -11,51 +16,69 @@ const (
...
+	Type         ServiceType `json:"type"`
+	Start        time.Time   `json:"start"`
+	End          time.Time   `json:"end"`
 	// Label selector to identify pods/containers targeted by this service
 	// Maps label keys to values (e.g., {"app": "nginx", "tier": "frontend"})
 	// Pods with matching labels will receive traffic from this service
-	Selector map[string]string `json:"selector,omitempty"`
-	// Lifecycle tracking
-	DurationSeconds Measurement `json:"durationSeconds"` // Duration service existed within measurement window
+	Selector         map[string]string `json:"selector,omitempty"`
```
bbylovas, line 66: nit - would change `Selector` to `Selectors`

core/pkg/model/kubemodel/dcgm.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// @bingen:generate:DCGMPod
+type DCGMPod struct {
```
bbylovas, line 25: missing json tags

core/pkg/model/kubemodel/pvc.go
```diff
@@ -7,49 +7,47 @@ import (
...
-	DurationSeconds       Measurement `json:"durationSeconds,omitempty"`
+	UID                 string    `json:"uid"`
+	NamespaceUID        string    `json:"namespaceUid"`
+	Name                string    `json:"name"`
+	PersistentVolumeUID string    `json:"persistentVolumeUID,omitempty"`
+	StorageClass        string    `json:"storageClass"`
+	Start               time.Time `json:"start"`
+	End                 time.Time `json:"end"`
+	RequestedBytes      float64   `json:"requestedBytes"`
+	UsageBytesAvg       float64   `json:"usageBytesAvg"`
```
bbylovas, line 18: could be wrong but doesn't look like `UsageBytesAvg|Max` fields are populated anywhere yet

core/pkg/model/kubemodel/owner.go
Sean-Holcomb: They are the only one we use. It is used for determining the controller of the pod. If a pod has a replicaset owner then we check if the replicaset has a deployment as its controller

core/pkg/model/kubemodel/node.go
Sean-Holcomb: I also don't love this but this one is kind of a "Governors General" type situation. You aren't supposed to make the adjective plural. We generally use plurals for multiple things so that makes `ResourceCapacity/ResourceQuantity` or just `Capacity/Resource` (like it is in the node def of the k8s api) no good. The only other options I can think of are `Capacities/Allocatables` and `ResourceCapacities/ResourceAllocatables`. Both have the grammatical issue (They are currently have a red underlined because 'allocatables' isn't a word) and while the former feels better without resources there, I do like having resource in the field because it is descriptive.

core/pkg/model/kubemodel/service.go
Sean-Holcomb: The items in selector are always combined with a logical And operation. So this is a single selector with multiple criteria rather than multiple selectors.

core/pkg/model/kubemodel/pvc.go
Sean-Holcomb: That appears to be right

core/pkg/clustercache/clustercache.go
```diff
@@ -168,6 +174,14 @@ type ResourceQuota struct {
...
+type CronJob struct {
+	UID         types.UID
+	Name        string
+	Namespace   string
+	Labels      map[string]string
+	Annotations map[string]string
+}
+
 type Volume struct {
 }
```
nikovacevic, line 186: Delete this

Sean-Holcomb:
Closing In favor of https://github.com/opencost/opencost/pull/3875

PR #3654, opened 2026-03-13
don't use provider config, thereby reading from cloud-integration.json

ameijer:
## Release Notes
This update moves all cloud provider integration configuration to the new cloud-integrations.json system. Anyone using the old cloud provider credentials mechanism will need to upgrade to the new multi cloud configuration mechanism. 

## Description
fixes an issue where we were not reading from cloud-integration.json for cloud costs


Sean-Holcomb (approved):
LGTM

PR #3656, opened 2026-03-17
Remove broken needs-follow-up-label workflow

### LINKED ISSUE #3598: needs-follow-up-label action is broken
Every time I comment on my PR I get an email like this:

Image

The workflow log looks like this:

Image

Kush172005:
This PR removes the `needs-follow-up-label` GitHub Actions workflow.

The workflow is currently failing on every comment due to permission issues and deprecated configurations, which ends up creating unnecessary noise in the repo.

Based on the discussion in #3598, this automation doesn’t add much value since open PRs can already be assumed to need follow-up unless they are closed.

Removing it keeps things simpler and avoids repeated workflow failures.

Fixes #3598


ameijer (approved):

ameijer:
thanks @Kush172005 for your contribution!

Kush172005:
Thanks for the review!

PR #3668, opened 2026-03-20
Fix resource leak in OTC API pagination by moving resp.Body.Close()

peatey:
## Description
Moved `resp.Body.Close()` from a deferred statement to an explicit call after `loadPaginatedResponse()` completes. This ensures the response body is closed immediately after use rather than being deferred until the end of the function, preventing potential resource leaks in the pagination loop.

The change maintains the same cleanup behavior while improving resource management by closing the body as soon as it's no longer needed, rather than holding the resource open until the function returns.

## User Impact
http.Client with a 30s timeout. This is an internal optimization to improve resource management in the OTC pricing API client.

## Testing
Existing tests should continue to pass. The change is a refactoring of resource cleanup logic with no functional behavior changes.


pkg/cloud/otc/pricingapi.go
```diff
@@ -5,10 +5,13 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
 	"github.com/opencost/opencost/core/pkg/log"
 )
 
+var otcHTTPClient = &http.Client{Timeout: 30 * time.Second}
+
```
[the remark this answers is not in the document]
peatey: LLM babysitting complete. I should have written this myself.

CLAUDE.md
```diff
@@ -2,6 +2,10 @@
 
 This document provides guidance for AI assistants working with the OpenCost codebase.
 
+## AI Assistant Behaviour
+
+- Never include claude.ai session links or URLs in commit messages or pull request bodies.
```
ameijer, line 7: oh claude

ameijer (approved):

CLAUDE.md
peatey: its a f-ing mess.

peatey: should have just written this myself

image

ameijer: yipers

PR #3670, opened 2026-03-20
fix: actually flush caches when settings change

### LINKED ISSUE #3601: Settings change handler logs cache flush but doesn't do anything
**Describe the bug**
In **pkg/costmodel/settings.go** (lines 47-56), there's a goroutine that's meant to clear caches when pricing or discount settings change:

```
// Clear caches when custom pricing or discount changes
go func(a *Accesses) {
    costDataCacheCh := make(chan string)
    a.SubscribeToCustomPricingChanges(costDataCacheCh)
    a.SubscribeToDiscountChanges(costDataCacheCh)
    for {
        msg := <-costDataCacheCh
        log.Infof("Flushing cost data caches: %s", msg)
    }
}(a)
```
It picks up the change notification and logs "Flushing cost data caches" but doesn't actually flush anything. The old cached values stay.
Also noticed cache entries are set with different expirations — **cache.NoExpiration** on first set (line 100) vs **cache.DefaultExpiration** on updates (line 112). So values set during init might never expire, and the "flush" doesn't clear them either.

**To Reproduce**

1. Start OpenCost, let it initialize
2. Change custom pricing or discount settings via the API
3. Check logs — you'll see "Flushing cost data caches: ..."
4. Old cached values are still being used

**Expected behavior**
The handler should actually flush the cache when it gets a change notification, not just log about it.

**Screenshots**
N/A

v1.119.2 — main

**Additional context**
Not sure if this is intentional (maybe the flush was removed at some point?) or just not implemented yet. The inconsistent cache expiration makes it worse since init values use NoExpiration. Happy to add the flush call if it should be there.

vincent067:
## Description

Fixes #3601

The settings change handler in `InitializeSettingsPubSub` was logging 'Flushing cost data caches' but not actually doing anything. This meant that when custom pricing or discount settings changed, the old cached values would still be used.

## Changes

- Added a call to `DownloadPricingData()` when settings change
- Added error logging if the re-download fails

## Testing

The fix ensures that new pricing/discount settings take effect immediately by re-downloading pricing data when:
- Custom pricing settings change
- Discount settings change

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are well-commented
- [x] Fixes the reported issue

Thanks for maintaining OpenCost! 🙏


peatey:
Couple of things here -

1. @vincent067 - thank you for the contribution
2. there is no tests included with this PR
3. Copilot review will need to be addressed

Lastly, the NoExpiration bug from the issue is unaddressed
Issue #3601 also called out an inconsistent cache expiration bug in the same file:

```
settings.go:100 uses cache.NoExpiration on first set
settings.go:112 uses cache.DefaultExpiration on updates
```
This means after the DefaultExpiration elapses, the next poll will find no cached entry, re-set with NoExpiration, and return false — incorrectly treating an expired entry as "no change". The PR description says it fixes #3601 but this part is left hanging.

PR #3674, opened 2026-03-20
Kcm 5329 network detail

aaliomer-ibm:
Store  network detail independently of pods to allow for different costs for TrafficType.


ameijer (approved):

core/pkg/model/kubemodel/networktrafficdetail.go
```diff
@@ -0,0 +1,28 @@
...
+	TrafficTypeCrossRegion TrafficType = "CrossRegion"
+	TrafficTypeInternet    TrafficType = "Internet"
+	TrafficTypeNatGateway  TrafficType = "NatGateway"
+)
+
+// @bingen:generate:NetworkTrafficDetail
+type NetworkTrafficDetail struct {
+	PodUID           string           `json:"podUid"`
+	Endpoint         string           `json:"endpoint,omitempty"` // destination service/IP, e.g. "aws-s3", "10.0.1.5" (AKA Service Name)
+	TrafficDirection TrafficDirection `json:"trafficDirection"`   // "Egress" or "Ingress"
```
Sean-Holcomb: This makes a lot of sense to me. A lot easier to ingest and less clutter.

pkg/kubemodel/kubemodel.go
```diff
@@ -425,140 +417,125 @@ func (km *KubeModel) computePods(kms *kubemodel.KubeModelSet, start, end time.Ti
...
+
+	return nil
+}
+
+func (km *KubeModel) computeNetworkTraffic(kms *kubemodel.KubeModelSet, start, end time.Time) error {
+	grp := source.NewQueryGroup()
+	metrics := km.ds.Metrics()
+
+	netZoneGiBFuture := source.WithGroup(grp, metrics.QueryNetZoneGiB(start, end))
+	netRegionGiBFuture := source.WithGroup(grp, metrics.QueryNetRegionGiB(start, end))
```
Sean-Holcomb: @mbolt35 Is service always blank for these queries?

pkg/kubemodel/kubemodel.go
```diff
@@ -425,140 +417,125 @@ func (km *KubeModel) computePods(kms *kubemodel.KubeModelSet, start, end time.Ti
...
+	netInternetGiBFuture := source.WithGroup(grp, metrics.QueryNetInternetGiB(start, end))
```
Sean-Holcomb: I think we can delete this query for this, because the InternetServiceGIB already queries this data.

core/pkg/model/kubemodel/kubemodel.go
```diff
@@ -42,23 +43,24 @@ func NewKubeModelSet(start time.Time, end time.Time) *KubeModelSet {
...
+		Devices:                  map[string]*Device{},
+		DeviceUsages:             map[string]*DeviceUsage{},
+		Namespaces:               map[string]*Namespace{},
+		Nodes:                    map[string]*Node{},
+		Pods:                     map[string]*Pod{},
+		PersistentVolumeClaims:   map[string]*PersistentVolumeClaim{},
+		ResourceQuotas:           map[string]*ResourceQuota{},
+		Services:                 map[string]*Service{},
+		PersistentVolumes:        map[string]*PersistentVolume{},
+		PodNetworkTrafficDetails: map[string][]*NetworkTrafficDetail{},
```
Sean-Holcomb: I am kind of vacillating on moving this out of the pod. In my GPU work there is a similar case where we are adding metrics from a different source (DCGM exporter) for the container. Right now we have the DeviceUsage which exists on the premise the we have a lot of information about Devices (we do not). The reality is that we are going to have something keyed on podUID/containerName and we can either add this in with the container or put another map in the main `KubeModelSet` 

One thought I had was, if it is just being keyed on a UID of a specific object doesn't that just mean it belongs on that object? This object is describing usage of the pod just like other metrics

Sean-Holcomb (commented):
I think this change in structure does a better job of recording the metrics that are coming out of the daemonset, to support this there does need to be some changes to the way we are hydrating. 

I want to keep discussing the place for this in the KubeModelSet.

pkg/kubemodel/kubemodel.go
mbolt35: As it works today, I don't think there exists a real scenario where the `service` label is populated where `internet != "true"` but there's nothing really preventing this from happening. For example, a user could configure the S3 service IP to be classified as `cross-zone` which would be a metric with `same_region="true"` and `service="aws-s3"`. 

We can punt on this scenario for now I think and assume that the `service` label can only be populated when `internet="true"`, as I believe that is a relatively safe assumption. 

So, querying for InternetGiB AND InternetServiceGiB is technically the same aggregate of data, but `InternetGiB` doesn't separate data by the service label. You could technically only query `InternetServiceGiB` and derive `InternetGiB` by totaling all the returned metrics. 

The reason the Service queries were separate is because the service based queries were made via the network insights pipeline before.

pkg/kubemodel/kubemodel.go
mbolt35: Yeah, see the previous comment. I _think_ (as long as `QueryNetInternetServiceGiB` includes `service=""`), then you can derive the `QueryNetInternetGiB` results by summing all returned metrics data.

pkg/kubemodel/kubemodel.go
Sean-Holcomb: Okay in that case I think we may want to add the service label to each of these queries

Sean-Holcomb (approved):
Looks good, my one thought is that we could just add on ingress or egress query instead of 1 per traffic type per direction.

ameijer (approved):

PR #3675, opened 2026-03-20
Device Update

Sean-Holcomb:


core/pkg/model/kubemodel/container.go
```diff
@@ -7,24 +7,17 @@ import (
...
-	GPUAllocated          float64   `json:"gpuAllocated"`
-	GPUUsageAvg           float64   `json:"gpuUsageAvg"`
-	GPUUsageMax           float64   `json:"gpuUsageMax"`
-	GPURequest            float64   `json:"gpuRequest"`
-	Start                 time.Time `json:"start"`
-	End                   time.Time `json:"end"`
+	PodUID           string                          `json:"podUid"`
+	Name             string                          `json:"name"`
+	ResourceRequests ResourceQuantities              `json:"resourceRequests"`
+	ResourceLimits   ResourceQuantities              `json:"ResourceLimits"`
```
Sean-Holcomb: Using ResourceQuantities to capture requests for arbitrary resource names made available by device plugins

core/pkg/model/kubemodel/container.go
```diff
@@ -7,24 +7,17 @@ import (
 
 // @bingen:generate:Container
 type Container struct {
-	PodUID                string    `json:"podUid"`
-	Name                  string    `json:"name"`
-	CPUMilliCoreAllocated float64   `json:"cpuMilliCoreAllocated"`
-	CPUMilliCoreUsageAvg  float64   `json:"cpuMilliCoreUsageAvg"`
-	CPUMilliCoreUsageMax  float64   `json:"cpuMilliCoreUsageMax"`
-	CPUMilliCoreRequest   float64   `json:"cpuMilliCoreRequest"`
-	CPUMilliCoreLimit     float64   `json:"cpuMilliCoreLimit"`
```
Sean-Holcomb, line 16: Replace all milliCore with Core

core/pkg/model/kubemodel/device.go
```diff
@@ -1,89 +1,44 @@
 package kubemodel
 
 import (
-	"errors"
 	"fmt"
 	"time"
 )
 
+// @bingen:generate:Device
```
Sean-Holcomb: Device, representing device plugins are very tricky. Information about them very limited. Most information that we have comes from the DCGM exporter which is specific to Nvidia GPUs but also can report stats for DRAs.

core/pkg/model/kubemodel/node.go
```diff
@@ -10,18 +10,17 @@ import (
...
-	Start         time.Time         `json:"start"`
-	End           time.Time         `json:"end"`
+	UID                  string             `json:"uid"`
+	ProviderID           string             `json:"providerId"`
+	Name                 string             `json:"name"`
+	Labels               map[string]string  `json:"labels"`
+	InstanceType         string             `json:"instanceType"`
+	Preemptible          bool               `json:"preemptible"` // TODO unpopulated
+	ResourceCapacities   ResourceQuantities `json:"resourceCapacities"`
+	ResourcesAllocatable ResourceQuantities `json:"resourcesAllocatable"`
```
Sean-Holcomb, line 20: Using ResourceQuantities here to capture arbitrary resources added by device plugins

aaliomer-ibm: im not a fan of ResourceQuantity as a concept. I limits should be explicit fields. similar to network direction  and traffic type. however in this case, a ResourceQuantity is so generic it will make it a nightmare to deal with the code and understand it without jumping around

core/pkg/model/kubemodel/device.go
```diff
@@ -1,89 +1,44 @@
...
-	}
-	if d.PowerWattSeconds < 0 {
-		return fmt.Errorf("PowerWattSeconds cannot be negative, got %.2f", d.PowerWattSeconds)
-	}
-	if d.PowerWattMax < 0 {
-		return fmt.Errorf("PowerWattMax cannot be negative, got %.2f", d.PowerWattMax)
-	}
-	return nil
+// @bingen:generate:ContainerDeviceUsage
+type ContainerDeviceUsage struct {
```
aaliomer-ibm: Devices are somewhat implicit. need to think about how these resources/devices are defined (by name, etc..).
I understand the motivation about giving ID's to objects, but similar to containers, a device might be identified by multiple properties.

aaliomer-ibm (approved):

PR #3695, opened 2026-03-26
fix: attribute pod costs to deployments, statefulsets, and services

### LINKED ISSUE #3694: Deployment-level cost breakdowns missing: costs shown only as unallocated or idle
**Describe the bug**
Deployment-level (and service-level) cost breakdowns are not appearing in the UI. Most costs are being categorized as __unallocated__ or __idle__, and deployment-level costing (and some other breakdown types as well) are unavailable.

**To Reproduce**
- Deploy workloads across namespaces with multiple Deployments and controllers.
- Run cost collection and report generation per normal configuration (cluster sources, kube state, etc.).
- Open the cost breakdown UI and view by Deployment (or Service).

**Expected behavior**
- Costs are attributed to individual Deployments and controllers where possible.
- __unallocated__ and __idle__ remain only for resources that truly cannot be attributed.
- Deployment and Controller breakdown views show correct cost allocation.

**Screenshots**

Image

Image

1.119.2 

**Additional context**

**Additional context**
I am using it on a GCP cluster with the following configuration of custom costing:
```yaml
customPricing:
  enabled: true
  provider: "custom"
  createConfigmap: true 
  costModel:
    description: Modified pricing configuration (unrealistically large values).
    CPU: 1.5
    spotCPU: 0.3
    RAM: 0.5
    spotRAM: 0.3
    GPU: 2.5
    storage: 0.7
    zoneNetworkEgress: 0.4
    regionNetworkEgress: 0.4
    internetNetworkEgress: 0.4
```

NETIZEN-11:
All pod costs were incorrectly appearing as `__unallocated__` or `__idle__` in the OpenCost UI, with no breakdown by Deployment, StatefulSet, or Service, even on a correctly configured cluster with kube-state-metrics and Prometheus.

After analyzing the allocation pipeline, I identified multiple issues that together break cost attribution. In many cases, even a single issue could silently disrupt attribution for an entire scrape cycle.

OpenCost relies on internally emitted Prometheus metrics (`deployment_match_labels`, `statefulSet_match_labels`, and `service_selector_labels`) and matches them against `kube_pod_labels` to attribute costs. If any of these metrics are missing or selector logic fails, pods are categorized as `__unallocated__`.

This PR fixes the following problems:

* Fixed a nil pointer panic in `TransformDeployment` by adding a guard before accessing `Spec.Selector`. Certain deployments (e.g., created by operators) may have nil selectors, which previously caused cache refresh failures and wiped deployment mappings.

* Added support for `matchExpressions` in `KubecostDeploymentCollector`. Previously, only `MatchLabels` were considered, causing deployments using `matchExpressions` to be ignored and not emitted as metrics.

* Applied the same `matchExpressions` fallback logic to statefulsets in `KubecostStatefulsetCollector`.

* Fixed a nil pointer panic in `UpdateControllerSelectorsCache` by adding checks before accessing `ReplicaSet` and `StatefulSet` selectors. This prevented failures in pod label metric generation.

* Prevented empty selectors from matching all pods in `labelsToPodControllerMap`. Previously, empty selectors were treated as match-all, causing incorrect controller assignment and overwriting valid mappings.

* Fixed the same issue in `applyServicesToPods`, where services with empty selectors incorrectly matched all pods in a namespace.

Changes include updates in:

* `core/pkg/clustercache/clustercache.go`
* `pkg/metrics/deploymentmetrics.go`
* `pkg/metrics/statefulsetmetrics.go`
* `pkg/metrics/podlabelmetrics.go`
* `pkg/costmodel/allocation_helpers.go`

Added test coverage for all fixes:

* Deployment collector with matchExpressions
* StatefulSet collector with matchExpressions
* Nil selector handling in controller cache

All new and existing tests pass successfully. No breaking changes introduced.

This fix restores correct cost attribution across deployments, statefulsets, and services, ensuring only genuinely unallocatable resources are categorized as `__unallocated__` or `__idle__`.

Closes #3694

Signed-off-by: NETIZEN-11 [[niteshkumar121411@gmail.com](mailto:niteshkumar121411@gmail.com)](mailto:niteshkumar121411@gmail.com)


peatey (changes requested):
Thanks for this detailed PR — the nil-selector guards (`TransformDeployment`, `UpdateControllerSelectorsCache`) and the empty-selector skips in `labelsToPodControllerMap` / `applyServicesToPods` are all correct and well-motivated. Two issues need addressing before merge.

---

### Bug 1 — `Exists` operator handling is semantically wrong (and the test uses invalid Kubernetes state)

**Files:** `pkg/metrics/deploymentmetrics.go`, `pkg/metrics/statefulsetmetrics.go`

```go
if len(expr.Values) == 1 &&
    (expr.Operator == metav1.LabelSelectorOpIn ||
        expr.Operator == metav1.LabelSelectorOpExists) {
    selectorLabels[expr.Key] = expr.Values[0]
}
```

The [Kubernetes API spec](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) requires that `Exists` and `DoesNotExist` have an **empty** `Values` slice. A well-formed `Exists` expression always has `len(expr.Values) == 0`, so `len(expr.Values) == 1 && operator == Exists` is a condition that can never be true for any real cluster resource. The `Exists` arm is dead code.

Beyond that, `Exists` semantics are "any pod that carries this label key, regardless of value." That cannot be captured as a flat `key=value` equality pair — if you pick an arbitrary value you'll miss all pods whose label has a different value.

The test case `"deployment with matchExpressions Exists operator single value"` uses `Values: []string{"frontend"}` with `Exists`, which Kubernetes would reject at admission. It passes only because the test constructs the struct directly, bypassing validation.

**Suggested fix:** Remove `LabelSelectorOpExists` from the condition entirely. Only `In` with a single value can be soundly converted to an equality pair.

---

### Bug 2 — Partial synthesis produces false-positive pod attributions

**Files:** `pkg/metrics/deploymentmetrics.go`, `pkg/metrics/statefulsetmetrics.go`

Consider a deployment whose selector is:

```yaml
matchExpressions:
  - {key: app,   operator: In,    values: [myapp]}
  - {key: env,   operator: NotIn, values: [prod]}
```

The loop synthesizes `{app: myapp}` (only the `In` expression qualifies) and emits that as `deployment_match_labels`. OpenCost then builds an equality selector from `{app: myapp}` and matches **all** pods with `app=myapp` — including prod pods — to this deployment. The resulting attribution is more permissive than the real selector, and it contradicts the `NotIn` constraint the cluster operator expressed.

The safe rule is: only synthesize a label map when **every** expression in the list can be reduced to a single equality pair. If any expression is non-synthesizable (multi-value `In`, `NotIn`, `DoesNotExist`, zero-value `Exists`), skip the deployment rather than emitting a partial/incorrect metric.

```go
selectorLabels := make(map[string]string)
ok := true
for _, expr := range deployment.SpecSelector.MatchExpressions {
    if expr.Operator == metav1.LabelSelectorOpIn && len(expr.Values) == 1 {
        selectorLabels[expr.Key] = expr.Values[0]
    } else {
        ok = false
        break
    }
}
if !ok {
    selectorLabels = nil // skip this deployment
}
```

(Same change needed in `statefulsetmetrics.go`.)

---

Everything else looks good. The nil/empty-selector defensive guards are correct, the test structure is clean, and the `matchLabels` precedence logic is right. Fixing these two issues will make the `matchExpressions` path safe to ship.

---

peatey:
this is interesting. The review above was autogenerated by claude when I asked it to review. I do this as part of all PRs - but I never allow it to straight post, code or do anything without reiview as it is often wrong. Plus we also use colpilot as well for this, I personally do not like it wrtiting on my behalf. though @NETIZEN-11 we do appreciate the contrribution those are not my words. 

Just, interesting.

NETIZEN-11:
@peatey Thanks for the detailed review! I've addressed the issues you pointed out:

- Fixed the incorrect handling of the `Exists` operator  
- Resolved the second issue as suggested  

Please take another look when you have a moment.

PR #3706, opened 2026-03-30
Memory tweaks 2

ameijer:
## Description

implements a number of additional methods for file storage of objects


core/pkg/storage/azurestorage.go
```diff
@@ -254,6 +257,66 @@ func (b *AzureStorage) Read(name string) ([]byte, error) {
...
+	}
+
+	retryReader := downloadResponse.NewRetryReader(ctx, &azblob.RetryReaderOptions{
+		MaxRetries: int32(b.config.ReaderConfig.MaxRetryRequests),
+	})
+	return retryReader, nil
+}
+
+// ReadToLocalFile streams the specified blob at path to destPath on the local file system.
+func (b *AzureStorage) ReadToLocalFile(path, destPath string) error {
```
mbolt35, line 282: Did we still need this? Might be helpful debugging?

Totally fine with it, just wanted to make sure :)

ameijer: I figured I would keep it in, because specifically I am worried about making this work with bingen compression. gz does not work with the ranged queries. when we grow into compressed bingen, we may still need to download and decompress files before streaming them - leaving this impl here saves us work later on

core/pkg/storage/s3storage.go
```diff
@@ -503,37 +608,143 @@ func (s3 *S3Storage) getRange(ctx context.Context, name string, off, length int6
...
+	if chunkSize <= 0 {
+		chunkSize = defaultS3ReadChunkSize
+	}
+	return &s3ChunkReader{
+		size:      size,
+		chunkSize: chunkSize,
+		chunkOff:  -1,
+		fetch:     fetch,
+	}
+}
```
mbolt35: What is the point of this type? It looks like the call to: 

```
r, err := s3.client.GetObject(ctx, s3.name, path, *opts)
```
in the ReadToLocalFile returns an `io.Reader` can we not use that? Just trying to figue out what the difference is.

core/pkg/util/buffer_test.go
```diff
@@ -193,11 +192,11 @@ func TestBufferBytes(t *testing.T) {
...
-		t.Errorf("Expected int value to be -42, got %v", intVal)
+	if intVal != 42 {
+		t.Errorf("Expected int value to be 42, got %v", intVal)
 	}
-	if floatVal != -3.14 {
-		t.Errorf("Expected float value to be -3.14, got %v", floatVal)
+	if floatVal != 3.14 {
+		t.Errorf("Expected float value to be 3.14, got %v", floatVal)
 	}
 }
```
mbolt35, line 202: Looks like this was changed accidentally

core/pkg/util/buffer_test.go
```diff
@@ -284,87 +283,6 @@ func TestStringBytes(t *testing.T) {
...
-		t.Errorf("expected 123456, got: %d", ui64)
-	}
-	if i16 != int16(44) {
-		t.Errorf("expected 44, got: %d", i16)
-	}
-	if f32 != float32(5.0) {
-		t.Errorf("expected 5.0, got: %f", f32)
-	}
-}
-
```
mbolt35, line 367: Lost some tests here?

core/pkg/storage/s3storage.go
ameijer: it's an optimization. I added a comment from AI describing it. I think the reasons are compelling

mbolt35 (approved):
Looks good to me!

PR #3720, opened 2026-04-08
Suppress spot feed warnings when spot data feed not configured

thomasvn:
## Description
Suppresses AWS spot feed warning logs on default installations where spot data feed is not configured.

**Problem:** Default installations log warnings for every preemptible node:
```
WRN Node i-05d3ee5813b30da02 marked preemptible but we have no data in spot feed
```

**Solution:** Only log warnings when `SpotRefreshEnabled()` returns true.

## Related Issues
N/A

## User Impact
- Default installations: No spot feed warnings
- Configured spot feed: Warnings still appear when expected

## Testing
- All existing tests pass (11 test cases)


ameijer (approved):
LGTM

PR #3721, opened 2026-04-10
Fix unsafe type assertion in custom cost plugin loader

### LINKED ISSUE #3714: Unsafe Type Assertion in Plugin Loader
**Describe the bug**
In `pkg/customcost/ingestor.go `, there's an unsafe type assertion when dispensing a plugin:

`custCostSrc := raw.(ocplugin.CustomCostSource)`

If a plugin returns an unexpected type (due to a version mismatch, misconfiguration, or a buggy plugin), this panics and crashes the entire OpenCost server process.

**To Reproduce**

1. Build or use a custom plugin that returns a type not implementing CustomCostSource
2. Start OpenCost with the plugin configured
3. The server panics on plugin initialization
4. Expected behavior

**Expected behavior**
The server should log an error and skip the offending plugin gracefully, without crashing. Other plugins and the rest of the server should continue running normally.

Latest (main branch) - pkg/customcost/ingestor.go:197

**Additional context**
The fix is straightforward - replace the bare assertion with a checked one:

```

custCostSrc, ok := raw.(ocplugin.CustomCostSource)
if !ok {
    log.Errorf("plugin %s returned invalid type: expected CustomCostSource, got %T", domain, raw)
    return
}

```

peatey:
Replace bare type assertion with comma-ok pattern to prevent server panic when a plugin returns an unexpected type. The server now logs an error and skips the faulty plugin instead of crashing.

Fixes #3714

### DIFF
--- a/pkg/customcost/ingestor.go
+++ b/pkg/customcost/ingestor.go
@@ -21,6 +21,12 @@ import (
 	"github.com/opencost/opencost/pkg/env"
 )
 
+// pluginConnector abstracts *plugin.Client so buildSingleDomain can be tested with mocks.
+type pluginConnector interface {
+	Client() (plugin.ClientProtocol, error)
+	Kill()
+}
+
 // IngestorStatus includes diagnostic values for a given Ingestor
 type IngestorStatus struct {
 	Created     time.Time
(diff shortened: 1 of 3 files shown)


ameijer (approved):
I wouldn't hate a test here...

peatey:
> I wouldn't hate a test here...

Good call — I looked into adding a test for this path but it's tricky: buildSingleDomain operates on the concrete *plugin.Client type, so reaching the type assertion line requires a fully functional go-plugin server (separate process) that successfully handshakes and dispenses the wrong type. The existing TestIngestor_BuildWindow_WithPlugin covers the earlier error paths but can't reach line 197.
Two options to make this testable:
Extract an interface around the plugin client (e.g. pluginDispenser) so buildSingleDomain can accept a mock — clean but a bigger refactor
Create a test plugin binary that deliberately returns the wrong type — heavyweight for a one-line fix
Happy to go with option 1 if you think the added testability is worth the refactor. Let me know!

ameijer:
idk I kinda like the mocking and interface idea

ameijer (approved):

PR #3722, opened 2026-04-10
Sth/kcm 5437

Sean-Holcomb:
## Description
  Extends the NodeKey/NodePricing pricing model schema and implements a GCP Cloud Billing Catalog pricing source alongside the existing AWS and Azure sources.  
                                                                                                                                                                  
  Schema changes (core/pkg/model/pricingmodel):                                                                                                                   
  - Added Family string, Accelerator string, and Type NodePricingType fields to NodeKey to support component-based pricing (per-vCPU, per-GB RAM, per-device) used
   by GCP                                                                                                                                                         
  - Renamed TotalHourlyRate → HourlyRate on NodePricing to reflect that entries may represent a component rate rather than a total instance rate                
  - Added NodePricingType string alias with constants: Total, CPUCore, RamGB, Device                                                                              
  - Regenerated bingen codecs; manually fixed cross-package import path issue in generated file                                                                   
                                                                                               
  GCP pricing source (pkg/cloud/gcp/billingpricingsource.go):                                                                                                     
  - Fetches paginated SKUs from the GCP Cloud Billing Catalog API (cloudbilling.googleapis.com)                                                                   
  - Emits per-vCPU and per-GB RAM entries keyed by family + region, and per-device entries keyed by accelerator label + region                                    
  - Reuses existing GCPPricing/NormalizeGPULabel types from the GCP provider package                                                                              
  - Skips commitment SKUs and zero-price reserved device SKUs                                                                                                     
                                                                                                                                                                  
  Pipeline config:                                                                                                                                                
  - Added GCPRunnerConfig (disabled by default; requires APIKey)                                                                                                  
  - Moved CurrencyCode to top-level PipelineConfig; all three sources share a single currency setting                                                             
  - AWS source now uses PriceListEC2PricePerUnit.ForCurrency() to select the configured currency, defaulting to USD                                               
                                                                                                                                                                  
  Bug fixes:                                                                                                                                                      
  - Fixed Azure Retail Pricing source sending unencoded $filter query parameter, causing 400 responses from Azure Front Door                                      
  - Added HTTP status code checks to Azure and GCP sources with body logging on error   

## Related Issues


## User Impact
The NodeKey and NodePricing types are updated. Consumers reading PricingModelSet data from storage will need to handle the renamed HourlyRate field and the new 
  Type/Family/Accelerator fields on NodeKey. The bingen wire format for NodeKey has changed (three additional string fields appended); existing cached files will
  fail to deserialize and should be deleted.                                                                                                                      
                                                                                                                                                                
  This is a breaking change to the NodeKey/NodePricing schema relative to feature/kubemodel.     

## Testing
  - Built and ran the AWS and Azure pricing sources end-to-end; confirmed pricing entries are populated correctly                                                 
  - Confirmed Azure 400 fix resolves the URL encoding issue
  - GCP source builds and follows the same pattern as Azure; requires a valid Billing API key for live testing

### DIFF
--- a/core/pkg/model/pricingmodel/node.go
+++ b/core/pkg/model/pricingmodel/node.go
@@ -10,7 +10,7 @@ const (
 	NodePricingTypeTotal   NodePricingType = "Total"
 	NodePricingTypeCPUCore NodePricingType = "CPUCore"
 	NodePricingTypeRamGB   NodePricingType = "RamGB"
-	NodePricingTypeGPU     NodePricingType = "GPU"
+	NodePricingTypeDevice  NodePricingType = "Device"
 )
 
 // @bingen:generate:NodeKey
--- a/core/pkg/model/pricingmodel/node.go
+++ b/core/pkg/model/pricingmodel/node.go
@@ -4,6 +4,7 @@ import (
 	"github.com/opencost/opencost/core/pkg/model/shared"
 )
 
+// @bingen:generate:NodePricingType
 type NodePricingType string
 
 const (
@@ -15,13 +16,13 @@ const (
 
 // @bingen:generate:NodeKey
 type NodeKey struct {
-	Provider    shared.Provider
-	Region      string
-	NodeType    string
-	UsageType   shared.UsageType
-	Family      string
-	Accelerator string
-	Type        NodePricingType
+	Provider   shared.Provider
+	Region     string
+	NodeType   string
+	UsageType  shared.UsageType
+	Family     string
+	DeviceType string
+	Type       NodePricingType
 }
 
 // @bingen:generate:NodePricing
--- a/core/pkg/model/pricingmodel/node.go
+++ b/core/pkg/model/pricingmodel/node.go
@@ -16,13 +16,13 @@ const (
 
 // @bingen:generate:NodeKey
 type NodeKey struct {
-	Provider   shared.Provider
-	Region     string
-	NodeType   string
-	UsageType  shared.UsageType
-	Family     string
-	DeviceType string
-	Type       NodePricingType
+	Provider    shared.Provider
+	PricingType NodePricingType
+	UsageType   shared.UsageType
+	Region      string
+	NodeType    string
+	Family      string
+	DeviceType  string
 }
 
 // @bingen:generate:NodePricing
--- a/core/pkg/model/pricingmodel/node.go
+++ b/core/pkg/model/pricingmodel/node.go
@@ -4,15 +4,27 @@ import (
 	"github.com/opencost/opencost/core/pkg/model/shared"
 )
 
+type NodePricingType string
+
+const (
+	NodePricingTypeTotal   NodePricingType = "Total"
+	NodePricingTypeCPUCore NodePricingType = "CPUCore"
+	NodePricingTypeRamGB   NodePricingType = "RamGB"
+	NodePricingTypeGPU     NodePricingType = "GPU"
+)
+
 // @bingen:generate:NodeKey
 type NodeKey struct {
-	Provider  shared.Provider
-	Region    string
-	NodeType  string
-	UsageType shared.UsageType
+	Provider    shared.Provider
+	Region      string
+	NodeType    string
+	UsageType   shared.UsageType
+	Family      string
+	Accelerator string
+	Type        NodePricingType
 }
 
 // @bingen:generate:NodePricing
 type NodePricing struct {
-	TotalHourlyRate float64
+	HourlyRate float64
 }
--- a/core/pkg/model/pricingmodel/pricingmodel_codecs.go
+++ b/core/pkg/model/pricingmodel/pricingmodel_codecs.go
@@ -13,9 +13,9 @@ package pricingmodel
 
 import (
 	"fmt"
+	"github.com/opencost/opencost/core/pkg/model/shared"
 	util "github.com/opencost/opencost/core/pkg/util"
 	"reflect"
-	"github.com/opencost/opencost/core/pkg/model/shared"
 	"strings"
 	"sync"
 	"time"
@@ -309,10 +309,10 @@ func (target *NodeKey) MarshalBinaryWithContext(ctx *EncodingContext) (err error
 	}
 	// --- [begin][write][alias](NodePricingType) ---
 	if ctx.IsStringTable() {
-		g := ctx.Table.AddOrGet(string(target.Type))
+		g := ctx.Table.AddOrGet(string(target.PricingType))
 		buff.WriteInt(g) // write table index
 	} else {
-		buff.WriteString(string(target.Type)) // write string
+		buff.WriteString(string(target.PricingType)) // write string
 	}
 	// --- [end][write][alias](NodePricingType) ---
 
@@ -447,7 +447,7 @@ func (target *NodeKey) UnmarshalBinaryWithContext(ctx *DecodingContext) (err err
 	w := x
 	u = w
 
-	target.Type = NodePricingType(u)
+	target.PricingType = NodePricingType(u)
 	// --- [end][read][alias](NodePricingType) ---
 
 	return nil
--- a/core/pkg/model/pricingmodel/pricingmodel_codecs.go
+++ b/core/pkg/model/pricingmodel/pricingmodel_codecs.go
@@ -13,12 +13,13 @@ package pricingmodel
 
 import (
 	"fmt"
-	"github.com/opencost/opencost/core/pkg/model/shared"
-	util "github.com/opencost/opencost/core/pkg/util"
 	"reflect"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/opencost/opencost/core/pkg/model/shared"
+	"github.com/opencost/opencost/core/pkg/util"
 )
 
 const (
@@ -607,10 +608,10 @@ func (target *PricingModelSet) MarshalBinaryWithContext(ctx *EncodingContext) (e
 	// --- [end][write][reference](time.Time) ---
 
 	if ctx.IsStringTable() {
-		b := ctx.Table.AddOrGet(target.Source)
+		b := ctx.Table.AddOrGet(string(target.SourceType))
 		buff.WriteInt(b) // write table index
 	} else {
-		buff.WriteString(target.Source) // write string
+		buff.WriteString(string(target.SourceType)) // write string
 	}
 	if target.NodePricing == nil {
 		buff.WriteUInt8(uint8(0)) // write nil byte
@@ -639,6 +640,12 @@ func (target *PricingModelSet) MarshalBinaryWithContext(ctx *EncodingContext) (e
 		}
 		// --- [end][write][map](map[NodeKey]NodePricing) ---
 
+	}
+	if ctx.IsStringTable() {
+		i := ctx.Table.AddOrGet(target.SourceKey)
+		buff.WriteInt(i) // write table index
+	} else {
+		buff.WriteString(target.SourceKey) // write string
 	}
 	return nil
 }
@@ -716,7 +723,7 @@ func (target *PricingModelSet) UnmarshalBinaryWithContext(ctx *DecodingContext) 
 		e = buff.ReadString() // read string
 	}
 	d := e
-	target.Source = d
+	target.SourceType = PricingSourceType(d)
 
 	if buff.ReadUInt8() == uint8(0) {
 		target.NodePricing = nil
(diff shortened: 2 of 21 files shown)


core/pkg/model/pricingmodel/node.go
```diff
@@ -4,15 +4,27 @@ import (
 	"github.com/opencost/opencost/core/pkg/model/shared"
 )
 
+type NodePricingType string
+
+const (
+	NodePricingTypeTotal   NodePricingType = "Total"
+	NodePricingTypeCPUCore NodePricingType = "CPUCore"
+	NodePricingTypeRamGB   NodePricingType = "RamGB"
+	NodePricingTypeGPU     NodePricingType = "GPU"
```
aaliomer-ibm: This should be be `Device`. Not a big fan of explicitly mentioning GPU. we need to generalize the concept of a device cost and not attach it explicitly to GPU (allow TPU, FPGA, Special networking, etc...) 
Ideally we should eliminate all GPU references from the code base and reference a device instead

ameijer:
@Sean-Holcomb can you fill out pr template

core/pkg/model/pricingmodel/node.go
peatey: +1 to @aaliomer-ibm's comment — `NodePricingTypeGPU` and the `Accelerator` field name both hard-code a GPU assumption. Renaming to something like `NodePricingTypeDevice` / `Device` would future-proof this for TPUs, FPGAs, custom ASICs, etc. The GCP Billing API already has non-GPU accelerator SKUs (e.g. TPU), so this isn't hypothetical.

---

core/pkg/model/pricingmodel/pricingmodel_codecs.go
```diff
@@ -289,6 +289,24 @@ func (target *NodeKey) MarshalBinaryWithContext(ctx *EncodingContext) (err error
...
+		buff.WriteInt(f) // write table index
+	} else {
+		buff.WriteString(target.Accelerator) // write string
+	}
+	if ctx.IsStringTable() {
+		g := ctx.Table.AddOrGet(string(target.Type))
+		buff.WriteInt(g) // write table index
+	} else {
+		buff.WriteString(string(target.Type)) // write string
+	}
```
peatey, line 319: Binary compatibility concern: the codec writes/reads the three new fields (`Family`, `Accelerator`, `Type`) unconditionally at the same `DefaultCodecVersion`. Any existing cached `PricingModelSet` files (e.g. from the AWS source) were written without these fields. When those cached files are loaded, `UnmarshalBinaryWithContext` will attempt to read past the end of the buffer and either panic or return corrupt data.

Consider either:
1. Bumping the codec version and adding a version check that handles the old format (skip the new fields if version < N), or
2. Invalidating existing caches on upgrade (e.g. by changing the cache file name/path).

---

pkg/cloud/gcp/billingpricingsource.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return url
+}
+
+type gcpBillingPage struct {
+	SKUs          []*GCPPricing `json:"skus"`
+	NextPageToken string        `json:"nextPageToken"`
+}
+
+func (g *GCPBillingPricingSource) parsePage(body io.Reader, pms *pricingmodel.PricingModelSet) (string, error) {
```
peatey, line 125: The API key ends up in the URL as a raw query parameter. Two concerns:

1. **URL injection / encoding** — if `pageToken` or `CurrencyCode` contain special characters (e.g. `&`, `=`), the URL breaks. The Azure source was nicely refactored to use `url.Values{}` for proper encoding — this should follow the same pattern for consistency and correctness.

2. **Key leakage in logs/errors** — the `err` from `http.Get(url)` on line 68 may embed the full URL (including the API key) in the error string. Consider redacting or using a separate variable for the loggable URL.

```go
func (g *GCPBillingPricingSource) buildURL(pageToken string) string {
	params := url.Values{}
	params.Set("key", g.config.APIKey)
	if g.config.CurrencyCode != "" {
		params.Set("currencyCode", g.config.CurrencyCode)
	}
	if pageToken != "" {
		params.Set("pageToken", pageToken)
	}
	return gcpBillingBaseURL + "?" + params.Encode()
}
```

---

pkg/cloud/gcp/billingpricingsource.go
```diff
@@ -0,0 +1,228 @@
...
+
+	url := g.buildURL("")
+	pageCount := 0
+
+	for url != "" {
+		resp, err := http.Get(url)
+		if err != nil {
+			return nil, fmt.Errorf("GCPBillingPricingSource: GET: %w", err)
+		}
+
```
peatey, line 83: Unlike the AWS pricing source which has `loadFromCache`/`saveToCache`, the GCP source re-fetches the entire billing catalog on every call. The GCP Billing Catalog API for Compute can be 50+ pages and relatively slow. Consider adding a similar caching mechanism to avoid re-downloading on every refresh interval when the data hasn't changed.

At minimum, if caching is deferred to a follow-up, it would be worth adding a TODO/comment here noting the intent.

---

pkg/cloud/gcp/billingpricingsource.go
```diff
@@ -0,0 +1,228 @@
...
+	if strings.Contains(strings.ToUpper(sku.Description), "RAM") {
+		pricingType = pricingmodel.NodePricingTypeRamGB
+	}
+
+	for _, region := range sku.ServiceRegions {
+		key := pricingmodel.NodeKey{
+			Provider:  shared.ProviderGCP,
+			Region:    region,
+			Family:    family,
+			UsageType: usageType,
```
peatey: Using description string matching to classify CPU vs RAM is fragile. If a GCP SKU description doesn't contain "RAM" (e.g. future SKU naming changes), it would be silently classified as CPU. 

The ResourceGroup already encodes the resource type for newer families — for example `N2Standard` has separate SKUs for "Ram" and "Core" that can be distinguished by the `usageUnit` field in the pricing expression (e.g. `GiBy.h` for RAM vs `h` for vCPU). Would checking `PricingExpression.UsageUnit` be more reliable here?

---

pkg/cloud/gcp/billingpricingsource.go
```diff
@@ -0,0 +1,228 @@
...
+		return 0, false
+	}
+	rates := sku.PricingInfo[0].PricingExpression.TieredRates
+	if len(rates) == 0 {
+		return 0, false
+	}
+	last := rates[len(rates)-1]
+	units, err := strconv.Atoi(last.UnitPrice.Units)
+	if err != nil {
+		units = 0
```
peatey, line 246: Silently swallowing the `Atoi` error could mask bad data from the API. Even if `units` is typically "0", a non-numeric value here (e.g. from an API change) would go unnoticed. Consider at least logging a warning:

```go
units, err := strconv.Atoi(last.UnitPrice.Units)
if err != nil {
    log.Warnf("GCPBillingPricingSource: non-numeric unit price '%s', defaulting to 0", last.UnitPrice.Units)
    units = 0
}
```

---

pkg/pricingmodel/pipeline.go
```diff
@@ -72,6 +75,23 @@ func NewPipeline(store corestorage.Storage, cfg PipelineConfig) (*Pipeline, erro
 		p.addSource(src, rc)
 	}
 
+	if cfg.GCPRunnerConfig.Enabled {
+		if cfg.GCPRunnerConfig.APIKey == "" {
+			log.Warnf("GCP Billing Catelog Pricing Source is enabled but no api key was provided")
```
peatey: Typo: "Catelog" → "Catalog".

Also, when the API key is empty, the source is still created and added to the pipeline. Every request will fail with a 403 from GCP. It would be cleaner to skip adding the source entirely when there's no key:

```go
if cfg.GCPRunnerConfig.Enabled {
    if cfg.GCPRunnerConfig.APIKey == "" {
        log.Warnf("GCP Billing Catalog Pricing Source is enabled but no API key was provided; skipping")
    } else {
        src := gcp.NewGCPBillingPricingSource(...)
        // ...
        p.addSource(src, rc)
    }
}
```

---

pkg/cloud/aws/pricelistapi.go
```diff
@@ -213,3 +213,15 @@ type PriceListEC2PricePerUnit struct {
...
+// back to USD if the code is unrecognized or the field is empty.
+func (p PriceListEC2PricePerUnit) ForCurrency(code string) string {
+	switch strings.ToUpper(code) {
+	case "CNY":
+		if p.CNY != "" {
+			return p.CNY
+		}
+	}
+	return p.USD
+}
```
peatey, line 227: `ForCurrency` only handles USD and CNY — if a user configures `CurrencyCode: "EUR"` at the pipeline level, this silently falls back to USD without any indication. The old code was explicit about China regions; the new generalized config creates an expectation that arbitrary currencies work.

Consider logging a warning when the requested currency isn't available:

```go
func (p PriceListEC2PricePerUnit) ForCurrency(code string) string {
    switch strings.ToUpper(code) {
    case "USD", "":
        return p.USD
    case "CNY":
        if p.CNY != "" {
            return p.CNY
        }
    }
    // Requested currency not available — fall back to USD
    return p.USD
}
```

Also worth noting in the godoc that only USD and CNY are supported by the AWS pricing list API.

---

pkg/cloud/gcp/billingpricingsource.go
```diff
@@ -0,0 +1,228 @@
...
+	for url != "" {
+		resp, err := http.Get(url)
+		if err != nil {
+			return nil, fmt.Errorf("GCPBillingPricingSource: GET: %w", err)
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			body, _ := io.ReadAll(resp.Body)
+			resp.Body.Close()
+			return nil, fmt.Errorf("GCPBillingPricingSource: unexpected status %d on page %d: %s", resp.StatusCode, pageCount, string(body))
```
peatey, line 87: Bare `http.Get()` without a timeout or context. If the GCP API hangs or is unreachable, this will block the runner goroutine indefinitely. Consider using an `http.Client` with a timeout, or accepting a `context.Context` to allow cancellation:

```go
client := &http.Client{Timeout: 60 * time.Second}
resp, err := client.Get(url)
```

This is a pre-existing pattern in the AWS/Azure sources, but since this is new code, it would be good to start with the right pattern.

---

peatey (changes requested):
## Review Summary

This PR adds a GCP Billing Catalog API pricing source, extends the `NodeKey`/`NodePricing` model to support per-resource pricing (CPU/RAM/GPU), refactors currency handling to the pipeline level, and improves Azure URL building and error handling. The overall direction is solid — decomposing pricing into per-vCPU/per-GB-RAM rates for GCP is the right approach since GCP doesn't publish per-instance-type totals the way AWS/Azure do.

### Must-fix

1. **Binary codec backward compatibility** — Adding 3 new fields to `NodeKey` without bumping the codec version will cause deserialization failures on existing cached data. This needs a version bump or cache invalidation strategy.

2. **GCP source runs with empty API key** — The pipeline warns but still adds the source, causing guaranteed 403s on every refresh. Should skip the source when no key is configured.

3. **URL building in GCP source** — Uses raw string concatenation (inconsistent with the `url.Values` fix applied to Azure in this same PR). Page tokens from GCP can contain `=` and `+` characters that will break without proper encoding. Also risks leaking the API key in error messages.

### Should-fix

4. **`NodePricingTypeGPU` / `Accelerator` naming** — Agree with @aaliomer-ibm that this should use device-agnostic naming to support TPUs, FPGAs, etc.

5. **No caching for GCP** — AWS has `loadFromCache`/`saveToCache`; the GCP Billing Catalog is large (50+ pages). Re-fetching on every interval is expensive.

6. **CPU vs RAM classification via description string** — `strings.Contains(description, "RAM")` is fragile. The `PricingExpression.UsageUnit` field (`GiBy.h` for RAM, `h` for vCPU) would be more reliable.

7. **`ForCurrency` silently falls back to USD** — If someone configures `CurrencyCode: "EUR"` at the pipeline level, AWS will silently return USD prices with no indication. Should at least document the limitation or log a warning.

### Nits

8. **Typo** in `pipeline.go`: "Catelog" → "Catalog"

9. **No `http.Client` timeout** on the GCP source — bare `http.Get()` can hang indefinitely. Pre-existing in AWS/Azure but worth fixing in new code.

10. **Silent `strconv.Atoi` error** in `gcpExtractHourlyRate` — should at least log a warning.

### Missing

- **No tests** — `gcpNormalizeFamily`, `gcpUsageType`, `gcpExtractHourlyRate`, `ForCurrency`, and the overall `processSKU` logic all need unit test coverage.
- **PR description is empty** — please fill in the Description, Related Issues, User Impact, and Testing sections.

---

peatey:
obviously - OC side we're using AI to check things over

peatey:
also my bad my PR filter isn't meant to pick up feature branch PR only into develop

aaliomer-ibm (approved):

PR #3724, opened 2026-04-12
docs(examples): add Python API client example

vincent067:
Hi team! 👋

I've been exploring the OpenCost API for a small internal automation project and noticed there isn't a standalone Python example in the repo. I thought I'd contribute one back in case it helps other developers get started faster.

This PR adds a minimal, self-contained Python client example under `examples/python-api-client/` that demonstrates:

- Connecting to the OpenCost API
- Querying **namespace-level** cost allocation
- Querying **deployment-level** cost allocation and surfacing the top spenders

**What's included:**
- `opencost_client.py` — a tiny reusable `OpenCostClient` class with clear docstrings
- `README.md` — setup instructions and expected sample output
- `requirements.txt` — just `requests`

**Why this matters:**
For teams building Python-based reporting or automation around OpenCost, having a starter example lowers the barrier to entry. I know the API docs are solid, but sometimes a working script is the fastest way to get going 😊

I'm happy to adjust the style, add more endpoints (e.g. `/assets` or `/cloudCost`), or move the files to a different location if you prefer. Just let me know!

Thanks for maintaining such a useful project 🙏


examples/python-api-client/opencost_client.py
```diff
@@ -0,0 +1,141 @@
...
+
+Usage:
+    export OPENCOST_URL="http://localhost:9003"
+    python opencost_client.py
+"""
+
+import os
+import sys
+from datetime import datetime, timedelta
+from urllib.parse import urljoin
```
peatey, line 19: Unused imports. `datetime` and `timedelta` are never referenced anywhere in the file.

```suggestion
from urllib.parse import urljoin
```

---

examples/python-api-client/opencost_client.py
```diff
@@ -0,0 +1,141 @@
...
+        Returns:
+            Parsed JSON response from the OpenCost API
+        """
+        params = {"window": window}
+        return self._get("assets", params=params)
+
+
+def fmt_cost(cost: float) -> str:
+    """Format a cost value as currency."""
+    return f"${cost:,.4f}"
```
peatey, line 79: This will crash with a `TypeError` when the API returns `null` for a cost field.

The OpenCost API serializes cost fields as `*float64` pointers (see `core/pkg/opencost/allocation_json.go`). When a value is `NaN` or `Inf`, `formatFloat64ForResponse` returns `nil`, which becomes JSON `null`. In Python, `metrics.get("cpuCost", 0.0)` will still return `None` when the key *exists* with a `null` value — `.get()` only falls back to the default when the key is **missing**, not when the value is `None`.

So the `None` will propagate into `fmt_cost(None)` and blow up on the f-string format spec.

```suggestion
def fmt_cost(cost: float) -> str:
    """Format a cost value as currency."""
    if cost is None:
        return "    n/a"
    return f"${cost:,.4f}"
```

---

examples/python-api-client/opencost_client.py
```diff
@@ -0,0 +1,141 @@
...
+        for window_data in dep_data:
+            for deployment, metrics in window_data.items():
+                total = metrics.get("totalCost", 0.0)
+                deployments.append((deployment, total))
+
+        deployments.sort(key=lambda x: x[1], reverse=True)
+        for deployment, total in deployments[:10]:
+            print(f"{deployment:<40} {fmt_cost(total):>12}")
+    except requests.RequestException as exc:
+        print(f"\n❌ Failed to fetch deployment data: {exc}")
```
peatey, line 134: Inconsistent error handling: the namespace allocation failure above (line 108) returns exit code `1`, but this deployment failure silently continues to print "Done!" and exit `0`. This would be confusing if someone uses the script in automation.

Either both should be fatal, or both should be non-fatal with a final aggregated exit code. For a simple example script, making both fatal is the easiest fix:

```suggestion
        print(f"\n❌ Failed to fetch deployment data: {exc}")
        return 1
```

---

examples/python-api-client/opencost_client.py
```diff
@@ -0,0 +1,141 @@
...
+
+        Args:
+            window: Time window (e.g. '1d', '7d')
+
+        Returns:
+            Parsed JSON response from the OpenCost API
+        """
+        params = {"window": window}
+        return self._get("assets", params=params)
+
```
peatey, line 75: `get_assets` is defined but never called in `main()`. For a minimal example, I'd remove it to avoid dead code — or add a short demo of it in `main()` to show it working.

---

examples/python-api-client/README.md
```diff
@@ -0,0 +1,65 @@
...
+__idle__                              $0.0234     $0.0156     $0.0390
+kube-system                           $0.0123     $0.0089     $0.0212
+default                               $0.0056     $0.0034     $0.0090
+
+📦 Top Deployments by Cost (Last 24h)
+Deployment                                   Total Cost
+-------------------------------------------------------
+kube-system/coredns                        $0.0034
+default/nginx-deployment                   $0.0021
+
```
peatey, line 58: The sample output doesn't match what the script would actually produce.

When aggregating by just `"deployment"`, the allocation key is only the controller name (e.g., `coredns`), **not** `namespace/deployment`. I verified this in the codebase — `allocationprops.go:389-395,551` builds the key from only the aggregation dimensions requested, then joins with `/`. With a single dimension, there's no prefix.

To get `kube-system/coredns` you'd need `aggregate="namespace,deployment"`. Either:
- Update the script to use `aggregate="namespace,deployment"`, or
- Fix the sample output to show just `coredns`, `nginx-deployment`, etc.

```suggestion
kube-system/coredns                        $0.0034
default/nginx-deployment                   $0.0021
```

(I'd lean toward option 1 since `namespace/deployment` is more useful information to display.)

---

examples/python-api-client/README.md
```diff
@@ -0,0 +1,65 @@
...
+✅ Done!
+```
+
+## Next Steps
+
+- See the full [OpenCost API documentation](https://www.opencost.io/docs/api) for additional endpoints such as `/assets`, `/cloudCost`, and `/aggregator/allocation`.
```
peatey, line 64: `/aggregator/allocation` is not a route served by this codebase. The actual registered routes (in `pkg/costmodel/router.go` and `pkg/cmd/costmodel/costmodel.go`) are:

- `/allocation` and `/allocation/summary`
- `/allocation/compute` and `/allocation/compute/summary`
- `/assets`
- `/cloudCost`

```suggestion
- See the full [OpenCost API documentation](https://www.opencost.io/docs/api) for additional endpoints such as `/allocation/compute/summary`, `/assets`, and `/cloudCost`.
```

---

peatey (changes requested):
Thanks for the contribution — having a standalone Python example is a nice idea for lowering the onboarding bar. The code is clean and well-structured overall. There are a few issues to address before this can merge though:

**Correctness issues:**

1. **`fmt_cost` will crash on `null` API values.** The OpenCost API serializes cost fields as `*float64` pointers. NaN/Inf values become JSON `null`, and Python's `.get("cpuCost", 0.0)` does NOT fall back to `0.0` when the key exists with a `None` value — it only falls back when the key is missing entirely. So `fmt_cost(None)` will `TypeError`. See inline comment for details.

2. **README sample output doesn't match actual API behavior.** Aggregating by `"deployment"` produces keys like `coredns`, not `kube-system/coredns`. The `namespace/` prefix only appears when you aggregate by `"namespace,deployment"`. Either update the script's aggregate parameter or fix the sample output.

3. **README references a non-existent endpoint** — `/aggregator/allocation` is not a route in this codebase.

**Code quality:**

4. **Unused imports** — `datetime` and `timedelta` are imported but never used.
5. **Inconsistent error handling** — namespace query failure returns exit 1, but deployment query failure prints an error and still exits 0 with "Done!".
6. **Dead code** — `get_assets()` is defined but never demonstrated.

**Process / project-level:**

7. **DCO check is failing** — the commit needs a `Signed-off-by` header (required by the project).
8. **New top-level `examples/` directory** — this doesn't exist yet in the repo. Maintainers should confirm this is the desired location for community examples.
9. **No `.gitignore`** — the README instructs users to create a `.venv` inside the example directory, which could be accidentally committed without a `.gitignore`.

---

ameijer:
@vincent067 we like this code, but can you move it to our docs website under maybe an examples section or something?

PR #3738, opened 2026-04-17
Hostnetwork network sharing

aaliomer-ibm:
Summary
Share host-network pod network usage across host-networked pods on the same node instead of attributing full node traffic to each pod.

Changes
add host-network metadata to cached pod specs and scraped pod info
propagate host-network state through network metric decoding
update allocation network totals logic to divide host-network RX/TX across host-networked pods per measurement
add simple unit tests covering normal, shared host-network, and mixed host/non-host-network cases
Validation
go test ./pkg/costmodel

### DIFF
--- a/core/pkg/source/decoders.go
+++ b/core/pkg/source/decoders.go
@@ -634,6 +634,7 @@ type PodInfoResult struct {
 	Pod          string
 	NamespaceUID string
 	NodeUID      string
+	HostNetwork  bool
 }
 
 func DecodePodInfoResult(result *QueryResult) *PodInfoResult {
@@ -642,6 +643,7 @@ func DecodePodInfoResult(result *QueryResult) *PodInfoResult {
 	pod, _ := result.GetPod()
 	namespaceUID, _ := result.GetString(NamespaceUIDLabel)
 	nodeUID, _ := result.GetString(NodeUIDLabel)
+	hostNetwork, _ := result.GetString("host_network")
 
 	return &PodInfoResult{
 		UID:          uid,
@@ -649,6 +651,7 @@ func DecodePodInfoResult(result *QueryResult) *PodInfoResult {
 		Pod:          pod,
 		NamespaceUID: namespaceUID,
 		NodeUID:      nodeUID,
+		HostNetwork:  hostNetwork == "true",
 	}
 }
 
@@ -1385,12 +1388,13 @@ func DecodeNetNatGatewayIngressGiBResult(result *QueryResult) *NetNatGatewayIngr
 }
 
 type NetReceiveBytesResult struct {
-	UID       string
-	Cluster   string
-	Namespace string
-	Pod       string
-	Container string
-	Data      []*util.Vector
+	UID         string
+	Cluster     string
+	Namespace   string
+	Pod         string
+	Container   string
+	HostNetwork bool
+	Data        []*util.Vector
 }
 
 func DecodeNetReceiveBytesResult(result *QueryResult) *NetReceiveBytesResult {
@@ -1399,23 +1403,26 @@ func DecodeNetReceiveBytesResult(result *QueryResult) *NetReceiveBytesResult {
 	namespace, _ := result.GetNamespace()
 	pod, _ := result.GetPod()
 	container, _ := result.GetContainer()
+	hostNetwork, _ := result.GetString("host_network")
 
 	return &NetReceiveBytesResult{
-		UID:       uid,
-		Cluster:   cluster,
-		Namespace: namespace,
-		Pod:       pod,
-		Container: container,
-		Data:      result.Values,
+		UID:         uid,
+		Cluster:     cluster,
+		Namespace:   namespace,
+		Pod:         pod,
+		Container:   container,
+		HostNetwork: hostNetwork == "true",
+		Data:        result.Values,
 	}
 }
 
 type NetTransferBytesResult struct {
-	UID       string
-	Cluster   string
-	Namespace string
-	Pod       string
-	Container string
+	UID         string
+	Cluster     string
+	Namespace   string
+	Pod         string
+	Container   string
+	HostNetwork bool
 
 	Data []*util.Vector
 }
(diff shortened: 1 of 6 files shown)


modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -381,6 +382,7 @@ func (ccs *ClusterCacheScraper) scrapePods(
 			source.PodLabel:          pod.Name,
 			source.NamespaceUIDLabel: string(nsUID),
 			source.NodeUIDLabel:      string(nodeUID),
+			"host_network":           strconv.FormatBool(pod.Spec.HostNetwork),
```
ameijer, line 385: should this have a const? kinda sticks out like this

core/pkg/source/decoders.go
```diff
@@ -642,13 +643,15 @@ func DecodePodInfoResult(result *QueryResult) *PodInfoResult {
 	pod, _ := result.GetPod()
 	namespaceUID, _ := result.GetString(NamespaceUIDLabel)
 	nodeUID, _ := result.GetString(NodeUIDLabel)
+	hostNetwork, _ := result.GetString("host_network")
```
Sean-Holcomb, line 646: can we add a constant for "host_network"

core/pkg/source/decoders.go
```diff
@@ -1399,23 +1403,26 @@ func DecodeNetReceiveBytesResult(result *QueryResult) *NetReceiveBytesResult {
...
-	Cluster   string
-	Namespace string
-	Pod       string
-	Container string
+	UID         string
+	Cluster     string
+	Namespace   string
+	Pod         string
+	Container   string
+	HostNetwork bool
```
Sean-Holcomb, line 1425: This fields isn't being populated. I also don't think you need it here

pkg/costmodel/allocation_helpers.go
```diff
@@ -915,67 +915,85 @@ func applyGPUsAllocated(podMap map[podKey]*pod, resGPUsRequested []*source.GPUsR
...
-			}
+	return pods
+}
+
+func getHostNetworkPodCountByNode(resPodInfo []*source.PodInfoResult) map[string]int {
+	hostNetworkPodCountByNode := map[string]int{}
+
+	for _, res := range resPodInfo {
+		if res.HostNetwork {
+			hostNetworkPodCountByNode[res.Cluster+","+res.NodeUID]++
```
Sean-Holcomb, line 992: NodeUID is a universal identifier. So you should just be able to key on that

pkg/costmodel/allocation_helpers.go
```diff
@@ -915,67 +915,85 @@ func applyGPUsAllocated(podMap map[podKey]*pod, resGPUsRequested []*source.GPUsR
...
-					thisPod, ok = podMap[uidKey]
-					if ok {
-						pods = append(pods, thisPod)
-					}
-				}
-			} else {
-				continue
+		divisor := float64(len(pods))
+		if hostNetwork {
+			nodePodCount := hostNetworkPodCountByNode[podKey.Cluster+","+podKey.Namespace]
```
Sean-Holcomb, line 929: The key here seems to be incorrect

pkg/costmodel/allocation_helpers.go
```diff
@@ -915,67 +915,85 @@ func applyGPUsAllocated(podMap map[podKey]*pod, resGPUsRequested []*source.GPUsR
...
 	}
+
+	for _, res := range resNetworkTransferBytes {
+		podKey, err := newResultPodKey(res.Cluster, res.Namespace, res.Pod)
+		if err != nil {
+			log.DedupedWarningf(10, "CostModel.ComputeAllocation: Network Transfer Bytes query result missing field: %s", err)
+			continue
+		}
+
+		applyNetworkMetric(podKey, res.Data[0].Value, res.HostNetwork, func(alloc *opencost.Allocation, value float64) {
```
Sean-Holcomb, line 951: HostNetwork wont be populated here

pkg/costmodel/allocation_helpers.go
```diff
@@ -915,67 +915,85 @@ func applyGPUsAllocated(podMap map[podKey]*pod, resGPUsRequested []*source.GPUsR
...
+
 	for _, res := range resNetworkReceiveBytes {
 		podKey, err := newResultPodKey(res.Cluster, res.Namespace, res.Pod)
 		if err != nil {
 			log.DedupedWarningf(10, "CostModel.ComputeAllocation: Network Receive Bytes query result missing field: %s", err)
 			continue
 		}
 
-		var pods []*pod
+		applyNetworkMetric(podKey, res.Data[0].Value, res.HostNetwork, func(alloc *opencost.Allocation, value float64) {
```
Sean-Holcomb, line 963: HostNetwork wont be populated here

Sean-Holcomb (changes requested):
There are a few things here.
1. There isn't anything here for the prom side of things metric emission or query.
2. The host network label will make it through on pod_info but not on the other metrics. The reason is that pod_info is a `info` type collector which aggregates on a label set but also includes the entire label set of the final update that the aggregator receives. the network tx and rx are `increase` aggregators and will only include their aggregate labels. I don't think we even need this label on tx and rx because the pods with host on them are identified by pod info.
3. The last one is a bit of a mess but I will do my best to explain it. The metrics that these queries are based off of are actually container level metrics. For some reason unknown to me (I have it on my personal todo list to look into this) we aggregate these numbers by pod then split them back out by container. So that just seems wrong to me, but in the context of this bug it raises even more questions.

If the base metrics are container based, are they having the same issue being reported to cldy for this host pods?

If the answer is yes, we are still looking at numbers that are being reported by container rather than by pod, so is it the case that we should be dividing the value by container count?

A general question: When hostnetwork is true, is the expectation that all pods on that node, with hostnetwork on, come back with the same value which is the sum of each of their individual traffics?

mbolt35:
I just want to weigh in a bit here and note that these specific metrics are not particularly useful for pricing network transfer. They include ALL socket transfers including anything over local host. These values have only historically been used on the kubecost side in determining whether something is an abandoned workload or not. We should always defer to using network-costs metrics for price or transfer accounting by pod.

mbolt35:
@Sean-Holcomb ```A general question: When hostnetwork is true, is the expectation that all pods on that node, with hostnetwork on, come back with the same value which is the sum of each of their individual traffics?```

This is ultimately why it's much easier to do this by IP address, which is what network-costs does. I still have no idea why we are continuing to use these somewhat unhelpful metrics. Is it because cloudy won't use the network costs daemonset? What's the deal?

PR #3741, opened 2026-04-20
Add STACKIT as a cloud provider

xtavras:
## Description

Add [STACKIT](https://stackit.com/) cloud provider support for OpenCost, enabling real-time cost allocation and cloud cost monitoring for STACKIT Kubernetes Engine (SKE) clusters.

**Provider detection:**
- `CLOUD_PROVIDER=STACKIT` environment variable
- Node ProviderID containing `stackit`
- Node label `node.stackit.cloud/ske` (SKE clusters)

**Cloud cost integration (Cost API):**
- Ingests actual billing data per service/project
- Requires service account with cost reader permissions
- Configured via `cloud-integration.json`

## Related Issues

N/A - new provider contribution

## User Impact

No breaking changes. Adds a new cloud provider option. Users on STACKIT can set `CLOUD_PROVIDER=STACKIT` or rely on auto-detection via node labels/ProviderID.

## Testing

- Manual deployment to SKE cluster, validated node pricing via `/allNodePricing` and `/pricingSourceStatus`
- Verified cloud cost integration with STACKIT Cost API and service account
- `go fmt ./...` and `go vet ./...` pass
- Unit tests


xtavras:
Hi, would be great if someone can look into PR, I already chatted with @rajith77 at KubeCon couple a weeks ago about contributing, the build is in use internally without issues during last 3 weeks.

ameijer:
@xtavras can you submit a PR to the website documenting the support and how to configure? we will review this

xtavras:
HI @ameijer, I've created PR for docs here https://github.com/opencost/opencost-website/pull/498 
Thanks for looking into it!

pkg/cloud/stackit/costconfiguration.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/opencost/opencost/core/pkg/opencost"
+	"github.com/opencost/opencost/pkg/cloud"
+)
+
+// CostConfiguration holds the configuration needed to connect to STACKIT Cost API
+type CostConfiguration struct {
+	CustomerAccountID     string `json:"customerAccountId"`
+	ProjectID             string `json:"projectId"`
+	ServiceAccountKeyPath string `json:"serviceAccountKeyPath,omitempty"`
```
Sean-Holcomb, line 15: Is this the only way that any user would want to authorize for this API.
It could be good to look at the authorizer interfaces in the other configurations

pkg/cloud/stackit/costintegration.go
```diff
@@ -0,0 +1,210 @@
...
+				Category:        category,
+				ProviderID:      sku,
+				Labels:          opencost.CloudCostLabels{},
+			}
+
+			cc := &opencost.CloudCost{
+				Properties: properties,
+				Window:     opencost.NewWindow(&periodStart, &periodEnd),
+				ListCost: opencost.CostMetric{
+					Cost: charge + discount,
```
Sean-Holcomb: Discount should only be included only in "net" Cost types

Sean-Holcomb (commented):
I like the Provider proper, I do have concerns about the CloudCost implementation. Particularly is this an appropriate API to be supplying CloudClosts. It seems like it is providing costs at the wrong level of granularity.

pkg/cloud/stackit/costintegration.go
```diff
@@ -0,0 +1,210 @@
...
+			periodStart, periodEnd := parsePeriod(tp.GetStart(), tp.GetEnd(), start, end)
+
+			properties := &opencost.CloudCostProperties{
+				Provider:        opencost.STACKITProvider,
+				AccountID:       ci.CustomerAccountID,
+				InvoiceEntityID: ci.CustomerAccountID,
+				RegionID:        "eu01",
+				Service:         serviceName,
+				Category:        category,
+				ProviderID:      sku,
```
ameijer, line 132: does sku == provider ID?

pkg/cloud/stackit/costconfiguration.go
xtavras: Hi @Sean-Holcomb, service account key file is standard authentication method, Token based auth was deprecated some time ago, there is new "workload identity federation" support as well, but not sure if it make sense to use in OpenCost case, need to play around with it, as I know it only for use in short CI pipelines. Is it fine to add support for WIF later in separate PR?

pkg/cloud/stackit/costintegration.go
xtavras: Yes, STACKIT Cost API returns a service type identifier in the sku field (e.g. ST-0009101). It's the most granular resource identifier available from this API, so we use it as ProviderID. Does it make sense?

pkg/cloud/stackit/costintegration.go
xtavras: Good call, fixed. `AmortizedCost` now uses the gross amount, consistent with `ListCost`. STACKIT doesn't have a commitment/reservation model yet (but it's planned and in progress AFAIK), so there's no true amortization.

xtavras:
Hi @Sean-Holcomb , @ameijer thank you for feedback, I've tried to comment on open question, added more tests (maybe trigger SonarQube againa) and fixed some CoPliot checks, please have a look, thanks!

pkg/cloud/stackit/pim.go
```diff
@@ -0,0 +1,321 @@
...
+			return nil, fmt.Errorf("decoding PIM response: %w", err)
+		}
+
+		allSKUs = append(allSKUs, searchResp.Data...)
+
+		if searchResp.Meta.NextCursor == "" || len(searchResp.Data) == 0 || len(searchResp.Data) < pimMaxPage {
+			break
+		}
+		cursor = searchResp.Meta.NextCursor
+	}
```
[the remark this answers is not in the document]
xtavras: PIM API returns a non-empty cursor even after all data is exhausted. Without this check, pagination makes hundreds of unnecessary requests. We verified that API always returns full pages until the last one, so a partial page reliably indicates the end of results

ameijer (approved):

xtavras:
Hi @ameijer Sorry, I force-pushed after rebasing onto develop. Could you re-enable auto-merge or merge manually?

NilsGriebner:
This would be super helpful for us.

PR #3755, opened 2026-04-22
Replace local storage cost queries with equivalent basic math

nikovacevic:
## Description
Local Storage Cost is currently populated by queries that contain basic math that can be done with other queries that we already make. This does that.

## Related Issues
Closes [KCM-5647](https://apptio.atlassian.net/browse/KCM-5647)

## User Impact
N/A

## Testing
In progress. Will update unit tests and will leave notes on manual testing results.


mbolt35 (approved):
Ship it!

mbolt35 (approved):
This looks good to me - I applied a divide by zero check suggested by copilot, but everything else looks ok!

PR #3764, opened 2026-04-28
implement autocomplete

ameijer:
## Description

implements autocomplete for assets, allocations, cloud costs


pkg/allocation/autocompletequeryservice.go
```diff
@@ -0,0 +1,172 @@
...
+
+	f := strings.ToLower(field)
+	switch f {
+	case "cluster", "namespace", "node", "controllerkind", "controllername", "pod", "container", "label", "namespacelabel":
+		return f, nil
+	}
+
+	if strings.HasPrefix(f, "label:") {
+		_, labelKey, _ := strings.Cut(f, ":")
+		return "label:" + labelKey, nil
```
mbolt35, line 108: you could do: 

```go
if labelKey, found := strings.CutPrefix(f, "label:"); found {
    return "label:" + labelKey, nil 
}
```

but this is also clear, so no complaints :)

pkg/allocation/autocompletequeryservice.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	case field == "namespacelabel":
+		return mapKeys(props.NamespaceLabels)
+	case strings.HasPrefix(strings.ToLower(field), "namespacelabel:"):
+		label := strings.TrimPrefix(field, "namespacelabel:")
+		if v, ok := props.NamespaceLabels[label]; ok {
+			return []string{v}
+		}
+	}
+	return nil
```
mbolt35, line 149: Was gonna suggest converting the field into an `AllocationProperty` type: https://github.com/opencost/opencost/blob/develop/core/pkg/opencost/allocationprops.go

since it has some helper functions for testing labels and annotations, but there appears to be a weird issue where not all of the AllocationProperty constants are actually of that type, so I think just keep as is? 

Also, it doesn't appear that namespace labels are supported in the AllocationProperty type.

pkg/cloudcost/repositoryquerier.go
```diff
@@ -2,13 +2,23 @@ package cloudcost
...
 	"github.com/opencost/opencost/core/pkg/opencost"
 )
 
+// ErrAutocompleteBadRequest indicates a client error in an autocomplete request.
+var ErrAutocompleteBadRequest = errors.New("autocomplete bad request")
+
+// IsAutocompleteBadRequest reports whether err is a client validation error.
+func IsAutocompleteBadRequest(err error) bool {
+	return errors.Is(err, ErrAutocompleteBadRequest)
+}
```
mbolt35, line 20: Big fan of this pattern, good stuff.

mbolt35 (approved):
This changeset looks good. The _only_ note-worthy thing that I would consider looking into would be the use of `AllocationProperty` and `AssetProperty` (not sure if there is a `CloudCostProperty`).  This may not even be the correct use case for those types, but there are some calls to things like `prom.SanitizeLabels(...)` as well as missing `namespacelabels:` handling. In short, if these types are applicable at all, we may want to consider making updates and leveraging them instead of replicating code here.

I'm still approving because I don't believe the above is necessary (may not even be a reasonable ask). 

Good stuff! Looks great.

PR #3789, opened 2026-05-18
Remove literal from description, which was failing the action

cpetersen5:
## Description

The literal in this description is causing the release Action to fail, as the comment is interpreted literally

## Related Issues

N/A

## User Impact

N/A


mbolt35 (approved):
LGTM!

Issue #2755, opened 2024-05-19
Can't get GCP billing data from EU region


SorsOps:
**Describe the bug**
Can't get GCP billing data for big query if not in location US

**To Reproduce**
1. Create a GCP account and export the billing info from the EU region
2. Attempt to read it 

**Expected behavior**
That it either should read the data or give me the ability to configure the region

**Screenshots**

```
64266Z INF CloudCost[<REDACTED>/<REDACTED>.<REDACTED>]: ingestor: building window [2024-05-01T00:00:00+0000, 2024-05-08T00:00:00+0000)
2024-05-19T19:05:58.753164267Z ERR CloudCost[<REDACTED>/<REDACTED>.<REDACTED>]: ingestor: build failed for window [2024-05-12T00:00:00+0000, 2024-05-19T00:00:00+0000): error retrieving CUD rates: GetFlexibleCUDRates: queryCUDAmountPayed: query error BigQueryQuerier: Query: error reading query results: googleapi: Error 404: Not found: Dataset<REDACTED>:d<REDACTED> was not found in location US, notFound
```

**Which version of OpenCost are you using?**

- ghcr.io/opencost/opencost-ui:1.110.0

mattray:
That looks like what I fixed in https://github.com/opencost/opencost/pull/2239

mattray:
This should be fixed in the next release

GaMartinelli:
Still seeing this issue on v1.111

Cluster is running in europe-west3 region. Not sure if something related to the region should be set during during the installation, just followed the steps in docs/configuration/gcp

`2024-09-20T02:31:35.326841108Z INF Starting cost-model version 1.111 (1dd2e52)
[...]
2024-09-20T02:31:47.013855318Z INF CloudCost[REDACTED/detailedbilling.gcp_billing_export_resource_v1_0121AC_C6F51B_690771]: ingestor: build[SLpFR]:  [2024-09-13T00:00:00+0000, 2024-09-20T00:00:00+0000) in 1.014658535s
2024-09-20T02:31:47.013881401Z INF CloudCost[REDACTED/detailedbilling.gcp_billing_export_resource_v1_0121AC_C6F51B_690771]: ingestor: building window [2024-09-06T00:00:00+0000, 2024-09-13T00:00:00+0000)
[...]

PR #3795, opened 2026-05-19
Add field `location` to `BigQueryConfiguration`

ioboi:
## Description

This PR adds the ability to configure the location for BigQuery.

## Related Issues

https://github.com/opencost/opencost/issues/2755 (not fixed)

## User Impact

This is not a breaking change. Users are now able to specify the default BigQuery location in their `cloud-integration.json` like:

```json
{
  "gcp": {
    "bigQuery": [
      {
        "projectID": "<GCP_PROJECT_ID>",
        "dataset": "detailedbilling",
        "table": "gcp_billing_export_resource_v1_0121AC_C6F51B_690771",
        "location": "EU"
        "authorizer": {
          "authorizerType": "GCPWorkloadIdentity"
        }
      }
    ]
  }
}
```

Before this change there was no way to specify the location and it was determined by the API like [this](https://docs.cloud.google.com/bigquery/docs/locations).

## Testing

Only local testing so far, but I can test the changes on a production environment if needed.

### DIFF
--- a/pkg/cloud/gcp/bigqueryconfiguration.go
+++ b/pkg/cloud/gcp/bigqueryconfiguration.go
@@ -77,6 +77,10 @@ func (bqc *BigQueryConfiguration) Equals(config cloud.Config) bool {
 		return false
 	}
 
+	if bqc.Location != thatConfig.Location {
+		return false
+	}
+
 	return true
 }
 
@@ -85,6 +89,7 @@ func (bqc *BigQueryConfiguration) Sanitize() cloud.Config {
 		ProjectID:  bqc.ProjectID,
 		Dataset:    bqc.Dataset,
 		Table:      bqc.Table,
+		Location:   bqc.Location,
 		Authorizer: bqc.Authorizer.Sanitize().(Authorizer),
 	}
 }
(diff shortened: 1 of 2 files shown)


ameijer:
@ioboi can you also update the GCP docs on the website?

ameijer (approved):

ioboi:
Website update: https://github.com/opencost/opencost-website/pull/510.

pkg/cloud/gcp/bigqueryconfiguration.go
```diff
@@ -106,7 +107,15 @@ func (bqc *BigQueryConfiguration) GetBigQueryClient(ctx context.Context) (*bigqu
...
 		return nil, err
 	}
-	return bigquery.NewClient(ctx, bqc.ProjectID, clientOpts...)
+
+	client, err := bigquery.NewClient(ctx, bqc.ProjectID, clientOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	client.Location = bqc.Location
```
ameijer, line 121: can we default to US just to guard existing behavior?

ioboi: I am not sure if it make sense because existing behavior is working like the official documentation describes: https://docs.cloud.google.com/bigquery/docs/locations

ameijer: ok, then should we consider simply not setting the variable if it is ""? just worried that setting an empty string will cause regressions

ioboi: Hmm. I don't think there is a difference, because before the location was never set and the zero value as well, see https://github.com/googleapis/google-cloud-go/blob/bigquery/v1.74.0/bigquery/bigquery.go#L81.

ameijer (approved):

PR #3805, opened 2026-05-20
Add external ID support for AWS assume role

Tomerfi1210:
## Summary
- Add optional externalID support to AWS AssumeRole authorizer
- Wire legacy Athena config externalID through custom pricing and conversion
- Add tests for externalID JSON handling and Athena conversion

## Test Plan
- go test ./pkg/cloud/aws


Awish021:
useful

PR #3824, opened 2026-05-29
move code into OC

ameijer:
## Description

moves method and makes some refactors to autocomplete code


Sean-Holcomb (approved):
In general looks good.

In terms of packaging it looks like a lot of the stuff in the pkg/... files are being imported by KCM. Can this be moved to `core/pkg/util/autocomplete/(allocation|asset|cloudcost)` or `core/pkg/(allocation|asset|cloudcost)`

Looking into the requests it seems like they are really all the same with AllocationAutoComplete having an optional label config and asset having TenantID (shouldn't they all have TenantID?). That makes me wonder, why 3 different request types, why 3 different normalize functions, why 3 different parsers

Issue #3533, opened 2026-01-12
Feature Request: Support LLM / GenAI Token Throughput & Cost-per-Token Metrics


AjayTripathy:
**Is your feature request related to a problem? Please describe.**
LLM and GenAI workloads consume expensive GPU resources but are currently measured only by cost ($/hour).   
 To understand *efficiency*, FinOps and platform teams need to link cost to *output*—tokens generated, trained, or evaluated. 

Example questions users can’t answer today: 

* How many tokens per second are we generating per GPU or pod?   
* What is the **cost per token** for inference vs. training?   
* Which model or phase gives the best efficiency per dollar? 

This feature would make OpenCost the central FinOps system for GenAI workloads by tracking **productivity per cost**, not just spend. 

Once implemented, users can visualize: 

* **Tokens/sec per GPU** to monitor utilization.   
* **Cost/token** to benchmark model efficiency.   
* **Phase-based breakdowns** (inference vs. training) for accurate budget allocation.   
* **Cluster or tenant summaries** of $/1M tokens. 

**Describe the solution you'd like**
*Accept external token & compute telemetry* 

Support ingestion or joining with standard OpenTelemetry GenAI metrics. There are no such metrics today, but I propose the following. We can edit these to reflect any future OTEL consensus 

| Metric  | Type  | Description  |
| ----- | ----- | ----- |
| llm\_tokens\_emitted\_total  | Counter  | Total tokens processed or emitted (input/output/label)  |
| llm\_gpu\_seconds\_total  | Counter  | GPU active seconds used for token generation/training  |
| llm\_cpu\_seconds\_total  | Counter  | Optional CPU usage counter  |

 

*Recognize key GenAI attributes for cost joins* 

| Attribute  | Example  | Purpose  |
| ----- | ----- | ----- |
| workflow.phase  | inference / training / finetune / eval  | Distinguish workload type  |
| gen\_ai.model.name  | gpt-5  | Attribute cost per model  |
| gen\_ai.model.version  | v1.3.2  | Version tracking  |
| tenant.id / cost\_center  | team-finops  | Multi-tenant allocation  |
| accelerator.type  | nvidia-h100  | GPU type for pricing  |
| gpu.uuid / mig.uuid  | GPU-32b0...  | Map tokens to physical device  |
| Kubernetes context  | cluster, namespace, pod, container  | Join key with OpenCost pod spend  |

*Update the agents to append and propagate gen-ai labels as attributes eg:* 

metadata: 

  labels: 

[...]

PR #3829, opened 2026-06-02
Design proposal for AI Inference APIs

simanadler:
Signed-off-by: Sima Nadler <sima@il.ibm.com>

## Description
Proposal for AI Inference Cost APIs 
Design document prior to implementation

## Related Issues
Relates to #3533 and [proposal](https://github.com/llm-d/llm-d/pull/1646)

## User Impact
This is not a breaking change and will not affect current OpenCost functionality.

### DIFF
--- a/docs/opencost-for-llm-d/inference-cost-api-design.md
+++ b/docs/opencost-for-llm-d/inference-cost-api-design.md
@@ -1,0 +1,745 @@
+# OpenCost Inference Cost API Design
+
+**Authors:** Sima Nadler (IBM)  
+**Status:** Design Proposal  
+**Based on:** [Inference Cost Proposal](https://github.com/llm-d/llm-d/pull/1646)
+
+## Executive Summary
+
+This document defines the REST APIs for OpenCost's new Inference Cost domain. Following the established patterns from OpenCost's [CloudCost](https://opencost.io/docs/integrations/api#cloudcost) and [CustomCost](https://opencost.io/docs/integrations/api#customcosttimeseries) APIs, this design introduces endpoints for querying AI inference costs with support for both **usage-based** and **allocation-based** cost metrics.
+
+### Core Metrics (Phase 1)
+
+We will focus initially on four primary metrics from the [proposal](https://github.com/llm-d/llm-d/pull/1646):
+
+1. **`llm_total_cost`** - Hourly cost per model
+2. **`llm_cost_per_million_tokens`** - Blended cost per 1M tokens
+3. **`llm_input_cost_per_million_tokens`** - Cost per 1M input tokens
+4. **`llm_output_cost_per_million_tokens`** - Cost per 1M output tokens
+
+Each metric is available in both **usage-based** and **allocation-based** variants.
+Additional metrics mentioned in the proposal will be added later.
+
+---
+
+## Design Principles
+
+### 1. Consistency with Existing OpenCost APIs
+
+Follow the established patterns from:
+- [`/cloudCost`](../../pkg/cloudcost/queryservice.go) - Query structure, filtering, aggregation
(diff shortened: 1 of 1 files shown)


docs/opencost-for-llm-d/inference-cost-api-design.md
```diff
@@ -0,0 +1,745 @@
...
+Follow the established patterns from:
+- [`/cloudCost`](../../pkg/cloudcost/queryservice.go) - Query structure, filtering, aggregation
+- [`/customCost`](../../pkg/customcost/queryservice.go) - Time series and totals patterns
+- [`/allocation`](../../pkg/costmodel/aggregation.go) - Window handling, accumulation
+
+### 2. Cost Basis Duality
+
+Support both cost perspectives as defined in the [proposal](https://github.com/llm-d/llm-d/pull/1646):
+
+**Usage-Based Cost:**
```
davidbreitgand, line 38: Do I understand correctly that calculating the usage-based cost aims at facilitating a fair comparison with SaaS, where you only pay-as-you-go for inference when you do the actual inference? So, you want to abstract away the infrastructure cost not related to _this_ inference and behavior of other workloads in the self-hosted cluster?

ameijer: @davidbreitgand the unsatisfying answer is "it depends". We have found that orgs have different policies when it comes to attributing cost. 

A contrived example: Let's say your LLM requests 10 GPUs (via DRA or plugin resource requests), consumed 5 input tokens, and produces 5 output tokens in a given period but only actually used 1 of those GPUs to do that work. Each GPU cost $1 during that time, and input and output tokens consume same resources. K8s holds those 10 GPUs for your app alone, and no other workloads.

Is your cost per token $1 (because you requested 10 GPUs that could not be used for other things to produce/consume your 10 tokens), or is it $0.10 (because you only actually used 1 of those GPUs to produce/consume the 10 tokens). What is 'fair' for comparison to a hosted models? The 10 GPUs that are consuming power/money and allocated to your app, or only the one you actually used? 

The answer will be different depending on accounting policy, company stage, engineering culture, and so on. If you want to encourage overprovisioning and only look at actual usage of the model, and don't care about waste (R&D, growth stage company, etc), then the usage based attribution is the best comparison. If you are trying to control costs, and minimize waste (finops cycles, production COGS control), then factoring in any requested but unused GPUs gives you a better comparison to a hosted model. 

All this is abstract in a hosted model (we have no idea how much of Anthropics GPUs are idle/wasted, nor do we care), but it is of great importance when we own the infrastructure. That's why in opencost we have a number of different cost attribution philosophies, and why we are proposing supporting both. 

I hope this helps

ameijer: To answer the second question you have, both allocation and usage based cost attribution ignores costs of workloads not related to the specific one you are looking at. If you do want to incorporate some set of overhead costs and other workloads to enrich your pricing model for a given LLM, there are mechanisms for that which are described in the sharing section

davidbreitgand: Thank you for the detailed answer, @ameijer .

docs/opencost-for-llm-d/inference-cost-api-design.md
```diff
@@ -0,0 +1,745 @@
...
+
+**Allocation-Based Cost:**
+- Measures full business cost (includes idle time, model availability)
+- Varies with utilization
+- Reconciles to infrastructure bills
+- Best for chargeback and total cost of ownership
+
+### 3. Multi-Dimensional Attribution
+
+Support aggregation and filtering by:
```
ev-shindin, line 52: In order to integrate this with WVA this should be attributed by Deployment/LWS name.

PR #3831, opened 2026-06-02
KCM 5801/3: AWS & Azure Pricing Module

alexsouthard:
## Description
Adds the AWS and Azure pricing module (because they're pretty similar) w/ Node/Volume support, as well as a fun little CLI extension that generates the json files as well as has a compare functionality. 🥳 

## Related Issues
https://apptio.atlassian.net/browse/KCM-5801
https://apptio.atlassian.net/browse/KCM-5803

## User Impact
Not really any user impact right now as it's the addition of the new pricing module

## Testing
Got some unit tests in with this (could probably stand to add more) and mostly just did the testing for it by running the CLI commands


modules/pricing/public/module.go
```diff
@@ -0,0 +1,199 @@
...
+type PricingModuleConfig struct {
+	BaseDir  string
+	Provider pricing.Provider
+	Currency unit.Currency
+}
+
+type PricingModule struct {
+	config     PricingModuleConfig
+	Providers  *ProviderPricing `json:"provider" yaml:"provider"`
+	pricingSet *pricing.PricingSet
```
nikovacevic, line 25: @alexsouthard do we have any sense of how big this `PricingSet` is?

I'm not opposed to holding it all in memory, this way, alongside the indexed version. Especially because, if I'm not mistaken, you're reusing the pointers from `pm.pricingSet` in the index, which is good.

The nice thing about holding the `PricingSet` in memory is that you can just use the `SliceReader` in `NewNodePricingReader` and `NewVolumePricingReader`. The irony, though, is that the `Reader` interface is meant to allow the possibility for reading buffered batches into a data sink. With the `SliceReader`, you're going to hold the whole slice in memory, anyways, on the upstream side. Which isn't a problem if it's not too big! But if it is too big, we'll have to think about a slight restructuring here.

modules/pricing/public/cmd/main.go
```diff
@@ -0,0 +1,143 @@
...
+		if err := os.WriteFile(outputPath, newData, 0644); err != nil {
+			return fmt.Errorf("failed to write output file: %w", err)
+		}
+		log.Infof("Wrote pricing data to %s", outputPath)
+		return nil
+	}
+
+	// Compute checksums
+	existingChecksum := computeChecksum(existingData)
+	newChecksum := computeChecksum(newData)
```
nikovacevic: Does the generator always canonicalize its output, so that computing checksums is deterministic?

pkg/cloud/aws/pricinglistpricingsource.go
```diff
@@ -43,43 +43,6 @@ func (p *PricingListPricingSource) cacheFilePath() (string, error) {
 	return filepath.Join(dir, pricingCacheFile), nil
 }
 
-func (p *PricingListPricingSource) loadFromCache() (*pricingmodel.PricingModelSet, bool) {
```
nikovacevic: No need to do this in your PR, but I think we can actually remove the _entire_ AWS `PricingListPricingSource` and Azure `RetailPricingSource` types from the `pkg/cloud/aws` and `pkg/cloud/azure` directories. Does that sound right @Sean-Holcomb ?

modules/pricing/public/azure/azurepricingsource.go
```diff
@@ -0,0 +1,309 @@
...
+// AzurePricingSourceConfig holds configuration for AzurePricingSource.
+type AzurePricingSourceConfig struct {
+	CurrencyCode string
+}
+
+var azureHTTPClient = &http.Client{Timeout: 60 * time.Second}
+
+// AzurePricingSource implements the PricingSource interface using the
+// Azure Retail Prices API (no auth required).
+type AzurePricingSource struct {
```
nikovacevic, line 33: Sean might prefer we keep the name `RetailPricingSource` in case there end up being other Azure pricing sources. Is that the idea @Sean-Holcomb ?

.github/workflows/pricing_update.yml
```diff
@@ -0,0 +1,38 @@
...
+      uses: actions/setup-go@v5
+      with:
+        go-version: 'stable'
+
+    - name: Run Pricing Generator
+      id: run_cli
+      run: |
+        cd modules/pricing/public
+        go run cmd/main.go fetch-pricing -c ${{ matrix.currency }} -p ${{ matrix.provider }} -x
+        # TODO: There should be more here, probably
```
nikovacevic: This is neat. I like the way you're scaffolding out the CLI and the GHA. I don't consider completing this TODO to be blocking for this PR, but let me know if you want to discuss what more we should do here!

modules/pricing/public/generator.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	log.Infof("Generated %d Azure node pricing entries", len(pricingSet.Nodes))
+	return pricingSet, nil
+}
+
+// GeneratePricingForProvider fetches pricing data for a specific provider
+// in the specified currency
+func GeneratePricingForProvider(provider pricing.Provider, currency unit.Currency) (*pricing.PricingSet, error) {
+	// NOTE: Could add an "all" flag/provider. Maybe it outputs a single frankensteined file for all providers
```
nikovacevic: I think that would be good. The `PricingSet` will allow multiple providers and multiple currencies. But whether we want that or not might depend on the ergonomics of the wider module. Things like: how do we generate the final module? how does the module get loaded into the application? etc.

nikovacevic (approved):

modules/pricing/public/azure/azurepricingsource.go
Sean-Holcomb: I don't think this is necessary anymore because they now live in a directory called public. Additional prcing sources would likely need to be configured and not live in here

modules/pricing/public/module.go
alexsouthard: Can definitely swap so the whole thing just isn't in memory but it was about 40MB for all of AWS

modules/pricing/public/cmd/main.go
alexsouthard: Updated it so it sorts it now, so it should be deterministic

.github/workflows/pricing_update.yml
alexsouthard: Swapped the implementation so it's a background process which runs on a timer instead of an action calling a cronjob. I can definitely add this back (because it took like 1 second) but there isn't the compare functionality of the CLI anymore (though I can probably just add that back too if we really want it)

modules/pricing/public/generator.go
alexsouthard: Added an 'all' provider which would just be Azure/Aws/GCP 👍

pkg/cloud/aws/pricinglistpricingsource.go
alexsouthard: Got a follow-up PR for GCP that I'm going to tear everything out in

Sean-Holcomb (approved):

PR #3836, opened 2026-06-03
Bingen 0.1.1 Streaming Writer Support

mbolt35:
## Description
* Regenerate bingen codecs with latest bingen 0.1.1 release that provides streaming writer support. 

## User Impact
* No direct impact on users yet, but we will slowly work towards enabling streaming writers in the exporters.


core/pkg/util/bufferhelper.go
```diff
@@ -491,3 +491,113 @@ func writeFloat64(w *bytes.Buffer, data float64) error {
...
+
+func writeBuffBool(w *bufio.Writer, data bool) error {
+	if data {
+		w.WriteByte(1)
+		return nil
+	}
+
+	w.WriteByte(0)
+	return nil
+}
```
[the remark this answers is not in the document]
ameijer: @mbolt35 this looks like a good idea?

core/pkg/util/bufferhelper.go
```diff
@@ -491,3 +491,113 @@ func writeFloat64(w *bytes.Buffer, data float64) error {
...
+
+func writeBuffInt8(w *bufio.Writer, data int8) error {
+	w.WriteByte(byte(data))
+	return nil
+}
+
+func writeBuffUint8(w *bufio.Writer, data uint8) error {
+	w.WriteByte(byte(data))
+	return nil
+}
```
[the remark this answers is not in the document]
ameijer: similar to above

core/pkg/util/buffer.go
```diff
@@ -147,216 +257,293 @@ func (b *Buffer) WriteString(i string) {
...
+		_ = b.w.Flush()
+	}
 }
 
 // this should be inlined
 func (b *Buffer) checkRO() {
-	if b.bw == nil {
-		panic("Buffer is set to read-only")
+	if b.IsReadOnly() {
+		panic("Tried to write to a Buffer that is set to read-only")
```
ameijer, line 325: we are OK to panic here? this would never happen after shipping right?

ameijer (approved):
LGTM

core/pkg/util/buffer.go
mbolt35: Right - copilot is just determining that "hey, something random could use this type this way and it would fail"

core/pkg/storage/s3storage.go
```diff
@@ -463,6 +462,32 @@ func (s3 *S3Storage) Write(name string, data []byte) error {
...
 }
 
+// Upload the contents of the reader as an object into the bucket. The returned `io.WriteCloser` must
+// be closed to finalize the write.
+func (s3 *S3Storage) WriteStream(name string) (io.WriteCloser, error) {
+	name = trimLeading(name)
+
+	log.Debugf("S3Storage::WriteStream::%s(%s)", s3.protocol(), name)
+
+	ctx := context.Background()
```
ameijer, line 472: kind of nitty, but I wonder if we should allow someone to pass a context through to this method so they can own timeouts etc

mbolt35: Definitely don't want to dismiss this as the proper way to handle async behavior, but it does also seem like we would want it across the board. I'll look into read/write contexts being externally passed. It's a little weird because not all storage impls are async but we can probably make something work

core/pkg/storage/memorystorage.go
```diff
@@ -132,6 +132,50 @@ func (ms *MemoryStorage) Write(path string, data []byte) error {
...
 
+// WriteStream creates a new relative path and returns the io.WriteCloser that can be used to
+// write into the storage path. Close() blocks until all data has been committed to storage.
+func (ms *MemoryStorage) WriteStream(path string) (io.WriteCloser, error) {
+	r, w := io.Pipe()
+	var wg sync.WaitGroup
+
+	wg.Go(func() {
+		data, err := io.ReadAll(r)
+		if err != nil {
```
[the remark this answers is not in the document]
mbolt35: False positive - WaitGroup does have the Go() func, and it's suggested by `go vet`

core/pkg/util/fileutil/writer.go
```diff
@@ -0,0 +1,42 @@
...
+
+	if err := LockFile(f); err != nil {
+		f.Close()
+		return nil, err
+	}
+
+	return &LockedFileWriter{
+		f: f,
+	}, nil
+}
```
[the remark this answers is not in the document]
mbolt35: This seems like a hallucination based on typical locking behavior, but unclear what the suggested approach is -- it's suggesting we need to ensure that we have a "lock" for a file descriptor that doesn't exist. 

The short description of the problem is that if I create two files simultaneously and write to one really fast, that the next call to os.Create() will truncate the file, then the original writer locks it. This seems _really_ unlikely, but maybe not impossible. Going to look further into it

core/pkg/exporter/encoder.go
```diff
@@ -101,17 +130,29 @@ func (gz *GZipEncoder[T]) Encode(data *T) ([]byte, error) {
...
+// binary encoding.
+func (gz *GZipEncoder[T]) EncodeTo(writer io.Writer, data *T) error {
+	gzWriter, err := gzip.NewWriterLevel(writer, gz.level)
+	if err != nil {
+		return fmt.Errorf("failed to create gzip writer: %w", err)
+	}
+	defer gzWriter.Close()
+
+	return gz.encoder.EncodeTo(gzWriter, data)
+}
```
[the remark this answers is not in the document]
ameijer: @mbolt35 this is kinda nit picky but might be worth accepting?

core/pkg/opencost/exporter/exporter_test.go
```diff
@@ -138,6 +139,10 @@ type UnknownSet struct{}
 func (u *UnknownSet) MarshalBinary() ([]byte, error) {
 	return []byte{}, nil
 }
+func (u *UnknownSet) MarshalBinaryTo(writer io.Writer) error {
+	return nil
+}
```
ameijer, line 144: 🤩

core/pkg/opencost/exporter/exporters.go
```diff
@@ -12,6 +13,28 @@ import (
 	"github.com/opencost/opencost/core/pkg/util/typeutil"
 )
 
+// ExportCompressionLevel is an enumeration value for allowing a streaming compute exporter to enable
+// compression at specific gzip levels.
+type ExportCompressionLevel int
```
ameijer, line 18: can we have an option to se this via env var?

core/pkg/storage/clusterstorage.go
```diff
@@ -339,6 +339,38 @@ func (c *ClusterStorage) Write(path string, data []byte) error {
...
+
+	fn := func(resp *http.Response) error {
+		return nil
+	}
+
+	args := map[string]string{
+		"path": path,
+	}
+
+	r, w := io.Pipe()
```
ameijer, line 353: does this need to get sized? can we write 15 gb to it and have that all go back to ram?

core/pkg/storage/clusterstorage_test.go
```diff
@@ -171,3 +172,114 @@ func TestClusterStorage_ReadStream(t *testing.T) {
...
+			}
+		}
+		if err = w.Close(); err != nil {
+			t.Fatalf("Close: %s", err)
+		}
+
+		if !bytes.Equal(received, want) {
+			t.Errorf("body mismatch: got %q, want %q", received, want)
+		}
+	})
```
ameijer, line 247: I think this test is good and puts a lot of my fears about cluster storage to rest. Should we consider running this with a non existent buffer or something? are you concerned about it just doing one write under the hood?

core/pkg/opencost/exporter/exporters.go
mbolt35: Seems like an implementation controlled env var (ie: finops agent). Kind of a code smell to add env var controlled vars to a core library

core/pkg/storage/clusterstorage.go
mbolt35: Are you saying you want it to block at a specific size? This is just a pass through pipe into the storage Put(). There is an internal buffer in the storage that is set to "auto" and I'm (pretty sure it blocks) that collects the data from the reader. 

If you're writing 15gb at once, that data is already in memory, unless I'm misunderstanding what you're saying.

core/pkg/storage/clusterstorage_test.go
mbolt35: One write is the worst case, and it's equivalent to exactly what we do without streaming. No matter what, there is an internal buffer in the storage copies that will throttle the writer. 

If you're asking if I am worried about writing 15GB copying all 15GB at once (30GB footprint), no I'm not. I think a few of the storage APIs allow you to set a custom chunk size for uploads, but I don't see any reason why we would need that - they're typically set to like 2-6mb

mbolt35: As it relates to cluster storage, I think it's a 4mb buffered io reader. I will look into that though.

core/pkg/exporter/encoder.go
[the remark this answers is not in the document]
mbolt35: Yeah, it's really obnoxious but I agree we should try to surface as many of these write errors as possible.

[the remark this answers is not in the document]
mbolt35: I don't know about accepting as is, but I'll look at it

core/pkg/storage/clusterstorage.go
ameijer: I guess what I am saying is that let's say your set is 1000 elements, taking up 1000 bytes of ram. Let's also say that the marshalled in-memory set is 500 byes. In the prior implementation, we would have 1500 bytes of memory used at a point in time (set is marshalled but not yet GC'd + original set). We are trying to have this take up less ram, so with streaming, we would hope that we would write, say 100 elements at a time, then GC them. Waht I am wondering about is, will the pipewriter simply accumulate all 1000 elements in the "stream" before sending them off? is that masking some lack of support for small chunked writes?

core/pkg/storage/clusterstorage_test.go
ameijer: 👍

core/pkg/storage/clusterstorage.go
mbolt35: The pipe itself is completely unbounded, but the entire reason we use a pipe is because storage puts accept an io.Reader. It's a push model similar to other go APIs. 

The pipe provides a mechanism for us to inverse the process so we can push data into the uploaders internal buffer (it will continuously read into a buffer of set size). 

I think the remaining question is whether or not writes over the pipe block if the internal upload buffer is full, which I am pretty sure it does.

PR #3845, opened 2026-06-11
AI Inference Costs v1

simanadler:
## Description
Implementation of AI Inference Cost metrics and APIs as described in the [proposal](https://github.com/llm-d/llm-d/pull/1646) and API design PR #3829 

Pending: 
* Integration with OpenCost UI
* Improved GPU efficiency 
* Workload and team-based attribution 
* (Optional) Optimization cost tracking and savings calculation (KV cache, prefill/decode, smart routing)
* Testing and validation framework

## Related Issues
Relates to https://github.com/opencost/opencost/issues/3533 

## User Impact
Adds new functionality without breaking existing capabilities.

## Testing
* Unit tests
* Manual and AI evaluation of following REST calls on cluster with 100+ GPUs and >30 AI models
```
# --- /inferenceCost/total ---

  # Last hour, all models (default allocation cost basis)
  curl -s "$BASE/inferenceCost/total?window=1h" | jq .

  # Last 24 hours, usage cost basis
  curl -s "$BASE/inferenceCost/total?window=24h&costBasis=usage" | jq .

  # Explicit RFC3339 window
  curl -s "$BASE/inferenceCost/total?window=2026-06-11T00:00:00Z,2026-06-11T12:00:00Z" | jq .

  # Aggregate by namespace only
  curl -s "$BASE/inferenceCost/total?window=24h&aggregate=namespace" | jq .

  # Aggregate by model_name and namespace together
  curl -s "$BASE/inferenceCost/total?window=24h&aggregate=model_name,namespace" | jq .

  # Filter to a specific namespace
  curl -s "$BASE/inferenceCost/total?window=24h&filter=namespace:llm-d-pic" | jq .

  # Filter to a specific model
  curl -s "$BASE/inferenceCost/total?window=24h&filter=model_name:gemma-4-31B" | jq .

  # Combined filter (AND): model + namespace
  curl -s "$BASE/inferenceCost/total?window=24h&filter=model_name:gemma-4-31B+namespace:llm-d-pic" | jq .

  # --- /inferenceCost/timeseries ---

  # Hourly breakdown over last 24 hours
  curl -s "$BASE/inferenceCost/timeseries?window=24h&accumulate=hour" | jq .

  # Daily breakdown over last week, usage basis
  curl -s "$BASE/inferenceCost/timeseries?window=7d&accumulate=day&costBasis=usage" | jq .

  # Daily breakdown aggregated by model_name
  curl -s "$BASE/inferenceCost/timeseries?window=7d&accumulate=day&aggregate=model_name" | jq .

  # Weekly breakdown for a specific namespace
  curl -s "$BASE/inferenceCost/timeseries?window=30d&accumulate=week&filter=namespace:llm-d-pic" | jq .
```


pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,544 @@
...
+		}
+	}
+
+	// Log the differences
+	for key, result := range results {
+		modelName, namespace := parseKey(key)
+		log.Debugf("InferenceCost: model=%s ns=%s alloc=$%.4f usage=$%.4f (%.1f%% of alloc)",
+			modelName, namespace, result.allocationTotalCost, result.usageTotalCost,
+			(result.usageTotalCost/result.allocationTotalCost)*100)
+	}
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,544 @@
...
+// which is the correct behaviour for pods that started mid-window.
+// Negative deltas (counter resets) are clamped to 0 per series before summing.
+func (c *Collector) queryCounterDelta(ctx context.Context, metric string, start, end time.Time) (map[string]float64, error) {
+	startUnix := start.Unix()
+	// Clamp end to now: last_over_time with a future @ timestamp returns no results.
+	effectiveEnd := end
+	if now := time.Now(); end.After(now) {
+		effectiveEnd = now
+	}
+	endUnix := effectiveEnd.Unix()
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,544 @@
...
+	// the window, while the @ pin ensures we don't extrapolate past end.
+	windowDuration := effectiveEnd.Sub(start)
+	windowMinutes := int(windowDuration.Minutes())
+	if windowMinutes < 2 {
+		windowMinutes = 2
+	}
+
+	// Query counter value at the end of the window.
+	endQuery := fmt.Sprintf(`sum by (model_name, namespace) (last_over_time(%s[%dm] @ %d))`, metric, windowMinutes, endUnix)
+	endVals, err := c.queryMetric(ctx, endQuery, end)
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,544 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("end-of-window query for %s: %w", metric, err)
+	}
+
+	// Query counter value at the start of the window.
+	// Use a narrow 2m lookback here: we want the value just before the window
+	// opens, not a stale value from much earlier that would undercount the delta.
+	startQuery := fmt.Sprintf(`sum by (model_name, namespace) (last_over_time(%s[2m] @ %d))`, metric, startUnix)
+	startVals, err := c.queryMetric(ctx, startQuery, end)
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/apitypes.go
```diff
@@ -0,0 +1,118 @@
...
+	cpmt := ic.CostPerMillionTokens[basis]
+	icpmt := ic.InputCostPerMillionTokens[basis]
+	ocpmt := ic.OutputCostPerMillionTokens[basis]
+
+	// Back-compute InputCost and OutputCost from the per-million rates and the
+	// token denominators used during cost calculation.
+	// The calculator uses EffectiveInputTokens for input and GenerationTokens
+	// for output, so we use the same denominators here.
+	inputCost := icpmt * ic.EffectiveInputTokens / 1_000_000
+	outputCost := ocpmt * ic.GenerationTokens / 1_000_000
```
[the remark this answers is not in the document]
simanadler: Updated documentation

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+The Collector fetches an `AllocationSet` from OpenCost's allocation layer for the requested window. It then attributes allocation costs to each model by matching pods that carry the `INFERENCE_MODEL_LABEL` pod label. Pods that carry `INFERENCE_SHARED_INFRA_LABEL=INFERENCE_SHARED_INFRA_LABEL_VALUE` (e.g. EPP, gateway) are treated as shared infrastructure: their costs are distributed across all models proportionally (by `AllocationTotalCost` weight) for `allocation` cost basis. For `usage` cost basis, shared infra costs are excluded.
+
+### Input/Output Cost Split
+
+OpenCost uses **compute-time based allocation** by default:
+
+1. Collects cumulative processing times from [vLLM](https://vllm.ai/):
+   - `vllm:request_prefill_time_seconds_total` — total time spent on input (prefill)
+   - `vllm:time_per_output_token_seconds_total` — total time spent on output (decode)
```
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Required [vLLM](https://vllm.ai/) Metrics
+
+| Metric | Required for |
+|--------|-------------|
+| `vllm:prompt_tokens_total` | Token counts, blended cost rate |
+| `vllm:generation_tokens_total` | Token counts, blended cost rate |
+| `vllm:request_prefill_time_seconds_total` | Compute-time allocation (input/output split) |
+| `vllm:time_per_output_token_seconds_total` | Compute-time allocation (input/output split) |
+| `vllm:prefix_cache_hits_total` | KV cache block correction (optional) |
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/exporter_test.go
```diff
@@ -0,0 +1,234 @@
...
+			t.Errorf("%s Help string should mention reconciliation, got: %q", name, help)
+		}
+	}
+}
+
+// TestExporter_Values verifies that exported gauge values match InferenceCost fields.
+func TestExporter_Values(t *testing.T) {
+	e, reg := newTestExporter(t)
+	ic := sampleMetric(AllocationMethodComputeTime)
+	e.Export([]*InferenceCost{ic})
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/exporter_test.go
```diff
@@ -0,0 +1,234 @@
...
+				if !floatEq(val, ic.UsageTotalCost) {
+					t.Errorf("llm_total_cost usage want %f got %f", ic.UsageTotalCost, val)
+				}
+			}
+		}
+	}
+
+	// llm_cost_per_million_tokens
+	_ = reg
+}
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/queryservice_helper.go
```diff
@@ -0,0 +1,132 @@
...
+	case opencost.AccumulateOptionHour:
+		return time.Hour
+	case opencost.AccumulateOptionDay:
+		return timeutil.Day
+	case opencost.AccumulateOptionWeek:
+		return timeutil.Week
+	case opencost.AccumulateOptionMonth:
+		// Use a calendar-aligned approach: step by 30 days (consistent with
+		// how OpenCost's CustomCost and CloudCost handle monthly accumulation).
+		return timeutil.Day * 30
```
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+### Full Environment Variable Reference
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `INFERENCE_COST_ENABLED` | `false` | Enable inference cost tracking |
+| `INFERENCE_MODEL_LABEL` | `llm-d.ai/model` | Pod label whose value is the vLLM model name |
+| `INFERENCE_SHARED_INFRA_LABEL` | `llm-d.ai/inference-serving` | Pod label key identifying shared infra pods (EPP, gateway) |
+| `INFERENCE_SHARED_INFRA_LABEL_VALUE` | `true` | Label value that marks a pod as shared infra |
+| `INFERENCE_KV_CACHE_BLOCK_SIZE` | `0` | Tokens per KV cache block; must match vLLM `--block-size`. Set to `0` to disable KV cache correction |
+| `INFERENCE_COLLECTION_INTERVAL` | `2m` | Background collection interval |
```
ameijer, line 32: can you add a column here that links to LLM-d docs describing these settings?

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Enabling Inference Cost Tracking
+
+Set the following environment variable on the OpenCost deployment:
+
+```bash
+INFERENCE_COST_ENABLED=true
+```
+
+OpenCost reads `PROMETHEUS_SERVER_ENDPOINT` for both the core metrics and the [vLLM](https://vllm.ai/) metric queries, so no separate Prometheus configuration is needed.
```
ameijer, line 22: we should also talk about adding this support to prom-less deployments

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+```yaml
+env:
+  - name: INFERENCE_COST_ENABLED
+    value: "true"
+  - name: INFERENCE_MODEL_LABEL
+    value: "llm-d.ai/model"
+  - name: INFERENCE_KV_CACHE_BLOCK_SIZE
+    value: "16"   # match your vLLM --block-size
+```
```
ameijer, line 42: let's be sure to have a complementary helm chart PR to default these settings for folks

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Prometheus Metrics
+
+When `INFERENCE_COST_ENABLED=true`, OpenCost registers and emits four gauge metrics every collection interval. All metrics carry `model_name`, `model_version`, `namespace`, and `cost_basis` labels.
+
+### `llm_total_cost`
+
+**Hourly infrastructure cost** attributed to a model.
+
+**Labels:** `model_name`, `model_version`, `namespace`, `cost_basis`
```
ameijer: consider adding to all 4 metrics: 
- workload_type (always inference for now)
- I wonder if we can take in a user-specified label for cost center? similarly to how you read in llm-d.ai/model to determine the model type to use, I wonder if we let the user give us a label to use, then surface it in that? and if cost center label unset, we simply do not emit it?

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+This is an instantaneous hourly rate ($/hour), not a cumulative counter.
```
ameijer, line 69: llm_total_hourly_cost? not advocating it, just raising for discussion

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+```promql
+# Current hourly cost for a model
+llm_total_cost{model_name="Qwen/Qwen3-32B", cost_basis="allocation"}
+
+# Estimated 24-hour cost (if rate stays constant)
+llm_total_cost{model_name="Qwen/Qwen3-32B", cost_basis="allocation"} * 24
```
ameijer: we can also do an average of the total cost over the last 24h and multiply by 24 for a bit more accuracy

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+llm_cost_per_million_tokens{model_name="Qwen/Qwen3-32B", cost_basis="allocation"}
+
+# Average over the past 24 hours
+avg_over_time(llm_cost_per_million_tokens{model_name="Qwen/Qwen3-32B"}[24h])
+
+# Compare models side-by-side
+llm_cost_per_million_tokens{cost_basis="allocation"}
+```
+
+### `llm_input_cost_per_million_tokens`
```
ameijer: do you see an advantage here of having this as a separate metric, vs adding 'phase' or something to llm_cost_per_million_tokens? phase could be the set of [prompt, generation]

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+**Cost per 1M effective input (prompt) tokens.** When KV cache block size is configured, cached tokens are excluded from the denominator (cost reflects tokens that actually required compute).
```
ameijer: should this behavior be toggle-able?

ameijer: edit - I see that you basically do have this via allocation_method. excellent!

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+**Labels:** `model_name`, `model_version`, `namespace`, `cost_basis`, `allocation_method`
+
+`allocation_method` values:
+- `compute_time` — input/output split is based on [vLLM](https://vllm.ai/) prefill time; KV cache correction applied
+- `compute_time_uncorrected` — based on [vLLM](https://vllm.ai/) prefill time but KV cache data was unavailable
+- `multiplier` — fixed output/input cost ratio used (timing metrics unavailable; default ratio 2.5×)
```
ameijer: I appreciate that when I get to reading the implementation, this will become clear but maybe worth summarizing how we handler these different cases

ameijer: for example, will metrics with all 3 values always be present, and if we cannot get prefill times, then compute_time will be 0? or will that metric not get published?

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+**Response shape:**
+
+```json
+{
+  "data": {
+    "inferenceCosts": {
+      "Qwen/Qwen3-32B:llm-d-namespace": {
+        "properties": {
+          "modelName": "Qwen/Qwen3-32B",
+          "namespace": "llm-d-namespace"
```
ameijer: namespace would not appear in the results, when only aggregating by model_name, unless the qwen model happens to only occur in the llm-d-namespace

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+Returns a single aggregated `InferenceCostSet` covering the full requested window.
+
+**Query parameters:**
+
+| Parameter | Required | Description |
+|-----------|----------|-------------|
+| `window` | Yes | Time window: RFC3339 `start,end` or named range (e.g. `7d`, `24h`, `2025-01-01T00:00:00Z,2025-01-02T00:00:00Z`) |
+| `costBasis` | No | `allocation` (default) or `usage` |
+| `aggregate` | No | Comma-separated dimensions: `model_name`, `model_version`, `namespace`, `cluster` |
```
ameijer: I think we could really benefit from also including pod name, controller name, and container name here too

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+| `accumulate` | No | Step size within the window: `hour`, `day`, `week`, `month` (results are then accumulated into one total) |
+| `filter` | No | `prop:value` pairs joined with `+` for AND logic, e.g. `namespace:default+model_name:llama3` |
```
ameijer:
```suggestion
| `filter` | No | `prop:value` pairs joined with `+` for AND logic, e.g. `namespace:default+model_name:llama3` |
| `idle` | No | `shareWeighted/separate/shareEven` |
```

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+        "totalCost": 42.50,
+        "promptTokens": 12000000,
+        "generationTokens": 3000000,
+        "totalTokens": 15000000,
+        "costPerMillionTokens": 2.83,
+        "inputCost": 28.40,
+        "outputCost": 14.10,
+        "inputCostPerMillionTokens": 2.37,
+        "outputCostPerMillionTokens": 4.70,
+        "allocationMethod": "compute_time"
```
ameijer, line 223: should this be taken is as an API param?

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+    expr: llm_cost_per_million_tokens{cost_basis="allocation"} > 10
+    for: 5m
+    labels:
+      severity: warning
+    annotations:
+      summary: "High inference cost for {{ $labels.model_name }}"
+      description: "Model {{ $labels.model_name }} in {{ $labels.namespace }} costs ${{ $value }}/M tokens"
+```
+
+## REST API Endpoints
```
ameijer, line 176: what other interesting cost related metrics are there? SHould we expose for example KV cache stuff?

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+| **Runner** | `runner.go` | Drives periodic background collection for the Prometheus exporter |
+| **Types / API Types** | `types.go`, `apitypes.go` | Internal and HTTP-facing data models |
+
+Integration point: `pkg/cmd/costmodel/costmodel.go` registers the HTTP routes and initialises the collector, calculator, exporter, and runner when `INFERENCE_COST_ENABLED=true`.
+
+## Cost Calculation Methodology
+
+### Infrastructure Cost Collection
+
+The Collector fetches an `AllocationSet` from OpenCost's allocation layer for the requested window. It then attributes allocation costs to each model by matching pods that carry the `INFERENCE_MODEL_LABEL` pod label. Pods that carry `INFERENCE_SHARED_INFRA_LABEL=INFERENCE_SHARED_INFRA_LABEL_VALUE` (e.g. EPP, gateway) are treated as shared infrastructure: their costs are distributed across all models proportionally (by `AllocationTotalCost` weight) for `allocation` cost basis. For `usage` cost basis, shared infra costs are excluded.
```
ameijer, line 332: so, here's a Big Idea - what if we can get LLM-d to add pod UID labels to their metrics, and start exporting these on a per pod basis? that would let us very easily and accurately get cost per token

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Input/Output Cost Split
+
+OpenCost uses **compute-time based allocation** by default:
+
+1. Collects cumulative processing times from [vLLM](https://vllm.ai/):
+   - `vllm:request_prefill_time_seconds_sum` — total time spent on input (prefill)
+   - `vllm:time_per_output_token_seconds_sum` — total time spent on output (decode)
+2. Allocates infrastructure cost proportionally: `InputCost = TotalCost × (PrefillTime / TotalTime)`
```
ameijer, line 341: why split based on time, and not prompt vs generation tokens?

docs/inference-cost-tracking.md
simanadler: OK.  Will add to to-do list for future PR

docs/inference-cost-tracking.md
simanadler: I removed the need to define kv cache block size.  It can differ from model to model, and is now identified dynamically based on vllm deployment config

docs/inference-cost-tracking.md
simanadler: Definitely.  It's on the to do list for a future PR

docs/inference-cost-tracking.md
simanadler: Updated and improved in recent commit

docs/inference-cost-tracking.md
simanadler: Added percentage of kv cache hits

docs/inference-cost-tracking.md
ameijer: ok - but as long as there are still variable, for example the ENABLED var, we need a way for helm users to set that and that

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,420 @@
...
+Input:      ($3.20 × 0.5) / 11,200,000 × 1,000,000 = $0.143/M effective input tokens
+Output:     ($3.20 × 0.5) / 3,000,000 × 1,000,000 = $0.533/M output tokens
+```
+
+## Required [vLLM](https://vllm.ai/) Metrics
+
+| Metric | Required for |
+|--------|-------------|
+| `vllm:prompt_tokens_total` | Token counts, blended cost rate |
+| `vllm:generation_tokens_total` | Token counts, blended cost rate |
```
ameijer, line 381: can these be used to split costs?

docs/inference-cost-tracking.md
simanadler: The reason is that tokens are not equally expensive. Using raw token counts to split cost implicitly assumes input and output tokens cost the same amount of GPU work, which is
  false:

  - Prefill (input) is parallelizable — all prompt tokens are processed simultaneously in one forward pass. 1000 input tokens ≈ 1 forward pass.
  - Decode (output) is sequential — each output token requires its own forward pass, one at a time. 1000 output tokens = 1000 forward passes.

  In practice, decode is typically 5–20× more GPU-time-intensive per token than prefill. If you split cost by raw token counts, you'd grossly overcharge for input and undercharge
  for output.

  Using prefillTime / (prefillTime + decodeTime) captures the actual GPU time consumed by each phase, so the split reflects real resource consumption regardless of batch size,
  sequence length, or hardware.

  KV cache hits complicate this further: a cache hit skips the prefill computation for those tokens entirely, so they consume almost no GPU time. The time-based split naturally
  handles this — if cache hits reduce prefill time to near zero, inputFraction shrinks accordingly, and less cost is attributed to input. You don't need to explicitly subtract
  cached tokens from the split; the timing already reflects it.

docs/inference-cost-tracking.md
simanadler: done

docs/inference-cost-tracking.md
simanadler: done

docs/inference-cost-tracking.md
simanadler: Changed as per your suggestion

docs/inference-cost-tracking.md
simanadler: Added to REST APIs but not to prometheus metrics.  Pod names can create high cardinality in metrics, especially with auto-scaling. This could impact Prometheus performance if exposed as metric labels.

docs/inference-cost-tracking.md
simanadler: It's determined at runtime based on the metrics available in prometheus

docs/inference-cost-tracking.md
simanadler: Great idea!  I will look into it with the llm-d team

pkg/inferencecost/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+			[]string{"model_name", "model_version", "namespace", "cost_basis"},
+		),
+		inputCostPerMillionTokens: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "llm_input_cost_per_million_tokens",
+				Help: "Infrastructure cost per 1M effective input tokens (prompt tokens that required compute; " +
+					"cached tokens excluded when KV cache block size is configured). " +
+					"allocation_method=compute_time: split by vLLM prefill time with cache correction. " +
+					"allocation_method=compute_time_uncorrected: split by vLLM prefill time, no cache correction. " +
+					"allocation_method=multiplier: fixed output/input ratio used (timing metrics unavailable).",
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+			},
+			[]string{"model_name", "model_version", "namespace", "cost_basis", "allocation_method"},
+		),
+		outputCostPerMillionTokens: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "llm_output_cost_per_million_tokens",
+				Help: "Infrastructure cost per 1M output (generation) tokens. " +
+					"allocation_method=compute_time: split by vLLM decode time. " +
+					"allocation_method=compute_time_uncorrected: split by vLLM decode time, no cache correction. " +
+					"allocation_method=multiplier: fixed output/input ratio used (timing metrics unavailable).",
```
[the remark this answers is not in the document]
simanadler: fixed

docs/swagger.json
```diff
@@ -313,6 +477,155 @@
...
+          },
+          "outputCostPerMillionTokens": {
+            "type": "number",
+            "description": "Infrastructure cost per 1M output (generation) tokens, in USD."
+          },
+          "allocationMethod": {
+            "type": "string",
+            "enum": ["compute_time", "compute_time_uncorrected", "multiplier"],
+            "description": "Method used to split costs between input and output tokens. `compute_time`: split by vLLM prefill/decode time with KV cache correction. `compute_time_uncorrected`: split by vLLM prefill/decode time without KV cache correction. `multiplier`: fixed output/input ratio (fallback when timing metrics are unavailable)."
+          }
```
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,418 @@
...
+The Collector fetches an `AllocationSet` from OpenCost's allocation layer for the requested window. It then attributes allocation costs to each model by matching pods that carry the `INFERENCE_MODEL_LABEL` pod label. Pods that carry `INFERENCE_SHARED_INFRA_LABEL=INFERENCE_SHARED_INFRA_LABEL_VALUE` (e.g. EPP, gateway) are treated as shared infrastructure: their costs are distributed across all models proportionally (by `AllocationTotalCost` weight) for `allocation` cost basis. For `usage` cost basis, shared infra costs are excluded.
+
+### Input/Output Cost Split
+
+OpenCost uses **compute-time based allocation** by default:
+
+1. Collects cumulative processing times from [vLLM](https://vllm.ai/):
+   - `vllm:request_prefill_time_seconds_sum` — total time spent on input (prefill)
+   - `vllm:time_per_output_token_seconds_sum` — total time spent on output (decode)
+2. Allocates infrastructure cost proportionally: `InputCost = TotalCost × (PrefillTime / TotalTime)`
```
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,418 @@
...
+
+## Required [vLLM](https://vllm.ai/) Metrics
+
+| Metric | Required for |
+|--------|-------------|
+| `vllm:prompt_tokens_total` | Token counts, blended cost rate |
+| `vllm:generation_tokens_total` | Token counts, blended cost rate |
+| `vllm:request_prefill_time_seconds_sum` | Compute-time allocation (input/output split) |
+| `vllm:time_per_output_token_seconds_sum` | Compute-time allocation (input/output split) |
+| `vllm:prefix_cache_hits_total` | KV cache block correction (optional) |
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,683 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/opencost"
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+)
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,683 @@
...
+	as, err := c.allocationQuerier.ComputeAllocation(start, end)
+	if err != nil {
+		return nil, err
+	}
+
+	opts := &opencost.AllocationAggregationOptions{
+		ShareIdle:    opencost.ShareWeighted,
+		ShareSplit:   opencost.ShareWeighted,
+		SharedLabels: map[string][]string{c.config.SharedInfraLabel: {c.config.SharedInfraLabelValue}},
+	}
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+			prometheus.GaugeOpts{
+				Name: "llm_input_cost_per_million_tokens",
+				Help: "Infrastructure cost per 1M effective input tokens (prompt tokens that required compute; " +
+					"cached tokens excluded when KV cache block size is configured). " +
+					"allocation_method=compute_time: split by vLLM prefill time with cache correction. " +
+					"allocation_method=compute_time_uncorrected: split by vLLM prefill time, no cache correction. " +
+					"allocation_method=multiplier: fixed output/input ratio used (timing metrics unavailable).",
+			},
+			[]string{"model_name", "model_version", "namespace", "cost_basis", "allocation_method"},
+		),
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+		outputCostPerMillionTokens: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "llm_output_cost_per_million_tokens",
+				Help: "Infrastructure cost per 1M output (generation) tokens. " +
+					"allocation_method=compute_time: split by vLLM decode time. " +
+					"allocation_method=compute_time_uncorrected: split by vLLM decode time, no cache correction. " +
+					"allocation_method=multiplier: fixed output/input ratio used (timing metrics unavailable).",
+			},
+			[]string{"model_name", "model_version", "namespace", "cost_basis", "allocation_method"},
+		),
```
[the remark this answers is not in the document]
simanadler: fixed

docs/swagger.json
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,420 @@
...
+The Collector fetches an `AllocationSet` from OpenCost's allocation layer for the requested window. It then attributes allocation costs to each model by matching pods that carry the `INFERENCE_MODEL_LABEL` pod label. Pods that carry `INFERENCE_SHARED_INFRA_LABEL=INFERENCE_SHARED_INFRA_LABEL_VALUE` (e.g. EPP, gateway) are treated as shared infrastructure: their costs are distributed across all models proportionally (by `AllocationTotalCost` weight) for `allocation` cost basis. For `usage` cost basis, shared infra costs are excluded.
+
+### Input/Output Cost Split
+
+OpenCost uses **compute-time based allocation** by default:
+
+1. Collects cumulative processing times from [vLLM](https://vllm.ai/):
+   - `vllm:request_prefill_time_seconds_sum` — total time spent on input (prefill)
+   - `vllm:time_per_output_token_seconds_sum` — total time spent on output (decode)
+2. Allocates infrastructure cost proportionally: `InputCost = TotalCost × (PrefillTime / TotalTime)`
```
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
```diff
@@ -0,0 +1,420 @@
...
+```
+
+## Required [vLLM](https://vllm.ai/) Metrics
+
+| Metric | Required for |
+|--------|-------------|
+| `vllm:prompt_tokens_total` | Token counts, blended cost rate |
+| `vllm:generation_tokens_total` | Token counts, blended cost rate |
+| `vllm:request_prefill_time_seconds_sum` | Compute-time allocation (input/output split) |
+| `vllm:time_per_output_token_seconds_sum` | Compute-time allocation (input/output split) |
```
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
simanadler: You are correct.  Fixed

pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,718 @@
+package inferencecost
```
ameijer, line 1: @simanadler , what do you think about restructuring this a bit? I know we had talked about making inference costs a first class citizen, which would I think be reinforced by putting its code alongside the regular container costing code where possible. 

For example, the collectors in OC are a separate go mod so we can manage their dependencies separately, see  `opencost/modules/prometheus-source`. thoughts on moving this code out there? then that sets us up for a prom-less port

simanadler: Thanks for the suggestion — the direction makes sense and aligns with the "inference costs as first-class citizen" goal.
  
  One thing worth noting: collector.go has two distinct data sources baked into one type — Prometheus queries for the vLLM token/timing/cache metrics, and the AllocationQuerier
  interface (backed by CostModel) for infrastructure costs. The Prometheus side could move to modules/prometheus-source cleanly, but the AllocationQuerier dependency flows the wrong
  way for that module. So the move would require splitting Collector first: the Prometheus-metric half goes to the new module, and the allocation join stays in the main module.
  That's actually the right design for the prom-less port you're describing — you'd swap the Prometheus half independently while the join logic stays untouched.
  
  Given that it's a prerequisite restructure rather than a file move, is this something you'd want to tackle now as part of this PR/feature, or would it make more sense to land the
  current implementation and file it as follow-up work once the feature is more stable?

pkg/inferencecost/collector.go
```diff
@@ -0,0 +1,741 @@
...
+	// Merge results: allocation costs from first query, usage costs from second
+	results := make(map[string]*allocationResult)
+	for key, allocResult := range allocationCosts {
+		results[key] = &allocationResult{
+			allocationTotalCost: allocResult.allocationTotalCost,
+			usageTotalCost:      0, // Will be filled from usageCosts
+			namespace:           allocResult.namespace,
+			cluster:             allocResult.cluster,
+		}
+	}
```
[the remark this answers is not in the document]
simanadler: fixed

docs/inference-cost-tracking.md
simanadler: I started looking into a prom-less deployment.  However,the inference package is reliant on vllm metrics which are only available via prometheus.  So, I don't think a prometheus-less deployment is feasible for inference.

docs/inference-cost-tracking.md
simanadler: [PR](https://github.com/opencost/opencost-helm-chart/pull/370) opened for helm

docs/inference-cost-tracking.md
ameijer: wanted to get your feedback on this one

docs/inference-cost-tracking.md
ameijer: interesting!

pkg/inferencecost/aggregate.go
```diff
@@ -0,0 +1,289 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/opencost/opencost/core/pkg/opencost"
+)
+
+// supportedAggregateProperties lists the InferenceCostProperties dimensions
+// that the collector actually populates in Phase 1.
+var supportedAggregateProperties = map[string]bool{
```
ameijer, line 12: setting everything to true makes me wonder if this would be better as a slice?

pkg/inferencecost/collector.go
ameijer: I'd go ahead and do it now. its important to get the file structure right before we do something that makes it harder to do

simanadler: The inference cost package requires prometheus because that's the only way to get the vllm metrics, so prom-less is not an option.

Are you proposing though that I move pkg/inferencecost into modules/inference-cost/pkg/inferencecost because inference is optional?    I.e. Create modules/inference-cost/ with its own go.mod?

docs/inference-cost-tracking.md
simanadler: Changed to llm_total_hourly_cost

docs/swagger.json
```diff
@@ -137,6 +137,170 @@
...
+              }
+            }
+          },
+          "400": {
+            "description": "Bad request — missing or invalid parameters (including missing `accumulate`)"
+          },
+          "501": {
+            "description": "Inference cost tracking is not enabled (`INFERENCE_COST_ENABLED` is not set to `true`)"
+          }
+        }
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/aggregate.go
simanadler: Good catch! I considered this, but opted for the map because it's used for O(1) membership checks in two hot paths:

Line 41 in aggKey() - validates each aggregation dimension
Line 254 in parseFilter() - validates each filter property
With a slice, we'd need either linear search (O(n) performance regression) or converting to a map anyway for lookups.

The map[string]bool pattern is idiomatic Go for set membership testing. If the "all true" values feel redundant, we could use map[string]struct{} instead (common Go idiom for sets, uses zero bytes per value), but I think the current approach is clear and performant.

Happy to change if you feel strongly about it though! Let me know your thoughts.

pkg/inferencecost/collector.go
simanadler: re-organized as per meeting yesterday

docs/inference-cost-tracking.md
simanadler: changed as suggested

docs/inference-cost-tracking.md
simanadler: workload_type added

pkg/inferencecost/collector_test.go
```diff
@@ -0,0 +1,648 @@
...
+		{name: "no activity", endVal: 500, startVal: 500, want: 0},
+		{name: "counter reset clamped to zero", endVal: 100, startVal: 900, want: 0},
+		{name: "new pod (no start sample)", endVal: 400, startVal: 0, want: 400},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			delta := tt.endVal - tt.startVal
+			if delta < 0 {
+				delta = 0
+			}
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/collector_test.go
```diff
@@ -0,0 +1,648 @@
...
+func TestQueryCounterDelta_Formula(t *testing.T) {
+	tests := []struct {
+		name     string
+		endVal   float64
+		startVal float64
+		want     float64
+	}{
+		{name: "normal increase", endVal: 1000, startVal: 200, want: 800},
+		{name: "no activity", endVal: 500, startVal: 500, want: 0},
+		{name: "counter reset clamped to zero", endVal: 100, startVal: 900, want: 0},
```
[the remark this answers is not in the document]
simanadler: fixed

modules/prometheus-source/pkg/prom/inference_queries.go
```diff
@@ -0,0 +1,378 @@
...
+// then subtracts. This avoids the extrapolation inflation produced by
+// increase(metric[Xm]) when a series has fewer samples than the window
+// (e.g. a pod that restarted mid-window, or a sum across many replicas
+// where Prometheus extrapolates each series independently before summing).
+//
+// last_over_time(metric[2m] @ t) fetches the most recent sample within 2
+// minutes of t. 2 minutes covers the default 30s scrape interval with margin.
+// Series with no sample near start get a start-value of 0 (treated as new),
+// which is the correct behaviour for pods that started mid-window.
+// Negative deltas (counter resets) are clamped to 0 per series before summing.
```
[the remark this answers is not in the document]
simanadler: fixed

core/pkg/source/datasource.go
```diff
@@ -144,6 +144,25 @@ type MetricsQuerier interface {
 
 	// Data Coverage Query
 	QueryDataCoverage(limitDays int) (time.Time, time.Time, error)
+
+	// Inference Metrics (vLLM) - relevant when INFERENCE_METRICS_ENABLE is set to true
```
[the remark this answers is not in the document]
simanadler: fixed

modules/prometheus-source/pkg/prom/inference_queries.go
```diff
@@ -0,0 +1,378 @@
...
+		results := mapToQueryResults(values)
+		resultsChan <- &source.QueryResults{Results: results}
+	}()
+	
+	return source.NewFuture(decodeInferenceProcessingTimeResult, resultsChan)
+}
+
+// QueryInferenceCachedTokens implements MetricsQuerier.QueryInferenceCachedTokens
+func (pds *PrometheusMetricsQuerier) QueryInferenceCachedTokens(start, end time.Time) *source.Future[source.InferenceTokensResult] {
+	const queryName = "QueryInferenceCachedTokens"
```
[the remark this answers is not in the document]
simanadler: fixed

modules/prometheus-source/pkg/prom/inference_queries.go
```diff
@@ -0,0 +1,378 @@
...
+		results := mapToQueryResults(values)
+		resultsChan <- &source.QueryResults{Results: results}
+	}()
+	
+	return source.NewFuture(decodeInferenceProcessingTimeResult, resultsChan)
+}
+
+// QueryInferenceOutputProcessingTime implements MetricsQuerier.QueryInferenceOutputProcessingTime
+func (pds *PrometheusMetricsQuerier) QueryInferenceOutputProcessingTime(start, end time.Time) *source.Future[source.InferenceProcessingTimeResult] {
+	const queryName = "QueryInferenceOutputProcessingTime"
```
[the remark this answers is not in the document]
simanadler: fixed

modules/prometheus-source/pkg/prom/inference_queries.go
```diff
@@ -0,0 +1,378 @@
...
+		results := mapToQueryResults(values)
+		resultsChan <- &source.QueryResults{Results: results}
+	}()
+	
+	return source.NewFuture(decodeInferenceTokensResult, resultsChan)
+}
+
+// QueryInferenceInputProcessingTime implements MetricsQuerier.QueryInferenceInputProcessingTime
+func (pds *PrometheusMetricsQuerier) QueryInferenceInputProcessingTime(start, end time.Time) *source.Future[source.InferenceProcessingTimeResult] {
+	const queryName = "QueryInferenceInputProcessingTime"
```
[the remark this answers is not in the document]
simanadler: fixed

modules/prometheus-source/pkg/prom/inference_queries.go
```diff
@@ -0,0 +1,378 @@
...
+		results := mapToQueryResults(values)
+		resultsChan <- &source.QueryResults{Results: results}
+	}()
+	
+	return source.NewFuture(decodeInferenceTokensResult, resultsChan)
+}
+
+// QueryInferenceGenerationTokens implements MetricsQuerier.QueryInferenceGenerationTokens
+func (pds *PrometheusMetricsQuerier) QueryInferenceGenerationTokens(start, end time.Time) *source.Future[source.InferenceTokensResult] {
+	const queryName = "QueryInferenceGenerationTokens"
```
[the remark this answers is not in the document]
simanadler: fixed

modules/prometheus-source/pkg/prom/inference_queries.go
```diff
@@ -0,0 +1,378 @@
...
+	"time"
+
+	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/source"
+	"github.com/opencost/opencost/core/pkg/util"
+)
+
+// QueryInferencePromptTokens implements MetricsQuerier.QueryInferencePromptTokens
+func (pds *PrometheusMetricsQuerier) QueryInferencePromptTokens(start, end time.Time) *source.Future[source.InferenceTokensResult] {
+	const queryName = "QueryInferencePromptTokens"
```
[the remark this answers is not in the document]
simanadler: fixed

pkg/inferencecost/collector_test.go
```diff
@@ -0,0 +1,648 @@
...
+	}
+	if _, ok := remappedKeys["google/gemma-4-31B:llm-d-pic"]; !ok {
+		t.Error("google/gemma-4-31B:llm-d-pic should be in remappedKeys")
+	}
+}
+
+// TestCollector_BuildQueryWindow verifies that buildQueryWindow generates
+// correct Prometheus time range selectors based on CollectionInterval.
+// TestQueryCounterDelta_Formula verifies the delta = end - start subtraction
+// and that negative deltas (counter resets) are clamped to zero.
```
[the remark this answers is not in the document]
simanadler: fixed

core/pkg/source/datasource.go
```diff
@@ -433,6 +441,25 @@ type MetricsQuerier interface {
...
+	QueryInferenceInputProcessingTime(start, end time.Time) *Future[InferenceProcessingTimeResult]
+
+	// QueryInferenceOutputProcessingTime returns output processing time in seconds by model_name and namespace
+	QueryInferenceOutputProcessingTime(start, end time.Time) *Future[InferenceProcessingTimeResult]
+
+	// QueryInferenceCachedTokens returns KV cache hit counts by model_name and namespace
+	QueryInferenceCachedTokens(start, end time.Time) *Future[InferenceTokensResult]
+
+	// QueryInferenceCacheConfig returns cache configuration (prefix caching enabled) by model_name and namespace
+	QueryInferenceCacheConfig(t time.Time) *Future[InferenceCacheConfigResult]
```
ameijer, line 462: thank you! amazing!

core/pkg/source/decoders.go
```diff
@@ -2134,6 +2134,64 @@ type DCGMDeviceContainerUsageResult struct {
...
+
+	return &InferenceCacheConfigResult{
+		Configs: map[string]*InferenceCacheConfig{
+			key: {
+				PrefixCachingEnabled: prefixCachingEnabled > 0,
+			},
+		},
+	}
+}
+
```
ameijer, line 2194: unit tests alongside, please

ameijer (approved):
lgtm, lets see how it burns in anyway

core/pkg/source/decoders.go
simanadler: done

ameijer (approved):

PR #3857, opened 2026-06-16
Add GPU saturation model to the Allocation API

cliffcolvin:
## Intent

Add the **data model** for GPU saturation signals (USE method) to the Allocation API. This is the foundation of a small stack that brings vendor-neutral GPU saturation/device metrics to the kubemodel path.

**This PR is the model only — nothing produces these values yet.** Producers (prometheus-source / collector-source) and the kubemodel hydration follow in stacked PRs.

## Problem

OpenCost reports GPU *cost* and basic *utilization*, but has no model for GPU **saturation** — the USE-method "how overloaded is this resource" signal. Operators running GPU workloads need to see throttling, framebuffer (memory) pressure, error events, and memory-bandwidth/compute activity to know when a GPU is the bottleneck. There is currently nowhere in the domain model to carry these signals.

## What this adds

- **`GPUSaturation`** (core/pkg/opencost): throttle violation/reason ratios (the reason map is decoded from the DCGM clock-throttle-reasons bitmask), framebuffer occupancy and memory-pressure ratio, XID error counts, memory-bandwidth (DRAM) and compute (SM) activity, and host/peer (PCIe/NVLink) link rates.
  - **Absence is explicit:** nil pointers / empty maps mean "telemetry unavailable", never zero. `SanitizeNaN` normalizes NaN to absent so `Equal` and codec roundtrips are well-defined.
- **`GPUAllocation.Saturation`** optional field (bingen field version 26). Older payloads decode with `Saturation` nil.
- **`GPU_MEMORY_SATURATION_THRESHOLD`** (core env, shared by both data sources): framebuffer-occupancy ratio above which a GPU counts as memory-pressured. Values outside `(0, 1]` fall back to the `0.9` default.

## Tests

`GPUSaturation` binary roundtrip, `Equal`, NaN sanitization, throttle-reason bitmask decode, and the env threshold (hermetic — the env var is set for every case including the unset/default path).

## Stack

1. **→ This PR** — saturation data model
2. Produce saturation/device metrics from both data sources
3. Hydrate DCGM device saturation in the kubemodel pipeline

Split out of the larger GPU saturation branch (#3847) for focused review. See #3847 for the full breakdown.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/pkg/env/gpu.go
+++ b/core/pkg/env/gpu.go
@@ -1,0 +1,28 @@
+package env
+
+const (
+	// GPUMemorySaturationThresholdEnvVar configures the framebuffer
+	// occupancy ratio above which a GPU counts as memory-pressured.
+	GPUMemorySaturationThresholdEnvVar = "GPU_MEMORY_SATURATION_THRESHOLD"
+)
+
+// DefaultGPUMemorySaturationThreshold is the framebuffer occupancy ratio
+// above which a GPU is considered memory-pressured when no valid override
+// is configured.
+const DefaultGPUMemorySaturationThreshold = 0.9
+
+// GetGPUMemorySaturationThreshold returns the configured framebuffer
+// occupancy threshold for GPU memory pressure. Values outside (0, 1] are
+// rejected in favor of the default.
+//
+// This lives in core (rather than per data-source module) because both the
+// prometheus-source query builder and the collector-source aggregator must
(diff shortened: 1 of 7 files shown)


cliffcolvin:
won't do, see https://github.com/opencost/opencost/pull/3915

PR #3865, opened 2026-06-17
fix: guard cloudcost Status coverage read and stop ClusterMap ticker

vtushar06:
## Description

Two small concurrency/cleanup fixes I came across while reading through the ingestion code:

1. **cloudcost `ingestor.Status()` data race.** It read `ing.coverage` without holding `coverageLock`, while the build loop reassigns `ing.coverage` under that lock (`ContractStart` / `ExpandStart`). `coverage` is an `opencost.Window` (two `*time.Time` fields), so a `Status()` read concurrent with a build is a torn read. Same shape as the customcost one in #3838, just a struct instead of a map (so it's a race rather than a hard crash). Fixed by reading the value under the lock.

2. **ClusterMap refresh ticker never stopped.** The refresher goroutine in `clustermap.go` creates `time.NewTicker(refresh)` but never calls `Stop()` before it returns on `<-cm.stop`. Added `defer ticker.Stop()`.

## Related Issues

No existing issue for either, found them reading the code. Happy to open one if you'd prefer.

## User Impact

No behavior change. `Status()` returns the same data; the ticker change just releases the ticker when the refresher exits. The cloudcost fix removes a data race that `go test -race` flags.

## Testing

- Added `TestIngestor_Status_ConcurrentWithCoverageWrite` (cloudcost), which runs `Status()` against `expandCoverage()` concurrently. It trips the race detector on the old code and passes with the fix. One caveat: it only fails under `-race` (a struct torn read can't be asserted deterministically), so it's mainly a guard for local `-race` runs.
- `go test ./pkg/cloudcost/ -race` green; full `go build ./...`, `go vet`, and the prometheus-source module build + vet all clean.

### DIFF
--- a/modules/prometheus-source/pkg/prom/clustermap.go
+++ b/modules/prometheus-source/pkg/prom/clustermap.go
@@ -47,6 +47,7 @@ func newPrometheusClusterMap(contextFactory *ContextFactory, cip clusters.Cluste
 
 		// Tick on interval and refresh clusters
 		ticker := time.NewTicker(refresh)
+		defer ticker.Stop()
 		for {
 			select {
 			case <-ticker.C:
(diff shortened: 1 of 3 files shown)


vtushar06:
Hey @ameijer, Can you have a review on this PR.

ameijer (approved):
lgtm thanks @vtushar06

PR #3866, opened 2026-06-17
fix(mcp): sanitize non-finite floats before SDK marshal

vtushar06:
## Description

The MCP integration test `TestMCPAssetVsHTTP` (historical assets) has been failing the merge queue with:

```
asset_mcp_vs_http_test.go:174: Failed to get MCP tool response:
MCP error (code 0): marshaling output: json: unsupported value: NaN
```

The MCP SDK marshals tool output with the standard library `encoding/json`, which returns an error for non-finite floats (`NaN`, `+Inf`, `-Inf`). Upstream cost calculations can produce these, for example a 0/0 breakdown or overhead fraction on an asset with zero cost or zero capacity. The HTTP `/assets` path doesn't hit this because it serializes through OpenCost's own JSON library, but the MCP path uses stdlib json, so a single non-finite float anywhere in the response fails the whole tool call.

This adds a small sanitization pass in `ProcessMCPRequest` that walks the response and replaces any non-finite float with 0 before it reaches the SDK. It runs for every query type (allocation, asset, cloud cost, efficiency), so it covers the asset case that's failing now and prevents the same class of failure elsewhere. The walk is best effort: any reflection panic is recovered and logged so the sanitizer can never break a request.

This is independent of #3838 and #3865 (their code is correct and approved) — it's a pre-existing bug that fails the merge queue for everyone, so fixing it here unblocks the queue.

## Related

Surfaced by the merge queue failures on #3838 and #3865, and the scheduled `develop` integration run hits it too.

## User Impact

No API or schema change. MCP responses that previously errored out on a non-finite float now return that field as 0 and marshal successfully. The HTTP API path is untouched.

## Testing

- Added `pkg/mcp/sanitize_test.go`: builds an `AssetResponse` with `NaN`/`Inf` in nested fields (base floats, a `*float64`, a nested `*NodeOverhead`, a nested `*AssetBreakdown`, inside a `map[string]*AssetSet`), asserts `encoding/json` rejects it before sanitization and accepts it after, that non-finite values are zeroed, and that finite values are preserved. Plus a nil-safety case.
- `go build ./...`, `go vet ./pkg/mcp/`, and the full `pkg/mcp` package tests pass locally.

I can't run the live integration stack from here, so this is verified via the unit test that reproduces the exact marshal failure rather than a full stack run.

cc @ameijer — this is the second, unrelated blocker on the merge queue we discussed (the NaN one, separate from the conn-reset retry in opencost/opencost-integration-tests#106).

### DIFF
--- a/pkg/mcp/sanitize_test.go
+++ b/pkg/mcp/sanitize_test.go
@@ -1,0 +1,70 @@
+package mcp
+
+import (
+	"encoding/json"
+	"math"
+	"testing"
+)
+
+// TestSanitizeNonFiniteFloatsAssetResponseMarshals reproduces the integration
+// failure (TestMCPAssetVsHTTP: "marshaling output: json: unsupported value:
+// NaN") and verifies the sanitizer fixes it: encoding/json must reject the
+// response before sanitization and accept it after, with non-finite floats
+// zeroed and finite ones preserved.
+func TestSanitizeNonFiniteFloatsAssetResponseMarshals(t *testing.T) {
+	usedBytes := math.NaN()
+	resp := &AssetResponse{
+		Assets: map[string]*AssetSet{
+			"assets": {
+				Name: "assets",
+				Assets: []*Asset{
+					{
+						Type:          "Node",
+						Minutes:       math.NaN(),
+						Adjustment:    math.Inf(1),
+						TotalCost:     math.Inf(-1),
+						CPUCost:       math.NaN(),
+						GPUCost:       5.0, // finite, must be preserved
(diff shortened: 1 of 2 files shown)


vtushar06:
@ameijer addressed the review: the sanitizer now returns the scrubbed value so value-type outputs are covered too (and switched to reflect.Ptr). Ready for re-review.

vtushar06:
@ameijer bumping this one again, still green and waiting on a review whenever you have a sec

ameijer (approved):
LGTM, thanks @vtushar06

PR #3871, opened 2026-06-19
fix: update Premium SSD pricing logic to ignore Disk Mount

### LINKED ISSUE #3870: bug(azure): Premium SSD PVs underpriced 16x — Disk Mount meter overwrites Disk meter in convertMeterToPricings
**Describe the bug**
`managed-csi-premium` PVs are underpriced by ~16× in v1.120.3. PR #3778 correctly fixed SKU detection (`skuname` lookup), but the `premium_ssd` price itself is wrong — `convertMeterToPricings()` matches both `P4 LRS Disk` ($5.28/mo) and `P4 LRS Disk Mount` ($0.32/mo) via the `"P4 "` prefix filter. The mount meter overwrites the disk meter, so only the mount fee is used for pricing.

Standard SSD is unaffected because Azure does not publish a separate `E4 LRS Disk Mount` meter.

**To Reproduce**
1. Run OpenCost on AKS with PVs using the `managed-csi-premium` StorageClass (`disk.csi.azure.com`, `skuname: Premium_LRS`)
2. Query `pv_hourly_cost` for any of those PVs

**Expected behavior**
`pv_hourly_cost` reflects the `P4 LRS Disk` rate. For Central US: $5.2795/mo → $2.26e-04/GiB/hr (~$0.165/GiB/month).

**Screenshots**
Here you can see the price change when we implemented `v1.120.3` for the skuname fix.
Image
Image

Here are the rates as shown on Azure Pricing.
Image

1.120.3

**Additional context**

- Cloud: Azure (AKS, Central US)
- Observed rate: `1.40e-05/GiB/hr` ($0.010/GiB/month) — matches `P4 LRS Disk Mount` ($0.32/mo ÷ 730 ÷ 32)
- Correct rate: `2.26e-04/GiB/hr` — from `P4 LRS Disk` ($5.2795/mo ÷ 730 ÷ 32)

sawyerward:
## Description
Fix Azure Premium SSD PV pricing from the legacy rate card by ignoring `P4 LRS Disk Mount` meters when deriving the `premium_ssd` storage class price.

Previously, `convertMeterToPricings()` matched both `P4 LRS Disk` and `P4 LRS Disk Mount` via the `"P4 "` prefix. Because the mount meter can be processed after the disk meter, it could overwrite the real disk price and underprice Premium SSD PVs.

## Related Issues
Fixes #3870

## User Impact
Azure AKS users with Premium SSD managed disk PVs, including `managed-csi-premium` StorageClasses using `skuname: Premium_LRS`, will see PV hourly costs based on the actual `P4 LRS Disk` meter instead of the much lower disk mount fee.

This is not a breaking change. It corrects underreported Azure Premium SSD PV costs.

## Testing
- Added a regression test covering `P4 LRS Disk` followed by `P4 LRS Disk Mount`, matching the overwrite scenario.
- Verified the resulting `centralus,premium_ssd` price uses the disk meter rate.
- Ran:

```bash
go test ./pkg/cloud/azure

### CODE REVIEW

File: pkg/cloud/azure/provider.go [hunk 1]
```diff
@@ -1005,7 +1005,7 @@ func convertMeterToPricings(info commerce.MeterInfo, regions map[string]string,
 	if strings.Contains(meterCategory, "Storage") {
 		if strings.Contains(meterSubCategory, "HDD") || strings.Contains(meterSubCategory, "SSD") || strings.Contains(meterSubCategory, "Premium Files") {
 			var storageClass string = ""
-			if strings.Contains(meterName, "P4 ") {
+			if strings.Contains(meterName, "P4 ") && !strings.Contains(meterName, "Disk Mount") {
```
ameijer: @sawyerward thanks for the good docs of the problem. Are you concerned that with this change, we will now miss disk mount pricing? should that be combined with disk pricing? what would you suggest?

sawyerward: Hmm good point. The mount costs only come into the equation when the disk is a [shared disk](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-shared). I supposed a broader fix would be to track and calculate the number of shares. I'm not as familiar with shared disk but am happy to look into it.

ameijer: remind me here - the meter is for your actually incurred costs, right? as in, this function isn't called UNLESS you actually have P4 disks in your cluster, etc? 

If that is the case, maybe a quick fix here is to catch the disk mount case and emit a warning log. This seems edge casey and would hate to send you off on some large undertaking for something we don't even know is actually used

sawyerward: yeah the meters are for incurred cost. They're used for when the app tries to look up the rate for something. Honestly for the purpose of kubernetes, shared disks seem like a very small edge case.

I would be of the opinion that this get merged to fix the issue we have now where the P4 disk rates are incorrect (grabbing per-mount cost instead of base disk price). If the disk mount share is worth adding, we do it as a separate issue/PR. This PR fixes the current broken state of P4 pricing. 😄 

Some info on the billing implications. 
https://learn.microsoft.com/en-us/azure/virtual-machines/disks-shared#billing-implications

A screenshot from the [Azure Pricing](https://azure.microsoft.com/en-us/pricing/details/managed-disks/). We see the base disk rate as well as the additional per-mount rate if that volume is shared across multiple VMs. Again, for kubernetes, this seems like an edge case or something overly burdensome to enable. I imagine most disks through kubernetes are via PVC and don't have the multi-mount features configured.
Screenshot 2026-06-22 at 8 30 18 AM

ameijer: I agree with you - I'm just wondering if a good "compromise" is a WARN or ERROR log letting the user know we don't support that? hopefulyl not more than 5 mins to add?

sawyerward: yeah, the easiest way to add this is to log something when the rate cards are being processed and we skip over the mount costs.

as a side note, the more I dig into the storage pricing, the more I see some issues. The whole rate card matching uses those P4, E4, S4 rates for those storage types even though other tiers (P20, P30, etc are billed at different per-GB rates than P4). Also, it calculates PVC/storage by multiplying the per-GB rate by the PVC size though on Azure's side, you get billed at the tier rate.

I'll do a bit more thinking on how to correct this at a higher level. So far I've just been chasing bugs as I come across them but even this fix is covering up some issues.

sawyerward: @ameijer here is the new PR with the pricing overhaul. A bigger change than I was hoping to do with my first contribution, but it should resolve all PV pricing issues for Azure.
https://github.com/opencost/opencost/pull/3920

PR #3875, opened 2026-06-22
KubeModel Changes

Sean-Holcomb:
Description
  
  Advances the KubeModel feature by completing the data model, compute pipeline, and storage/export infrastructure needed to produce a fully-populated KubeModelSet from live cluster data.

  Model (core/pkg/model/kubemodel/) — Replaces the previous generic Device/DeviceUsage/Owner types with properly-typed resource structs for every Kubernetes workload and storage kind: Deployment,
  StatefulSet, DaemonSet, Job, CronJob, ReplicaSet, Service, PersistentVolume, PersistentVolumeClaim, ResourceQuota, and DCGM device. Removes the protobuf-generated codecs and replaces them with a
  bingen-based binary codec (kubemodel_codecs.go) that supports both in-memory and streaming serialization with a pre-pass string table for compression.

  Compute pipeline (core/pkg/compute/kubemodel/) — Splits the monolithic kubemodel.go compute file into one file per resource type (cluster, node, namespace, pod, container, deployment, statefulset,
  daemonset, job, cronjob, replicaset, service, pv, pvc, resourcequota, dcgm). Each file owns the query fan-out and assembly for that resource.
  
  New MetricsQuerier endpoints — Adds QueryKMPVInfo, QueryPVCBytesUsedAverage, and QueryPVCBytesUsedMax to the interface, with implementations in both the Prometheus and collector data sources.
  
  Storage layer (pkg/kubemodel/) — Adds a Querier for reading stored KubeModelSet windows, a Janitor for pruning old files by retention policy, and a Pipeline that wires compute → export.
  
  Metrics (pkg/metrics/kubemodel.go) — Exposes KubeModel-backed Prometheus metrics for nodes, namespaces, pods, deployments, statefulsets, daemonsets, jobs, cronjobs, replicasets, and resource quotas.
  
  Exporter — NewComputePipelineExporter now accepts a ComputeExporterConfig struct. NewPipelinesExportConfig gains exportLegacy and exportKubeModel boolean flags for explicit pipeline opt-in.
  
  Related Issues

  Relates to KCM-5295
  
  User Impact

  No user-facing API changes. The KubeModel pipeline is opt-in via the exportKubeModel flag in NewPipelinesExportConfig and is not enabled by default. 
  
  Testing

  - Unit tests added for all new model types (Register/Validate for each resource kind)
  - Round-trip codec tests for KubeModelSet covering both the in-memory (MarshalBinary/UnmarshalBinary) and streaming pre-pass (MarshalBinaryTo/UnmarshalBinaryFromReader) paths
  - Janitor and Querier unit tests in pkg/kubemodel/
  - Exporter integration tests updated to reflect the new ComputeExporterConfig API
  - All existing unit tests pass (just test-core, just test-opencost)


pkg/cloud/provider/cloud_test.go
Sean-Holcomb, line 1: Removed repetitive ClusterCache Mock

pkg/costmodel/allocation.go
Sean-Holcomb, line 1: The names of these queries were updated to better match their actual functionality

core/pkg/compute/kubemodel/container.go
```diff
@@ -0,0 +1,172 @@
...
+			log.Warnf("container %s/%s has not been initialized to add RAM usage max", res.UID, res.Container)
+			continue
+		}
+		if len(res.Data) > 0 {
+			container.RAMBytesUsageMax = res.Data[0].Value
+		}
+	}
+
+	for _, container := range containerMap {
+		err := kms.RegisterContainer(container)
```
ameijer, line 165: I remember registering is to build indexes and etc. do we actually use these? can these be removed?

core/pkg/compute/kubemodel/cronjob.go
```diff
@@ -0,0 +1,71 @@
+package kubemodel
```
ameijer, line 1: package wide ask - it would be great to see unit tests with mocked out prom responses/futures, and watch those get built into respective kubemodel objects

core/pkg/compute/kubemodel/cronjob.go
```diff
@@ -0,0 +1,71 @@
...
+	for _, res := range cronJobLabelsResult {
+		cronJob, ok := cronJobMap[res.UID]
+		if !ok {
+			log.Warnf("cronjob with UID '%s' has not been initialized to add labels", res.UID)
+			continue
+		}
+		cronJob.Labels = res.Labels
+	}
+
+	cronJobAnnotationsResult, _ := cronJobAnnotationsResultFuture.Await()
```
ameijer, line 53: general question - we allow people to disable annotations for performance reasons. should we provide an off/short-circuit mechanism here for that?

pkg/metrics/nodemetrics_test.go
```diff
@@ -436,19 +436,3 @@ func TestGetConditions(t *testing.T) {
 		})
 	}
 }
-
-// FakeNodeCache implements ClusterCache interface for testing
-type FakeNodeCache struct {
```
ameijer, line 441: nice LoC savings from this mock.

pkg/metrics/kubemetrics.go
```diff
@@ -65,6 +71,12 @@ func InitKubeMetrics(clusterCache clustercache.ClusterCache, metricsConfig *Metr
 			)
 		}
 
+		prometheus.MustRegister(KubeModelCollector{
+			KubeClusterCache: clusterCache,
+			ClusterInfo:      clusterInfo,
+			metricsConfig:    *metricsConfig,
+		})
```
ameijer, line 78: is this so that we can support the prom collector? does this introduce another cycle of us reading from prom our own metrics?

pkg/costmodel/cluster.go
```diff
@@ -764,7 +764,7 @@ func pvCosts(
 
 		// TODO niko/assets storage class
```
ameijer, line 765: unrelated to this - but @nikovacevic is this comment dead?

pkg/cloud/provider/cloud_test.go
ameijer: love it

core/pkg/source/datasource.go
```diff
@@ -31,19 +34,33 @@ type MetricsQuerier interface {
 	QueryNodeIsSpot(start, end time.Time) *Future[NodeIsSpotResult]
 	QueryNodeRAMSystemPercent(start, end time.Time) *Future[NodeRAMSystemPercentResult]
 	QueryNodeRAMUserPercent(start, end time.Time) *Future[NodeRAMUserPercentResult]
+	QueryNodeResourceCapacities(start, end time.Time) *Future[ResourceResult]
+	QueryNodeResourcesAllocatable(start, end time.Time) *Future[ResourceResult]
```
ameijer, line 249: should we also grab a nodes public IP? that is a known cloduy feature KCM doesn't have

core/pkg/source/datasource.go
```diff
@@ -10,17 +10,20 @@ import (
...
-	QueryPVUsedAverage(start, end time.Time) *Future[PVUsedAvgResult]
-	QueryPVUsedMax(start, end time.Time) *Future[PVUsedMaxResult]
 
 	// Local Cluster Disks
 	QueryLocalStorageActiveMinutes(start, end time.Time) *Future[LocalStorageActiveMinutesResult]
 	QueryLocalStorageUsedAvg(start, end time.Time) *Future[LocalStorageUsedAvgResult]
 	QueryLocalStorageUsedMax(start, end time.Time) *Future[LocalStorageUsedMaxResult]
 	QueryLocalStorageBytes(start, end time.Time) *Future[LocalStorageBytesResult]
 
+	QueryKMLocalStorageUsedAvg(start, end time.Time) *Future[NodeUIDValueResult]
```
ameijer, line 231: can you add comment to this section explaining what this group is?

core/pkg/source/datasource.go
```diff
@@ -70,19 +87,79 @@ type MetricsQuerier interface {
...
+	QueryDCGMDeviceInfo(start, end time.Time) *Future[DCGMDeviceInfoResult]
+	QueryDCGMDeviceUptime(start, end time.Time) *Future[DCGMDeviceUptimeResult]
+	QueryDCGMContainerUsageAvg(start, end time.Time) *Future[DCGMDeviceContainerUsageResult]
+	QueryDCGMContainerUsageMax(start, end time.Time) *Future[DCGMDeviceContainerUsageResult]
+
 	// PVC
 	QueryPodPVCAllocation(start, end time.Time) *Future[PodPVCAllocationResult]
 	QueryPVCBytesRequested(start, end time.Time) *Future[PVCBytesRequestedResult]
 	QueryPVCInfo(start, end time.Time) *Future[PVCInfoResult]
+	QueryKMPVCInfo(start, end time.Time) *Future[PVCInfoResult]
```
ameijer, line 313: this func sig looks confusing. a different function with the same return type and params?

core/pkg/model/kubemodel/dcgm.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// DCGMDevice holds recording from the DCGM exporter which provides identification and usage metrics for
+// Nvidia gpu. These Nvidia devices can be incorporated into the cluster via k8s Device Plugin API or DRAs.
+// While the DCGM exporter does provide unique identifiers for the containers that it is reporting metrics on,
+// It is split out here to provide some isolate from the rest of the KubeModel which represent universal structures
```
ameijer:
```suggestion
// It is split out here to provide some isolation from the rest of the KubeModel which represent universal structures
```

core/pkg/model/kubemodel/node.go
```diff
@@ -5,99 +5,60 @@ import (
...
-	CapacityBytes    Measurement `json:"capacityBytes"`    // Total capacity of the volume in bytes
-	UsageByteSeconds Measurement `json:"usageByteSeconds"` // Cumulative usage (Byte × seconds) over measurement window
-	VolumeType       string      `json:"volumeType"`       // "root" for primary disk, "persistent" for additional PVs
-	ProviderID       string      `json:"providerId"`       // Cloud provider volume ID (e.g., "vol-xxxxx" for AWS EBS)
-	DurationSeconds  Measurement `json:"durationSeconds"`  // Duration the volume was attached during measurement window in seconds
+// @bingen:generate:FileSystem
+// FileSystem records information for a nodes local storage
+type FileSystem struct {
+	CapacityBytes float64 `json:"capacityBytes"` // Total capacity of the volume in bytes
+	UsageByteAvg  float64 `json:"usageByteAvg"`
```
ameijer, line 28: should usage be tracked at the container level? or is this for like OS usage etc

core/pkg/model/kubemodel/dcgm.go
```diff
@@ -0,0 +1,66 @@
...
+// from the k8s API. It is left to the end user to interpret the relationships to the rest of the cluster based on
+// container unique identifiers
+// @bingen:generate:DCGMDevice
+type DCGMDevice struct {
+	UUID      string             `json:"uuid"`
+	Start     time.Time          `json:"start"`
+	End       time.Time          `json:"end"`
+	Device    string             `json:"device"`
+	ModelName string             `json:"modelName"`
+	PodUsages map[string]DCGMPod `json:"podUsages"`
```
ameijer, line 21: For discussion - it feels like kubecost mostly uses a 'many to one' graph where each pod has a pointer to its namespace. DCGM device in particular (but can be seen in other objects) feels like it is 'one to many', i.e., we need to traverse this data structure to get pointers, which could be many, then use them. 

I'm just trying to envision the use case of "ok, how much GPU did my pod use, and how much available GPU did there exist"? Becuase that is how we start to root out idle

core/pkg/model/kubemodel/dcgm.go
```diff
@@ -0,0 +1,66 @@
...
+
+// @bingen:generate:DCGMPod
+type DCGMPod struct {
+	ContainerUsages map[string]DCGMContainer `json:"container-usages"`
+}
+
+// @bingen:generate:DCGMContainer
+type DCGMContainer struct {
+	UsageAvg float64 `json:"usageAvg"`
+	UsageMax float64 `json:"usageMax"`
```
ameijer, line 32: I wonder if we convert these to histogram types. We know that GPUs are going to be red hot here. We also know that while utilization perhaps could be a 2^64 different values between 0 and 1, I think we would be OK saying it could be 100 discrete values. What if we changed this to a type that is a 100 long array of 8 bit numbers(or some optimized data structure containing that data) and collector that sampled the GPU 250 times over the 10 min interval. the bin corresponding to the % utilization would get incremented.

core/pkg/model/kubemodel/dcgm.go
ameijer: to an extent, it feels like we are asking the sama data structure to track its usage AND capacity.

core/pkg/model/kubemodel/dcgm.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// DCGMDevice holds recording from the DCGM exporter which provides identification and usage metrics for
+// Nvidia gpu. These Nvidia devices can be incorporated into the cluster via k8s Device Plugin API or DRAs.
+// While the DCGM exporter does provide unique identifiers for the containers that it is reporting metrics on,
+// It is split out here to provide some isolate from the rest of the KubeModel which represent universal structures
+// from the k8s API. It is left to the end user to interpret the relationships to the rest of the cluster based on
+// container unique identifiers
+// @bingen:generate:DCGMDevice
+type DCGMDevice struct {
```
ameijer, line 15: could we have some pseudocode explaining how the fields in the product interface @ https://ibm-my.sharepoint.com/:w:/r/personal/alexander_meijer_ibm_com/Documents/High%20Level%20Kubemodel%20Device%20Thoughts.docx?d=w2335cd9be49e48ceafc4b7b4307b7e1b&csf=1&web=1&e=0YFV6h can be derived from the metrics in this object ?

core/pkg/compute/kubemodel/kubemodel.go
```diff
@@ -0,0 +1,148 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/model/kubemodel"
+	"github.com/opencost/opencost/core/pkg/source"
+)
+
+const logTimeFmt string = "2006-01-02T15:04:05"
```
ameijer: feels like an odd place for this, is there some log/utils package we should be using?

core/pkg/compute/kubemodel/kubemodel.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// 2.1 Compute Cluster
+	err = km.computeCluster(kms, start, end)
+	if err != nil {
+		kms.Error(err)
+		return kms, fmt.Errorf("error computing kubemodel.Cluster for (%s, %s): %w", start.Format(logTimeFmt), end.Format(logTimeFmt), err)
+	}
+
+	// 2.2 Compute Nodes
+	err = km.computeNodes(kms, start, end)
```
ameijer: I wonder, for maintainability, if you keep this funcs in a list an loop over them? save a few repetitive loc?

modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -33,24 +35,50 @@ func newClusterCacheScraper(clusterCache clustercache.ClusterCache) Scraper {
...
+	pods := ccs.clusterCache.GetAllPods()
+	pvcs := ccs.clusterCache.GetAllPersistentVolumeClaims()
+	pvs := ccs.clusterCache.GetAllPersistentVolumes()
+	services := ccs.clusterCache.GetAllServices()
+	statefulSets := ccs.clusterCache.GetAllStatefulSets()
+	daemonSets := ccs.clusterCache.GetAllDaemonSets()
+	jobs := ccs.clusterCache.GetAllJobs()
+	cronJobs := ccs.clusterCache.GetAllCronJobs()
+	replicaSets := ccs.clusterCache.GetAllReplicaSets()
+	resourceQuotas := ccs.clusterCache.GetAllResourceQuotas()
```
ameijer, line 51: @mbolt35 is there any opportunity here to escape the concurrent map access issues we see in OC at a certain scale? my heart says yes but my head says no

ameijer: what if we even went so far as to instantiate another kube client?

modules/collector-source/pkg/scrape/index.go
```diff
@@ -0,0 +1,48 @@
...
+func buildPVCIndex(pvcs []*clustercache.PersistentVolumeClaim) map[pvcKey]types.UID {
+	m := make(map[pvcKey]types.UID, len(pvcs))
+	for _, pvc := range pvcs {
+		m[pvcKey{name: pvc.Name, namespace: pvc.Namespace}] = pvc.UID
+	}
+	return m
+}
+
+// buildPVIndex returns a map from PV name to UID.
+func buildPVIndex(pvs []*clustercache.PersistentVolume) map[string]types.UID {
```
ameijer, line 42: neat. I like this paradigm

core/pkg/model/kubemodel/owner.go
```diff
@@ -16,43 +18,30 @@ const (
...
+	default:
+		log.Warnf("failed to find owner kind for '%s'", kind)
+		return OwnerKind(strings.ToLower(kind))
 	}
+}
 
-	return nil
+// @bingen:generate:Owner
+// Owner represents a Kubernetes resource owner (workload controller)
+type Owner struct {
```
ameijer, line 43: I'm happy we kept this!

core/pkg/model/kubemodel/pod.go
```diff
@@ -5,43 +5,61 @@ import (
...
-	DurationSeconds      Measurement       `json:"durationSeconds"`
-	NetworkTransferBytes Measurement       `json:"networkTransferBytes"`
-	NetworkReceiveBytes  Measurement       `json:"networkReceiveBytes"`
-	Start                time.Time         `json:"start,omitempty"` // Pod creation/start timestamp
-	End                  time.Time         `json:"end,omitempty"`   // Pod deletion/end timestamp (nil if still running)
+	UID                   string                 `json:"uid"`
+	NamespaceUID          string                 `json:"namespaceUid"`
+	NodeUID               string                 `json:"nodeUid"`
+	Name                  string                 `json:"name"`
+	Owners                []Owner                `json:"owners"`
```
ameijer, line 20: is this a list, not a hierarchy?

core/pkg/model/kubemodel/networktrafficdetail.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// @bingen:generate:NetworkTrafficDetail
+type NetworkTrafficDetail struct {
+	PodUID           string           `json:"podUid"`
+	Endpoint         string           `json:"endpoint,omitempty"`
+	TrafficDirection TrafficDirection `json:"trafficDirection"`
+	TrafficType      TrafficType      `json:"trafficType"`
+	IsNatGateway     bool             `json:"isNatGateway"`
+	Bytes            float64          `json:"bytes"`
```
ameijer, line 27: interesting. so network insights has been fused into kube model. I am supportive

core/pkg/compute/kubemodel/cronjob.go
Sean-Holcomb: What is the name of the config

pkg/metrics/kubemetrics.go
Sean-Holcomb: Yes it is that. That has always been a requirement of prom. I believe it has been more problematic in the past when we have emitted metrics based on metrics being read in from prometheus

core/pkg/model/kubemodel/pod.go
Sean-Holcomb: It is a list, that is how it is defined in the K8s spec. Realistically I have only ever seen one owner with `controller==true`

core/pkg/model/kubemodel/networktrafficdetail.go
Sean-Holcomb: Yeah we are getting rid of the Network Insight pipeline too. Not that anyone necessarily cares, but once all these PR's go, in reconciled network insights will be trivial to implement.

core/pkg/compute/kubemodel/container.go
nikovacevic: I think that the `container` doesn't actually get set on the `KubeModelSet` at all until it's registered.

core/pkg/env/core.go
```diff
@@ -61,3 +64,11 @@ func IsPProfEnabled() bool {
...
 	return Get(InstallNamespaceEnvVar, def)
 }
+
+func GetExportLegacyDataModel() bool {
+	return GetBool(ExportLegacyDataModelEnvVar, true)
+}
+
+func GetExportKubeModel() bool {
+	return GetBool(ExportKubeModelEnvVar, true)
+}
```
nikovacevic: Default here `false`?

nikovacevic: Or should we talk that over?

core/pkg/model/kubemodel/bingen.go
```diff
@@ -20,4 +20,6 @@ package kubemodel
 //
 ////////////////////////////////////////////////////////////////////////////////
 
-//go:generate bingen -package=kubemodel -version=1 -buffer=github.com/opencost/opencost/core/pkg/util
+// @bingen:define[string]:github.com/opencost/opencost/core/pkg/model/shared.Provider
+
+//go:generate bingen -package=kubemodel -version=2 -buffer=github.com/opencost/opencost/core/pkg/util
```
nikovacevic, line 25: Trying to remember... any side-effects to `-version=1` > `-version=2` that we should be aware of?

core/pkg/model/kubemodel/container.go
```diff
@@ -5,105 +5,52 @@ import (
...
-	VolumeStorageByteUsageMax  map[string]Measurement `json:"volumeStorageByteUsageMax,omitempty"`
-	CpuMillicoreLimitSeconds   Measurement            `json:"cpuMillicoreLimitSeconds,omitempty"`
-	RAMByteSecondsLimit        Measurement            `json:"ramByteSecondsLimit,omitempty"`
-	Start                      time.Time              `json:"start"`
-	End                        time.Time              `json:"end"`
+	PodUID            string             `json:"podUid"`
+	Name              string             `json:"name"`
+	ResourceRequests  ResourceQuantities `json:"resourceRequests"`
+	ResourceLimits    ResourceQuantities `json:"resourceLimits"`
+	CPUCoresAllocated float64            `json:"cpuCoresAllocated"`
```
nikovacevic: `CPUCoresAllocated` -> `CPUCoreAllocationAvg`? This is an average right? We would need `CoreHours` again to make it not an average?

core/pkg/model/kubemodel/container.go
```diff
@@ -5,105 +5,52 @@ import (
...
+	CPUCoreUsageAvg   float64            `json:"cpuCoreUsageAvg"`
+	CPUCoreUsageMax   float64            `json:"cpuCoreUsageMax"`
+	RAMBytesAllocated float64            `json:"ramBytesAllocated"`
```
nikovacevic: Ditto above

core/pkg/model/kubemodel/daemonset.go
```diff
@@ -0,0 +1,59 @@
...
+
+// @bingen:generate:DaemonSet
+// DaemonSet represents a Kubernetes DaemonSet resource
+type DaemonSet struct {
+	UID              string            `json:"uid"`
+	NamespaceUID     string            `json:"namespaceUid"`
+	Name             string            `json:"name"`
+	Labels           map[string]string `json:"labels,omitempty"`
+	Annotations      map[string]string `json:"annotations,omitempty"`
+	DevicePluginInfo map[string]string `json:"devicePluginInfo"`
```
nikovacevic: Any docs on this? Does Cloudy use it? Or is it for GPUs?

nikovacevic (commented):
Not quite done reviewing, but dropping what comments I have for now.

core/pkg/model/kubemodel/bingen.go
Sean-Holcomb: The only concern is that if the agent is producing v2 then the ingestor needs to know about v2

core/pkg/env/core.go
Sean-Holcomb: It sounds like people want false. The one thing that concerns me about is that right now it is a simple straight forward control, turn it off and the kubemodel stops turn it on and you get kubemodel. The issue is that we are already producing kubemodel for resource quotas. So default false will turn that off.

core/pkg/model/kubemodel/node.go
Sean-Holcomb: We could track this at the container level, but we don't actually have any reason to at this point.

core/pkg/clustercache/helper.go
```diff
@@ -0,0 +1,15 @@
...
+
+func GetLoadBalancerIngressAddress(service *Service) []string {
+	var addresses []string
+	for _, loadBalancerIngress := range service.Status.LoadBalancer.Ingress {
+		address := loadBalancerIngress.IP
+		// Some cloud providers use hostname rather than IP
+		if address == "" {
+			address = loadBalancerIngress.Hostname
+		}
+		addresses = append(addresses, address)
```
mbolt35, line 11: Naive question here, but do all Services contain LoadBalancers? Is there are possibility of `LoadBalancer.Ingress` throwing a nil panic?

core/pkg/model/kubemodel/dcgm.go
Sean-Holcomb: I need a little more elaboration on what you are proposing here.

core/pkg/compute/kubemodel/cronjob.go
mbolt35: There are configs I think on whether or not we emit annotation metrics: 

https://github.com/opencost/opencost/blob/1741baaf3682445acba7dfd28c5c5a2a01d4006b/pkg/env/costmodel.go#L66-L67

Not sure if we want to look at these on consume or not. There also may exist a config I'm not familiar with.

core/pkg/clustercache/helper.go
Sean-Holcomb: `LoadBalancer` here is while "optional" in the k8s spec is not a pointer. So no risk of a nil panic

core/pkg/model/kubemodel/daemonset.go
Sean-Holcomb: This Is GPU stuff, there is a particular type of GPU that requires us to know an arg on the the container for the daemon set.

core/pkg/env/core.go
```diff
@@ -61,3 +64,11 @@ func IsPProfEnabled() bool {
 func GetInstallNamespace(def string) string {
 	return Get(InstallNamespaceEnvVar, def)
 }
+
+func GetExportLegacyDataModel() bool {
+	return GetBool(ExportLegacyDataModelEnvVar, true)
+}
+
+func GetExportKubeModel() bool {
```
mbolt35: Nit: `bool` return values should use an `Is` prefix instead of `Get`

core/pkg/compute/kubemodel/cronjob.go
ameijer: // IsEmitNamespaceAnnotationsMetric returns true if cost-model is configured to emit the kube_namespace_annotations metric
// containing the namespace annotations
func IsEmitNamespaceAnnotationsMetric() bool {
	return env.GetBool(EmitNamespaceAnnotationsMetricEnvVar, false)
}

// IsEmitPodAnnotationsMetric returns true if cost-model is configured to emit the kube_pod_annotations metric containing
// pod annotations.
func IsEmitPodAnnotationsMetric() bool {
	return env.GetBool(EmitPodAnnotationsMetricEnvVar, false)
}

Sean-Holcomb: Okay so there are a few things here. Those env vars are related to general metric emission. I don't think we really need to do anything with the consumption side of things if we want to say that these env vars relate to both prom metrics and collector scrapes. I like the idea. It also gets at something I have been thinking about which is that there should be some sort of concrete struct that traverses clustercache and produces metrics/updates which is shared by pkg `metrics` and the collector's scraper that definitely out of scope for now but what do you guys think

Sean-Holcomb: This one is turning into a bit of a rabbit hole. For now I am just adding a filter which responds to environment variables. But this really is the type of thing that should be in a config map (which we have half an implementation for) I have set all the new label and annotation metrics to false by default. I wont be adding these env vars to the helm chart

core/pkg/source/datasource.go
Sean-Holcomb: We discussed and we are punting on it.

core/pkg/model/kubemodel/dcgm.go
Sean-Holcomb: Adding ignore to dcgm field for now

nikovacevic (approved):
Thanks, @Sean-Holcomb. Huge effort. Great job seeing it through.

PR #3876, opened 2026-06-22
Pricing changeset

Sean-Holcomb:
## Description

This PR introduces a new modular pricing architecture to OpenCost, replacing the existing monolithic cloud-provider-specific pricing logic with a composable, interface-driven system.

**New `core/pkg/pricing` package** — defines the canonical pricing interfaces and data types:
- `PricingSource` / `PricingModule` interfaces covering cluster, network, node, persistent volume, and service pricing
- `PricingSet` with stable `Checksum()` support for drift detection
- Type definitions: `NodePricing`, `PersistentVolumePricing`, `NetworkPricing`, `ClusterPricing`, `ServicePricing`
- Supporting types: `Price`, `Prices`, `VolumeType`, `CommitmentType`, `ProvisioningType`

**New `core/pkg/reader` package** — generic streaming reader abstractions:
- `Reader[T]` interface with `SliceReader`, `FuncReader`, and `JSONLinesReader` implementations
- Benchmarked `JSONLinesReader` for high-throughput pricing data ingestion

**New `core/pkg/unit` package** — typed `Currency` and dimensional unit helpers

**New `modules/pricing/basic` module** — a provider-agnostic `PricingModule` implementation backed by a `PricingStore`, with default fallback pricing

**New `modules/pricing/public` module** — fetches live pricing from AWS, Azure, and GCP APIs and serialises to JSONL:
- AWS Price List API integration (`pkg/cloud/aws/pricelistapi.go` also added to the main module)
- Azure Retail Prices API integration
- GCP Cloud Billing Catalog API integration
- `fetch-pricing` CLI (`cmd/main.go`) with `--currency` and `--compare` flags for both USD and CNY outputs
- Pre-bundled `usd/` and `cny/` JSONL artefacts (nodes + persistent volumes) for offline use

**New CI workflow** — `compare-pricing-data.yaml` runs daily on a cron schedule to detect upstream pricing drift across all three providers and both currencies, exiting `2` on drift vs `1` on infrastructure errors.

**Small supporting additions:**
- `core/pkg/model/pricingmodel/pricingmodel.go` — shared pricing model helpers
- `core/pkg/model/shared/usagetype.go` / `provider.go` — additional usage-type and provider values
- `core/pkg/pipelines/name.go` — pipeline name constants
- `core/pkg/source/decoders.go` — decoder updates

## User Impact

This is an additive change — no existing APIs or configuration are removed. The new pricing module is not yet wired into the main cost model pipeline in this PR; that integration will follow.

The pre-bundled JSONL pricing data in `modules/pricing/public/usd/` and `modules/pricing/public/cny/` enables offline pricing lookups without any cloud provider credentials at runtime.

No breaking changes.

## Testing

- Unit tests added for all new `core/pkg/pricing` types (`pricingset_test.go`, `memory_test.go`, `storage_test.go`, `mock_test.go`)
- Unit tests added for `core/pkg/reader` (`func_test.go`, `jsonlines_test.go`, `slice_test.go`)
- Unit tests added for `core/pkg/unit` (`currency_test.go`, `unit_test.go`)
- Unit tests added for each cloud provider pricing source (`awspricingsource_test.go`, `azurepricingsource_test.go`, `gcppricingsource_test.go`, `pricelistapi_test.go`)
- `modules/pricing/basic` has full module-level tests (`module_test.go`)
- YAML fixture pricing data (`core/pkg/pricing/test/{aws,azure,gcp,default}.yaml`) used as test inputs
- Benchmarks added for `JSONLinesReader`
- Manual: `fetch-pricing --currency USD --compare` validated against bundled JSONL artefacts


core/pkg/cloud/provider.go
```diff
@@ -2,6 +2,10 @@ package shared
 
 import "strings"
 
+// TODO: reconsider "shared" as a package name
+// TODO: for this file, maybe core/pkg/model/cloud?
+// TODO: maybe even core/pkg/cloud?
+
```
Sean-Holcomb, line 8: Do we want to change this?

core/pkg/model/pricingmodel/pricingmodel.go
Sean-Holcomb, line 1: removing

core/pkg/model/shared/usagetype.go
Sean-Holcomb, line 1: removing

core/pkg/cloud/provider.go
nikovacevic: I would change it, but I don't consider it absolutely necessary. Options I would consider:
* `core/pkg/cloud` -- `cloud.Provider` (does it belong in a `/model/` dir?)
* `core/pkg/model/cloud` -- `cloud.Provider`
* `core/pkg/model` -- `model.CloudProvider`

core/pkg/pricing/commitment.go
```diff
@@ -0,0 +1,8 @@
+package pricing
+
+type CommitmentType string
```
nikovacevic: This is unused. I recommend we keep it -- but we should be aware that everything needs to be `CommitmentNone` for now!

core/pkg/pricing/node.go
```diff
@@ -0,0 +1,86 @@
...
+func (np *NodePricing) String() string {
+	return np.Properties.String() + "|" + np.Prices.canonical()
+}
+
+type NodePricingProperties struct {
+	Provider     shared.Provider   `json:"provider,omitempty" yaml:"provider,omitempty"`
+	Region       string            `json:"region,omitempty" yaml:"region,omitempty"`
+	InstanceType string            `json:"instanceType,omitempty" yaml:"instanceType,omitempty"`
+	Provisioning ProvisioningType  `json:"provisioning,omitempty" yaml:"provisioning,omitempty"`
+	Commitment   CommitmentType    `json:"commitment,omitempty" yaml:"commitment,omitempty"`
```
nikovacevic: Similar to `Commitment`, `Provisioning` also needs to be `OnDemand`. If we think these are too risky to keep in without using them, we can remove. But it might be nice to have them plug-and-play for the future. cc @Sean-Holcomb

core/pkg/pricing/source.go
```diff
@@ -0,0 +1,44 @@
...
+
+type NodePricingSource interface {
+	NewNodePricingReader(ctx context.Context) (reader.Reader[*NodePricing], error)
+	GetNodePricing(ctx context.Context, props NodePricingProperties) (*NodePricing, error)
+}
+
+type PersistentVolumePricingSource interface {
+	NewPersistentVolumePricingReader(ctx context.Context) (reader.Reader[*PersistentVolumePricing], error)
+	GetPersistentVolumePricing(ctx context.Context, props PersistentVolumePricingProperties) (*PersistentVolumePricing, error)
+}
```
nikovacevic, line 35: Nit (and certainly my fault) -- these `Get` funcs are after the `NewReader` funcs, but elsewhere they're in the other order. Also... we don't even use these today.

cc @Sean-Holcomb worth keeping? Or should we remove?

The `GetXPricing` is actually kind of ambiguous. It's not clear if they should perfectly match, or do a "best match" as we do elsewhere. Possibly worth a quick convo.

core/pkg/source/decoders.go
```diff
@@ -2134,13 +2134,31 @@ type DCGMDeviceContainerUsageResult struct {
...
-// Inference Metrics Decoders
+func DecodeDCGMDeviceContainerUsageResult(result *QueryResult) *DCGMDeviceContainerUsageResult {
+	uuid, _ := result.GetString(UUIDLabel)
+	podUID, _ := result.GetString(PodUIDLabel)
+	container, _ := result.GetString(ContainerLabel)
+	var value float64
+	if len(result.Values) > 0 {
+		value = result.Values[0].Value
+	} else {
+		log.Warnf("Error decoding DCGM Device Container Udage Result for device '%s': empty value returned", uuid)
```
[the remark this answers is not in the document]
nikovacevic: Good catch, here.

core/pkg/unit/currency.go
```diff
@@ -0,0 +1,60 @@
+package unit
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Currency string
+
+// TODO which of these are supported in the pricing APIs??
```
nikovacevic: We can remove these TODOs -- they are moot now.

core/pkg/unit/unit.go
```diff
@@ -0,0 +1,63 @@
+package unit
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Unit string
```
nikovacevic, line 8: @Sean-Holcomb do you still have aspirations to unify the codebase around a single `Currency` type? Should we discuss?

modules/pricing/basic/default.go
```diff
@@ -0,0 +1,136 @@
...
+const DefaultNetworkNATGatewayIngressPricePerGiB float64 = 0.045
+
+const DefaultNodePricePerVCPUHour float64 = 0.031611
+const DefaultNodePricePerRAMGiBHour float64 = 0.004237
+const DefaultNodePricePerGPUHour float64 = 0.95
+const DefaultNodePricePerLocalDiskGiBHour float64 = 0.0001096
+
+const DefaultPersistentVolumePricePerGiBHour float64 = 0.00005479452
+
+const DefaultServicePricePerHour float64 = 0.025
```
nikovacevic, line 24: Now that Basic Pricing is (going to be) configurable, I think it's OK to leave these hard-coded. But we could consider putting them in a file. Probably no advantage to that IMO.

modules/pricing/basic/go.mod
```diff
@@ -0,0 +1,117 @@
+module github.com/opencost/modules/pricing/basic
```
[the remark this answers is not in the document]
nikovacevic: Very good catch! I'll correct this one.

modules/pricing/basic/store.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/opencost/opencost/core/pkg/pricing"
+	"github.com/opencost/opencost/core/pkg/storage"
+)
+
+const basicPricingFilePath = "basic-pricing.json"
+
+func DefaultBasicPricingStore(ctx context.Context) (pricing.PricingStore, error) {
+	storage := storage.NewFileStorage(env.GetConfigPath())
+
+	store, err := pricing.NewStoragePricingStore(ctx, storage, basicPricingFilePath)
```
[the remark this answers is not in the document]
nikovacevic: Good suggestion. I'll pick this up.

modules/pricing/public/aws/awspricingsource.go
```diff
@@ -0,0 +1,204 @@
...
+			return
+		}
+
+		// TODO: handle currency?
+		// Parse the currency from config, default to USD if invalid
+		// currency, err := unit.ParseCurrency(p.config.CurrencyCode)
+		// if err != nil {
+		// 	log.Warnf("invalid currency code '%s', defaulting to USD: %s", p.config.CurrencyCode, err.Error())
+		// 	currency = unit.USD
+		// }
```
nikovacevic: We can drop this now, right @alexsouthard ?

modules/pricing/public/aws/pricelistapi.go
```diff
@@ -0,0 +1,241 @@
...
+		return fmt.Errorf("bogus fetch of \"%s\": %w", pricingURL, err)
+	}
+
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Warnf("failed to close response body: %v", err)
+		}
+	}()
+
+	dec := json.NewDecoder(resp.Body)
```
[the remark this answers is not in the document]
nikovacevic: Probably worth handling this one @alexsouthard

modules/pricing/public/aws/pricelistapi.go
```diff
@@ -0,0 +1,241 @@
...
+	for {
+		t, err := dec.Token()
+		if err == io.EOF {
+			log.Infof("done loading \"%s\"\n", resp.Request.URL.String())
+			break
+		} else if err != nil {
+			log.Errorf("error parsing response json %v", resp.Body)
+			break
+		}
```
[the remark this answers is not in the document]
nikovacevic: Not as confident in this one. But also it's not as big of a deal, as we're currently planning to check hashes at time of generation. Could take it or leave it.

modules/pricing/public/aws/pricelistapi.go
```diff
@@ -0,0 +1,241 @@
...
+				if err != nil {
+					return err
+				}
+				product := &PriceListEC2Product{}
+
+				err = dec.Decode(&product)
+				if err != nil {
+					log.Errorf("Error parsing response from \"%s\": %v", resp.Request.URL.String(), err.Error())
+					break
+				}
```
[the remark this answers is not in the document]
nikovacevic: Ditto, above. Take it or leave it. Will catch issues at generation time.

modules/pricing/public/aws/pricelistapi.go
```diff
@@ -0,0 +1,241 @@
...
+						return err
+					}
+					offerTerm := &PriceListEC2Term{}
+					err = dec.Decode(&offerTerm)
+					if err != nil {
+						log.Errorf("Error decoding AWS Offer Term: %s", err.Error())
+					}
+
+					handleTerm(offerTerm)
+
```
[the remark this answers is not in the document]
nikovacevic: Probably agree with this one. Would suggest we `continue` if we encounter this.

modules/pricing/public/azure/azurepricingsource.go
```diff
@@ -0,0 +1,302 @@
...
+			continue
+		}
+
+		// TODO: handle currency?
+		// Parse the currency from config, default to USD if invalid
+		// currency, err := unit.ParseCurrency(a.config.CurrencyCode)
+		// if err != nil {
+		// 	log.Warnf("invalid currency code '%s', defaulting to USD: %s", a.config.CurrencyCode, err.Error())
+		// 	currency = unit.USD
+		// }
```
nikovacevic: Drop this too @alexsouthard ?

modules/pricing/public/azure/azurepricingsource.go
```diff
@@ -0,0 +1,302 @@
...
+		if volumeType == pricing.VolumeTypeNil {
+			continue
+		}
+
+		// TODO: handle currency?
+		// currency, err := unit.ParseCurrency(a.config.CurrencyCode)
+		// if err != nil {
+		// 	log.Warnf("invalid currency code '%s', defaulting to USD: %s", a.config.CurrencyCode, err.Error())
+		// 	currency = unit.USD
+		// }
```
nikovacevic: Ditto

modules/pricing/public/azure/azurepricingsource.go
```diff
@@ -0,0 +1,305 @@
...
+func (a *AzurePricingSource) includeItem(item AzurePricingAttributes) bool {
+	if item.ArmSkuName == "" || item.ArmRegionName == "" {
+		return false
+	}
+	if strings.Contains(item.ProductName, "Windows") {
+		return false
+	}
+	skuLower := strings.ToLower(item.SkuName)
+	productLower := strings.ToLower(item.ProductName)
+	if strings.Contains(skuLower, "low priority") {
```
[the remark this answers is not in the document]
nikovacevic: Probably worth accepting

modules/pricing/public/cmd/main.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	currency string
+	compare  bool
+)
+
+// Assumes execution from the /cmd directory
+const outputDir = "../%s"
```
nikovacevic: Nit, as we'll be running this internally... but I wonder if there's a way around hard-coding the output dir this way?

modules/pricing/public/gcp/gcppricingsource.go
```diff
@@ -0,0 +1,340 @@
...
+			g.parseVolumeSKU(sku, volumeCosts)
+			continue
+		}
+
+		if isComputeResource(resourceGroup) {
+			g.parseComputeSKU(sku, usageType, nodeCPUCosts, nodeRAMCosts)
+			continue
+		}
+
+		// TODO: Add GPU pricing support
```
nikovacevic, line 148: Seems OK to leave this TODO. Not required in the first pass.

modules/pricing/public/gcp/gcppricingsource.go
```diff
@@ -0,0 +1,340 @@
...
+		// Skip if we don't have both CPU and RAM costs
+		if cpuCost == 0 || ramCost == 0 {
+			continue
+		}
+
+		// Determine provisioning type
+		// NOTE: Both preemptible and spot map to spot type
+		provisioning := pricing.ProvisioningOnDemand
+		if key.UsageType == "preemptible" || key.UsageType == "spot" {
+			provisioning = pricing.ProvisioningSpot
```
nikovacevic: It strikes me that we may bust Checksums a lot, if spot instances change pricing frequently. Which... I imagine they do? Open to others' thoughts here.

1. I think it's totally OK to update spot pricing every release. It's just an estimated figure.
2. Alternatively, we could ignore spot now and just support on demand? Spot could be a more "live" module in the future.

cc @Sean-Holcomb

modules/pricing/public/generator.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/opencost/opencost/core/pkg/log"
+	"github.com/opencost/opencost/core/pkg/pricing"
+	"github.com/opencost/opencost/core/pkg/unit"
+	"github.com/opencost/opencost/modules/pricing/public/aws"
+	"github.com/opencost/opencost/modules/pricing/public/azure"
+	"github.com/opencost/opencost/modules/pricing/public/gcp"
+)
+
+// GenerateAWSPricing fetches AWS pricing data in the specified currency
+func GenerateAWSPricing(currency unit.Currency) (*pricing.PricingSet, error) {
```
nikovacevic, line 16: Would love to see a demo of generation! cc @alexsouthard maybe we can set up time to do that this week?

pkg/cloud/aws/fargate.go
```diff
@@ -73,7 +73,7 @@ func (f *FargatePricing) Initialize(nodeList []*clustercache.Node) error {
 		return fmt.Errorf("pricing download failed: status=%d", resp.StatusCode)
 	}
 
-	var pricing AWSPricing
+	var pricing PriceListEC2Response
```
nikovacevic: I'm surprised this changed. Anyone know why? Doesn't look like we updated the HTTP URL that we're pinging to hydrate Fargate prices. Was the old struct out-of-date, and the new one better decodes data from the same URL?

pkg/cloud/aws/pricelistapi.go
```diff
@@ -0,0 +1,227 @@
...
+) error {
+	pricingURL := getListPriceURL("AmazonEC2", region)
+
+	log.Infof("starting download of \"%s\", which is quite large ...", pricingURL)
+	resp, err := http.Get(pricingURL)
+	if err != nil {
+		return fmt.Errorf("Bogus fetch of \"%s\": %w", pricingURL, err)
+	}
+
+	dec := json.NewDecoder(resp.Body)
```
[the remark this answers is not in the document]
nikovacevic: Definitely want to `Close()` -- probably want to check `StatusCode == http.StatusOK` too.

pkg/env/cloudcost.go
```diff
@@ -22,6 +22,8 @@ const (
 	CustomCostEnabledEnvVar         = "CUSTOM_COST_ENABLED"
 	CustomCostQueryWindowDaysEnvVar = "CUSTOM_COST_QUERY_WINDOW_DAYS"
 
+	PricingModelEnabledEnvVar = "PRICING_MODEL_ENABLED"
```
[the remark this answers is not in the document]
nikovacevic: @Sean-Holcomb was the plan to use this env var? For what?

nikovacevic (commented):
A few things to clean up here, but in general it's looking good!

pkg/env/cloudcost.go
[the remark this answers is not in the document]
Sean-Holcomb: This was for when pricing was still a pipeline that existed in the cloudcost pod

pkg/cloud/aws/pricelistapi.go
```diff
@@ -0,0 +1,227 @@
...
+	for {
+		t, err := dec.Token()
+		if err == io.EOF {
+			log.Infof("done loading \"%s\"\n", resp.Request.URL.String())
+			break
+		} else if err != nil {
+			log.Errorf("error parsing response json %v", resp.Body)
+			break
+		}
```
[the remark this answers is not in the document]
Sean-Holcomb: I removed this file since we copied it into public pricing module

pkg/cloud/aws/pricinglistpricingsource.go
```diff
@@ -0,0 +1,149 @@
...
+			return
+		}
+
+		if attr.OperatingSystem != "" && attr.OperatingSystem != "NA" && attr.OperatingSystem != "Linux" {
+			return
+		}
+
+		if attr.PreInstalledSw != "" && attr.PreInstalledSw != "NA" {
+
+		}
```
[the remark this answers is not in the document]
Sean-Holcomb: Removed this file and addressed in replacement

pkg/cloud/aws/pricelistapi.go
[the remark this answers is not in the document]
Sean-Holcomb: This file doesn't exist anymore, adding the error code handling to where it was moved. Closing already occurs

pkg/cloud/gcp/billingpricingsource.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (g *GCPBillingPricingSource) buildURL(pageToken string) string {
+	url := fmt.Sprintf("%s?key=%s", gcpBillingBaseURL, g.apiKey)
+	if g.currencyCode != "" {
+		url += "&currencyCode=" + g.currencyCode
+	}
+	if pageToken != "" {
+		url += "&pageToken=" + pageToken
+	}
```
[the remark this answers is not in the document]
Sean-Holcomb: file no longer exist. Fixed at new location.

pkg/cloud/aws/pricelistapi.go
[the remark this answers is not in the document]
Sean-Holcomb: applied changes in public pricing module

modules/pricing/public/azure/azurepricingsource_test.go
```diff
@@ -0,0 +1,351 @@
...
+	if source.config.CurrencyCode != "USD" {
+		t.Errorf("CurrencyCode = %v, want USD", source.config.CurrencyCode)
+	}
+}
+
+// Helper function to check if a string contains a substring
+func contains(s, substr string) bool {
+	return len(s) >= len(substr) && (s == substr || len(substr) == 0 ||
+		(len(s) > 0 && (s[0:len(substr)] == substr || contains(s[1:], substr))))
+}
```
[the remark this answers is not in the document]
Sean-Holcomb: using strings.Contains instead

modules/pricing/public/azure/azurepricingsource_test.go
```diff
@@ -0,0 +1,351 @@
...
+	if source.config.CurrencyCode != "USD" {
+		t.Errorf("CurrencyCode = %v, want USD", source.config.CurrencyCode)
+	}
+}
+
+// Helper function to check if a string contains a substring
+func contains(s, substr string) bool {
+	return len(s) >= len(substr) && (s == substr || len(substr) == 0 || 
+		(len(s) > 0 && (s[0:len(substr)] == substr || contains(s[1:], substr))))
+}
```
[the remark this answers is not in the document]
Sean-Holcomb: use strings.Contains instead.

pkg/cloud/aws/pricelistapi.go
```diff
@@ -0,0 +1,227 @@
...
+				if err != nil {
+					return err
+				}
+				product := &PriceListEC2Product{}
+
+				err = dec.Decode(&product)
+				if err != nil {
+					log.Errorf("Error parsing response from \"%s\": %v", resp.Request.URL.String(), err.Error())
+					break
+				}
```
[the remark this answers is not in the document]
Sean-Holcomb: applied in public pricing module

pkg/cloud/aws/pricelistapi.go
```diff
@@ -0,0 +1,227 @@
...
+					// SKUOndemand
+					_, err = dec.Token()
+					if err != nil {
+						return err
+					}
+					offerTerm := &PriceListEC2Term{}
+					err = dec.Decode(&offerTerm)
+					if err != nil {
+						log.Errorf("Error decoding AWS Offer Term: %s", err.Error())
+					}
```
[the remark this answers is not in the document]
Sean-Holcomb: updated in public module

pkg/cloud/aws/fargate.go
Sean-Holcomb: I extracted the API from the provider code for reuse but we just ended up copying it into the module. I have deleted this changes and left them as they are before this commit. We will just delete those implementations when the time comes.

pkg/cloud/aws/pricelistapi.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// PriceListEC2PriceDimension encodes data about the price of a product
+type PriceListEC2PriceDimension struct {
+	Unit         string                   `json:"unit"`
+	PricePerUnit PriceListEC2PricePerUnit `json:"pricePerUnit"`
+}
+
+func (pd *PriceListEC2PriceDimension) String() string {
+	return fmt.Sprintf("{unit: %s, pricePerUnit: %v", pd.Unit, pd.PricePerUnit)
```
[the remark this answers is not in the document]
Sean-Holcomb: removed string functions

modules/pricing/public/aws/pricelistapi.go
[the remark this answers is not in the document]
Sean-Holcomb: handled

modules/pricing/public/aws/pricelistapi.go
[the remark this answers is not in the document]
Sean-Holcomb: updated

modules/pricing/public/aws/pricelistapi.go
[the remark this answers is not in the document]
Sean-Holcomb: updated

modules/pricing/public/aws/pricelistapi.go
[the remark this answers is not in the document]
Sean-Holcomb: Since this isn't running live I will just have this return

core/pkg/pricing/node.go
Sean-Holcomb: I think Provisioning makes a bit more sense. We need to both be able to identify a specific "spot" price and nodes that are "spot". It works because spot is a permanent characteristic of a pod through out its lifetime. Commitment based pricing is something that is ephemerally applied to on-demand nodes. The two ways that I could see commitment based pricing being applied to nodes is api which identify the price of nodes via a node identifier or through a blended cost approach on a specific node type or family.

core/pkg/pricing/source.go
Sean-Holcomb: The subset of props actually used is up to the implementation. 
My understanding is that the `GetXPricing` is for the Opencost use-case of needing to price data in go. It implies the existence of O(1) lookup available in the module which we would want to initialize for a specific process not have readily available in a module itself, especially when we just need the reader.

Maybe what we need is a `NewXPricingGetter` along with
```
type XpricingGetter interface {
    GetXPricing(...)...
}
```
That way we can still enforce the definition of the matching contract without forcing its initialization

core/pkg/unit/unit.go
Sean-Holcomb: aspirations yes, motivation no

Sean-Holcomb (commented):
I dealt with all the copilot comments. I will make a final pass once some of the discussions are done.
@alexsouthard can you take a pass, and address the issues where you are mentioned
@nikovacevic I responded in a few comments, let me know what you think.

modules/pricing/public/gcp/gcppricingsource.go
Sean-Holcomb: Spot could have its own module, if the mechanism only repriced item that had their source module checksum invalidated it could be this might not be so bad. Ie if spot pricing gets updated only nodes that were spot priced before get repriced

core/pkg/cloud/provider.go
Sean-Holcomb: going with `core/pkg/cloud` this is useful outside of model I think

modules/pricing/public/generator.go
alexsouthard: Yeah if we get this in today I might throw some showcase on Friday, I want to set up the secrets in the repo still which might require some help from Christian and that just might be a simple way of doing it because it'll basically just run automatically as a demo (though I can show locally also, it just takes a little bit depending on my download speed 😆)

modules/pricing/public/gcp/gcppricingsource.go
Sean-Holcomb: We are ignoring spot and preemptible for now

core/pkg/pricing/node.go
nikovacevic: Happy to drop `Commitment`. We'd just have to make sure to do it everywhere, including in KCM.

core/pkg/pricing/source.go
nikovacevic: Let's remove them.

nikovacevic (approved):
Took a final pass live with @Sean-Holcomb and @alexsouthard . Resolved outstanding comments. Looking good!

PR #3879, opened 2026-06-23
core: fix potential error overwrite in InsertExternalAllocations loop

### LINKED ISSUE #3878: core: Fix potential error overwrite in InsertExternalAllocations loop
**Describe the bug**
In the method [InsertExternalAllocations](file:///Users/rucha/LFX/OpenCost/opencost/core/pkg/opencost/summaryallocation.go#L1634), the inner loop iterates over external allocations and calls `sas.Insert(externalSA)`. If an insertion fails and returns an error, the loop currently continues and overwrites the `err` variable with subsequent insertions. If the final insertion in the loop succeeds, the function returns a `nil` error, hiding the previous insertion failure entirely.

**To Reproduce**
1. Invoke `InsertExternalAllocations` with an `AllocationSetRange` containing one invalid/nil allocation and a valid allocation.
2. The method returns `nil` (success) if the valid allocation is processed last, despite failing to insert the invalid one.

**Expected behavior**
The loop should check for insertion errors, break early, and return the error immediately to prevent error overwrites.

Current develop branch.

**Additional context**
This is a critical bug in the core library's public API for inserting external allocations.

Rucha0901:
## Description
This PR fixes a bug in [InsertExternalAllocations](file:///Users/rucha/LFX/OpenCost/opencost/core/pkg/opencost/summaryallocation.go#L1634) where errors returned from `sas.Insert(...)` inside the inner loop could be overwritten by subsequent successful insertions. The inner loop now checks the returned error and returns it early to prevent error suppression.

## Related Issues
Fixes #3878 

## User Impact
No user-facing changes or breaking API changes. Improves safety and error reporting correctness for integrations using external allocations.

## Testing
- Added a new unit test [TestSummaryAllocationSetRange_InsertExternalAllocations](file:///Users/rucha/LFX/OpenCost/opencost/core/pkg/opencost/summaryallocation_test.go#L1130) to [summaryallocation_test.go](file:///Users/rucha/LFX/OpenCost/opencost/core/pkg/opencost/summaryallocation_test.go#L1130).
- Ran all tests successfully in the core module.

### DIFF
--- a/core/pkg/opencost/summaryallocation.go
+++ b/core/pkg/opencost/summaryallocation.go
@@ -1636,6 +1636,11 @@ func (sasr *SummaryAllocationSetRange) InsertExternalAllocations(that *Allocatio
 		return fmt.Errorf("cannot insert range into nil AllocationSetRange")
 	}
 
+	// Providing an empty or nil set range is a no-op
+	if that == nil {
+		return nil
+	}
+
 	// keys maps window to index in range
 	keys := map[string]int{}
 	for i, as := range sasr.SummaryAllocationSets {
@@ -1666,7 +1671,7 @@ func (sasr *SummaryAllocationSetRange) InsertExternalAllocations(that *Allocatio
 		for _, alloc := range thatAS.Allocations {
 			externalSA := NewSummaryAllocation(alloc, true, true)
 			if err := sas.Insert(externalSA); err != nil {
-				return err
+				return fmt.Errorf("InsertExternalAllocations: failed inserting into window %s: %w", thatAS.Window.String(), err)
 			}
 		}
 	}
--- a/core/pkg/opencost/summaryallocation.go
+++ b/core/pkg/opencost/summaryallocation.go
@@ -1650,34 +1650,28 @@ func (sasr *SummaryAllocationSetRange) InsertExternalAllocations(that *Allocatio
 		return nil
 	}
 
-	var err error
 	for _, thatAS := range that.Allocations {
-		if thatAS == nil || err != nil {
+		if thatAS == nil {
 			continue
 		}
 
 		// Find matching AllocationSet in asr
 		i, ok := keys[thatAS.Window.String()]
 		if !ok {
-			err = fmt.Errorf("cannot merge AllocationSet into window that does not exist: %s", thatAS.Window.String())
-			continue
+			return fmt.Errorf("cannot merge AllocationSet into window that does not exist: %s", thatAS.Window.String())
 		}
 		sas := sasr.SummaryAllocationSets[i]
 
 		// Insert each Allocation from the given set
 		for _, alloc := range thatAS.Allocations {
 			externalSA := NewSummaryAllocation(alloc, true, true)
-			// This error will be returned below
-			// TODO:CLEANUP should Each have early-error-return functionality?
-			err = sas.Insert(externalSA)
-			if err != nil {
+			if err := sas.Insert(externalSA); err != nil {
 				return err
 			}
 		}
 	}
 
-	// err might be nil
-	return err
+	return nil
 }
 
 func (sasr *SummaryAllocationSetRange) TotalCost() float64 {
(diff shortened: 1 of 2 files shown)


KrishnaParihar1:
Hey @Rucha0901, your pr fixes the issue, but in `core/pkg/opencost/summaryallocation_test.go` — `TestSummaryAllocationSetRange_InsertExternalAllocations` ,the test only exercises the nil allocation skip (`as1.Allocations["invalid"] = nil`), which was never broken. The actual bug — error overwrite across iterations, is untested. can you add a case where the first `sas.Insert()` fails and a second succeeds, then assert the first error is returned.

Rucha0901:
Hi @KrishnaParihar1, I've pushed updates addressing all the feedback:
- Added a regression test verifying that an initial `sas.Insert()` error is correctly returned and not overwritten by subsequent successful operations.
- Added a nil-guard check for the input `that` to prevent panics.
- Wrapped the returned insertion errors for better debugging context.
- Removed stale comments and cleaned up the test notes.

Ready for another review when you have a moment!

vtushar06 (commented):
hey @Rucha0901 I ran the full `core/pkg/opencost` suite plus your two new tests, all green, gofmt clean. The fix reads great, returning on the first error instead of letting a later success overwrite it is exactly right, and the nil guard + wrapping the insert error with the window context are nice touches 👍

One optional thought on the overwrite test: since the inner loop ranges a map, the iteration order isn't fixed, so I reran it against the pre-fix code a bunch and it caught the bug ~27/30 (a few slipped through when the valid insert happened to land last). Rock solid for the fixed code, just slightly non-deterministic as a regression guard, so asserting the returned error actually wraps the insert failure (`errors.Is` or a message check) would make it catch a regression every time. Totally optional, the fix itself is spot on. Nice work by the way

core/pkg/opencost/summaryallocation.go
ameijer, line 1634: actually, I think all this code can be removed. @Rucha0901 can you investigate if this is actually used? I think this was a precursor to cloud costs . 


Being able to remove all this would be a hugely impactful contribution

Rucha0901:
Hi @ameijer,

I investigated and performed a repository-wide search (git grep) for InsertExternalAllocations.

The method was only defined in core/pkg/opencost/summaryallocation.go and exercised in core/pkg/opencost/summaryallocation_test.go. There are no external callers, references, or interface implementations for this method anywhere in the repository.

I have went ahead and removed the dead method and its associated unit tests. The core suite still compiles and all remaining tests pass cleanly. Thanks for the catch and the recommendation to clean this up!

Rucha0901:
Hi @vtushar06,

Thank you for the review and the excellent suggestion on making the overwrite test deterministic (asserting error wrapping)!

Following @ameijer's feedback above, we investigated the usage of InsertExternalAllocations and confirmed it was dead code with no external references. We have now removed this method and its associated unit tests entirely from the codebase. I appreciate your feedback and help on this PR!

Rucha0901:
Hi @ameijer, I removed the unused InsertExternalAllocations method and its associated tests as discussed. The branch is ready for another review whenever you have time. Thanks!

core/pkg/opencost/summaryallocation.go
ameijer: @Rucha0901 we would love to see everything related to external allocations removed

ameijer (changes requested):

Rucha0901:
@ameijer ,Thanks for the clarification! I understand this is a longer-term direction for the project. I’ll keep this PR focused on fixing the bug in InsertExternalAllocations and avoid introducing broader changes. One doubt ,does this mean this PR is no more requied?

ameijer:
@Rucha0901 yes, I think since we are looking for someone to delete the concept of external allocations from the project, this is essentially throwaway work. if you want to re-purpose this PR to removing external allocations as a concept, I would certainly not object :)

PR #3880, opened 2026-06-23
fix(gcp): add nil config guard for cluster info

### LINKED ISSUE #3868: nil pointer dereference in GCP.ClusterInfo() when config path is not writable
**Describe the bug**

When running OpenCost on GKE, if the config directory (`CONFIG_PATH`, default `/var/configs`) is not writable, the pod crashes with a `nil pointer dereference` panic instead of falling back to default behavior.

The startup log shows:

```
ERR Error opening config: Failed to prepare path: mkdir /var/configs: permission denied
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x398 pc=0x43d5c13]
goroutine 147 [running]:
github.com/opencost/opencost/pkg/cloud/gcp.(*GCP).ClusterInfo(0x12b2acf38500)
    /pkg/cloud/gcp/provider.go:302
...
github.com/opencost/opencost/modules/prometheus-source/pkg/prom.(*PrometheusClusterMap).loadClusters(...)
```

**Root cause**

In `GCP.ClusterInfo()` ([`pkg/cloud/gcp/provider.go:298-304`](https://github.com/opencost/opencost/blob/v1.120.3/pkg/cloud/gcp/provider.go#L298-L304)):

```go
c, err := gcp.GetConfig()
if err != nil {
    log.Errorf("Error opening config: %s", err.Error())
}
if c.ClusterName != "" {   // <- c is nil here, panics
    attribute = c.ClusterName
}
```

`GetConfig()` returns `(nil, err)` when `GetCustomPricingData()` fails (`provider.go:140-142`). The error from the config write/`mkdir` propagates up, `c` is `nil`, and the subsequent `c.ClusterName` dereference panics. The error is only logged (not returned), so execution continues into the nil dereference.

Note that `ProviderConfig.loadConfig()` actually returns a non-nil `*CustomPricing` together with the error, but `GetConfig()` discards it via `if err != nil { return nil, err }`.

**To Reproduce**

1. Deploy OpenCost on GKE (or run with `CLOUD_PROVIDER=gcp`) where the process cannot create/write `CONFIG_PATH` (default `/var/configs`) — e.g. a read-only root filesystem or non-root user without a writable config mount.
2. Start the pod.
3. The config write fails with `mkdir /var/configs: permission denied`, and the pod panics with a nil pointer dereference in `GCP.ClusterInfo()`.

**Expected behavior**

OpenCost should not panic. When the config cannot be loaded, it should fall back to default pricing / a default cluster name and continue, or at least fail with a clear error message.

v1.120.3 (the same code path exists on `develop`).

**Additional context**

- Cloud provider: GCP (GKE)
- The minimal fix is a nil guard at `provider.go:302` (`if c != nil && c.ClusterName != ""`); arguably `GetConfig()` should also return the default pricing on error instead of `nil`.

KrishnaParihar1:
## Description
Fix nil pointer dereference panic in `GCP.ClusterInfo()` when `CONFIG_PATH` is not writable (e.g. read-only root filesystem or non-root user on GKE).

`GetConfig()` was returning `nil, err` on config failure, but `ProviderConfig.loadConfig` always returns a non-nil default `*CustomPricing` alongside any error. The fix passes that default through instead of discarding it. A nil guard on the `c.ClusterName` dereference in `ClusterInfo()` is added as a defensive backstop.

## Related Issues
Fixes #3868

## User Impact
Non-breaking. OpenCost no longer crashes on GKE when `CONFIG_PATH` is unwritable — it falls back to default pricing and a default cluster name and continues running.

## Testing
- Updated `TestGCP_GetConfig` to assert the error path returns a non-nil `*CustomPricing`
- Updated `TestGCP_ClusterInfo` to assert no panic occurs on config failure and the fallback cluster name is used
- Added `mockConfigError` to simulate a permission-denied config backend


pkg/cloud/gcp/provider.go
```diff
@@ -140,7 +140,9 @@ type multiKeyGCPAllocation struct {
 func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
 	c, err := gcp.Config.GetCustomPricingData()
 	if err != nil {
-		return nil, err
+		// loadConfig always returns a non-nil default alongside the error;
+		// propagate both so callers can use defaults instead of panicking.
+		return c, err
 	}
```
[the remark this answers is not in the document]
thomasvn: Invalid Copilot review. I think this diff is doing the exact opposite. Safe to discard.

pkg/cloud/gcp/provider.go
```diff
@@ -139,10 +140,8 @@ type multiKeyGCPAllocation struct {
 
 func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
 	c, err := gcp.Config.GetCustomPricingData()
-	if err != nil {
-		// loadConfig always returns a non-nil default alongside the error;
-		// propagate both so callers can use defaults instead of panicking.
```
thomasvn: I would leave some kind of comment here still stating that even if a `nil` CustomPricing is returned or an error is returned, we still want to use the following defaults.

pkg/cloud/gcp/provider.go
```diff
@@ -292,16 +293,20 @@ func (gcp *GCP) UpdateConfig(r io.Reader, updateType string) (*models.CustomPric
 func (gcp *GCP) ClusterInfo() (map[string]string, error) {
 	remoteEnabled := env.IsRemoteEnabled()
 
-	attribute, err := gcp.MetadataClient.InstanceAttributeValue("cluster-name")
-	if err != nil {
-		log.Infof("Error loading metadata cluster-name: %s", err.Error())
+	var attribute string
+	if gcp.MetadataClient != nil {
```
thomasvn, line 310: Can you help me understand in what scenario this is needed? When would `gcp.MetadataClient.InstanceAttributeValue` be `nil`?

KrishnaParihar1:
Hey @thomasvn , `MetadataClient` can be nil in a couple of cases:

- Unit tests: GCP{} is usually created without a MetadataClient because *metadata.Client is a concrete type (not an interface), so it can't be easily mocked. Tests just leave it as nil.
- Non-GKE deployments: When OpenCost is running on a cluster that isn't GKE (for example, a self-managed Kubernetes cluster on GCP), the provider factory in pkg/cloud/provider/provider.go creates the GCP struct but only sets MetadataClient if it successfully detects GKE metadata. If that detection doesn't happen, MetadataClient stays nil.

Without the nil check, ClusterInfo() will panic in both of these cases before it ever falls back to the config-based path.

pkg/cloud/gcp/provider.go
```diff
@@ -152,7 +162,7 @@ func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
 	if c.CurrencyCode == "" {
 		c.CurrencyCode = "USD"
 	}
-	return c, nil
+	return c, err
```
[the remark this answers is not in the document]
KrishnaParihar1: The behavior in `parsePages` and `DownloadPricingData` is pre-existing and out of scope for this PR, which specifically targets the nil pointer dereference in `ClusterInfo()`. Happy to address the broader error-handling in a follow-up issue.

pkg/cloud/gcp/provider.go
```diff
@@ -139,9 +139,19 @@ type multiKeyGCPAllocation struct {
 }
 
 func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
+	if gcp.Config == nil {
+		return &models.CustomPricing{
+			Discount:           "30%",
+			NegotiatedDiscount: "0%",
+			CurrencyCode:       "USD",
```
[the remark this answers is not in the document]
KrishnaParihar1: resolved

pkg/cloud/gcp/provider.go
```diff
@@ -139,9 +139,15 @@ type multiKeyGCPAllocation struct {
 }
 
 func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
-	c, err := gcp.Config.GetCustomPricingData()
```
thomasvn, line 142: Could you add a function comment associated with your changes? We are now always returning a non-nil `models.CustomPricing` in order to avoid downstream callers from accessing a nil pointer.

thomasvn (approved):

pkg/cloud/digitalocean/provider_test.go
thomasvn, line 1: Minor comment, but these tests seem out of scope for this PR?

.gitignore
thomasvn, line 1: Undo this?

pkg/cloud/gcp/provider.go
KrishnaParihar1: Hey i added the comments

pkg/cloud/gcp/provider.go
```diff
@@ -139,9 +139,15 @@ type multiKeyGCPAllocation struct {
 }
 
 func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
```
thomasvn, line 144: Minor comment, redoing the conditional logic like this could read a little cleaner.

```go
  func (gcp *GCP) GetConfig() (*models.CustomPricing, error) {
  	c := &models.CustomPricing{}
  	var err error
  	if gcp.Config == nil {
  		err = fmt.Errorf("GCP provider config is nil")
  	} else if pc, e := gcp.Config.GetCustomPricingData(); pc != nil {
  		c, err = pc, e
  	} else {
  		err = e
  	}
  	// ...fill defaults, return c, err
  }
```

.gitignore
KrishnaParihar1: Do you want me to check and remove any unnecessary test?

thomasvn: I don't think we need `go1.26.4.linux-amd64.tar.gz` added to the `.gitignore`? Unless you think otherwise?

pkg/cloud/gcp/provider.go
KrishnaParihar1: fixed it thanks for the suggestion

.gitignore
KrishnaParihar1: Ah, that .gitignore line was an accidental leftover from my local setup — removing it now, it's not needed for this PR. The test changes (TestGCP_GetConfig, TestGCP_ClusterInfo) are intentional and cover the nil-guard fix directly, so I'll leave those in unless you see a specific one that's redundant.

KrishnaParihar1: fixed it

KrishnaParihar1:
@thomasvn , can you review the pr

PR #3886, opened 2026-06-28
aws: parse disk region dynamically from AvailabilityZone in findCostF…

### LINKED ISSUE #3885: aws: parse disk region dynamically from AvailabilityZone in findCostForDisk
**Describe the bug**
In the AWS provider, the function [findCostForDisk](file:///Users/rucha/LFX/OpenCost/opencost/pkg/cloud/aws/provider.go#L2175) constructs the pricing lookup key using the global `aws.ClusterRegion` (`key := aws.ClusterRegion + "," + class`). This assumes that all persistent volumes and disks reside in the cluster's default region. However, a cluster might manage resources or disks in other regions or availability zones.

**To Reproduce**
1. Run OpenCost in an AWS cluster that has orphaned disks in a region other than the main `ClusterRegion`.
2. Inspect the calculated monthly cost of those disks. They will fail to match pricing keys since it uses the cluster region rather than the disk's region.

**Expected behavior**
The function should extract the region dynamically from the disk's `AvailabilityZone` using `regionRx` (e.g., matching `us-east-1` from `us-east-1a`) and fall back to the cluster region only if not found.

Current develop branch.

**Additional context**
This improves accuracy of AWS orphaned resource calculations when dealing with multi-region AWS environments.

Rucha0901:
## Description
This PR resolves a limitation in the AWS provider's [findCostForDisk](file:///Users/rucha/LFX/OpenCost/opencost/pkg/cloud/aws/provider.go#L2175) function, which previously relied on the global `aws.ClusterRegion` for disk pricing lookups. We now parse the region dynamically from the disk's `AvailabilityZone` using `regionRx` and fall back to `aws.ClusterRegion` if parsing fails.

## Related Issues
Fixes #3885

## User Impact
Improves AWS disk pricing lookup accuracy for clusters utilizing multi-region persistent disks. No breaking changes or backwards compatibility issues.

## Testing
- Added a new unit test [TestAWS_findCostForDisk](file:///Users/rucha/LFX/OpenCost/opencost/pkg/cloud/aws/provider_test.go#L1368) to [provider_test.go](file:///Users/rucha/LFX/OpenCost/opencost/pkg/cloud/aws/provider_test.go#L1368).
- Verified that all AWS provider tests pass locally.

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -2177,9 +2177,8 @@ func (aws *AWS) GetOrphanedResources() ([]models.OrphanedResource, error) {
 }
 
 func (aws *AWS) findCostForDisk(disk *ec2Types.Volume) (*float64, error) {
-	// todo: use AWS pricing from all regions
 	if disk.AvailabilityZone == nil {
-		return nil, fmt.Errorf("nil region")
+		return nil, fmt.Errorf("nil AvailabilityZone")
 	}
 	if disk.Size == nil {
 		return nil, fmt.Errorf("nil disk size")
@@ -2187,7 +2186,12 @@ func (aws *AWS) findCostForDisk(disk *ec2Types.Volume) (*float64, error) {
 
 	class := volTypes[string(disk.VolumeType)]
 
-	key := aws.ClusterRegion + "," + class
+	region := regionRx.FindString(*disk.AvailabilityZone)
+	if region == "" {
+		region = aws.ClusterRegion
+	}
+
+	key := region + "," + class
 
 	pricing, ok := aws.Pricing[key]
 	if !ok {
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -17,7 +17,6 @@ import (
 	"time"
 
 	"github.com/aws/smithy-go"
-	"github.com/opencost/opencost/pkg/cloud/httputil"
 	"github.com/opencost/opencost/pkg/cloud/models"
 	"github.com/opencost/opencost/pkg/cloud/utils"
 
@@ -859,10 +858,7 @@ func (aws *AWS) getRegionPricing(nodeList []*clustercache.Node) (*http.Response,
 	}
 
 	log.Infof("starting download of \"%s\", which is quite large ...", pricingURL)
-	// This file is large and can take a while to stream, so the streaming client
-	// bounds connect/TLS/response-header time but not the total body read - enough
-	// to bail on a hung endpoint without truncating a legitimate slow download.
-	resp, err := httputil.StreamingGet(context.Background(), pricingURL)
+	resp, err := http.Get(pricingURL)
 	if err != nil {
 		log.Errorf("Bogus fetch of \"%s\": %v", pricingURL, err)
 		return nil, pricingURL, err
@@ -1356,7 +1352,7 @@ func (aws *AWS) spotPricingFromHistory(k models.Key) (*SpotPriceHistoryEntry, bo
 
 	price, err := aws.SpotPriceHistoryCache.GetSpotPrice(region, instanceType, availabilityZone)
 	if err != nil {
-		log.Debugf("Failed to get spot price history for instance %s: %s", k.ID(), err.Error())
+		log.DedupedWarningf(10, "Failed to get spot price history for instance %s: %s", k.ID(), err.Error())
 		return nil, false
 	}
 	return price, true
@@ -1487,7 +1483,7 @@ func (aws *AWS) createNode(terms *AWSProductTerms, usageType string, k models.Ke
 			UsageType:    PreemptibleType,
 		}, meta, nil
 	} else if aws.isPreemptible(key) { // Preemptible but we don't have any data in the pricing report.
-		log.Debugf("Node %s marked preemptible but no spot feed data available; falling back to other pricing sources", k.ID())
+		log.DedupedWarningf(5, "Node %s marked preemptible but no spot feed data available; falling back to other pricing sources", k.ID())
 
 		// Try to get spot pricing from DescribeSpotPriceHistory API
 		if historyEntry, ok := aws.spotPricingFromHistory(k); ok {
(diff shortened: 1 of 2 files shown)


pkg/cloud/aws/provider.go
```diff
@@ -2177,17 +2177,21 @@ func (aws *AWS) GetOrphanedResources() ([]models.OrphanedResource, error) {
...
+		return nil, fmt.Errorf("nil AvailabilityZone")
 	}
 	if disk.Size == nil {
 		return nil, fmt.Errorf("nil disk size")
 	}
 
 	class := volTypes[string(disk.VolumeType)]
 
-	key := aws.ClusterRegion + "," + class
+	region := regionRx.FindString(*disk.AvailabilityZone)
```
thomasvn: Thanks for the PR! I generally agree with CoPilot here. What happens if `regionRx.FindString()` returns an invalid cluster region? That would result in an invalid `key` and an invalid `pricing`.

Rucha0901:
Thanks for the review, @thomasvn! 

To address this, I have updated the lookup logic to validate whether the parsed region exists in the pricing map. If the dynamically parsed region does not have pricing data loaded (i.e. the key `region + "," + class` is not found in `aws.Pricing`), we now fall back to using `aws.ClusterRegion` instead of failing the disk lookup.

I have also added unit tests covering this fallback behavior to verify that:
1. A valid region that doesn't have pricing data loaded falls back to `aws.ClusterRegion`.
2. A completely invalid region parsed from the AvailabilityZone falls back to `aws.ClusterRegion`.

Rucha0901:
Thanks for the review, @thomasvn!

I’ve addressed the feedback by improving the region parsing logic, adding the fallback to aws.ClusterRegion when needed, and expanding the unit tests to cover the new scenarios.

I’d appreciate another review when you have a chance. Thanks!

Rucha0901:
Hi @thomasvn, I’ve addressed the feedback by improving the region parsing, adding the fallback logic, and expanding the tests to cover the discussed scenarios. I’d appreciate another review when you have time. Thanks!

pkg/cloud/aws/provider.go
```diff
@@ -859,10 +858,7 @@ func (aws *AWS) getRegionPricing(nodeList []*clustercache.Node) (*http.Response,
 	}
 
 	log.Infof("starting download of \"%s\", which is quite large ...", pricingURL)
-	// This file is large and can take a while to stream, so the streaming client
-	// bounds connect/TLS/response-header time but not the total body read - enough
-	// to bail on a hung endpoint without truncating a legitimate slow download.
-	resp, err := httputil.StreamingGet(context.Background(), pricingURL)
+	resp, err := http.Get(pricingURL)
```
thomasvn: Why did you make this change? Let's please undo this. I would like this PR to be scoped specifically to parsing of Disk Availability Zone.

pkg/cloud/aws/provider.go
```diff
@@ -1356,7 +1352,7 @@ func (aws *AWS) spotPricingFromHistory(k models.Key) (*SpotPriceHistoryEntry, bo
 
 	price, err := aws.SpotPriceHistoryCache.GetSpotPrice(region, instanceType, availabilityZone)
 	if err != nil {
-		log.Debugf("Failed to get spot price history for instance %s: %s", k.ID(), err.Error())
+		log.DedupedWarningf(10, "Failed to get spot price history for instance %s: %s", k.ID(), err.Error())
```
thomasvn: Let's undo these logging changes? Let's keep this PR scoped to Disk Availability Zone?

pkg/cloud/aws/provider.go
```diff
@@ -2177,19 +2173,37 @@ func (aws *AWS) GetOrphanedResources() ([]models.OrphanedResource, error) {
...
+				cost := price * timeutil.HoursPerMonth * float64(*disk.Size)
+				return &cost, nil
+			}
+		}
+	}
 
-	pricing, ok := aws.Pricing[key]
+	// Fallback to ClusterRegion
+	key = aws.ClusterRegion + "," + class
+	pricing, ok = aws.Pricing[key]
```
thomasvn: Could this logic be simpler and cleaner using an `else` statement for the fallback? This function now has two locations that it's calculating and returning cost. Seems a little overly complex and difficult to maintain, for what we're trying to achieve?

ameijer:
@Rucha0901 looks like you have outstanding review from copilot and @thomasvn

Rucha0901:
@thomasvn , I did the changes can u plz review them as u will get time?

thank you!!

pkg/cloud/aws/provider.go
```diff
@@ -17,7 +17,6 @@ import (
 	"time"
 
 	"github.com/aws/smithy-go"
-	"github.com/opencost/opencost/pkg/cloud/httputil"
 	"github.com/opencost/opencost/pkg/cloud/models"
 	"github.com/opencost/opencost/pkg/cloud/utils"
```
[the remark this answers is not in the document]
thomasvn: Stale

pkg/cloud/aws/provider.go
```diff
@@ -2177,17 +2177,23 @@ func (aws *AWS) GetOrphanedResources() ([]models.OrphanedResource, error) {
...
 	}
 
 	class := volTypes[string(disk.VolumeType)]
 
-	key := aws.ClusterRegion + "," + class
+	region := parseRegionFromAZ(*disk.AvailabilityZone)
+
+	key := region + "," + class
+	_, ok := aws.Pricing[key]
+	if region == "" || !ok {
```
thomasvn, line 2258:
```suggestion
	if _, ok := aws.Pricing[key]; !ok {
```

Minor stylistic suggestion. But overall looks good!

pkg/cloud/aws/provider.go
```diff
@@ -2119,7 +2119,7 @@ func (aws *AWS) GetOrphanedResources() ([]models.OrphanedResource, error) {
 				zone = *volume.AvailabilityZone
 			}
 			var region, url string
-			region = regionRx.FindString(zone)
+			region = parseRegionFromAZ(zone)
```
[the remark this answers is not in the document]
thomasvn: I think this is a false positive. Safe to ignore.

thomasvn (approved):
Thanks @Rucha0901! LGTM.

ameijer (approved):
merging due to @thomasvn and @vtushar06 approval

PR #3887, opened 2026-06-29
KCM-5392: Configmap External label implementation and connect it to Default Collector

avrodrigues5:
## Description

Adds a new core/pkg/external package that enables external node labels — sourced from a Kubernetes ConfigMap — to be injected into the collector data source pipeline.

## Related Issues

EPIC: https://apptio.atlassian.net/browse/KCM-5392
Story: https://apptio.atlassian.net/browse/KCM-6023
## User Impact

Users can now configure a ConfigMap (by name, optional namespace, optional YAML key + route) via environment variables. The collector scraper will automatically pick up those labels and attach them to node metrics on every reconciliation, making them available in allocation and asset data.

## Testing

Unit test cases, please refer ibm-finops-chart for testing details and opencost-chart for testing details


core/pkg/externallabels/configmap.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to start configMap external labels provider: %w", err)
+	}
+
+	log.Infof("ExternalLabels: ConfigMapProvider: started in namespace %q with selector %q", cmp.namespace, ExternalLabelsSelector)
+	informer.Run(ctx.Done())
+	return nil
+}
```
mbolt35: We already have code that watches config maps. Any reason why you've reinvented it here?

core/pkg/externallabels/configmap.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Labels returns a copy of the currently cached external labels.
+func (cmp *ConfigMapProvider) Labels(_ context.Context) (map[string]string, error) {
+	cmp.mu.RLock()
+	defer cmp.mu.RUnlock()
+	out := make(map[string]string, len(cmp.labels))
+	for k, v := range cmp.labels {
+		out[k] = v
+	}
+	return out, nil
```
mbolt35: `maps.Clone(cmp.labels)`

core/pkg/externallabels/configmap.go
```diff
@@ -0,0 +1,127 @@
...
+func (cmp *ConfigMapProvider) onConfigMapUpdate(obj any) {
+	cm, ok := obj.(*v1.ConfigMap)
+	if !ok {
+		return
+	}
+	cmp.mu.Lock()
+	defer cmp.mu.Unlock()
+	for k, v := range cm.Data {
+		cmp.labels[k] = v
+	}
```
mbolt35: This only overwrites existing labels. If labels were removed, it won't be reflected.

core/pkg/externallabels/provider.go
```diff
@@ -0,0 +1,8 @@
+package externallabels
+
+import "context"
+
+type Provider interface {
+	Start(ctx context.Context) error
```
mbolt35: It feels unusual for `Start` to exist on a provider interface.

modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -161,6 +165,14 @@ func (ccs *ClusterCacheScraper) scrapeNodes(nodes []*clustercache.Node) []metric
...
 		nodeLabels := util.ToMap(labelNames, labelValues)
 
+		// Add external labels to existing node labels
+		if ccs.externalLabelsProvider != nil {
+			externalLabels, err := ccs.externalLabelsProvider.Labels(context.Background())
+			if err != nil {
+				log.DedupedErrorf(10, "failed to apply external labels to nodes: %s", err)
+			}
+			nodeLabels = applyExternalLabelsToNodeLabels(nodeLabels, externalLabels)
+		}
```
mbolt35: It seems like we should get the map at the beginning of the scrape, and since we _don't_ change the map, we can just continuously use the same return value to copy strings to the node labels. 

Additionally, it does feel weird that we have something called "external labels" that we are only applying to nodes.

core/pkg/externallabels/configmap.go
```diff
@@ -0,0 +1,48 @@
...
+	"sync"
+
+	"github.com/opencost/opencost/core/pkg/log"
+)
+
+// ConfigMapProvider maintains a key/value map of external labels sourced from a
+// ConfigMap. It is intended to be wired up as a ConfigMapWatcher WatchFunc —
+// the caller is responsible for registering it against the appropriate ConfigMap
+// name via ConfigMapWatchers.AddWatcher.
+type ConfigMapProvider struct {
```
avrodrigues5: I'm leaning toward naming the package external (although I'm not sure if that's too broad) and calling the type NodeLabelProvider.

If we add support for another type of external label provider in the future, I think it would mostly involve implementing another watcher, similar to the ConfigMap watcher, that returns a map[string]string.

core/pkg/externallabels/configmap.go
avrodrigues5: fixed!

core/pkg/externallabels/provider.go
avrodrigues5: fixed!

core/pkg/externallabels/configmap.go
avrodrigues5: You were absolutely write there.

modules/collector-source/pkg/scrape/clustercache.go
avrodrigues5: https://github.com/opencost/opencost/pull/3887#discussion_r3540311573 Im open to opinion i feel like external and then Node Label Provider makes sense. But is it too broad is my fear.

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,60 @@
...
+			return nil, fmt.Errorf("route segment %q: parent is not a map in ConfigMap %s", segment, cms.cfg.ConfigMapName)
+		}
+		current, ok = m[segment]
+		if !ok {
+			return nil, fmt.Errorf("route segment %q not found in ConfigMap %s", segment, cms.cfg.ConfigMapName)
+		}
+	}
+
+	// Convert the reached node to map[string]string.
+	m, ok := current.(map[string]interface{})
```
mbolt35: Small nit:


```suggestion
	m, ok := current.(map[string]any)
```

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,60 @@
...
+	// yaml.v3 always unmarshals maps as map[string]interface{}, no dual-type handling needed.
+	var doc map[string]interface{}
+	if err := yaml.Unmarshal([]byte(raw), &doc); err != nil {
+		return nil, fmt.Errorf("failed to parse YAML from key %q in ConfigMap %s: %w", cms.cfg.Key, cms.cfg.ConfigMapName, err)
+	}
+
+	// Traverse the dot-separated route.
+	var current interface{} = doc
+	for _, segment := range strings.Split(cms.cfg.Route, ".") {
+		m, ok := current.(map[string]interface{})
```
mbolt35: Nit


```suggestion
		m, ok := current.(map[string]any)
```

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,60 @@
...
+		return data, nil
+	}
+
+	// Block-scalar ConfigMap — extract the YAML document from data[Key].
+	raw, ok := data[cms.cfg.Key]
+	if !ok {
+		return nil, fmt.Errorf("key %q not found in ConfigMap %s", cms.cfg.Key, cms.cfg.ConfigMapName)
+	}
+
+	// yaml.v3 always unmarshals maps as map[string]interface{}, no dual-type handling needed.
```
mbolt35:
```suggestion
	// yaml.v3 always unmarshals maps as map[string]any, no dual-type handling needed.
```

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,60 @@
...
+	var doc map[string]interface{}
```
mbolt35:
```suggestion
	var doc map[string]any
```

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,60 @@
...
+	if err := yaml.Unmarshal([]byte(raw), &doc); err != nil {
+		return nil, fmt.Errorf("failed to parse YAML from key %q in ConfigMap %s: %w", cms.cfg.Key, cms.cfg.ConfigMapName, err)
+	}
+
+	// Traverse the dot-separated route.
+	var current interface{} = doc
```
mbolt35:
```suggestion
	var current any = doc
```

core/pkg/external/merge.go
```diff
@@ -0,0 +1,18 @@
...
+	}
+	out := make(map[string]string, len(base)+len(external))
+	for k, v := range external {
+		out[k] = v
+	}
+	for k, v := range base {
+		out[k] = v // base overwrites external on conflict
+	}
+	return out
+}
```
mbolt35: I have two pieces of feedback here:

1. If we are going to make this func public, let's make it generic and put in the `core/pkg/util/maputil/maputil.go` utility with the following definition: 
```go
// Merge accepts two compatible maps and merges them into a single new map instance.
func Merge[M1 ~map[K]V, M2 ~map[K]V, K comparable, V any](m1 M1, m2 M2) map[K]V {
	size := len(m1) + len(m2)
	if size == 0 { 
		return map[K]V{}
	}

	result := make(map[K]V, size)
	maps.Copy(result, m1)
	maps.Copy(result, m2)
	return result
}
```

2. I'm wondering if we should omit the pre-allocation of the new map (This line: `result := make(map[K]V, len(m1)+len(m2))`). This is preferable when you don't expect collisions, but if we assume that the input will collide more often than not, we would actually be _over_ allocating (by 2x in the worst case). The question is whether or not it's worth caring about at all. It's tough to say whether counting collisions would be worth it, where the new allocation would be something like: 
```go
collisions := 0
for k, v := range m1 {
    if _, ok := m2[k]; ok { 
        collisions += 1
    }
}
result := make(map[K]V, len(m1)+len(m2) - collisions)
// ... 
```
My guess is that the time spent counting would cancel out any gains from using precise allocation. Just something to think about. I think, overall, the overallocation is the safest (in that it potentially prevents many smaller reallocations if you were to omit it).

core/pkg/external/nodelabelprovider.go
```diff
@@ -0,0 +1,39 @@
...
+	"maps"
+	"sync"
+
+	"github.com/opencost/opencost/core/pkg/log"
+)
+
+// NodeLabelProvider maintains a key/value map of external labels sourced from any
+// watcher function. It is intended to be wired up to a WatchFunc such as ConfigMapWatcher
+// the caller is responsible for registering it.
+type NodeLabelProvider struct {
```
mbolt35, line 13: Love it!

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	// Convert the reached node to map[string]string.
+	m, ok := current.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("route %q does not point to a map in ConfigMap %s", cms.cfg.Route, cms.cfg.ConfigMapName)
+	}
+	labels := make(map[string]string, len(m))
+	for k, v := range m {
+		labels[k] = fmt.Sprintf("%v", v)
```
mbolt35: I'm not sure that this is the best way to convert `any` to string. After looking around a bit, yaml v3 will turn scalers into `string` but will error on lists. To me, this feels like the correct way to deal with this, but open to discussion. ie:
```go
    var doc map[string]string
	if err := yaml.Unmarshal([]byte(raw), &doc); err != nil {
		return nil, fmt.Errorf("failed to parse YAML from key %q in ConfigMap %s: %w", cms.cfg.Key, cms.cfg.ConfigMapName, err)
	}
```

instead of parsing the yaml into `map[string]any`, then trying to convert the `any` into a `string`. My reasoning is that we don't want things like `map["foo": []MyStruct{ 44, &Foo{ ... }}]` in the labels. If someone puts an object in the yaml under their labels, then we just error instead of trying to forcefully turn that into a string. 

The other way we could handle `any ` is by doing a go type switch:
```go
for k, v := range m {
    switch val := v.(type) { 
    case string: 
        labels[k] = val
    case bool: 
        labels[k] = strconv.ParseBool(val) // ?? No idea if this is a real func
    case int, int64, uint, uint64: // etc....
        labels[k] = strconv.ParseInt(...)
    default: 
        // error
    }
}
```

`nil` may also be handled differently here, but I'd need to look into it more deeply. I'm _pretty_ sure that `fmt.Sprintf("%v", nil)` will yield the string `<nil>` which is probably not a value we want in labels. 

Let's discuss further or loop in Chip if he has opinions. If I recall, he wanted to very carefully validate the input and error if it's bad.

core/pkg/external/configmapsource_test.go
```diff
@@ -0,0 +1,172 @@
...
+	s := newSource(&Config{
+		ConfigMapName: "my-cm",
+		Key:           "config.yaml",
+		Route:         "labels",
+	})
+
+	yaml := `
+labels:
+  priority: 42
+  replicas: 3
```
mbolt35, line 195: I like the yaml test cases with scalars, but we should add tests with all sorts of yaml types here: 
```yaml
labels:
  priority: 42
  replicas: 3
  other: 
    - "hello"
    - "this"
    - "is"
    - "a"
    - "test"
  inner-map:
    my-key: "value"
    other-key: "value"
```

This is the test case that I think should fail because of the `slice` and `map` types nested inside the labels. I think we should delegate all of the validation responsibility to the `yaml.v3` package.

core/pkg/external/configmap_source.go
avrodrigues5: Ah i always do this interface is any now!

core/pkg/external/merge.go
avrodrigues5: Fixed. I kept it simple but i do agree over allocation is prioritized here

core/pkg/external/configmap_source.go
avrodrigues5: I went through yaml.v3 package today. When we use https://pkg.go.dev/gopkg.in/yaml.v3#Node, we can make only scalar values be used. I have added further test cases to reiterate this. Please have a look. 

Only scalar such as int, float, bool and string proceed now and if there is SequenceNode, aliasNode or MapNode they get rejected with error.

core/pkg/external/configmapsource_test.go
avrodrigues5: This is awesome feedback, i went ahead and use yaml.v3 package more effectively now.

pkg/costmodel/router.go
```diff
@@ -501,6 +501,7 @@ func Initialize(router *httprouter.Router, additionalConfigWatchers ...*watcher.
 				clusterInfoProvider,
 				k8sCache,
 				nodeStatClient,
+				nil,
 			)
```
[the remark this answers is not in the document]
avrodrigues5: The external labels features is not an opencost feature. It's used in the ibm-finops-agent.

modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -157,8 +170,12 @@ func (ccs *ClusterCacheScraper) scrapeNodes(nodes []*clustercache.Node) []metric
 			}
 		}
 
+		// Merge external labels into node labels; node labels win on conflict.
+		// mapUtil.Merge handles that when node.Labels is the second comparable map.
+		mergedLabels := maputil.Merge(externalLabels, node.Labels)
+
```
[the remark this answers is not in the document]
avrodrigues5: this is fixed. Good feedback!

modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -157,8 +171,15 @@ func (ccs *ClusterCacheScraper) scrapeNodes(nodes []*clustercache.Node) []metric
 			}
 		}
 
+		// Merge external labels into node labels; node labels win on conflict.
+		// mapUtil.Merge handles that when node.Labels is the second comparable map.
+		mergedLabels := stdmap.Clone(node.Labels)
+		if externalLabels != nil {
+			mergedLabels = maputil.Merge(externalLabels, node.Labels)
```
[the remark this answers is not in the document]
avrodrigues5: ah, missed saving this. TY!

modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -84,6 +88,15 @@ func (ccs *ClusterCacheScraper) GetScrapeNodes(nodes []*clustercache.Node) Scrap
 func (ccs *ClusterCacheScraper) scrapeNodes(nodes []*clustercache.Node) []metric.Update {
 	var scrapeResults []metric.Update
 
+	// get external labels
+	var externalLabels map[string]string
+	var err error
+	if ccs.externalLabelProvider != nil {
+		externalLabels, err = ccs.externalLabelProvider.Labels()
+		if err != nil {
+			log.Errorf("failed to get external labels to nodes: %s", err)
```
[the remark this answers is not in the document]
avrodrigues5: https://github.com/rs/zerolog/blob/master/event.go#L140. String print formatter is used when its %v. Does not make sense to make this change.

modules/collector-source/pkg/scrape/clustercache_test.go
```diff
@@ -200,6 +201,202 @@ func Test_kubernetesScraper_scrapeNodes(t *testing.T) {
...
+
+	const (
+		testSource             = "mock"
+		testExternalLabelKey   = "externalLabelKey"
+		testExternalLabelValue = "externalLabelValue"
+	)
+	mockLabelProvider := external.NewNodeLabelProvider()
+	err := mockLabelProvider.Update(testSource, map[string]string{testExternalLabelKey: testExternalLabelValue})
+	if err != nil {
+		t.Fatalf("failed to get test node labels: %s", err)
```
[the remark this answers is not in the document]
avrodrigues5: above explanation

core/pkg/external/source.go
```diff
@@ -0,0 +1,50 @@
...
+			return fmt.Errorf("nil LabelProvider")
+		}
+	}
+	return func(name string, data map[string]string) error {
+		labels, err := src.ExtractNodeLabels(data)
+		if err != nil {
+			return err
+		}
+		return provider.Update(name, labels)
+	}
```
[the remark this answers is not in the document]
avrodrigues5: done

modules/collector-source/pkg/scrape/clustercache.go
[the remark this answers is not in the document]
avrodrigues5: https://go.dev/play/p/jN4NbottJ2D

Its the same, format directive %s and %v once you check err != nil is the same.

modules/collector-source/pkg/scrape/clustercache_test.go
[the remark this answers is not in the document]
avrodrigues5: https://go.dev/play/p/jN4NbottJ2D

Its the same, format directive %s and %v once you check err != nil is the same.

Sean-Holcomb (commented):
This seems like it fulfill the requirement. I do wonder if merging in these labels is better done at while computing kubemodel/assets rather than scrape. The concern being that adding a static label set one to every node in every scrape could add size to the wal and slow down the update/scrape process which is sensitive to time. Less importantly this would enable this feature for prom users, which are bypassed. 

At a higher level I wonder if this could be configured via the aggregator and be applied during ingestion since as mentioned in one of the comments this is a feature exclusively for the agent. Maybe a little too late for that thought though

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,223 @@
...
+	route := nlCfg.Route()
+	// Traditional ConfigMap — labels live directly in data.
+	if key == "" && route == "" {
+		return maps.Clone(data), nil
+	}
+
+	// route is optional for block scalar. A root yaml node can be the map of node labels.
+	if key == "" && route != "" {
+		return nil, fmt.Errorf("key must be set for block scalar configMap")
+	}
```
mbolt35: Is it possible to use the default key AND a specific route?

mbolt35:
Regarding the use of `_` in filenames in go, I've done my best to communicate this, but it does seem that the desire to break away from conventions is high.

> Go source filenames do not follow strict rules, but the prevailing convention is to use lowercase flatcase (e.g., `nethttp.go` rather than `net_http.go`). Using underscores in standard filenames is generally avoided to reserve them for build constraints, operating systems, or test suffixes. Special filename conventions to be aware of include `:_test.go`: Suffix strictly used for unit tests. `_linux.go` / `_windows.go`: Suffixes used for OS-specific build constraints. Leading Underscores: Files or packages that begin with an _ are completely ignored by the Go toolchain.

The standard for go is to _AVOID_ using `_` in filenames unless you are specifying a compiler build directive. (`_test.go`, `_linux.go`, `_darwin.go`, etc...)

This is not constrained to reserved words only, as you can set compile flags to include any custom strings as well. 

If the AI is creating the file names, I would suggest adding a line to your MD file that includes the convention for go files.

pkg/costmodel/router.go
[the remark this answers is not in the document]
mbolt35: @avrodrigues5 Is there anything preventing us from adding this to the opencost implementation as well?

core/pkg/external/configmap_source.go
```diff
@@ -0,0 +1,223 @@
...
+	// Block-scalar ConfigMap — extract the YAML document from data[Key].
+	raw, ok := data[key]
+	if !ok {
+		return nil, fmt.Errorf("key %q not found in ConfigMap %s", key, cm)
+	}
+
+	// yaml.v3 always unmarshals to a yaml.Node type.
+	// yaml.Unmarshal creates a root node whose first child is the
+	// actual YAML document.
+	var root yaml.Node
```
mbolt35: Does this imply that unmarshalling into `map[string]string` doesn't work the exact same way, but with much less overhead?

mbolt35: It's much better in my opinion to do the following: 

1. Parse full yaml as `map[string]any`
2. Traverse routes checking that:
    - Route is valid key in the current map
    - Value at route key is a `map[string]any` 
3. Marshal the route result with `resultBytes, err := yaml.Marshal(targetRoute)`
4. Unmarshal the `resultBytes` as `map[string]string` 

For example (with steps labeled):
```go
func ParseRoute(input []byte, routes []string) (map[string]string, error) {
	// 1. parse as map[string]any
	var m map[string]any
	err := yaml.Unmarshal(input, &m)
	if err != nil {
		return nil, fmt.Errorf("failed to parse root yaml: %w", err)
	}

	// 2. traverse the yaml based on the route. error if any of the routes don't exist
	for _, route := range routes {
		// 2a. Validate the route exists 
		value, ok := m[route]
		if !ok {
			return nil, fmt.Errorf("failed to locate route: %s within yaml", route)
		}

		// 2b. Validate the resulting value is a map[string]any
		innerMap, ok := value.(map[string]any)
		if !ok {
			return nil, fmt.Errorf("route at %s was not a map", route)
		}

		// update the node to the route 
		m = innerMap
	}

	// 3. Now that we've traversed the route, our `m` value can be marshalled back into yaml,
	// and then unmarshalled regularly
	resultBytes, err := yaml.Marshal(m)
	if err != nil {
		return nil, fmt.Errorf("failed to marshal route yaml block: %w", err)
	}

	// 4. finally unmarshal as map[string]string to take advantage of all the yaml conversion and validation
	var labels map[string]string
	err = yaml.Unmarshal(resultBytes, &labels)
	if err != nil {
		return nil, fmt.Errorf("failed to parse yaml: %w", err)
	}

	return labels, nil
}
```

mbolt35: There are a few cases where we use tricks like this to parse a specific subsection of a configurations (Storage configs do something similar to this). While I would never take advantage of this approach in a hot path, my opinion is that it's much more appropriate in this context. This lets us take advantage of all the yaml v3 validation without having to reproduce it line for line in our code base.

mbolt35 (changes requested):
Please review the feedback on yaml v3 parsing as `map[stirng]string`. Lets discuss if you have any questions.

core/pkg/external/configmap_source.go
avrodrigues5: I dont think we can assume default key. The default key becomes the yaml that is passed we can overwrite it by saying config.yaml

```
>kubectl create configmap external-labels-alias-without-configname --from-file=alias.yaml
```
Creates 

```
(arn:aws:eks:us-east-2:297945954695:cluster/qa-eks3/alan-external-config-test) alanrodrigues@Alans-MacBook-Pro in /Users/alanrodrigues
>k get cm external-labels-alias -oyaml                   
apiVersion: v1
data:
  config.yaml: |
    defaultEnvironment: &defaultEnvironment Production
    defaultTeam: &defaultTeam Infra

    labels:
      environment: *defaultEnvironment
      team: *defaultTeam
kind: ConfigMap
metadata:
  creationTimestamp: "2026-07-16T15:44:55Z"
  name: external-labels-alias
  namespace: alan-external-config-test
  resourceVersion: "405485109"
  uid: a72bc784-6f65-43d6-a804-465250430aea
```

```
>kubectl create configmap external-labels-alias --from-file=config.yaml=alias.yaml   
```

```
>k get cm external-labels-alias-without-configname -oyaml      
apiVersion: v1
data:
  alias.yaml: |
    defaultEnvironment: &defaultEnvironment Production
    defaultTeam: &defaultTeam Infra

    labels:
      environment: *defaultEnvironment
      team: *defaultTeam
kind: ConfigMap
metadata:
  creationTimestamp: "2026-07-16T16:49:19Z"
  name: external-labels-alias-without-configname
  namespace: alan-external-config-test
  resourceVersion: "405513977"
  uid: 6a29e717-f7ff-4501-aea8-4d5aac00d067
```

core/pkg/external/configmap_source.go
avrodrigues5: Excellent suggestion. All of the test cases passed.

One test case caused slight deviation from my initial implementation https://github.com/opencost/opencost/pull/3887/changes#diff-3a04c27f217bd385e14e433eb8ec36d859e6b31b4d8c8b8b2905096b9881cc1bR246.

When yaml have null values im appending empty label value. This is also the behavior with kubernetes

```
k label node ip-192-168-116-155.us-east-2.compute.internal testemptystring=""
```

I see that the label get populated in the node. For transparency i mention this deviation

Screenshot 2026-07-16 at 10 18 16 AM

core/pkg/external/configmapsource.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// we have nested routes, parse route
+	if len(routes) > 0 {
+		// parse with routes
+		return parseRoute(input, routes)
+	} else {
+		// no routes, just parse yaml as is
+		return parseNormally(input)
+	}
+}
```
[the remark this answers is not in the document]
avrodrigues5:
```

routes := parts[:0]
--
for _, p := range parts {
if p != "" {
routes = append(routes, p)
}
}


```

This isnt needed. I took other suggestion however

When user give prometheusK8s..externalLabels. It's better to fail than remove the accidental dot added.

pkg/costmodel/router.go
```diff
@@ -484,6 +485,49 @@ func Initialize(router *httprouter.Router, additionalConfigWatchers ...*watcher.
...
+		elNamespace := nlCfg.Namespace()
+		// If configmap is in the same namespace as the finops agent we can just use the same configmap watcher.
+		if elNamespace == "" {
+			configWatchers.Add(nlCfg.ConfigMapName(), external.WatchFunc(elSource, elProvider))
+		} else {
+			elWatchers := watcher.NewConfigMapWatchers(kubeClientset, elNamespace)
+			elWatchers.Add(nlCfg.ConfigMapName(), external.WatchFunc(elSource, elProvider))
+			elWatchers.Watch()
+		}
+	}
```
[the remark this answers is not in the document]
avrodrigues5: This does not add anything to have nil elProvider. It doesnt impact much

mbolt35 (approved):
Looks great! Thanks for the patience while we got all of this squared away. Excellent tests!

mbolt35:
> This seems like it fulfill the requirement. I do wonder if merging in these labels is better done at while computing kubemodel/assets rather than scrape. The concern being that adding a static label set one to every node in every scrape could add size to the wal and slow down the update/scrape process which is sensitive to time. Less importantly this would enable this feature for prom users, which are bypassed.
> 
> At a higher level I wonder if this could be configured via the aggregator and be applied during ingestion since as mentioned in one of the comments this is a feature exclusively for the agent. Maybe a little too late for that thought though

I considered all of what you're saying. The label "sanitization" stuff that happens was already obnoxiously over-allocating, which I think we fixed. This feature felt core to the collector itself (similar to prom external label appending), and ultimately, the stringtable formatting will remove the wasted space/load time in bingen encoded wals, and the extra pointer space will likely not be significant. Do I think there is a case where someone could create a truly problematic set of external labels? Yes, 100%.

And as for prom, you can do this already with label configurations in prom itself. This feature is meant to add this functionality to the collector.

PR #3890, opened 2026-06-30
CLDYON 7496: Pricing Overhaul + GCP + Small Cleanup

alexsouthard:
## Description
Does 3️⃣ primary things:
1. Updates the pricing generation to use a cronjob, uses the new data types, and embeds a pricing file as part of the package (which have now been split out into usd/cny). USD is the only currency which all three providers support since AWS only does USD normally and CNY for the chinese region. 🇨🇳 CNY isn't even fully supported since GCP doesn't use it, but I just kept it in because the logic was already there and the logic is separate when it comes to fetching the data
1. Adds in the GCP pricing module. This was just a copy paste from an existing PR: https://github.com/opencost/opencost/pull/3851 (Added here just to make things easier to get things reviewed/not out of sync)
1. Removes some of the old pricing files. Least important part but just cuts them out 🤷 

For the pipeline action to work it will need 3 secrets in the repo
1. GCP_API_KEY
1. SLACK_BOT_TOKEN
1. SLACK_PRICING_CHANNEL_ID

I tested by setting a GCP_API_KEY locally from the cloudability account but I doubt that that's what kubecost wants to use and also I don't have a way of setting it here because im not a repo owner.

I can probably whip together something for the SLACK_BOT_TOKEN and perhaps make a new channel just for this, but I think those are sort of gated by the ability to create these secrets in the first place (and I don't believe i can check what existing secrets there are so hopefully they don't overlap 😓). I haven't done an auto slack alert before so it will probably be a tiny secondary problem on setting that up.

## Related Issues
https://apptio.atlassian.net/browse/CLDYCON-7496

## User Impact
No real impact

## Testing
Manual testing on the actual cli
Unit tests

Tests should start working once its hooked up into the actual sources


.github/workflows/compare-pricing-data.yaml
```diff
@@ -0,0 +1,95 @@
+name: Compare Pricing Data
+
+on:
+  schedule:
+    - cron: '0 10 * * *'
```
alexsouthard, line 5: Runs every day, don't know if that's huge overkill but it seemed like an easy interval to me. Another good option is like once a week on like Mondays or something

nikovacevic: Every day is totally fine. Huge improvements from every time the agent starts, on every agent, all the time.

modules/pricing/public/cmd/main.go
```diff
@@ -91,3 +74,36 @@ func run(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+// comparePricing comparesa fresh pricing set against the existing pricing-data.json
```
nikovacevic: `comparesa` => `compares a`

modules/pricing/public/cny/module.go
```diff
@@ -0,0 +1,156 @@
...
+	"fmt"
+
+	"github.com/opencost/opencost/core/pkg/pricing"
+	"github.com/opencost/opencost/core/pkg/reader"
+	"github.com/opencost/opencost/core/pkg/unit"
+	"github.com/opencost/opencost/modules/pricing/public"
+)
+
+// PricingModule must satisfy the pricing.PricingModule interface
+var _ pricing.PricingModule = (*PricingModule)(nil)
```
nikovacevic, line 15: Excellent

modules/pricing/public/usd/pricing-data.json
```diff
@@ -0,0 +1,7 @@
+{
+  "nodePricing": [],
+  "persistentVolumePricing": [],
+  "clusterPricing": [],
+  "networkPricing": [],
+  "servicePricing": []
+}
```
nikovacevic: Seems like, in order to use the post-generation module, where such a module embeds the JSON file and hydrates Go pricing structs from it, we'll need to have this JSON filled out and checked into Go, right? Or did you come up with another plan for pulling in a versioned public pricing module as a dep?

modules/pricing/public/usd/module.go
```diff
@@ -0,0 +1,156 @@
...
+	return nil, fmt.Errorf("service pricing not yet implemented")
+}
+
+func (pm *PricingModule) GetPricingSet(ctx context.Context) (*pricing.PricingSet, error) {
+	return pm.store.GetPricingSet(ctx)
+}
+
+func (pm *PricingModule) SourceKind() string {
+	return "public-usd"
+}
```
nikovacevic, line 175: I think we want `public` here for source kind. Then source name can be `public-usd`. What do you say @Sean-Holcomb ?

modules/pricing/public/usd/module.go
```diff
@@ -0,0 +1,156 @@
...
+	return nil, fmt.Errorf("network pricing not yet implemented")
+}
+
+func (pm *PricingModule) NewNetworkPricingReader(ctx context.Context) (reader.Reader[*pricing.NetworkPricing], error) {
+	return nil, fmt.Errorf("network pricing not yet implemented")
+}
+
+func (pm *PricingModule) GetServicePricing(ctx context.Context, props pricing.ServicePricingProperties) (*pricing.ServicePricing, error) {
+	return nil, fmt.Errorf("service pricing not yet implemented")
+}
```
nikovacevic, line 127: @Sean-Holcomb should we file a follow-on ticket to implement these? Even if we can't draw them from the APIs, perhaps we could hard-code them for the time being. Or maybe we should leave them empty and fall back on the basic pricing module, which users can configure.

core/pkg/embeddedstorage/embeddedstorage.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/opencost/opencost/core/pkg/storage"
+)
+
+// EmbeddedStorage implements storage.Storage by reading directly from embedded bytes
+// without copying them to memory
+type EmbeddedStorage struct {
+	data []byte
+	path string
+}
```
nikovacevic: Love it. Nit: could we move this to `core/pkg/embeddedstorage.go` so that others can leverage it?

core/pkg/embeddedstorage/embeddedstorage.go
```diff
@@ -0,0 +1,104 @@
...
+	}, nil
+}
+
+// Read returns the embedded data directly without copying
+func (es *EmbeddedStorage) Read(path string) ([]byte, error) {
+	if path != es.path {
+		return nil, storage.DoesNotExistError
+	}
+	return es.data, nil
+}
```
nikovacevic: I think this is OK for now... certainly for our purposes. But if we do move to `pkg/core`, I think reusing the embedded byte array could cause some issues. Or we need to make it clear that what gets `Read` is read-only -- cannot be modified directly.

People who try to do that will find out quickly, at any rate.

modules/pricing/public/embeddedstorage.go
```diff
@@ -0,0 +1,104 @@
...
+		path: path,
+	}
+}
+
+func (es *EmbeddedStorage) String() string {
+	return string(es.StorageType())
+}
+
+func (es *EmbeddedStorage) StorageType() storage.StorageType {
+	return storage.StorageTypeMemory
```
nikovacevic: Let's make a new type here in `core/pkg/storage/storagetype.go`?

```
const (
>>> StorageTypeEmbedded StorageType = "embedded" <<<
	StorageTypeMemory      StorageType = "memory"
	StorageTypeFile        StorageType = "file"
	StorageTypeCluster     StorageType = "cluster"
	StorageTypeBucketS3    StorageType = "bucket|s3"
	StorageTypeBucketGCS   StorageType = "bucket|gcs"
	StorageTypeBucketAzure StorageType = "bucket|azure"
)
```

core/pkg/embeddedstorage/embeddedstorage.go
```diff
@@ -0,0 +1,104 @@
...
+// NewEmbeddedStorage creates a storage that serves embedded data without copying it
+func NewEmbeddedStorage(data []byte, path string) *EmbeddedStorage {
+	return &EmbeddedStorage{
+		data: data,
+		path: path,
+	}
+}
+
+func (es *EmbeddedStorage) String() string {
+	return string(es.StorageType())
```
nikovacevic: Probabth `es.path` here?

nikovacevic: `FileStorage` returns `baseDir`, e.g.

modules/pricing/public/embeddedstorage.go
```diff
@@ -0,0 +1,104 @@
...
+	return path == es.path, nil
+}
+
+func (es *EmbeddedStorage) List(path string) ([]*storage.StorageInfo, error) {
+	if path == "" || path == "." || path == "/" {
+		return []*storage.StorageInfo{
+			{
+				Name:    es.path,
+				Size:    int64(len(es.data)),
+				ModTime: time.Now(),
```
nikovacevic: Can't imagine it matters, but `time.Zero()` might be more appropriate. Or just leave it unset.

modules/pricing/public/cny/module.go
```diff
@@ -0,0 +1,156 @@
...
+
+//go:embed pricing-data.json
+var embeddedPricingData []byte
```
nikovacevic: When you actually generate the JSON, what is the file size? Do we know if it'll be a problem to retain the byte array in memory? Seems like this design would do that. With the way we've designed the `PricingModule` interfaces, I'm not even sure there's a great way around it that doesn't involve re-parsing the whole embedded file into a new `[]byte` on each operation. Hopefully it's just not an issue, but we should try to get some numbers wrapped around that.

nikovacevic (changes requested):
Couple of recommendations here, but this is shaping up nicely!

modules/pricing/public/usd/pricing-data.json
alexsouthard: Yeah in this current implementation the JSON has to be filled out and checked in. I originally had it as part of this PR but I wanted to make another after it's merged in that includes the data just not to scare anyone by the sheer number of lines 😱 

Another option is that the package includes a version or tag that points to a file inside an S3 bucket where the consumer then pulls the data from, which removes the need for the pricing data to live in memory while allowing us to bump pricing data by just increasing a version in the consumer, but this isn't the path that I went with just for complexity reasons. This could be a totally valid use case still and maybe more programmatic, I just didn't have a strong sense of the gravity or the direction that I wanted to take this, so I'm open to thoughts 😄

modules/pricing/public/cny/module.go
alexsouthard: USD: 61Mb
CNY: 43Mb

Yeah this is sort of a gray zone for me, the data set is embedded in the package but it's specifically not retaining the object in memory because that seems like it would overlap. So each time it reparses the whole thing out which doesn't sound amazing on paper but I'm not really positive how else we would be able to tie the data to specific versions while also keeping the memory footprint low (other than like a remote storage option like I mentioned earlier or using a different data format that isn't a super inefficient JSON). 

I'm totally up for changing things here because I'm not in love with this solution but it's a little difficult for me to pinpoint what we really want out of this

modules/pricing/public/usd/module.go
Sean-Holcomb: I think that later is correct. Hard coding feels like it would break the functionality of the basic pricing module in a way. In that you would no longer be able to set these prices on providers that this is hard coded

modules/pricing/public/cny/module.go
Sean-Holcomb: When is this loaded into memory from the embedded file, on start up or when the model is used?

core/pkg/embeddedstorage/embeddedstorage.go
Sean-Holcomb: This does feel like a pretty big departure from the intention of this interface. This makes it seem `NewStoragePricingStore` doesn't actually needs a full storage interface. It needs a different interface which could wrap a storage and also have an embedded implementation.

modules/pricing/public/usd/pricing-data.json
Sean-Holcomb: The draw back of the bucket idea is that it breaks the module in air gapped environments. 

One thought that comes to mind is that we could convert the pricing set Json to bingen and save it to a file in a scratch director on start up. Rather than re loading the whole thing into memory each time we could stream the bingen, while searching for matches and also include a cache for property sets that have already been queried. Not saying we do that but just pointing out that there are options

Sean-Holcomb (commented):
I don't really like where embedded storage, but if we want to push forward I wont block

modules/pricing/public/cny/module.go
alexsouthard: Just checked it out a little more, looks like I was wrong about saying they both are in RAM when imported. They apparently both have virtual-memory allocations but only the package that's pulled in/unmarshalled/whatever operation of us actually hydrating the data will put it into physical memory, which is better than I previously thought

core/pkg/embeddedstorage/embeddedstorage.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/opencost/opencost/core/pkg/storage"
+)
+
+// EmbeddedStorage implements storage.Storage by reading directly from embedded bytes
+// without copying them to memory. This should be read-only!
+type EmbeddedStorage struct {
+	data []byte
+	path string
+}
```
nikovacevic: Instead of `EmbeddedStorage` perhaps we can make a new `pricing.Store` that is:

```go
// jsonstore.go
package pricing

type JSONPricingStore struct {
	data  []byte
	mu    sync.RWMutex
}
```

modules/pricing/public/usd/module.go
```diff
@@ -0,0 +1,156 @@
...
+	ctx := context.Background()
+	
+	// Create an embedded storage that serves the embedded data directly without copying
+	embeddedStorage := embeddedstorage.NewEmbeddedStorage(embeddedPricingData, "pricing-data.json")
+	
+	// Create a pricing store backed by the embedded storage
+	store, err := pricing.NewStoragePricingStore(ctx, embeddedStorage, "pricing-data.json")
+	if err != nil {
+		return nil, fmt.Errorf("creating USD pricing store: %w", err)
+	}
```
nikovacevic: Option 2: this becomes a `MemoryPricingStore` and the embedded JSON file is used ephemerally, read into `[]byte` array, and then those bytes are GC'd.

within `NewPricingModule()`...
```go
//go:embed pricing-data.json
var embeddedPricingData []byte

// unmarshal the JSON into *pricing.PricingSet -- return err if that fails
var pricing PricingSet
err = json.Unmarshal(data, &pricing)
if err != nil {
	return nil, fmt.Errorf("decoding pricing: %w", err)
}

store := NewMemoryPricingStore()
store.SetPricingSet(pricingSet)
```

nikovacevic: We think this is the better option for now, but it does force us to pay the memory penalty of holding Go objects and cloning.

Sean-Holcomb (approved):
This addresses all of my issues

modules/pricing/public/cmd/main.go
```diff
@@ -44,50 +47,120 @@ func run(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	if err := writeJSONL(dir+"/persistentvolumes.jsonl", ps.PersistentVolumePricing); err != nil {
+		return err
+	}
+	return nil
+}
+
+// writeJSONL marshals each item in items as a single line and writes them to path.
+func writeJSONL[T any](path string, items []T) error {
```
nikovacevic, line 79: Nice!

modules/pricing/public/usd/module.go
```diff
@@ -0,0 +1,192 @@
...
+func NewPricingModule() (*PricingModule, error) {
+	return &PricingModule{}, nil
+}
+
+func (pm *PricingModule) newNodeReader() (reader.Reader[*pricing.NodePricing], error) {
+	f, err := embeddedFS.Open("nodes.jsonl")
+	if err != nil {
+		return nil, fmt.Errorf("opening embedded nodes.jsonl: %w", err)
+	}
+	return reader.NewJSONLinesReader[*pricing.NodePricing](f), nil
```
nikovacevic, line 32: 😍

nikovacevic (approved):
Looks good. Maybe we can find some time later in the week to run these scripts and generate the JSONL data? I'd love to see the actual results! It would also be very interesting, as a test, to run it multiple times and make sure that the checksums match.

PR #3906, opened 2026-07-07
fix: default handling in custom provider

### LINKED ISSUE #3905: Custom(on-prem): warning messages keep showing up when certain fields are not set
**Describe the bug**
In on-prem(Nutanix Kubernetes Platform) environment where the overrides for custom pricing is not specified, there could be a bunch of WARN log message keep emitting:

1. WRN Error getting LoadBalancer cost: strconv.ParseFloat: parsing "": invalid syntax --  every minute
1. WRN Failed to load 'name' field for ClusterInfo -- every 5 minute

Currently, we need workaround like [this](https://github.com/opencost/opencost/issues/2925#issuecomment-2705033852) to make it less noisy, but that only addresses #1.

For #2, in AWS and other provider, the cluster name will fall back to something else but for custom provider, there is just no fall back.

The values.yaml we use is [here](https://github.com/mesosphere/kommander-applications/blob/release-2.18/applications/opencost/2.5.14/release/cm.yaml#L8-L91)

**To Reproduce**
1. deploy using `custom` provider
2. do not override `opencost.customPricing.costModel`
3. look at opencost pod log
4. See the 2 log messages

**Expected behavior**
the defaults should work without emitting warnings

**Screenshots**

1.118.0

cwyl02:
## Description

to eliminate:

WRN Failed to load 'name' field for ClusterInfo
WRN Error getting LoadBalancer cost: strconv.ParseFloat: parsing "": invalid syntax

For the LB metrics, I provided default because I already see error handling.

For the Cluster Name warning, the ClusterInfo metric is produced without a proper fallback default like other clouds. 

## Related Issues

Fixes: https://github.com/opencost/opencost/issues/3905

## User Impact

N/A

## Testing

built a image and tested on my live cluster manually

### DIFF
--- a/configs/default.json
+++ b/configs/default.json
@@ -11,8 +11,5 @@
     "regionNetworkEgress": "0.01",
     "internetNetworkEgress": "0.12",
     "natGatewayEgress": "0.045",
-    "natGatewayIngress": "0.045",
-    "firstFiveForwardingRulesCost": "0.0",
-    "additionalForwardingRuleCost": "0.0",
-    "LBIngressDataCost": "0.0"
+    "natGatewayIngress": "0.045"
 }
--- a/configs/default.json
+++ b/configs/default.json
@@ -11,5 +11,8 @@
     "regionNetworkEgress": "0.01",
     "internetNetworkEgress": "0.12",
     "natGatewayEgress": "0.045",
-    "natGatewayIngress": "0.045"
+    "natGatewayIngress": "0.045",
+    "firstFiveForwardingRulesCost": "0.0",
+    "additionalForwardingRuleCost": "0.0",
+    "LBIngressDataCost": "0.0"
 }
--- a/pkg/cloud/provider/customprovider.go
+++ b/pkg/cloud/provider/customprovider.go
@@ -137,16 +137,25 @@ func (cp *CustomProvider) ClusterInfo() (map[string]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	m := make(map[string]string)
-	if conf.ClusterName != "" {
-		m["name"] = conf.ClusterName
-	} else {
-		m["name"] = coreenv.GetClusterID()
+
+	const defaultClusterName = "Custom Cluster"
+	clusterID := coreenv.GetClusterID()
+	if clusterID == "" {
+		clusterID = "default-cluster"
 	}
+	clusterName := conf.ClusterName
+	if clusterName == "" {
+		if clusterName = coreenv.GetClusterID(); clusterName == "" {
+			clusterName = defaultClusterName
+		}
+	}
+
+	m := make(map[string]string)
+	m["name"] = clusterName
 	m["provider"] = opencost.CustomProvider
 	m["region"] = cp.ClusterRegion
 	m["account"] = cp.ClusterAccountID
-	m["id"] = coreenv.GetClusterID()
+	m["id"] = clusterID
 	return m, nil
 }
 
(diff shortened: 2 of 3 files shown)


pkg/cloud/provider/customprovider.go
```diff
@@ -140,6 +140,8 @@ func (cp *CustomProvider) ClusterInfo() (map[string]string, error) {
 	m := make(map[string]string)
 	if conf.ClusterName != "" {
 		m["name"] = conf.ClusterName
+	} else {
+		m["name"] = coreenv.GetClusterID()
```
peatey: If `CLUSTER_ID` is also unset, this sets `name` to the empty string. Empty label values are dropped from the `kubecost_cluster_info` metric, so the Prometheus clustermap loader (`modules/prometheus-source/pkg/prom/clustermap.go`) will still fail `result.GetString("name")` and drop the cluster — the warning this PR aims to eliminate would still fire in that case.

Suggest mirroring the AWS provider's chain (`pkg/cloud/aws/provider.go` `ClusterInfo`): `ClusterName` → `CLUSTER_ID` → a static default, e.g.:

```go
const defaultClusterName = "Custom Cluster"
clusterName := conf.ClusterName
if clusterName == "" {
	if clusterName = coreenv.GetClusterID(); clusterName == "" {
		clusterName = defaultClusterName
	}
}
m["name"] = clusterName
```

---

configs/default.json
```diff
@@ -11,5 +11,8 @@
     "regionNetworkEgress": "0.01",
     "internetNetworkEgress": "0.12",
     "natGatewayEgress": "0.045",
-    "natGatewayIngress": "0.045"
+    "natGatewayIngress": "0.045",
+    "firstFiveForwardingRulesCost": "0.0",
+    "additionalForwardingRuleCost": "0.0",
```
peatey: Nit: the file loses its trailing newline here.

Bigger point: adding these keys to the shipped `default.json` only helps deployments that use the file untouched. Any Helm-managed custom pricing configmap that omits these keys (e.g. the Kommander one linked in #3905) still hits `strconv.ParseFloat: parsing ""` in `CustomProvider.LoadBalancerPricing()` every minute. Making the parser tolerate unset fields fixes the warning for all configs — see the suggestion in the review summary.

---

peatey (commented):
Thanks for filing #3905 with such a clear reproduction, and for following up with a fix!

I traced both warnings through the code and I think this PR fixes the reported symptom for stock deployments but leaves both root causes partially in place:

1. **`ParseFloat` warning**: the parse failure lives in `CustomProvider.LoadBalancerPricing()` (`pkg/cloud/provider/customprovider.go`), which hard-fails on any unset field. Adding the keys to `default.json` covers the default file, but any custom pricing configmap that omits them (including the Kommander one linked in the issue) still warns every minute — and `GetLBCost` aborts on the first error, so *all* LB costs are dropped rather than priced at zero. Suggest treating `""` as `0` in `LoadBalancerPricing()`, e.g. via a small helper:

   ```go
   func parsePriceOrZero(field, value string) (float64, error) {
   	if value == "" {
   		return 0, nil
   	}
   	price, err := strconv.ParseFloat(value, 64)
   	if err != nil {
   		return 0, fmt.Errorf("invalid custom pricing value %q for %s: %w", value, field, err)
   	}
   	return price, nil
   }
   ```

   While there, falling back to the existing `defaultLBPrice` config field when `firstFiveForwardingRulesCost` is unset would make the custom provider consistent with Alibaba/STACKIT/Oracle, which already honor it.

2. **ClusterInfo `name` warning**: falling back to `CLUSTER_ID` is right, but when that env var is also unset the label ends up empty and the clustermap loader still drops the cluster (see inline comment) — a final static default closes that hole.

Also note the DCO check is failing — commits need a `Signed-off-by` line (`git commit -s`, then force-push after amending).

Happy to see this land — with the parser-side fix plus a test for the empty-config case this would fully resolve #3905.

---

configs/default.json
cwyl02: done

pkg/cloud/provider/customprovider.go
```diff
@@ -137,14 +137,25 @@ func (cp *CustomProvider) ClusterInfo() (map[string]string, error) {
...
+	clusterID := coreenv.GetClusterID()
+	if clusterID == "" {
+		clusterID = "default-cluster"
+	}
+	clusterName := conf.ClusterName
+	if clusterName == "" {
+		if clusterName = coreenv.GetClusterID(); clusterName == "" {
+			clusterName = defaultClusterName
+		}
 	}
```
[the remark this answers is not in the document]
cwyl02: want to save a else

pkg/cloud/provider/customprovider.go
cwyl02: done

pkg/cloud/provider/customprovider.go
```diff
@@ -325,20 +336,39 @@ func (cp *CustomProvider) NetworkPricing() (*models.Network, error) {
...
-	fffrc, err := strconv.ParseFloat(cpricing.FirstFiveForwardingRulesCost, 64)
+
+	firstFiveForwardingRulesCostField := "firstFiveForwardingRulesCost"
+	firstFiveForwardingRulesCost := cpricing.FirstFiveForwardingRulesCost
+	if firstFiveForwardingRulesCost == "" && cpricing.DefaultLBPrice != "" {
+		firstFiveForwardingRulesCostField = "defaultLBPrice"
+		firstFiveForwardingRulesCost = cpricing.DefaultLBPrice
+	}
+
+	fffrc, err := parsePriceOrZero(firstFiveForwardingRulesCostField, firstFiveForwardingRulesCost)
```
ameijer, line 363: logDedupedInfoF("could not find pricing for XYZ in load balancer, defaulting to 0")

pkg/cloud/provider/customprovider.go
```diff
@@ -137,14 +137,25 @@ func (cp *CustomProvider) ClusterInfo() (map[string]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	m := make(map[string]string)
-	if conf.ClusterName != "" {
-		m["name"] = conf.ClusterName
+
+	const defaultClusterName = "Custom Cluster"
```
thomasvn, line 141: Is this backwards incompatible? Are CustomProvider users now getting a new default cluster name, if one is not set via `coreenv.GetClusterID()`.

I'm ok with this! But would just need to call this out in release notes.

ameijer (approved):
LGTM, legacy case is marginal

PR #3908, opened 2026-07-08
AI Inference: Namespace uid label

simanadler:
## Description
Adds the k8s namespace UID to the 3 inference metrics as per request from @ameijer 

## User Impact
Non-breaking change.

UIDs are used to link resources. The use case will be to eventually tie these to specific GPUs for optimization logic

## Testing
Unit testing
Full live integration testing to be performed after PR #3901 is merged, since the testing cluster needs that fix.


Sean-Holcomb (changes requested):
I was hoping to get namespace uid introduced at the initial metrics emission level. Digging deeper I am seeing that the metrics are coming from vllm. Is it the case that each vllm container is exposing its own metrics endpoint? ideally we could configure prom or vllm to include uid labels. 

I have some general concerns about how decoders result types are being used I am also interested in what other labels are present on these labels. 

Overall namespace_uid which can be used to uniquely identify namespaces if they are recreated in the same time window aren't particularly useful if they are populated after window aggregation. That being the case I think we can just close this one.

simanadler:
> I was hoping to get namespace uid introduced at the initial metrics emission level. Digging deeper I am seeing that the metrics are coming from vllm. Is it the case that each vllm container is exposing its own metrics endpoint? ideally we could configure prom or vllm to include uid labels.
> 
> I have some general concerns about how decoders result types are being used I am also interested in what other labels are present on these labels.
> 
> Overall namespace_uid which can be used to uniquely identify namespaces if they are recreated in the same time window aren't particularly useful if they are populated after window aggregation. That being the case I think we can just close this one.

Agreed on closing. 

To confirm the mechanism: the UID doesn't come from vLLM (vLLM has no namespace-UID concept and there's no clean way to inject one via scrape config) — it's joined from OpenCost's own namespace_info by namespace name in combineMetrics, after per-window aggregation. 

As you note, that means it can't disambiguate a same-name/same-window namespace recreation, which is the only case where the UID would earn its keep. So the label adds cost-cardinality without delivering the identity guarantee. 

Closing.

Separately — can you point me at the specific Decode*Result usage you were uneasy about? I want to make sure that concern is captured even though we're dropping the UID label itself.

PR #3915, opened 2026-07-09
Inference saturation metrics: collect Model Server Protocol gauges in both sources, land in KubeModel

cliffcolvin:
## What

Collects the model-server scheduler telemetry standardized by the [Gateway API Inference Extension Model Server Protocol](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/003-model-server-protocol) (KV-cache utilization, queue depth, running requests) in **both** data sources, and lands it in the **KubeModel** as a new `InferenceEngine` entity. This builds on the AI Inference Costs work (#3845), which collects token and timing metrics but not the capacity signals.

| Signal | vLLM metric | Aggregations |
|---|---|---|
| KV-cache utilization (0 to 1) | `vllm:kv_cache_usage_perc` | avg, p95, max |
| Queue depth (requests waiting) | `vllm:num_requests_waiting` | avg, p95, max |
| Running requests (achieved batch) | `vllm:num_requests_running` | avg, p95, max |
| Preemptions (KV thrash / pressure) | `vllm:num_preemptions_total` | delta over window |

The avg/p95/max triple summarizes each gauge's window distribution. Quantiles compute identically from both data sources (Prometheus `quantile_over_time` and a `QuantileOverTime` aggregator in the collector); bucketed histograms would need one subquery per bucket per metric on the Prometheus side.

## Why

Host-level GPU metrics cannot tell you whether an LLM serving deployment is actually working for a living. A serving engine preallocates its memory budget (the way a database takes all the RAM it is given) and any nonempty batch keeps SMs busy, so `DCGM_FI_PROF_GR_ENGINE_ACTIVE` (which OpenCost already collects) and VRAM occupancy read 90+ percent for every healthy deployment regardless of load. In field measurements across concurrent deployments under identical load, hosts reported 93 to 97 percent busy while true capacity consumption ranged from 1.7 to 43 percent.

The engine's own scheduler state is the honest signal:

- **KV-cache utilization** reports how much of the engine's KV block budget the live request set holds.
- **Sustained queue depth above zero** means the replica is at or past saturation and latency is growing.
- **Running requests** gives the achieved batch, and its window maximum reveals the effective concurrent-sequence ceiling.

KV utilization alone does not answer "is this replica at capacity". Two independent constraints bound a model server and either can bind first: the KV-cache budget, and the concurrent-sequence limit `max_num_seqs`, which caps the running batch no matter how much KV memory is free. Short-context, high-concurrency serving exhausts batch slots while KV utilization stays low, since each sequence holds few KV blocks; long-context serving does the reverse. A consumer treating KV utilization as the only capacity signal will read a batch-saturated replica as nearly idle.

vLLM publishes no metric for `max_num_seqs`, and no scheduler-config or cache-config info metric carries it, so the denominator of batch occupancy cannot be collected directly. It is observable instead: while the queue is non-empty the engine admits every sequence it can, so the running gauge sits pinned at its effective limit, and the window maximum is that ceiling. That is why running requests are collected as avg, p95 and max rather than as an average alone.

Two properties make these the right raw material for a cost tool:

1. **They are an emerging cross-engine standard.** The Model Server Protocol requires exactly these signals for a server to be routable by an inference gateway, with published mappings for vLLM, SGLang, and Triton TensorRT-LLM; llm-d scrapes them today for endpoint picking. OpenCost consuming the same contract means no bespoke integration per engine.
2. **They stay valid under MIG.** Each MIG instance runs its own engine sized to its slice, so KV usage and queue depth are already expressed relative to the slice's capacity. Meanwhile `DCGM_FI_DEV_GPU_UTIL` is not supported at all when MIG is enabled, so under MIG the engine-level signals are the only reliable capacity telemetry.

## The entity: `InferenceEngine`

Named for the **inference engine instance**: the process that admits requests, batches them, and manages the KV-cache budget. That is what these signals describe. KV-cache utilization is a property of the engine's block allocator; queue depth and running requests are its admission scheduler; preemptions are its eviction events. A pod has none of those, and neither does a Deployment; the pod is where the engine runs, and the workload is the axis a consumer rolls up to.

The name also stays honest as engines are added. Field semantics are normalized to the Model Server Protocol, which defines the per-engine metric mapping externally, so SGLang and Triton TensorRT-LLM are additive mappings rather than new types, and the `Engine` field records which mapping produced an entry.

**Identity is the pod UID**, matching how every other entity in the KubeModel joins. One flat entry per engine, stored as `map[podUID]*InferenceEngine`, carrying `NamespaceUID` and `ModelName` as fields. `PodUID` indexes `kms.Pods` and `NamespaceUID` indexes `kms.Namespaces` directly. There is deliberately no model-level grouping entity; a rollup by served model is a view a consumer computes from `ModelName`.

## How

**`core/pkg/source`**
- Ten new `MetricsQuerier` methods (`QueryInferenceKVCacheUsageAvg/Max/P95`, `QueryInferenceQueueDepthAvg/Max/P95`, `QueryInferenceRunningRequestsAvg/Max/P95`, `QueryInferencePreemptions`), bringing the inference surface to 16
- New `InferenceEngineMetricResult` (`model_name`, `pod_uid`, `namespace_uid`, value) with a shared decoder used by both sources
- `mock`/`noop`/`record` implementations

**`modules/prometheus-source`**
- Window aggregations grouped by `(model_name, pod_uid, namespace_uid, <cluster label>)`, pinned to the window end, consistent with the clamping in `queryCounterDelta` from #3845
- `pod_uid` comes from the scrape config (one relabel rule on `__meta_kubernetes_pod_uid`, the same route dcgm-exporter series already use). There is no meta-label for a namespace UID, so `namespace_uid` is joined in from OpenCost's own `namespace_info`. That join is **non-fatal**: a deployment without `namespace_info` keeps its series with the UID unset rather than losing them, and both sides of the match carry the cluster label so a federated Prometheus does not produce a duplicate match.

**`modules/collector-source`** (Prometheus-free path)
- New `InferenceScraper`: discovers model-server pods by the `INFERENCE_MODEL_LABEL` pod label (default `llm-d.ai/model`), scrapes their metrics endpoints directly (port from the `prometheus.io/port` annotation when present, else `INFERENCE_SCRAPE_PORT`, default 8000). Clusters with no labelled pods produce zero targets and no scrape traffic, same as the DCGM scraper on GPU-free clusters.
- Serving engines do not self-report their Kubernetes identity, so the scraper attaches `namespace`, `namespace_uid`, `pod` and `pod_uid` at scrape time, and fills in `model_name` for series the engine does not label with it (`vllm:cache_config_info` carries the cache configuration, not the model). The engine's own `model_name` always wins.
- 16 declarative `MetricCollector`s. The ten saturation collectors group by `(model_name, pod_uid, namespace_uid)`, matching the DCGM collectors.
- The six token, timing and cache-config queries that previously returned "inference metrics not supported by collector source" are now implemented, so the whole inference surface works without Prometheus at no extra scrape traffic.

**KubeModel**
- New `InferenceEngine` entity, split out from the universal k8s structures the same way `DCGMDevice` is
- `computeInferenceEngines` compute function (all queries degrade gracefully to an empty map)
- `protos/kubemodel/inference.proto` alongside `gpu.proto`
- bingen codecs regenerated at version 3: `DCGMDevices` and `InferenceEngines` move from `@bingen:field[ignore]` to `@bingen:field[version=3]`, so both serialize with the KubeModelSet; older payloads (version 2 and below) continue to decode with the new fields skipped by the version gates
- `computeDCGMDevices` and `computeInferenceEngines` are enabled in `computeFuncs` (the DCGM compute function existed but was not wired in; its orchestration test already carried the overrides, and the previously commented assertion is now active)

## Design alignment with the kubemodel device direction

This intentionally does not introduce a generic paradigm alongside the per-source device modeling direction:

- **An engine is not a device.** It is the capacity manager between the workload and the device(s). `InferenceEngine` is a sibling of `DCGMDevice` (source-scoped, split out from the universal k8s structures), not a generalization of device modeling.
- **Normalization is by an upstream contract, not a lowest-common-denominator type.** A shared concrete type is safe here, where it is not for GPU vendors (a field like "power" means different things for AMD versus NVIDIA), because the Model Server Protocol defines the per-engine mapping and semantics externally. Worth noting explicitly: this is the one respect in which the `DCGMDevice` analogy does **not** carry, since there is no equivalent cross-vendor contract for GPU telemetry.
- **Engine-to-device linkage is the `GetParent` analog and is deliberately out of scope.** Relating an engine to the physical device or MIG instance it occupies belongs to the DRA/device-plugin requests join.
- **Distributions via quantiles**, since they compute identically from both data sources.

## Scope: raw materials only

This PR does not compute derived efficiency numbers (saturation scores, waste rates, headroom extrapolations) or make recommendations. Those belong to consumers of the model. The goal is that any tool reading the KubeModel can compute capacity-aware cost efficiency from neutral, standardized measurements.

## Known limitations, documented in `docs/inference-cost-tracking.md`

- **These are decode-stage signals.** An engine serving pooling models (embedding, classification, reward) runs no decode loop and reports these gauges absent or degenerate. That reflects the workload's mode, not a collection failure.
- **One engine per pod is the common case, not a guarantee.** Under vLLM data parallelism each DP rank runs as its own core engine process inside the same pod. Keyed by pod UID, such a deployment collapses to one entry per pod. Extending identity to `(pod UID, engine)` is the shape that addresses it.
- **Per-replica signals are engine-relative.** Under GPU time-slicing (no hard capacity boundary) the KV denominator reflects the engine's self-declared budget rather than isolated hardware; MIG and whole-GPU are the clean cases.
- **Which constraint binds is a property of traffic shape**, so a consumer should compare KV utilization against the running-batch ceiling rather than assume either is the limit.

## Follow-ups (separate PRs)

- SGLang and Triton TensorRT-LLM metric mappings per the Model Server Protocol tables
- Request outcome counters as additional raw material for cost-per-outcome analysis
- Surfacing these fields through the `inferenceCost` API/exporter

## Testing

- `core`: `TestComputeInferenceEngines` (mirrors `TestComputeDCGMDevices`: empty data, single engine with all gauges and counters, multi-engine grouping, cross-namespace separation, missing-identity skipping); the `ComputeKubeModelSet` orchestration test now asserts `DCGMDevices` and `InferenceEngines` are populated; codec round-trip suite passes at version 3
- `collector-source`: `TestCollectorMetricsQuerier_QueryInferenceSaturation` and `TestCollectorMetricsQuerier_QueryInferenceCost` exercise the full store, aggregation and decode path with exact expected values; scraper tests cover identity attachment and the `model_name` fallback; `QuantileOverTime` has aggregator unit tests plus a guard for coarse grouping
- `prometheus-source`: query-shape tests for every gauge, plus `TestInferenceQueriesJoinIsNonFatalAndClusterScoped` and `TestInferenceQueriesApplyClusterFilter`
- `go build ./...` and `go test ./...` pass on core, opencost, prometheus-source and collector-source

## Companion chart change

The Prometheus path needs `pod_uid` on the model-server series, which is a scrape-config concern the OpenCost chart did not previously cover. Draft chart PR: opencost/opencost-helm-chart#379 (an opt-in PodMonitor that attaches `namespace`, `pod` and `pod_uid`). Operators who already scrape their model servers can instead add the single `pod_uid` relabel rule to their existing job.


modules/collector-source/pkg/metric/aggregator/quantileovertime.go
```diff
@@ -0,0 +1,75 @@
...
+// QuantileOverTime returns a MetricAggregatorFactory that computes the
+// phi-quantile (0 <= phi <= 1) of all values observed in the window.
+func QuantileOverTime(phi float64) MetricAggregatorFactory {
+	return func(labelValues []string) MetricAggregator {
+		return &quantileOverTimeAggregator{
+			labelValues: labelValues,
+			phi:         phi,
+		}
+	}
+}
```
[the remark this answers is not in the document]
cliffcolvin: Confirmed, and this was a real panic: with 3+ samples and phi > 1, rank = phi*(n-1) has a floor beyond the last index, and the previous bounds check only clamped the upper index. The existing test masked it by using exactly 2 samples. Reproduced with a 3-sample regression test (panicked as described), then fixed by clamping phi to [0, 1] at construction as suggested; the now-dead index clamps were removed. Fixed in f51c8e10.

core/pkg/model/kubemodel/inference.go
```diff
@@ -0,0 +1,126 @@
...
+// sits between the workload and the device(s). The type is concrete and
+// source-scoped: field semantics are normalized to the Model Server Protocol
+// (an upstream contract that defines the per-engine metric mapping), and the
+// Engine field preserves which engine's metrics populated the entry, so no
+// per-engine meaning is erased by the normalization. The replica-to-device
+// linkage (the GetParent analog of a MIG instance pointing at its physical
+// device) is deliberately not collected here; it belongs to the DRA/device
+// plugin requests join, which also relates replicas to MIG instances.
+// Utilization distributions (histograms) are a planned follow-up; this
+// version carries window avg/max scalars.
```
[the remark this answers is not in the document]
cliffcolvin: Right, the comment predated the p95 and preemption fields. Updated to describe the avg/p95/max distribution summaries and preemption counts. Fixed in 077f7223.

core/pkg/compute/kubemodel/inferenceserver.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/opencost/opencost/core/pkg/model/kubemodel"
+	"github.com/opencost/opencost/core/pkg/source"
+)
+
+// computeInferenceServers builds InferenceServer entries from the model-server
+// scheduler telemetry queries (Gateway API Inference Extension Model Server
+// Protocol gauges: KV-cache utilization, queue depth, running requests).
+// Servers are keyed by "model_name:namespace" with one replica entry per pod.
+// All five queries degrade gracefully: a data source with no model-server
+// telemetry produces an empty map.
```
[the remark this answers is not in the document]
cliffcolvin: Right, it is eight queries now. Reworded to name the signal set instead of a count so it will not drift again. Fixed in 077f7223.

modules/collector-source/pkg/metric/aggregator/quantileovertime.go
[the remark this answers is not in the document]
cliffcolvin: Follow-up: the old bounds check had the mirror-image gap for phi < 0 with 3+ samples (negative ceil(rank) index, unclamped from below), and the original negative-phi test had the same 2-sample masking flaw. The construction clamp covers both directions; 165d783a pins the negative side with a 3-sample regression case as well.

modules/collector-source/pkg/metric/aggregator/quantileovertime.go
```diff
@@ -0,0 +1,78 @@
...
+	"math"
+	"sort"
+	"sync"
+	"time"
+)
+
+// quantileOverTimeAggregator retains the values observed in the window and
+// computes the phi-quantile with linear interpolation between samples,
+// matching Prometheus quantile_over_time semantics.
+type quantileOverTimeAggregator struct {
```
Sean-Holcomb, line 39: Not that this is necessarily important with the collectors in place but this aggregator does not take into account aggregation over the label (non-temporal) dimension. 

Updates come at fixed timestamps if we said to aggregate by just names space we might get readings 

|  |  0:00 |  0:30 |  1:00 |
| ------ |  ------ | ----- | ----- |
| Pod1:Namespace1  |       2   |      4  |    1    |
| Pod2:Namespace1 |     3  |    5    |     1   |
| Namespace1 (agged) | 5 | 9 | 2 |

With these reading and the current implementation aggregated on namespace the result would be 4.75 however if we take the p95 of the samples aggregated by namespace then we get 8.6

modules/collector-source/pkg/scrape/inference.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+		}
+
+		url := fmt.Sprintf("http://%s:%d/metrics", pod.Status.PodIP, port)
+		log.Debugf("Inference: found target: %s", url)
+
+		targets = append(targets, inferenceTarget{
+			target:    target.NewUrlTarget(url),
+			namespace: pod.Namespace,
+			pod:       pod.Name,
```
Sean-Holcomb: including pod_uid here is trivial and makes these metrics fit into kubemodel in a much more straight forward way.

modules/collector-source/pkg/collector/metricsquerier.go
```diff
@@ -821,3 +821,41 @@ func (c *collectorMetricsQuerier) QueryInferenceCacheConfig(t time.Time) *source
 	}
 	return source.NewFuture(decoder, ch)
 }
+
```
Sean-Holcomb, line 911: With the scrape target implemented, should we also implement the other queryInference queries

modules/collector-source/pkg/collector/collector.go
```diff
@@ -1601,6 +1609,186 @@ func NewDCGMContainerUsageMaxMetricCollector() *metric.MetricCollector {
...
+//			vllm:kv_cache_usage_perc[1h]
+//		)
+//	) by (model_name, namespace, pod, cluster_id)
+
+func NewInferenceKVCacheUsageAvgMetricCollector() *metric.MetricCollector {
+	return metric.NewMetricCollector(
+		metric.InferenceKVCacheUsageAvgID,
+		metric.VLLMKVCacheUsagePerc,
+		[]string{
+			source.InferenceModelNameLabel,
```
Sean-Holcomb, line 1642: ideally we would be aggregating just by pod_uid here for all of these metrics

modules/collector-source/pkg/scrape/inference.go
Sean-Holcomb: additionally with namespace_uid could be populated if that is still considered relevant

modules/collector-source/pkg/scrape/inference.go
```diff
@@ -0,0 +1,176 @@
...
+	// model-server pod and whose value is the served model name. The default
+	// matches the label used by the inference cost feature and llm-d.
+	inferenceModelLabelEnv     = "INFERENCE_MODEL_LABEL"
+	inferenceModelLabelDefault = "llm-d.ai/model"
+
+	// inferenceScrapePortEnv overrides the default metrics port used when a
+	// model-server pod does not carry a prometheus.io/port annotation. The
+	// default matches the vLLM OpenAI-compatible server port.
+	inferenceScrapePortEnv     = "INFERENCE_SCRAPE_PORT"
+	inferenceScrapePortDefault = 8000
```
Sean-Holcomb: These should be moved to the env dirrectory

core/pkg/model/kubemodel/inference.go
```diff
@@ -0,0 +1,127 @@
...
+// linkage (the GetParent analog of a MIG instance pointing at its physical
+// device) is deliberately not collected here; it belongs to the DRA/device
+// plugin requests join, which also relates replicas to MIG instances.
+// Each gauge carries a window distribution summary (avg, p95, max) rather
+// than per-bucket histograms, since quantiles compute identically from both
+// data sources; preemption counts round out the pressure signals.
+// @bingen:generate:InferenceServer
+type InferenceServer struct {
+	ModelName string `json:"modelName"`
+	Namespace string `json:"namespace"`
```
Sean-Holcomb: I am not really seeing the importance of namespace here. Could we not just map this by pod_uid?

core/pkg/compute/kubemodel/inferenceserver.go
```diff
@@ -0,0 +1,96 @@
...
+				server = &kubemodel.InferenceServer{
+					ModelName: res.ModelName,
+					Namespace: res.Namespace,
+					// The querier contract is currently implemented with the
+					// vLLM metric mapping in both data sources; when further
+					// Model Server Protocol mappings are added, engine
+					// provenance must ride the query results instead.
+					Engine:   kubemodel.EngineVLLM,
+					Start:    start,
+					End:      end,
```
Sean-Holcomb: If these are just going to be the provided start and end they should be removed since they will just be the window of the KubeModelSet

Sean-Holcomb (commented):
Looks good overall. I was talking to Alex the other day about the other inference work and mentioned that this is another great example of 3rd party metrics that we need to fit into a schema defined for modeling kubernetes. 

The most important thing is to properly fit into the ecosystem the metrics themselves need to have UIDs. This is easy with collector as mentioned but is a bit more challenging with Prometheus. We are already updating the scrape config to add on namespace (and pod?) We need to show that we can do this for pod_uid (and namesapce_uid?) This will allow these metrics to cleanly interact with the rest of kubemodel.

My big question in regards to inference metrics what is the role of namespace here? Is it just part of the fully qualified pod identifier or is it an important dimension on which these models are judged. What is the name inferenceserver refering too, is that an attempt of an abstraction on the metrics that vllm provides?

dev404ai:
The new per-replica counter-delta path here also exposes a correctness gap in the existing inference-cost counters on current `develop` (`a7427ef5d590c77bb605c5ad890f7f5de2c287bb`).

I exercised the production `queryCounterDelta` with two replicas of one model over the same window:

- replica A: 900 → 100 after a reset, so the endpoint-only lower bound is 100;
- replica B: 100 → 1,000, so its delta is 900.

The reset-aware model total is therefore 1,000. The current queries first `sum by (model_name, namespace)` at each endpoint and then subtract, so they see 1,000 → 1,100 and return 100. A scripted-Prometheus test exercised the actual helper with the aggregate endpoint values induced by that two-replica scenario and reproduced the undercount deterministically; the native `modules/prometheus-source/pkg/prom` package remains green without the probe.

This affects prompt tokens, generation tokens, prefill time, output-token time, and cache hits because all five use the same helper. A reset masked by another replica's growth undercounts the token/time denominator and can inflate the derived cost per million tokens.

The smallest safe direction seems to be a shared helper that computes and reset-corrects deltas per `(model_name, namespace, pod)`, then re-aggregates them to the existing `map[string]float64` model/namespace contract. `queryCacheConfigs` already relies on `pod` for `prompt_tokens_total`, but I have not found the same label guarantee for all five cost counters. Is there a supported vLLM or Prometheus configuration where any of those counters lacks a stable `pod` label?

cliffcolvin:
@dev404ai Confirmed, and thank you for the depth here. I drove the real helper with your two-replica case and it returns 100 where the reset-aware total is 1,000, with the guard never firing because the aggregate delta is positive.

I raised #4005: per-replica reset-corrected deltas, folded back into the existing `model_name:namespace` contract, so no caller or result type changes. Your list of five affected callers was correct.

On your direct question about the `pod` label, yes, there are supported configurations where it is absent, and our own docs are ambiguous about whether it is required. `pod` is never self-reported by vLLM; it comes entirely from Kubernetes SD relabeling, so it is missing under non-Kubernetes service discovery, under `write_relabel_configs` or `labeldrop` that strip it for cardinality, and under federation of pre-aggregated recording rules like `sum by (model_name, namespace) (...)`. Rather than making `pod` a hard dependency, those series now collapse into a single bucket that reproduces today's behaviour.

Splitting it out of this PR because it is in the merged cost path from #3845 and affects `develop` users today, independent of the saturation work here, which makes it cherry-pickable on its own.

@Sean-Holcomb would you be willing to take a look at #4005? It touches the inference cost counters rather than the saturation work in this PR, and I would value your read on the degraded path in particular.

modules/collector-source/pkg/metric/aggregator/quantileovertime.go
cliffcolvin: Both of your numbers reproduce exactly. Flat pool `[1,1,2,3,4,5]` gives 4.75; summing per timestamp first gives `[2,5,9]` and 8.6.

I did not change the aggregator's arithmetic. I changed the grouping so the precondition holds, and then guarded it. All three collectors registered against `QuantileOverTime` now group by `(model_name, pod_uid, namespace_uid, engine)`, which makes one-series-per-aggregator structural rather than assumed.

The reason for per-series-then-combine rather than sum-then-quantile is parity with the Prometheus source, which runs `max by (...) (quantile_over_time(...))` for these same signals. Also, summing is only meaningful for additive quantities: `vllm:kv_cache_usage_perc` is a fraction, so two engines at 0.4 and 0.5 must not become 0.9.

You were right that a documented precondition is not enforcement. `Update` already received a timestamp and discarded it, and within one series each scrape timestamp appears once, so a repeat is direct evidence of pooling. It now warns once per aggregator on the first repeat (`64e9cb88`).

Worth noting that guard would have fired on a data-parallel deployment before `1c2e4b4a`, since two engine cores in one pod pooled into one aggregator. That is now structurally unreachable for inference, so the guard exists for the next caller rather than this one.

modules/collector-source/pkg/scrape/inference.go
cliffcolvin: Done. `inferenceTarget` carries both from Kubernetes discovery, and every kept sample gets `namespace`, `namespace_uid`, `pod` and `pod_uid`.

A missing namespace UID does not drop the sample: it is recoverable from `kms.Pods[PodUID]` and is carried only for convenience. `TestInferenceScraperGetTargetsUnknownNamespace` pins that.

Chasing this found a real bug in the same function, fixed in `4cea6b6f`. The scraper attached Kubernetes identity but not `model_name`, and `vllm:cache_config_info` is labelled with the cache configuration rather than the model (which is exactly why the Prometheus path joins it against a token metric). The collector filtered on `model_name != ""`, so every real sample was dropped and prefix-caching detection was silently dead on the Prometheus-free path.

It passed review because the test fixture invented a `model_name` on that series. The fixture now matches what an engine actually emits.

modules/collector-source/pkg/collector/metricsquerier.go
cliffcolvin: Yes. All sixteen `QueryInference*` methods are implemented; no `"not supported by collector source"` stub remains.

No extra scrape traffic: same endpoint, same pod-selection predicate. The keep-list grew by six names applied to a response body the scraper already fetches and parses.

One thing I would rather flag than have you find: counter-reset semantics still differ between the two sources for the same metric. For samples `100, 110, restart, 5, 15, 25`, Prometheus' own `increase()` reports 35. The Prometheus path reported 25 and now reports 35 (`8145d523`). The collector path reports 30, because the shared `Increase` aggregator credits nothing for the step where a counter goes backwards.

Converging it is one line, but `Increase` is shared by every counter collector in the module, not just the inference ones, and `increase_test.go` pins the current behaviour deliberately. That reads to me like a repo-wide counter-semantics decision rather than something to slip into an inference PR. Happy to do it in a separate PR, or here if you would rather. Your call.

modules/collector-source/pkg/collector/collector.go
cliffcolvin: Done, and you were right for a stronger reason than consistency.

The ten saturation collectors now group by `(model_name, pod_uid, namespace_uid, engine)`. The six cost collectors do too, with the rollup to `model_name:namespace` moved into the querier (`7f0e9624`), so the reported key is unchanged.

The cost ones mattered most. Sharing one `Increase` aggregator across replicas produces two errors: a replica restarting drags the pooled total below its previous value, so the whole cycle's increase is discarded including replicas that never reset; and a replica discovered mid-window folds its entire cumulative counter in as one cycle's growth. Per-pod aggregators remove both.

Known cost: a fresh aggregator does not credit its first sample, so a new pod now costs one scrape interval instead of one per model. During a rolling deploy that is one interval per replacement pod. Much smaller than the errors above, and pinned by test.

On why it is four labels rather than just `pod_uid`: `MetricCollector.Get` rebuilds the result label map only from the group-by list, so a label left out is erased from the output. The namespace name in particular is what the querier rebuilds the result key from.

modules/collector-source/pkg/scrape/inference.go
cliffcolvin: Moved. `INFERENCE_MODEL_LABEL` and `INFERENCE_SCRAPE_PORT` now live in `modules/collector-source/pkg/env/collectorenv.go` with `GetInferenceModelLabel` and `GetInferenceScrapePort`, following the shape of the existing accessors. Defaults unchanged. The scraper keeps only `prometheusPortAnnotation`, which is a Kubernetes annotation key rather than a setting.

Two you may want the other way. `INFERENCE_MODEL_LABEL` is unprefixed deliberately, because the scraper must select pods by the same label the inference cost feature reads. `INFERENCE_SCRAPE_PORT` is collector-source-only, so `COLLECTOR_` would arguably fit better. Say the word and I will prefix it.

Also worth naming: `llm-d.ai/model` is now a literal in two packages, since the root module requires collector-source and so collector-source cannot import root `pkg/env`. `core/pkg/env` is importable by both if you want one source of truth.

core/pkg/model/kubemodel/inference.go
cliffcolvin: Done, and it was worse than the one field you flagged. I checked the whole package at the commit you reviewed: `Namespace` as a *name* appeared exactly once in non-test kubemodel code, on this type. Every other namespace-scoped entity carries `NamespaceUID`. The replica map was also keyed by pod *name*, while `DCGMDevice.PodUsages`, the entity I cited as precedent, was already keyed by pod UID.

It is now one flat entry per engine. `InferenceServerReplica` is gone. `NamespaceUID` survives as a field rather than as identity, because it is the cost and ownership axis consumers aggregate on, and it is not required by validation since it is recoverable from `kms.Pods[PodUID]`.

One correction to my own answer: pod UID alone turned out not to be sufficient either. See the reply on your review comment.

core/pkg/compute/kubemodel/inferenceserver.go
cliffcolvin: Removed. The first branch of your condition applied: they were assigned verbatim from the compute function's window arguments, so they only restated `KubeModelSet.Window`.

The `computeDCGMDevices` branch does not apply as a fix. That one derives real coverage because it has a dedicated uptime query carrying first and last sample timestamps. Every inference query is a window aggregation returning one value per series, and the decoder takes the last value and discards timestamps, so nothing in the inference surface carries sample coverage.

The consequence is that the entity has no coverage signal at all: a consumer cannot distinguish an engine that served the whole window from one that started near the end. If you want that, an inference uptime query mirroring `QueryDCGMDeviceUptime` is the shape, and I would rather do it as a follow-up than widen this further.

cliffcolvin:
@Sean-Holcomb thanks for this round. Pushed, so everything below is on the branch now.

Three answers, and one place where I was wrong.

**Where I was wrong: the `DCGMDevice` precedent I leaned on does not carry.** I cited it to justify a shared concrete type normalized across engines. But device modeling avoids a normalized cross-vendor type precisely because no external contract fixes what a field means, since "power" is a different measurement on AMD and NVIDIA. That is the argument for per-vendor structs, and quoting DCGM as support for the opposite was sloppy. What actually justifies a shared type here is the Model Server Protocol, which defines the per-engine mapping and semantics upstream. I have rewritten the design note to say that, and to carry the constraint it implies: a signal with no Model Server Protocol mapping does not belong on this type.

**The role of namespace.** Not a judged dimension, and it should not have been identity. vLLM emits `model_name` and nothing about where it runs; every Kubernetes label on these series is attached from outside. Namespace was doing two jobs and only one was real: disambiguating the same model served in two places, which UID does structurally and better, and being the cost and ownership axis, which is why `NamespaceUID` survives as a field. The old composite key was not even sufficient for the first job, since two Deployments of one model in one namespace collapsed into a single entry.

**What the name refers to, and it is now `InferenceEngine`.** The old name described a thing the KubeModel does not otherwise model, which I think is what prompted your question. The subject of these signals is the engine instance: KV-cache utilization is a property of its block allocator, queue depth and running requests are its admission scheduler, preemptions are its eviction events. A pod has none of those and neither does a Deployment. So the type is named for what is measured rather than where it runs. `Engine` still records which protocol mapping produced the values.

**On UIDs, which is where your instinct went further than my fix.** Both sources carry them now, and for Prometheus `pod_uid` is one relabel rule on `__meta_kubernetes_pod_uid`, documented, with a companion chart PR at opencost/opencost-helm-chart#379 adding an opt-in PodMonitor (it turned out the chart had no model-server scrape job at all, so there was no existing rule list to extend).

But pod UID alone was not enough. vLLM labels every metric with `["model_name", "engine"]`, where `engine` is the stringified core index, and under data parallelism each rank is its own engine process inside one pod. Keyed on pod UID, a two-rank pod produced one entry: the second rank's values overwrote the first's while keeping the first's identity, nondeterministically, since collector results iterate a Go map. Identity is now `(pod UID, engine index)` end to end (`1c2e4b4a`), in one commit because splitting it either way reproduces that overwrite.

Two more caveats I would rather state than have you find. The `namespace_uid` join was an inner join, so a deployment without `namespace_info` lost every saturation series rather than one label; it is now non-fatal and cluster-scoped, since two clusters with a namespace named `prod` were a duplicate match that failed the whole query. And the token, timing and cache-config results are still keyed by name, inherited from #3845, so the inference surface is half UID-keyed and half name-keyed. The additive fix is a parallel per-engine result type; I did not want to re-key a shipped contract inside this PR.

Still draft on purpose: I would like your read on the counter-semantics question in the `metricsquerier.go` thread before I mark it ready.

PR #3919, opened 2026-07-09
fix(azure): close response body and validate HTTP status codes

### LINKED ISSUE #3918: bug(azure): response body not closed and HTTP status check never triggers
## Describe the bug

I found two issues in the Azure retail price lookup implementation (**pkg/cloud/azure/provider.go**):

1. The HTTP response body returned by **client.Get()** in **getRetailPrice()** is never closed, which can **leak** HTTP connections and unnecessarily consume resources over time.
2. The HTTP status code validation uses the following condition:

```go
if resp.StatusCode < 200 && resp.StatusCode > 299
```

Since a response status code cannot be both **less than 200** and **greater than 299** at the **same time**, this condition never evaluates to **true**. As a result, non-2xx responses are not handled correctly, and the code attempts to parse the error response instead of returning a clear HTTP error.

## Expected behavior

- Always close the HTTP response body after the request completes.
- Properly detect and handle non-2xx HTTP responses before attempting to parse the response body.

## Additional context

I have investigated the issue will raise a pr soon

Kush172005:
## Summary

Fixes #3918

This PR fixes two issues in the Azure retail price lookup implementation.

### Changes

* Close the HTTP response body after the request completes to avoid leaking HTTP connections.
* Fix the HTTP status code validation logic so non-2xx responses are handled correctly before attempting to parse the response.

### Why

The response body was never closed, which could lead to unnecessary resource usage and leaks over time.

The existing status code check could never evaluate to true because it used `&&` instead of `||`. As a result, HTTP error responses were not handled correctly and could lead to parsing errors.

### Testing

* Verified the changes build successfully.
* Confirmed the response body is closed after each request.
* Confirmed non-2xx HTTP responses are now detected and returned as errors.

### DIFF
--- a/pkg/cloud/azure/provider.go
+++ b/pkg/cloud/azure/provider.go
@@ -317,8 +317,9 @@ func getRetailPrice(region string, skuName string, currencyCode string, spot boo
 	if err != nil {
 		return "", fmt.Errorf("failed to fetch retail price with URL \"%s\": %w", pricingURL, err)
 	}
+	defer resp.Body.Close()
 
-	if resp.StatusCode < 200 && resp.StatusCode > 299 {
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
 		return "", fmt.Errorf("retail price responded with error status code %d", resp.StatusCode)
 	}
 


ameijer (approved):
LGTM

PR #3920, opened 2026-07-09
fix: Azure managed disk tier pricing

### LINKED ISSUE #3870: bug(azure): Premium SSD PVs underpriced 16x — Disk Mount meter overwrites Disk meter in convertMeterToPricings

sawyerward:
## Description
Refactors Azure managed disk pricing to be tier-aware and fixes incorrect pricing behavior in rate-card ingestion. Current pricing assumes a linear per-GB across disk sizes, but that is not accurate for Microsoft pricing.

## Related Issues
Fixes #3870

## User Impact
Azure AKS users with managed disks (especially Premium SSD classes like `Premium_LRS` ) will get more accurate PV/disk pricing based on actual disk-tier meters, with correct fallback behavior when specific tier meters are missing.

Windows VM pricing keys are preserved during pricing-map cleanup.

This is not a breaking change; it is a correctness improvement for Azure cost attribution/pricing behavior.

## Testing
- Added regression tests for:
  - disk mount overwrite prevention,
  - managed-disk key filtering preserving Windows keys,
  - tier-aware PV pricing and fallback logic,
  - tier-aware disk cost lookup.
- Added focused unit tests in `disktiers_test.go` for meter parsing, SKU resolution, tier selection, conversion helpers, and managed-disk key detection.
- Ran:

```bash
go test ./pkg/cloud/azure

### DIFF
--- a/pkg/cloud/azure/disktiers.go
+++ b/pkg/cloud/azure/disktiers.go
@@ -1,0 +1,256 @@
+package azure
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/opencost/opencost/core/pkg/clustercache"
+	"github.com/opencost/opencost/core/pkg/util/timeutil"
+)
+
+// Azure managed disk size tiers (GiB), ordered ascending.
+// Source: https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types
+//
+// Follow-ups intentionally not modeled here:
+//   - Shared disk "Disk Mount" fees (meters are skipped with a warning)
+//   - Premium SSD performance-tier upgrades without resize
+//   - Premium SSD v2 / Ultra Disk (per-GiB + IOPS + throughput)
+//   - Azure Files Standard usage-based billing (still $0 class workaround)
+var (
+	premiumSSDTiers = []diskTier{
+		{Name: "P1", SizeGiB: 4},
+		{Name: "P2", SizeGiB: 8},
+		{Name: "P3", SizeGiB: 16},
+		{Name: "P4", SizeGiB: 32},
+		{Name: "P6", SizeGiB: 64},
+		{Name: "P10", SizeGiB: 128},
+		{Name: "P15", SizeGiB: 256},
+		{Name: "P20", SizeGiB: 512},
+		{Name: "P30", SizeGiB: 1024},
+		{Name: "P40", SizeGiB: 2048},
+		{Name: "P50", SizeGiB: 4096},
+		{Name: "P60", SizeGiB: 8192},
+		{Name: "P70", SizeGiB: 16384},
+		{Name: "P80", SizeGiB: 32767},
+	}
+
+	standardSSDTiers = []diskTier{
+		{Name: "E1", SizeGiB: 4},
+		{Name: "E2", SizeGiB: 8},
+		{Name: "E3", SizeGiB: 16},
+		{Name: "E4", SizeGiB: 32},
+		{Name: "E6", SizeGiB: 64},
+		{Name: "E10", SizeGiB: 128},
+		{Name: "E15", SizeGiB: 256},
+		{Name: "E20", SizeGiB: 512},
+		{Name: "E30", SizeGiB: 1024},
+		{Name: "E40", SizeGiB: 2048},
+		{Name: "E50", SizeGiB: 4096},
+		{Name: "E60", SizeGiB: 8192},
+		{Name: "E70", SizeGiB: 16384},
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/cloud/azure/disktiers.go [hunk 1]
```diff
@@ -0,0 +1,284 @@
...
+		{Name: "S6", SizeGiB: 64},
+		{Name: "S10", SizeGiB: 128},
+		{Name: "S15", SizeGiB: 256},
+		{Name: "S20", SizeGiB: 512},
+		{Name: "S30", SizeGiB: 1024},
+		{Name: "S40", SizeGiB: 2048},
+		{Name: "S50", SizeGiB: 4096},
+		{Name: "S60", SizeGiB: 8192},
+		{Name: "S70", SizeGiB: 16384},
+		{Name: "S80", SizeGiB: 32767},
```
ameijer, line 61: @sawyerward are you concerned about these going out of date/becoming stale?

ameijer (commented):
@sawyerward have you tested this on your clusters? we do not have an OC azure test cluster to play around with

pkg/cloud/azure/disktiers.go
sawyerward: It is possible, but right now the biggest one is 32 TiB so they'd have to release a drive bigger than that. It's unfortunate that they don't have a linear per-GB price so we have to maintain some sort of mapping.

sawyerward:
> @sawyerward have you tested this on your clusters? we do not have an OC azure test cluster to play around with

We have it running in some of our sandbox clusters. The numbers are looking good but I'll likely let it run over the weekend to get better numbers.

sawyerward:
I haven't had time to wrap up all my testing but here's the results for StandardSSD.

Here we are testing the `managed` and `managed-csi` storage classes which both map to `StandardSSD_ZRS`. We tested in `eastus2` region. This shows how when a PV isn't using the full disk size (e.g. 65 GiB/128 GiB), the total cost when multiplied by the PV size equals the total disk tier size (128 GiB = $14.40) and is the same total monthly cost as if you're using the full 128 GiB PV. 

Azure bills based on the provisioned tier size not on the PV size itself. The effective rate is computed per-requested-GiB so that allocation multiplied by tier rate equals the full tier price.

https://azure.microsoft.com/en-us/pricing/details/managed-disks/

Screenshot 2026-07-10 at 11 36 27 AM
Screenshot 2026-07-10 at 11 37 28 AM

ameijer (approved):
going to give this a shot @sawyerward , can you pull out of draft so we can merge?

sawyerward:
> going to give this a shot @sawyerward , can you pull out of draft so we can merge?

I could do that. I'm currently trying to test the ZRS edge case. It's not showing up in the rate card API so I'd have to set up an SP with permissions for the price sheet API, etc. Wouldn't get to that until next week. If you're okay without those specific test results, let's get it merged. Right now the ZRS lookups log like:

```2026-07-10T18:48:29.977567696Z WRN Persistent Volume tier pricing not found for eastus2,premium_ssd,ZRS,P6 size 64 (ZRS); falling back to class rate```

PR #3925, opened 2026-07-10
Fix race condition in GenericStore.Replace

### LINKED ISSUE #3924: bug(clustercache): GenericStore.Replace releases lock between clearing and repopulating the store
### Describe the bug

The `GenericStore.Replace()` method in `pkg/clustercache/store.go` has a race condition caused by releasing the write lock between clearing the items map and re-adding items.

On line 98-100, the mutex is acquired, the map is reset to an empty map, and then the mutex is **immediately released**. The subsequent loop (lines 102-107) calls `s.Add()` for each item, which re-acquires the lock individually per item.

During the window between the unlock on line 100 and the completion of all `Add()` calls, any concurrent `GetAll()` call (which acquires `RLock`) will observe an **empty or partially populated store**, leading to incorrect or incomplete cost calculations downstream.

Additionally, `s.onInit` is read and written on lines 110-112 **outside any lock**, while `Watch()` on line 51 also writes to `s.onInit` concurrently from a different goroutine. This is a data race on `s.onInit`.

### To Reproduce

1. Open `pkg/clustercache/store.go`.
2. Navigate to the `Replace()` method (line 97).
3. Observe that `s.mutex.Lock()` is called on line 98, and `s.mutex.Unlock()` is called on line 100 — **before** the loop that re-adds items via `s.Add()` on lines 102-107.
4. Observe that `s.Add()` delegates to `s.Update()` (line 60), which acquires its own separate lock (line 65).
5. Observe that `GetAll()` (line 86) uses `RLock`, meaning it can execute between the `Unlock()` on line 100 and any individual `Add()` call, observing a partially populated map.
6. Observe that `s.onInit` (lines 110-112) is accessed without holding any lock, while `Watch()` (line 51) writes to the same field from a different goroutine.

### Expected behavior

The `Replace()` method should hold the lock for the entire duration of the clear-and-repopulate operation, so that concurrent readers via `GetAll()` never observe an intermediate (empty or partial) state of the store. The `s.onInit` field should also be accessed under the lock to avoid a data race.

### Which version of OpenCost are you using?

Current `develop` branch.

### Additional context

- `GenericStore` is used as the backing store for the V2 cluster cache (`KubernetesClusterCacheV2` in `pkg/clustercache/clustercache2.go`), which manages 16 separate stores for Pods, Nodes, Deployments, Services, etc.
- `Replace()` is called by the Kubernetes `cache.Reflector` whenever it re-lists resources from the API server.
- `GetAll()` is called by the cost model during every allocation computation cycle to retrieve cached Kubernetes objects.
- This means the race window between `Replace()` and `GetAll()` is hit in normal production operation — not just in contrived scenarios.
- The bug could be confirmed with `go test -race` if a test exercised concurrent `Replace()` and `GetAll()` calls.

sapnilbiswas:
## Description
This PR fixes a race condition in `GenericStore.Replace`. 

Previously, the write lock was released immediately after clearing the items map, and items were re-added one-by-one via `Add()`, which acquired its own lock per item. This created a window where concurrent `GetAll()` calls could observe an empty or partially populated store, leading to incorrect cost calculations.

Additionally, `s.onInit` was read and written outside any lock, creating a data race with `Watch()` which sets `onInit` from a different goroutine.

This PR fixes both issues by:
1. Holding the lock for the entire clear-and-repopulate operation.
2. Inlining the transform logic to avoid deadlock (`sync.RWMutex` is not reentrant).
3. Capturing `onInit` under the lock before calling it after unlock to prevent potential deadlocks.

## Related Issues
Fixes #3924

## User Impact
No breaking changes. This ensures that cost calculations using the V2 cluster cache do not intermittently use empty or partial Kubernetes resource data during cache repopulation.

## Testing
- `go test ./pkg/clustercache/...` passes
- `go vet ./pkg/clustercache/...` passes
- `go build ./pkg/clustercache/...` succeeds

### DIFF
--- a/pkg/clustercache/store.go
+++ b/pkg/clustercache/store.go
@@ -20,6 +20,7 @@ type GenericStore[Input UIDGetter, Output any] struct {
 	// storing this cyclic reflector allows us to defer watching
 	reflector *cache.Reflector
 	onInit    func()
+	isWatching bool
 }
 
 // NewGenericStore creates a new instance of GenericStore.
@@ -49,6 +50,11 @@ func CreateStore[Input UIDGetter, Output any](
 
 func (s *GenericStore[Input, Output]) Watch(stopCh <-chan struct{}, onInit func()) {
 	s.mutex.Lock()
+	if s.isWatching {
+		s.mutex.Unlock()
+		return
+	}
+	s.isWatching = true
 	s.onInit = onInit
 	s.mutex.Unlock()
 
@@ -97,17 +103,19 @@ func (s *GenericStore[Input, Output]) GetAll() []Output {
 
 // Replace replaces the current list of items in the store.
 func (s *GenericStore[Input, Output]) Replace(list []any, _ string) error {
+	newItems := make(map[types.UID]Output, len(list))
+	for _, o := range list {
+		item := o.(Input)
+		newItems[item.GetUID()] = s.transformFunc(item)
+	}
+
 	var onInit func()
 	
 	func() {
 		s.mutex.Lock()
 		defer s.mutex.Unlock()
 		
-		s.items = make(map[types.UID]Output, len(list))
-		for _, o := range list {
-			item := o.(Input)
-			s.items[item.GetUID()] = s.transformFunc(item)
-		}
+		s.items = newItems
 		
 		// Capture onInit under the lock to avoid a data race with Watch(),
 		// but call it after releasing the lock to prevent potential deadlocks.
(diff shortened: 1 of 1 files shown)


mbolt35:
This is definitely safer than what was there previously. However, this still isn't quite right. The new map of items can be created outside of the lock, and then set inside the lock.

mbolt35:
Also, `onInit` was really ok before, but ok, if we want to be compulsive here, we should really be restricting any further calls to Watch() while already watching. In this case, you'd not enter a race with onInit. Most of the time, you'd have exactly one caller of Watch, and the onInit isn't really racing.

sapnilbiswas:
I've updated Replace() to build the new items map entirely outside the lock. I also added an isWatching check to Watch() to restrict multiple concurrent watch calls, completely eliminating any potential race with onInit.

PR #3934, opened 2026-07-15
feat(cloud): dynamic AWS & GCP load balancer pricing lookup

### LINKED ISSUE #3935: Dynamic AWS & GCP Load Balancer Pricing Lookup via LBKey
### Description
Currently, AWS and GCP cloud provider implementations in OpenCost return hardcoded load balancer costs (e.g., a flat $0.025/hour for AWS or a basic rule calculation for GCP). Now that the provider interface accepts a `models.LBKey` parameter, we should upgrade the AWS and GCP providers to dynamically parse the key features/ID and query the downloaded pricing cache using keys that target specific load balancer types.

### Scope
- `pkg/cloud/aws/provider.go`
- `pkg/cloud/gcp/provider.go`

### Requirements
1. AWS:
   - Parse `lbKey.Features()` and `lbKey.ID()` in `LoadBalancerPricing` to check for NLB, ALB, or CLB tags.
   - Cache specific hourly rates matching these types during bulk pricing load.
   - Query the cache using keys targeting these specific load balancer types.
2. GCP:
   - Decode `lbKey` features/ID to differentiate between External HTTP(S), Internal HTTP(S), Network/TCP, and SSL Proxy load balancers.
   - Query the downloaded pricing cache using keys targeting forwarding rule SKU descriptions matching these types.
3. Fallback:
   - Fall back to standard defaults if the lookup returns no exact match.

### Acceptance Criteria
- AWS and GCP `LoadBalancerPricing` return distinct, correct hourly costs for different load balancer types.
- Unit tests mock various `LBKey` features and verify that correct rates are resolved.

Rucha0901:
Fixes #3935

### Description
This PR upgrades the AWS and GCP provider implementations of `LoadBalancerPricing` to dynamically parse the `models.LBKey` features and ID and select specific hourly costs based on the load balancer configuration.

### Changes
- **AWS**:
  - Modified `populatePricing` to check and register specific load balancer usage types (Network, Application, Classic) in the pricing cache.
  - Upgraded `LoadBalancerPricing` to parse the features/ID of `lbKey` and query the cache for specific SKUs matching these types, falling back to the standard generic regional pricing if no match is found.
- **GCP**:
  - Modified `parsePage` to check if a product is a load balancer forwarding rule, classify it (External HTTP(S), Internal HTTP(S), Network/TCP, SSL Proxy), and store it in the pricing cache under region-specific keys.
  - Implemented `getGCPPricingRate` helper to calculate hourly prices from GCPPricing tiered rates.
  - Upgraded `LoadBalancerPricing` to inspect `lbKey` features/ID and query the specific cached forwarding rule rate, falling back to the standard default calculation if there is no exact match.

### Verification
- Added table-driven unit tests for both AWS and GCP packages verifying that various `LBKey` feature sets resolve to their distinct mock hourly rates.
- Verified all unit tests pass locally:
  - `go test ./pkg/cloud/aws/...`
  - `go test ./pkg/cloud/gcp/...`

### DIFF
--- a/pkg/cloud/aws/provider.go
+++ b/pkg/cloud/aws/provider.go
@@ -1198,25 +1198,27 @@ func (aws *AWS) populatePricing(resp *http.Response, inputkeys map[string]bool) 
 						Sku:          product.Sku,
 						LoadBalancer: &models.LoadBalancer{},
 					}
-					key := product.Attributes.RegionCode + ",LoadBalancerUsage"
-					aws.Pricing[key] = productTerms
-					skuToPricingKeyMap[product.Sku] = key
-					aws.ValidPricingKeys[key] = true
-
-					var specKey string
+					genericKey := product.Attributes.RegionCode + ",LoadBalancerUsage"
+					pricingKey := genericKey
 					switch product.Attributes.Operation {
 					case "LoadBalancing:Network":
-						specKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Network"
+						pricingKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Network"
 					case "LoadBalancing:Application":
-						specKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Application"
+						pricingKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Application"
 					case "LoadBalancing":
-						specKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Classic"
+						pricingKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Classic"
 					}
+					aws.Pricing[pricingKey] = productTerms
+					skuToPricingKeyMap[product.Sku] = pricingKey
+					aws.ValidPricingKeys[pricingKey] = true
 
-					if specKey != "" {
-						aws.Pricing[specKey] = productTerms
-						skuToPricingKeyMap[product.Sku] = specKey
-						aws.ValidPricingKeys[specKey] = true
+					// Preserve a stable generic fallback key (prefer Network when available).
+					if pricingKey == genericKey || product.Attributes.Operation == "LoadBalancing:Network" {
+						aws.Pricing[genericKey] = productTerms
+						aws.ValidPricingKeys[genericKey] = true
+					} else if _, exists := aws.Pricing[genericKey]; !exists {
+						aws.Pricing[genericKey] = productTerms
+						aws.ValidPricingKeys[genericKey] = true
 					}
 				}
 			}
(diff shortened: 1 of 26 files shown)


pkg/cloud/aws/provider.go
```diff
@@ -1192,6 +1192,8 @@ func (aws *AWS) populatePricing(resp *http.Response, inputkeys map[string]bool)
 					skuToPricingKeyMap[product.Sku] = key
 					aws.ValidPricingKeys[key] = true
 					aws.ValidPricingKeys[spotKey] = true
+				} else if strings.Contains(product.Attributes.UsageType, "LoadBalancerUsage") {
+					// Parse different types of load balancers when available.
 				} else if strings.Contains(product.Attributes.UsageType, "LoadBalancerUsage") && product.Attributes.Operation == "LoadBalancing:Network" {
```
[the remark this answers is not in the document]
thomasvn: @Rucha0901 Looks like you should fix this?

pkg/cloud/aws/provider.go
```diff
@@ -1204,6 +1206,22 @@ func (aws *AWS) populatePricing(resp *http.Response, inputkeys map[string]bool)
...
+						specKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Application"
+					case "LoadBalancing":
+						specKey = product.Attributes.RegionCode + ",LoadBalancerUsage:Classic"
+					}
+
+					if specKey != "" {
+						aws.Pricing[specKey] = productTerms
+						skusToKeys[product.Sku] = specKey
+						aws.ValidPricingKeys[specKey] = true
+					}
```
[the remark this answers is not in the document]
thomasvn: @Rucha0901 This too?

pkg/cloud/aws/provider.go
```diff
@@ -1463,15 +1477,40 @@ func (aws *AWS) NetworkPricing() (*models.Network, error) {
 	}, nil
 }
 
-func (aws *AWS) LoadBalancerPricing() (*models.LoadBalancer, error) {
-	// TODO: determine key based on function arguments
-	// this is something that should be changed in the Provider interface
+func (aws *AWS) LoadBalancerPricing(lbKey models.LBKey) (*models.LoadBalancer, error) {
+	aws.DownloadPricingDataLock.RLock()
+	defer aws.DownloadPricingDataLock.RUnlock()
```
thomasvn, line 1485: This is a very expensive call to make, as the AWS pricing file can be very large. Not sure if it's worth it to make this call just for LoadBalancer pricing data? This call is already made for NodePricing. Can that be reused in any way?

thomasvn: Sorry! Realizing this is just acquiring the lock. Could you add a comment explaining why the lock needed to be acquired.

pkg/cloud/models/network.go
```diff
@@ -1,9 +1,38 @@
...
+
+// CustomNetworkKey is a default implementation of NetworkKey
+type CustomNetworkKey struct {
+	NetworkID       string
+	NetworkFeatures string
+}
+
+func (k *CustomNetworkKey) ID() string         { return k.NetworkID }
+func (k *CustomNetworkKey) Features() string   { return k.NetworkFeatures }
+func (k *CustomNetworkKey) IsNetworkKey() bool { return true }
```
thomasvn, line 35: Scope creep for this PR? I think this PR wanted to address dynamic LoadBalancer pricing, not NetworkCosts. I think this change resulted in many other interfaces getting changed but not used. We should reduce scope to just LoadBalancers?

PR #3936, opened 2026-07-16
Fix handler early returns on invalid window parameter

### LINKED ISSUE #3923: docs/swagger.json: /allocation, /assets, /cloudCost document only 200 (no 400 for required window), and Window is inline-duplicated
Went through `docs/swagger.json` (OpenAPI 3.0.1) on `develop` and hit two internal inconsistencies that make it harder to generate clients or contract tests from the spec. Both are in the checked-in spec only — no behavior change implied.

**1. Error responses are documented on the inferenceCost endpoints but not the older ones.**

`GET /inferenceCost/total` and `GET /inferenceCost/timeseries` document `400` and `501` responses. `GET /allocation`, `GET /assets`, and `GET /cloudCost` document only `200`, even though `window` is `required: true` on all five operations and an invalid/missing `window` is a documented `400` in the newer ones. A client or contract test generated from the spec will assume `/allocation`, `/assets`, and `/cloudCost` never return a 4xx, so the error path (the `code`/`status` body on a bad `window`) is invisible. Adding the `400` (and `501` where applicable) to those three would make the five endpoints describe the same error contract.

**2. `Window` is a component schema, but two of its uses inline-duplicate it.**

`components/schemas/Window` (`{ start, end: string/date-time }`) is `$ref`'d correctly by `InferenceCostResponse.window` and `InferenceCostSet.window`. But `Allocation.window` and `Asset.window` define the identical object inline instead of `$ref`-ing `Window`. Code generators then emit two extra one-off window types, and any future change to the canonical `Window` (a `step`/duration field, or a timestamp-format change) silently skips the two inline copies — the spec would describe two different `window` shapes for one concept.

Happy to open a PR pointing the two inline `window` properties at `#/components/schemas/Window` and adding the missing response codes, if that is a welcome change.

peatey:
## Description
This PR fixes a bug where HTTP handlers (`ComputeAllocationHandler` and `ComputeAllocationHandlerSummary`) would continue executing after writing a 400 Bad Request response for invalid or missing `window` parameters. The handlers now properly return immediately after writing the error response, preventing potential panics or undefined behavior from executing with zero-value state.

Additionally, the Swagger/OpenAPI documentation has been updated to reflect proper error responses (400, 500, 501) across all allocation and cost query endpoints.

## Changes
- **pkg/costmodel/aggregation.go**: Added `return` statements after `http.Error()` calls in `ComputeAllocationHandler` and `ComputeAllocationHandlerSummary` to prevent continued execution after error responses
- **pkg/costmodel/aggregation_test.go**: Added comprehensive test coverage for invalid and missing window parameters in both handlers
- **docs/swagger.json**: Updated API documentation to include 400, 500, and 501 error responses for allocation, asset, cloud cost, and inference cost endpoints; refactored window schema to use `$ref` for consistency

## Related Issues
Fixes #3923

Also fixes a potential panic when handlers receive invalid window parameters and continue execution with nil/zero-value state.

## User Impact
- **Bug Fix**: Handlers now properly reject invalid requests without risk of panic
- **API Documentation**: Swagger documentation now accurately reflects all possible HTTP status codes returned by the endpoints
- **No Breaking Changes**: This is a bug fix that makes error handling more robust

## Testing
Added 4 new unit tests covering:
- `TestComputeAllocationHandler_InvalidWindow_Returns400`: Verifies 400 response with error message for invalid window
- `TestComputeAllocationHandler_MissingWindow_Returns400`: Verifies 400 response for missing window parameter
- `TestComputeAllocationHandlerSummary_InvalidWindow_Returns400`: Verifies 400 response with error message for invalid window in summary endpoint
- `TestComputeAllocationHandlerSummary_MissingWindow_Returns400`: Verifies 400 response for missing window parameter in summary endpoint

All tests verify that handlers return immediately with appropriate HTTP status codes and error messages.

### DIFF
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -78,6 +78,12 @@
                 }
               }
             }
+          },
+          "400": {
+            "description": "Bad request — missing or invalid parameters (e.g. `window`, `aggregate`, `step`, `accumulate`)"
+          },
+          "500": {
+            "description": "Internal server error — the allocation query failed"
           }
         }
       }
@@ -106,6 +112,12 @@
                 }
               }
             }
+          },
+          "400": {
+            "description": "Bad request — missing or invalid `window` parameter"
+          },
+          "500": {
+            "description": "Internal server error — computing or filtering assets failed"
           }
         }
       }
@@ -134,6 +146,15 @@
                 }
               }
             }
+          },
+          "400": {
+            "description": "Bad request — missing or invalid parameters"
+          },
+          "500": {
+            "description": "Internal server error — the cloud cost query failed"
+          },
+          "501": {
+            "description": "Cloud cost query service is not available. Note: when `CLOUD_COST_ENABLED` is not set to `true` this route is not registered at all and requests return 404."
           }
         }
       }
@@ -214,6 +235,9 @@
           },
           "400": {
             "description": "Bad request — missing or invalid parameters"
+          },
+          "500": {
+            "description": "Internal server error — the inference cost query failed"
           },
           "501": {
             "description": "Inference cost tracking is not enabled (`INFERENCE_COST_ENABLED` is not set to `true`)"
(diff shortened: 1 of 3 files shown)


ameijer (approved):
lgtm

PR #3941, opened 2026-07-19
fix(gcp): close billing API response body and validate HTTP status

vtushar06:
Was going through the cloud providers and noticed the GCP billing pricing path never closes the HTTP response body or checks the status code, while azure/oracle/aws/otc all already do both.

The pagination walks every page of the GCP Cloud Billing catalog, and each request did `httpClient.Get(...)` then handed `resp.Body` straight to `parsePage`, which only reads it. Nothing ever closed the body, so a full pricing refresh leaks a connection per page (dozens of them) and also blocks keep-alive reuse. The status code wasn't checked either, so a 4xx/5xx (expired key, quota, outage) went into the JSON parser as if it were pricing data.

Pulled the fetch+parse into a small `fetchAndParsePage` helper that defers `Body.Close()` and returns an error on a non-200 before parsing. Since it's its own method now, each page's body closes before the next request instead of stacking up until the whole catalog finishes.

Added two httptest unit tests (non-200 -> error mentioning the status code, 200 -> parses cleanly). Happy path is unchanged.

### DIFF
--- a/core/pkg/external/configmapsource_test.go
+++ b/core/pkg/external/configmapsource_test.go
@@ -333,11 +333,11 @@ labels:
 
 func TestFilterValidLabels_AllValid_ReturnedUnchanged(t *testing.T) {
 	in := map[string]string{
-		"env":                     "prod",
-		"region":                  "us-east-1",
-		"app.kubernetes.io/name":  "opencost",
-		"my-key":                  "my-value",
-		"a":                       "b",
+		"env":                    "prod",
+		"region":                 "us-east-1",
+		"app.kubernetes.io/name": "opencost",
+		"my-key":                 "my-value",
+		"a":                      "b",
 	}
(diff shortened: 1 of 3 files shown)


vtushar06:
hey @ameijer,  when you get a chance gcp billing pagination was leaking response bodies + not checking status codes on error pages, other providers already guard for both so brought this one in line. and format-check shows red here but that's actually [#3943 ](https://github.com/opencost/opencost/pull/3941)(pre-existing gofmt break on develop, unrelated to this), will rebase once that merges

vtushar06:
hey @ameijer, all green now (format-check fixed too), ready for review whenever you get a chance

vtushar06:
hey @mbolt35 can you please have a look, alex told me to ping u wherver need.

PR #3944, opened 2026-07-20
feat(gcp): add C3D machine family pricing support

aditya-rathi1-glean:
## What
Add GCP **C3D** machine-family pricing support to the GCP provider.

## Why
`parsePage` maps Cloud Billing Catalog SKU descriptions to instance types via hardcoded substring matches. C3D is absent, so C3D nodes get no catalog price and fall back to the default estimate. 

- **`parsePage`**: classify `cpu`/`ram` SKUs whose description contains `C3D` (and not `SOLE`, to skip sole-tenancy premiums) as `c3dstandard`.
- **`parseGCPInstanceTypeLabel`**: fold `c3d-highmem-*` / `c3d-highcpu-*` onto `c3dstandard` (C3D variants share per-vCPU/per-GB rates).
- **`sustainedUseDiscount`**: C3D is a newer flexible-CUD family with no sustained-use discount → SUD class `0.0`.

## Testing
- Added `parsePage` unit tests (on-demand + spot + sole-tenancy exclusion) with matching `test/skus.json` fixtures.
- `gofmt` clean.


peatey:

PR #3945, opened 2026-07-20
feat(gcp): add C4A machine family pricing support

aditya-rathi1-glean:
## What
Add GCP **C4A** (Arm/Axion) machine-family pricing support to the GCP provider.

## Why
Same class of gap as C3D (#3944): `parsePage` has no `C4A` match, so C4A nodes get no catalog price and fall back to the default estimate (spot billed at on-demand list). C4A is an Arm family, so the Core/RAM classification keys on the description containing `RAM` — note `ARM` (A-R-M) does not contain `RAM` (R-A-M), so the existing RAM-vs-Core split stays correct.

- **`parsePage`**: classify `cpu`/`ram` SKUs whose description contains `C4A` (and not `SOLE`) as `c4astandard`.
- **`parseGCPInstanceTypeLabel`**: fold `c4a-highmem-*` / `c4a-highcpu-*` onto `c4astandard`.
- **`sustainedUseDiscount`**: no sustained-use discount for C4A → SUD class `0.0`.

## Testing
- Added `parsePage` unit tests (on-demand + spot + sole-tenancy exclusion; asserts `ARM`≠`RAM`) with `test/skus.json` fixtures.
- `gofmt` clean.

### DIFF
--- a/pkg/cloud/gcp/provider.go
+++ b/pkg/cloud/gcp/provider.go
@@ -748,6 +748,13 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]models.Key, pvKeys m
 				if (instanceType == "ram" || instanceType == "cpu") && strings.Contains(strings.ToUpper(product.Description), "E2 INSTANCE") {
 					instanceType = "e2"
 				}
+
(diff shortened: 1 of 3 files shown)


peatey:

PR #3947, opened 2026-07-20
feat(gcp): add G2 machine family pricing support

aditya-rathi1-glean:
## What
Add GCP **G2** (NVIDIA L4 GPU) machine-family pricing support to the GCP provider.

## Why
Same `parsePage` gap as the other families. G2 is a GPU family: the L4 accelerator is already priced via the existing GPU path (`nvidia-l4` exists in `gpu.go`), but the **G2 vCPU/RAM** portion was unpriced because `G2` had no `parsePage` match — so the compute half of a G2 node fell back to the default estimate.

- **`parsePage`**: classify `cpu`/`ram` SKUs whose description contains `G2 INSTANCE` (and not `SOLE`) as `g2standard`. The `!SOLE` guard matters here — G2 is sole-tenancy-eligible, so GCP emits `"Sole Tenancy Premium for G2 Instance Core"`, which would otherwise overwrite the real vCPU rate.
- **`parseGCPInstanceTypeLabel`**: fold `g2-standard-*` onto `g2standard`.
- **`sustainedUseDiscount`**: no sustained-use discount for G2 → SUD class `0.0`.

The L4 GPU cost continues to flow through the existing GPU key path; the vCPU/RAM costs merge onto the same node key — no double-counting, and compute is no longer $0.

## Testing
- Added `parsePage` unit tests: L4 GPU attach + G2 Core/RAM merge onto one node key (asserts vCPU + RAM + L4, no double-count), plus spot handling, with `test/skus.json` fixtures.
- `gofmt` clean.

### DIFF
--- a/pkg/cloud/gcp/provider.go
+++ b/pkg/cloud/gcp/provider.go
@@ -748,6 +748,15 @@ func (gcp *GCP) parsePage(r io.Reader, inputKeys map[string]models.Key, pvKeys m
 				if (instanceType == "ram" || instanceType == "cpu") && strings.Contains(strings.ToUpper(product.Description), "E2 INSTANCE") {
 					instanceType = "e2"
 				}
+
+				// Sole-tenancy SKUs (including "Sole Tenancy Premium" surcharge
+				// SKUs) are excluded so they do not overwrite the standard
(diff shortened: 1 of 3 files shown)


peatey:

PR #3963, opened 2026-07-27
perf(prometheus): resolve pod timespans with first/last timestamp queries

mdog-px:
## Description

`ComputeAllocation` discovers pods with:

```
avg(kube_pod_container_status_running{...} != 0) by (pod, namespace, uid, <cluster>)[<window>:<resolution>]
```

This returns the full range vector for every pod: one sample per resolution step, e.g. 289 samples per pod for a 24h window at 5m resolution. However, the only consumer of these results, `calculateStartAndEnd` in `pkg/costmodel/allocation_helpers.go`, reads just the first and last timestamp of each series and discards everything in between.

Materializing the unused samples is expensive. Each response is fully buffered as raw JSON, unmarshalled into a generic `interface{}` tree (where every sample becomes a boxed `[]interface{}{float64, string}`), and then converted into per-point `*util.Vector` allocations. For a 10,000 pod cluster over a 24h window that is roughly 2.9M samples: tens of MB of raw JSON and several hundred MB for the intermediate decoded representations, nearly all of it thrown away immediately. Because this runs on every allocation computation (and per step for ranged queries), it drives the peak heap of the process, and therefore its resident memory, on large clusters.

This PR changes `QueryPods` and `QueryPodsUID` to instead issue two aggregated instant queries:

```
min_over_time(timestamp(avg(kube_pod_container_status_running{...} != 0) by (pod, namespace, uid, <cluster>))[<window>:<resolution>])
max_over_time(timestamp(avg(kube_pod_container_status_running{...} != 0) by (pod, namespace, uid, <cluster>))[<window>:<resolution>])
```

Prometheus now returns a single sample per pod per query (the first and last timestamp at which the pod was seen running), and the two result sets are merged client side into a synthetic two-point series per pod. `DecodePodsResult`, `applyPodResults`, and `calculateStartAndEnd` are unchanged and behave exactly as before. The response size and decoded result set drop from O(pods * window/resolution) to O(pods), about 150x smaller for a 24h window at 5m resolution.

Notes on equivalence and compatibility:

* Both queries evaluate at the same time parameter (the window end), so the subquery steps align with the steps of the previous range vector. The min and max of `timestamp(...)` over those steps are exactly the timestamps the old code read from positions 0 and len-1 of the range.
* Timestamps are rounded to the nearest 10 seconds, matching the rounding `parseDataPoint` applied to range vector timestamps.
* A pod present at only one step yields start == end, which `calculateStartAndEnd` already widens by one resolution, same as the old single-sample case.
* If a series appears in only one of the two results (possible if samples are ingested between the two query executions), the merge falls back to the timestamp of the side that is present.
* The `min_over_time(timestamp(...))` pattern is already used by the oldest/newest sample queries in the same file, and subqueries are already required by many other queries, so there is no new Prometheus version requirement.
* The two instant queries run concurrently. The server evaluates the same inner expression twice, but the transfer and client-side decode cost drops by roughly the window/resolution factor.

Other queries in `metricsquerier.go` use the same bare `[window:resolution]` pattern for uptime/active-minutes purposes (PVs, PVCs, load balancers, nodes, and so on). Those have much lower cardinality than pods, so this PR starts with the largest offender; the same treatment can be applied to the others as follow-up work if this approach is accepted.

## Related Issues

Relates to #2637

## User Impact

No API, output, or configuration changes. Allocation queries on large clusters should show substantially lower memory spikes and less GC pressure. Pod start/end times produced by the allocation pipeline are unchanged.

## Testing

* New unit tests for the merge logic (`podsquery_test.go`): matching series by cluster/namespace/pod/uid, timestamp rounding, one-sided results, end-before-start clamping, empty results, and empty inputs.
* Both new query shapes validated against the upstream Prometheus PromQL parser.
* `go test ./...` and `go vet ./...` pass on all four modules (core, opencost, prometheus-source, collector-source).
* Not yet verified against a live Prometheus; equivalence is argued from the consumer code above, and I would appreciate a check from anyone with a large-cluster staging setup.


fabioaraujopt:
Great work @mdog-px, thanks for you contribution!

PR #3983, opened 2026-07-31
feat(huawei): add Huawei Cloud provider support

alejandromore:
## Summary

Adds Huawei Cloud as a first-class OpenCost provider, following the same
integration pattern used for AWS/Azure/GCP: node/disk pricing via the
provider interface, billing ingestion via CloudCost, and a custom-cost
plugin for OBS. Cloud assets are further broken down by Huawei service
(ECS, EVS, OBS, RDS, DCS, ELB, NAT, VPC, EIP, DEW) instead of being
collapsed into a single "Cloud" bucket on the Infra Assets dashboard.

## What's included

- **Provider** (`pkg/cloud/huawei`): CCE cluster detection, ECS/EVS live
  pricing via the Huawei Cloud Pricing API, BSS-based billing for
  RDS/DCS/DEW, and ELB pricing.
- **Agency-based auth**: resolves credentials and project ID via IMDS
  instance metadata when not explicitly configured, with caching to avoid
  redundant metadata lookups per reconcile.
- **CloudCost ingestion** (`pkg/costmodel/assets.go`,
  `pkg/cloud/huawei/costintegration.go`): billing data ingested through the
  existing CloudCostIntegration interface, surfaced as `Cloud` assets.
- **Cloud asset sub-categorization** (`core/pkg/opencost/asset.go`): new
  `AssetType` values (ECS/EVS/OBS/RDS/DCS/ELB/NAT/VPC/EIP/DEW/OtherCloud)
  so the Infra Assets dashboard shows Huawei services as separate
  categories instead of one generic "Cloud" line. This includes a bingen
  codec version bump (Assets v21→v22) with a versioned field addition, so
  existing binary ETL/federation blobs remain readable (default to
  `CloudAssetType` for pre-v22 data).
- **OBS custom cost plugin** (`modules/huawei-obs-plugin`): standalone
  plugin following the OpenCost custom-cost-source plugin pattern for
  Object Storage Service billing.
- Config plumbing: `pkg/cloud/config`, `pkg/env/costmodel.go`,
  `pkg/cmd/costmodel/costmodel.go` wiring to register/enable the provider.

## Why

Huawei Cloud is not currently a supported OpenCost provider. This lets
clusters running on Huawei CCE get accurate node/disk pricing and cost
allocation without a custom fork, and gives operators the same
service-level cost breakdown (RDS vs DCS vs OBS, etc.) that other clouds
already have via native billing categories.

## Testing

- `go build ./...` (root and `core` modules) — clean.
- `go test ./pkg/cloud/huawei/... ./modules/huawei-obs-plugin/... ./core/pkg/opencost/...`
  — passing.
- Merged latest `develop` in; verified no breakage from the upstream
  pricing-module refactor.
- Manually validated against a live Huawei CCE cluster: CCE detection,
  ECS/EVS pricing, RDS/DCS/OBS billing ingestion, and the Infra Assets
  dashboard showing per-service sub-categories.

## Notes for reviewers

- The `AssetType` enum growth and the Cloud asset `typ` sub-type field are
  the most invasive part of this PR since `Cloud` is a bingen-generated,
  binary-serialized type used by ETL/federation storage. The field was
  added at the end of the struct with `@bingen:field[version=22]` per the
  project's own versioning convention, not inserted at the front, to avoid
  silently corrupting existing binary assets on rolling upgrade.
- Happy to split this into smaller PRs (provider core / OBS plugin / asset
  sub-categorization) if that's preferred for review.


alejandromore:
### On the failing Snyk check

The `security/snyk (ameijer)` check is red on this PR, so I looked into what it is reporting. Two findings this branch genuinely introduced are now fixed; everything left appears to be pre-existing on `develop`.

**Introduced by this PR, now fixed:**

- `go.mongodb.org/mongo-driver` resolved to v1.13.1 through `huaweicloud-sdk-go-v3`, carrying CVE-2026-2303 / GHSA-cp6g-7hqx-qxhp. Bumped to v1.17.7 in 07f9a005.
- `modules/huawei-obs-plugin` is a separate module, so it had resolved the `golang.org/x/*` family independently and landed several releases behind the root module (crypto v0.49.0 vs v0.53.0, net v0.52.0 vs v0.56.0, mod v0.34.0 vs v0.37.0, text v0.36.0 vs v0.40.0, sys v0.42.0 vs v0.46.0). Aligned in bbfbed93; `otel` and `grpc` moved past their high-severity advisories in 258652bd. That module's declared requirements now carry no known advisories.

**What is left.** Comparing the `require` set of the root `go.mod` on `develop` against this branch through the OSV API: 15 advisories on `develop`, the same 15 here. This PR adds none of them.

| module | version | advisories | fix |
|---|---|---|---|
| `github.com/aws/aws-sdk-go` | v1.55.8 | GO-2022-0635, GO-2022-0646 | **none available** |
| `github.com/golang-jwt/jwt/v4` | v4.5.0 | GHSA-mh63-6h87-95cp (high) + 3 | 4.5.2 |
| `go.opentelemetry.io/otel/sdk` | v1.41.0 | GHSA-hfvc-g4fc-pqhx (high) + 1 | 1.43.0 |
| `google.golang.org/grpc` | v1.79.3 | GHSA-hrxh-6v49-42gf (high) + 1 | 1.82.1 |
| `github.com/klauspost/compress` | v1.18.4 | GO-2026-5841 | 1.18.7 |
| `golang.org/x/mod` | v0.37.0 | GO-2026-6179, GO-2026-6180 | 0.40.0 |
| `golang.org/x/crypto` | v0.53.0 | GO-2026-5932 (`openpgp` unmaintained) | none |

The reason this surfaces here and nowhere else: every other open PR reports `No manifest changes detected in 4 projects` and passes without running. This is currently the only open PR that touches a `go.mod`, so it is the only one the check actually tests — and what it then reports is the root project's existing state.

Since two of those (`aws-sdk-go` v1) have no fixed version at all, I don't think this check can be brought to green from a contributor's side.

I can't see the report itself — `app.snyk.io/org/ameijer` is private — so the above is inferred from public data and I'm happy to be corrected by anyone who can open it. If you would like the fixable ones bumped as part of this PR I'll do it, though folding unrelated dependency churn into a provider PR seemed worse than raising it here.

Separately: `Build/Test`, `Format Check` and `Generate SBOM` are sitting in `action_required` and have never run on this branch, since workflow runs from a fork need a maintainer to approve them. Locally, `go build ./...` is clean, `gofmt` reports nothing across the 37 Go files this PR touches, and the tests pass for every package it changes. It would be useful to get a real CI run to confirm that.

alejandromore:
@Sean-Holcomb — following up here since the Slack thread went quiet.

Just pushed a merge of current `develop` (0d87236), which resolves the `go.sum` conflict from #4033 / #4035. Verified locally on linux (`golang:1.26`): `go test` + `go vet` exit 0 across all four modules, `pkg/cloud/huawei` at 80.6% coverage.

The three `pull_request` workflows (Build/Test, Format Check, Generate SBOM) have never run on this branch — they've been waiting on maintainer approval since Jul 31. Tushar confirmed on Slack (Aug 16) that it's just the fork-PR gate. Whenever you have a minute, "Approve and run" is all it needs.

Heads-up: Format Check will fail once approved, but not because of this PR — `pkg/cloud/digitalocean/provider.go` is unformatted on `develop` since 81e4474b (#4015). Same class of issue #4006 is fixing for `httpclient.go`.

alejandromore:
@peatey — you just reviewed and approved #4044 (IBM Cloud Usage Reports integration), which registers a new cloud cost integration through `pkg/cloud/config/configurations.go`, `statuses.go` and `pkg/cloudcost/integration.go`. This PR does the same thing for Huawei Cloud, through the same three files.

I've just merged current `develop` and resolved the overlap with #4044: the Huawei blocks sit directly after the IBM ones in each file, mirroring the pattern you approved there. `go test` + `go vet` exit 0 across all four modules on linux (`golang:1.26`); `pkg/cloud/huawei` at 80.6% coverage.

Two things it needs from a maintainer, neither of which I can do from a fork:

1. "Approve and run" on the three `pull_request` workflows — they've been at `action_required` since Jul 31 and have never run on this branch.
2. A review. Given you have #4044 fresh, you're probably the fastest person to tell whether this follows the right shape.

On the checks: `security/snyk` is red for the same reason it was red on #4044 — it tests the root `go.mod` whenever a PR touches it, and 3 of the 9 findings on `develop` have no fix available (`aws-sdk-go` v1, `x/crypto` openpgp). This PR adds none of them. Format Check will also fail once approved, but from `pkg/cloud/digitalocean/provider.go` being unformatted on `develop` since 81e4474b (#4015), not from this branch.

peatey:
I've read the full diff at bd0e2072 and verified the build locally.

**Verification:** `go build ./...` clean on root and `core`; `go vet` clean on `pkg/cloud/huawei`, `pkg/costmodel`, `pkg/cloud/provider`, `pkg/cloud/config`; `gofmt -l` empty on all touched files; `go test` passes for `pkg/cloud/huawei`, `pkg/costmodel`, `pkg/cloud/provider`, `core/pkg/opencost`, and `modules/huawei-obs-plugin`. The `pull_request` workflows have now run and pass (backend, format-check, generate-sbom). The author's claims check out.

## Blocking

**1. Cross-provider regression in `ComputeAssets` — `pkg/costmodel/assets.go:199-206`**

`pkg/cmd/costmodel/costmodel.go` sets `a.Model.CloudCostQuerier` whenever a cloud cost pipeline exists, for every provider. `ComputeAssets` then inserts every CloudCost line item as a `Cloud` asset. The dedup guard only covers Huawei:

```go
func isClusterCloudAsset(ca *opencost.Cloud, nodePools map[string]struct{}) bool {
	switch ca.Type() {
	case opencost.ECSCloudAssetType, opencost.EVSCloudAssetType:
	default:
		return false   // AWS/GCP/Azure always land here
	}
```

An existing AWS user with `CLOUD_COST_ENABLED=true` gets EC2 and EBS line items inserted alongside the Node/Disk assets already priced from metrics. `/assets` totals double-count cluster compute and storage. Silent regression for current users, shipped inside a Huawei PR.

**2. `Cloud.add()` drops `typ` — `core/pkg/opencost/asset.go:807-834`**

`add` merges properties, labels, window, cost, adjustment, credit, and never touches `typ`. Aggregating an RDS asset with a DCS asset yields a row labelled RDS carrying both costs. Invisible in the default path because `key()` includes `AssetTypeProp`, but `/assets?aggregate=account|service|cluster` hits it. Mismatched sub-types should collapse to `CloudAssetType`, the way `Any` degrades.

**3. 28 provider-specific `AssetType` values and a codec bump in `core`**

This is the call worth settling before the rest. `AssetProperties` already carries `Service` and `Category`, and `CloudCostProperties.Service` already holds the BSS service type. Sub-categorising the Infra Assets dashboard is a UI grouping change, not a core domain-type change. As written it burns Assets codec v21→v22 permanently, adds `MarshalBinary` to a type that had no binary form, and doesn't scale: nine providers × their service catalogues in one enum. I'd want a strong reason not to do this with `Service`/`Category` before accepting it.

**4. Layering**

`core/pkg/opencost/huaweiservices.go` puts a Huawei service catalogue and `HuaweiNodePoolLabel` in the provider-agnostic core package. `clusterNodePools`/`isClusterCloudAsset` put CCE node-pool naming conventions in generic `pkg/costmodel/assets.go`. Both belong in `pkg/cloud/huawei`.

The dedup heuristic is name-prefix matching (`<pool>-*`), and by its own comment an empty pool set keeps everything. A Huawei cluster whose nodes lack the CCE label silently double-counts its entire compute spend rather than failing loudly.

**5. The `ramGB` fix should be its own PR — `pkg/costmodel/costmodel.go:1245`**

`defaultRAM*ram` → `defaultRAM*ramGB` is a genuine, serious bug fix (bytes multiplied by a per-GB rate, ~1.07e9× inflation landing on GPU price). Unrelated to Huawei, affects every provider on the default-pricing fallback path, and deserves separate review and a backport. Please split it out; I'd merge that one quickly.

## Should fix before merge

- **`CostConfiguration.UnmarshalJSON` does blocking HTTP** to `169.254.169.254` with a 3s timeout (`costconfiguration.go:82`, `metadata.go:67`). Config unmarshalling shouldn't touch the network. Resolve the project ID lazily at integration time.
- **`context.Background()`** in `ClusterCloudCosts` (`assets.go:265`). The `/assets` request can't cancel the BSS query.
- **`modules/huawei-obs-plugin` isn't wired into anything.** Not in `just fmt`, `just test`, or any workflow; there's no `go.work`. Its `go.mod` is unchecked by any job and will rot. Separately, plugins live in `opencost/opencost-plugins`. No other plugin ships from this repo.
- **Internal references in shipped code**: `configs/huawei.json` description points at "the pricing-configs ConfigMap in mlops-infra helm/opencost"; `metadata.go:17` (both copies) references `mlops-infra/images/ces-exporter-image/ces_exporter.py`; `costconfiguration.go:78` references `values.yaml cloud.projectId`; three `Fase 3` markers in `pricingapi.go`. Meaningless outside the author's org.
- **BSS is hardwired to `bssintl` / `ap-southeast-1`** (`pricingapi.go:31`) with a TODO. Mainland-China accounts can't use this at all. Should be configurable, not a TODO.
- **Currency inconsistency**: `provider.go:517` defaults `CurrencyCode` to `"CNY"`, `configs/huawei.json` says `USD`, and the bssintl endpoint returns USD.
- **`LoadBalancerPricing` hardcodes `0.053`** (`provider.go:501`), a la-south-2 USD rate, as the global fallback. Belongs in `huawei.json`.
- **Supply chain**: the Huawei SDK pulls `go.mongodb.org/mongo-driver`, `tjfoc/gmsm`, and forces `json-iterator/go` from v1.1.12 to an untagged pseudo-version `v1.1.13-0.20220915233716-71ac16282d12` in the root module. That last one is shared with the k8s stack and worth a second look.
- **Provider detection ordering** (`pkg/cloud/provider/provider.go:321`): the new `os.name` prefix check now runs before OTC's `cce.cloud.com/cce-nodepool` check. The comment asserts these can't collide; since OTC is Huawei CCE technology, I'd want an OTC user to confirm before changing a shared detection path.
- **Sonar reports 19 new issues** on this PR. The GHAS hardcoded-IP finding on `metadata.go:19` is the IMDS address and is dismissable (other providers do the same), but the remaining 18 are worth a scan.

## Nits

`AssetType.MarshalBinary` truncates to one byte with no guard (`UnmarshalBinary` does check length) · `tryRefreshPricing` (`provider.go:333-335`) does RUnlock/RLock around `DownloadPricingData`, leaving the caller's deferred `RUnlock` unbalanced on panic · a single `*CloudCostProperties` (`costintegration.go:86`) is shared across every per-day `CloudCost` in a row · `Cloud.Credit` is never populated by the Huawei integration · nine `// TODO: Implement method` stubs in `provider.go:540-605`.

## Overall

The provider package itself is good work. The comments are unusually careful about *why*, test coverage is real (80.6% on `pkg/cloud/huawei`), and the author has clearly validated against a live account. The problem is scope: a provider addition bundled with a core domain-model change, a codec version bump, a cross-provider assets-pipeline change, an unrelated pricing bug fix, and an out-of-tree plugin. You already offered to split it, and I'd take you up on that:

1. The `ramGB` fix (merge fast)
2. Provider core + CloudCost integration (`pkg/cloud/huawei` + config plumbing). Mergeable on its own merit.
3. OBS plugin → `opencost-plugins`
4. Cloud asset sub-categorisation. Separate design discussion.

Snyk: findings are pre-existing on `develop`, 3 unfixable. Your analysis is correct.

alejandromore:
Thanks for the review — that's the most useful feedback this PR has had, and I've taken you up on the split.

**Out of this PR now** (d29555d):

1. `modules/huawei-obs-plugin` — removed. Goes to `opencost-plugins`. You were right that nothing built it here: not in `just fmt`, `just test`, or any workflow, and no `go.work`.
2. The 28 AssetType values, the `typ` field on `Cloud`, its `MarshalBinary`/`UnmarshalBinary` and the Assets codec v21→v22 — reverted. `core/pkg/opencost/{asset,asset_json,bingen,opencost_codecs}.go` are untouched again. Your point stands: sub-categorising the dashboard is a UI grouping question, and burning a permanent codec version inside a provider PR is the wrong place to settle it. Separate discussion.
3. The CloudCost→Cloud asset pipeline in `pkg/costmodel` (`ClusterCloudCosts`, `clusterNodePools`, `isClusterCloudAsset`) and its wiring in `pkg/cmd/costmodel` — removed, which takes the cross-provider regression with it. That one was a real bug and I'm glad you caught it before it shipped: the dedup guard only matched Huawei asset types, so any AWS/GCP/Azure user with cloud costs enabled would have had billed compute and storage inserted alongside the metrics-priced Node and Disk assets. `pkg/costmodel` is untouched by this PR now.
4. The `defaultRAM*ram` → `defaultRAM*ramGB` fix — now #4068, with a regression test.

**Layering (your point 4).** The service catalogue moved to `pkg/cloud/huawei/services.go` and now carries only the category mapping, resolving to the existing Compute/Storage/Network/Management categories — no new types. `core/pkg/opencost` carries one Huawei symbol, the `HuaweiProvider` constant in `assetprops.go`, next to every other provider's.

46 files / +5,163 → **23 files / +3,091**. `go test` and `go vet` exit 0 across all four modules on linux (`golang:1.26`).

One thing worth flagging on #4068: while writing the test for it I found that the existing `TestCustomProviderGPUNodeUsesDefaultHourlyPricing` does not actually cover that fallback — it passes with and without the fix. `CustomProvider`'s pricing map always yields a GPU price, so the GPU-set/GPUCost-empty shape that triggers the ratio-based fallback never occurs there. The new test fails on `develop` (GPU priced at 44,008,664/hr instead of 0.95) and passes with the change.

Still to do from your "should fix" list, on this branch: the blocking HTTP in `UnmarshalJSON`, the hardwired `bssintl`/`ap-southeast-1`, the CNY/USD inconsistency, the hardcoded load balancer rate, the internal `mlops-infra` references, and the Sonar findings. I'll push those next.

On provider detection ordering vs OTC — agreed that I shouldn't change a shared detection path on my own reasoning. Happy to gate the Huawei check behind explicit configuration instead, or wait for an OTC user to confirm; tell me which you'd prefer.

alejandromore:
Pushed 7e23566, covering the "should fix before merge" list and the Sonar findings.

**Configuration no longer touches the network.** `CostConfiguration.UnmarshalJSON` resolved the project ID from the instance metadata service; that moves to `resolveProjectID`, called from `GetCloudCost`. `projectID` is now optional in the config — `Validate` documents that it is resolved at integration time — and `Key` falls back to the region when it is absent.

**The BSS endpoint is configurable.** `bssRegionID` was a constant with a TODO; it is now `HUAWEICLOUD_BSS_REGION`, defaulting to the international endpoint. I turned the TODO into documentation rather than a knob that cannot work: a mainland-China account reaches BSS through the `services/bss/v2` package, not a different region on `bssintl`, so pointing this at `cn-north-1` would fail. Supporting that package is its own change; the comment now says so.

**Pricing defaults are consistent.** `GetConfig` defaulted `CurrencyCode` to CNY while `configs/huawei.json` declared USD and the international endpoint reports USD — now USD. `LoadBalancerPricing`'s hardcoded `0.053` moves to `defaultLBPrice` in `huawei.json`, read through `CustomPricing` the way `pkg/cloud/stackit` does, with `0.053` kept as the built-in default.

**Internal references removed:** the pricing-configs ConfigMap in `huawei.json`, `ces_exporter.py` and the `huaweiobs` plugin in comments, `values.yaml` in `costconfiguration.go`, and the three `Fase 3` markers.

**Sonar.** Pulled the list from the API — 19 issues, of which 5 were in the code the split removed (`asset.go`, three in `opencost_codecs.go`, and the catalogue, which moved). Of the rest: the nine `// TODO: Implement method` stubs are now one comment explaining that they satisfy `models.Provider` with zero values, as `pkg/cloud/stackit` does; the empty `ApplyReservedInstancePricing` says why it is empty; `DownloadPricingData`'s cognitive complexity (38) comes down via the extracted `resolveProviderProjectID`; and the catalogue lookup (19) folds its exact-match passes together.

One correction on the hardcoded-IP finding: you noted other providers do the same, but `169.254.169.254` appears nowhere else in the repo — no other provider queries an instance metadata service at all. It is still the standard link-local IMDS address and I think the finding is fair to dismiss, but not on that basis.

Five new tests: the `GetConfig` defaults, that they do not override configured values, the load balancer fallback from config, and the built-in default. Two updated: `Validate` no longer requires `projectID`, and the load balancer test now needs a config.

`go test` and `go vet` exit 0 across all four modules on linux (`golang:1.26`).

Still open from your review: whether to gate the Huawei node-label check in `pkg/cloud/provider/provider.go` behind explicit configuration, or leave the ordering as is until an OTC user can confirm. That one is your call.

PR #3987, opened 2026-08-01
feat: Add configuration validation and diagnostics API endpoint (GET /config/validate)

### LINKED ISSUE #3986: feat: Add configuration validation and diagnostics API endpoint (GET /config/validate)
---
name: OpenCost Feature request
about: Suggest an idea for this project
title: 'feat: Add configuration validation and diagnostics API endpoint (GET /config/validate)'
labels: 'feature'
assignees: ''

---

Operating OpenCost requires specifying complex environment configurations (Prometheus endpoints, cloud credentials, custom pricing schema files, etc.). If a config file or custom pricing CSV is malformed or invalid, OpenCost often falls back to default settings without providing a diagnostic endpoint for administrators to detect configuration issues.

Implement a diagnostic endpoint `GET /config/validate` (protected by `ADMIN_TOKEN` security middleware) that aggregates health and validation status across config sources:
1. Prometheus server endpoint ping status and latency.
2. Validation status of custom pricing CSV (parse status, valid row count, syntax error reporting).
3. Health/handshake status of registered CustomCost plugins.

Manually checking logs or attempting query requests to infer configuration errors, but a dedicated diagnostic endpoint provides instant visibility into configuration health.

**Additional context**
The endpoint returns a structured JSON payload detailing configuration health and syntax or connection errors.

Rucha0901:
## Description
This PR implements a diagnostic endpoint `GET /config/validate` in OpenCost protected by `ADMIN_TOKEN` security middleware (`adminAuthMiddleware`).

Key changes:
1. Registered `GET /config/validate` in `pkg/costmodel/router.go` wrapped with `adminAuthMiddleware`.
2. Implemented `ValidateConfig` handler on `*Accesses` struct aggregating:
   - Prometheus server endpoint ping status, latency, and connectivity errors (`validatePrometheusEndpoint`).
   - Custom pricing CSV file parsing validity, valid row count, and syntax errors (`validateCustomPricingCSV`).
   - Health and executable status of registered CustomCost plugins (`validateCustomCostPlugins`).
3. Added comprehensive unit tests in `pkg/costmodel/router_test.go` covering authentication (503/401/403/200), Prometheus endpoint ping, CSV validation, and plugin discovery.

## Related Issues
Closes #3986

## User Impact
Administrators and FinOps teams can query `GET /config/validate` with `ADMIN_TOKEN` bearer authentication to rapidly inspect configuration health across Prometheus, custom pricing CSV, and CustomCost plugins.

## Testing
1. Added unit tests for validation endpoint authentication, Prometheus ping, CSV parsing, and plugin discovery in `pkg/costmodel/router_test.go`.
2. Verified all tests pass using `CGO_ENABLED=0 go test ./pkg/costmodel/...`.
3. Verified code formatting (`go fmt`) and static analysis (`go vet`).


pkg/costmodel/router.go
```diff
@@ -680,3 +686,263 @@ func InitializeCustomCost(router *httprouter.Router) *customcost.PipelineService
...
+	}
+	if endpoint == "" {
+		return PrometheusValidation{
+			Endpoint: "",
+			Status:   "not_configured",
+			Error:    "PROMETHEUS_SERVER_ENDPOINT is not set",
+		}
+	}
+
+	client := &http.Client{
```
peatey, line 739: This bare `http.Client` bypasses OpenCost's entire Prometheus client configuration. `prom.NewPrometheusClient` (`modules/prometheus-source/pkg/prom/prom.go:386`) applies:

- basic auth / bearer token (`DB_BASIC_AUTH_USERNAME`, `DB_BASIC_AUTH_PW`, `DB_BEARER_TOKEN`)
- TLS settings — custom `RootCAs`, client certificates, `InsecureSkipVerify`
- the Cortex/Mimir `X-Scope-OrgID` header

None of that reaches this probe. Against an auth-protected Prometheus both requests get 401 and the endpoint reports `unreachable` while OpenCost is querying it perfectly well — the diagnostic tells an admin their working config is broken.

The inverse is also true: the root-URL fallback below reports `connected` for *any* HTTP server answering `<400`, so an endpoint typo'd onto some other service reads as healthy.

`prom.Validate` (`modules/prometheus-source/pkg/prom/validate.go:19`) already does this correctly — it runs the `up` query and is what `NewPrometheusDataSource` itself uses at startup. `Accesses.DataSource` is available on the receiver here.

---

pkg/costmodel/router.go
```diff
@@ -680,3 +686,263 @@ func InitializeCustomCost(router *httprouter.Router) *customcost.PipelineService
...
+			respDirect, errDirectDo := client.Do(reqDirect)
+			if errDirectDo == nil && respDirect.StatusCode < 400 {
+				respDirect.Body.Close()
+				return PrometheusValidation{
+					Endpoint: endpoint,
+					Status:   "connected",
+					PingTime: time.Since(start).String(),
+				}
+			}
+			if respDirect != nil {
```
peatey, line 772: **Response body leak.** When `/-/healthy` returns ≥400 (so `err == nil` and `resp != nil`) and the root fallback succeeds, this returns `connected` without ever closing `resp.Body`. The `if resp != nil { resp.Body.Close() }` further down is only reached on the `unreachable` path.

That's exactly the Thanos/Mimir/Cortex shape — no `/-/healthy`, root serves a UI — so it leaks a connection on *every* call to the endpoint in the configuration the fallback exists to support.

A `defer resp.Body.Close()` immediately after the `client.Do` (guarded on `err == nil`) fixes it and lets you drop the two manual closes.

---

pkg/costmodel/router.go
```diff
@@ -680,3 +686,263 @@ func InitializeCustomCost(router *httprouter.Router) *customcost.PipelineService
...
+
+	if csvPath == "" {
+		return CSVValidation{
+			Enabled: enabled,
+			Valid:   false,
+			Error:   "CSV provider is enabled but CSV_PATH environment variable is not set",
+		}
+	}
+
+	f, err := os.Open(csvPath)
```
peatey, line 822: `CSV_PATH` isn't necessarily a local file. `CSVProvider.CSVLocation` is set from `env.GetCSVPath()` (`pkg/cloud/provider/provider.go:135`), and `DownloadPricingData` branches on `strings.HasPrefix(c.CSVLocation, "s3://")` to fetch from S3 (`pkg/cloud/provider/csvprovider.go:80`).

So with `CSV_PATH=s3://bucket/pricing.csv` this reports `valid: false, "Failed to open CSV file: ... no such file or directory"` on a completely valid configuration.

---

pkg/costmodel/router.go
```diff
@@ -680,3 +686,263 @@ func InitializeCustomCost(router *httprouter.Router) *customcost.PipelineService
...
+			Enabled: enabled,
+			Path:    csvPath,
+			Valid:   false,
+			Error:   fmt.Sprintf("Failed to open CSV file: %v", err),
+		}
+	}
+	defer f.Close()
+
+	reader := csv.NewReader(f)
+	reader.FieldsPerRecord = -1
```
peatey, line 834: `FieldsPerRecord = -1` disables the one check the real provider performs. Compare `pkg/cloud/provider/csvprovider.go:118-122`:

```go
header, err := csvutil.Header(price{}, "csv")
fieldsPerRecord := len(header)
...
csvReader.FieldsPerRecord = fieldsPerRecord
dec, err := csvutil.NewDecoder(csvReader, header...)
```

The provider pins the column count to the `price` struct and decodes into it. This validator explicitly turns that off, so a CSV with the wrong number of columns — which the provider **will** reject — is reported `valid: true, rowCount: N`. It also never checks that the columns are the ones `price` expects.

Reporting green on a pricing file that won't load is worse than not checking at all. Reusing the provider's own header + decode path would make the diagnostic agree with the runtime by construction.

---

pkg/costmodel/router.go
```diff
@@ -680,3 +686,263 @@ func InitializeCustomCost(router *httprouter.Router) *customcost.PipelineService
...
+		Status:   "connected",
+		PingTime: time.Since(start).String(),
+	}
+}
+
+func validateCustomPricingCSV(csvPath string) CSVValidation {
+	if csvPath == "" {
+		csvPath = env.GetCSVPath()
+	}
+	enabled := env.IsUseCSVProvider() || csvPath != ""
```
peatey, line 804: `enabled` is inferred rather than reported. Setting `CSV_PATH` without `USE_CSV_PROVIDER=true` yields `enabled: true` even though the CSV provider is off and the file is never read.

Since this is a diagnostics payload, it'd be more useful to report the two facts separately — `env.IsUseCSVProvider()` as `enabled`, and the configured path alongside it — so an admin can see "path set but provider disabled", which is itself a likely misconfiguration worth surfacing.

---

pkg/costmodel/router.go
```diff
@@ -680,3 +686,263 @@ func InitializeCustomCost(router *httprouter.Router) *customcost.PipelineService
...
+	configDir := env.GetPluginConfigDir()
+	execDir := env.GetPluginExecutableDir()
+	enabled := env.IsCustomCostEnabled()
+
+	if configDir == "" {
+		configDir = "/var/configs/plugins"
+	}
+	if execDir == "" {
+		execDir = "/var/configs/plugins/exec"
+	}
```
peatey, line 881: These fallbacks are unreachable, and the values they'd assign are wrong. `pkg/env/cloudcost.go:79-85`:

```go
func GetPluginConfigDir() string {
	return env.Get(PluginConfigDirEnvVar, "/opt/opencost/plugin/config")
}

func GetPluginExecutableDir() string {
	return env.Get(PluginExecutableDirEnvVar, "/opt/opencost/plugin/bin")
}
```

Both already carry defaults and never return `""`, so neither branch fires — and `/var/configs/plugins` contradicts the real defaults, which would mislead anyone reading this later. Suggest dropping both blocks.

---

pkg/costmodel/router.go
```diff
@@ -680,3 +686,263 @@ func InitializeCustomCost(router *httprouter.Router) *customcost.PipelineService
...
+			ExecDir:   execDir,
+			Error:     fmt.Sprintf("Failed to read plugin config dir: %v", err),
+		}
+	}
+
+	for _, entry := range entries {
+		if entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
+			continue
+		}
+		parts := strings.Split(entry.Name(), "_")
```
peatey, line 907: This duplicates `getRegisteredPlugins` in `pkg/customcost/pipelineservice.go:33` — the same `<name>_config.json` convention and the same `execFmt` (`%s/%s.ocplugin.%s.%s`) are now hardcoded in a second place, so the two can drift.

They already behave differently. `getRegisteredPlugins` *errors* on a filename that doesn't match the convention (`pipelineservice.go:53-55`); this silently `continue`s. So a misnamed config file that will make the custom-cost pipeline fail to start is reported here as `count: 0, plugins: []` — a clean bill of health for a broken deployment.

Exporting the existing discovery (or factoring the scan out of it) and calling that would keep the diagnostic honest.

---

pkg/costmodel/router_test.go
```diff
@@ -115,3 +118,189 @@ func TestAdminAuthMiddleware(t *testing.T) {
...
+	})
+}
+
+func TestValidatePrometheusEndpoint(t *testing.T) {
+	t.Run("not configured when empty", func(t *testing.T) {
+		res := validatePrometheusEndpoint("")
+		if res.Status != "not_configured" {
+			t.Errorf("status = %q, want 'not_configured'", res.Status)
+		}
+	})
```
peatey, line 197: This subtest depends on ambient environment. `validatePrometheusEndpoint("")` falls back to `promenv.GetPrometheusServerEndpoint()`, so with `PROMETHEUS_SERVER_ENDPOINT` set it makes a live network call instead of returning `not_configured`.

That's the exact variable `AGENTS.md` tells contributors to export to run OpenCost locally, so the test fails on a normal dev machine:

```
$ PROMETHEUS_SERVER_ENDPOINT="http://127.0.0.1:9080" go test ./pkg/costmodel/ -run TestValidatePrometheusEndpoint
--- FAIL: TestValidatePrometheusEndpoint/not_configured_when_empty
    router_test.go:195: status = "unreachable", want 'not_configured'
```

It passes in CI only because the runner has the variable unset. Adding `t.Setenv(promenv.PrometheusServerEndpointEnvVar, "")` pins it.

---

pkg/costmodel/router_test.go
```diff
@@ -115,3 +118,189 @@ func TestAdminAuthMiddleware(t *testing.T) {
...
+
+	t.Run("200 when authorized with valid bearer token", func(t *testing.T) {
+		t.Setenv(env.AdminTokenEnvVar, testToken)
+		req := httptest.NewRequest(http.MethodGet, "/config/validate", nil)
+		req.Header.Set("Authorization", "Bearer "+testToken)
+		rec := httptest.NewRecorder()
+
+		handler := adminAuthMiddleware(a.ValidateConfig)
```
peatey, line 173: This subtest invokes the real `ValidateConfig`, so a "unit" test dials the network and reads the plugin config directory. It's fast today only because the CI runner has no Prometheus endpoint configured; pointed at a routable-but-dead address it burns the full 5s client timeout on every run.

The other three subtests here are the valuable part — they exercise `adminAuthMiddleware` and would work just as well against a trivial stub handler. Then `ValidateConfig`'s body can be covered separately, with the collaborators injected.

---

PR_DRAFT.md
peatey, line 1: This file and `ISSUE_DRAFT.md` look like local scaffolding for authoring the issue and PR description — they shouldn't be committed to the repo root. Please drop both.

(Minor: this draft says `Closes #3980` while the PR description says `Closes #3986`, which is the sort of drift that comes with keeping a second copy around.)

---

modules/pricing/public/httpclient/httpclient.go
peatey, line 1: Nit: this gofmt-only change is unrelated to the feature and adds diff noise. Worth splitting into its own PR (or dropping, if `develop` has since been formatted).

---

peatey (changes requested):
Thanks for this — a config diagnostics endpoint is a genuinely useful addition, and the plumbing is right. Route registration, `adminAuthMiddleware` wrapping, the CORS header, and the `WriteData` response shape all match the existing admin endpoints, there's no httprouter path conflict with the existing `/cloud/config/*` routes, and the 503/401/403/200 auth table is a good test to have.

My concern is with the three validators. Each re-implements a check that already exists elsewhere in the tree, and each diverges from the real implementation in a way that makes the diagnostic report the wrong answer. For an endpoint whose entire purpose is telling an admin whether their config works, that's the one failure mode it can't afford — and in several cases it fails toward a false "healthy", which is worse than no check at all.

### Blocking

1. **Prometheus check bypasses the real client config** — no auth, TLS, or `X-Scope-OrgID`, so an auth-protected Prometheus reports `unreachable` while OpenCost queries it fine. The root-URL fallback also reports `connected` for any HTTP server that answers. `prom.Validate` already does this correctly. (inline)
2. **Response body leak** in the `/-/healthy`-fails-then-fallback-succeeds path — exactly the Thanos/Mimir/Cortex shape. (inline)
3. **CSV validator disables the provider's own column-count check** (`FieldsPerRecord = -1`), so a file the provider will reject reports `valid: true`. (inline)
4. **`os.Open` on a path that may be an `s3://` URI** — S3-hosted pricing files are supported and report as broken. (inline)
5. **A test fails whenever `PROMETHEUS_SERVER_ENDPOINT` is set**, which is what `AGENTS.md` tells contributors to do to run locally. Reproduced inline; it passes in CI only because the runner is bare.
6. **`ISSUE_DRAFT.md` / `PR_DRAFT.md`** committed to the repo root. (inline)

### Should fix

7. `TestValidateConfigEndpointAuth/200 when authorized` does real network and filesystem I/O. (inline)
8. Plugin discovery duplicated from `pkg/customcost/pipelineservice.go` and divergent — a misnamed config that crashes the pipeline at startup reports `count: 0`. (inline)
9. Dead `if configDir == ""` fallbacks assigning paths that contradict the real defaults. (inline)
10. CSV `enabled` inferred from the path rather than read from `USE_CSV_PROVIDER`. (inline)
11. `AGENTS.md` has both an "Admin-protected endpoints" list and an "API Endpoints" table — `/config/validate` should be added to both.

### Nits

- The gofmt-only change to `modules/pricing/public/httpclient/httpclient.go` is unrelated. (inline)
- `PingTime` is a Go duration string (`"1.234ms"`) rather than a number, and in the fallback path it times both requests.
- `validatePrometheusEndpoint(endpoint string)`'s parameter is purely a test seam — the handler always passes `""`. Resolving the endpoint in the caller would be cleaner and would remove the ambient-env dependency behind finding 5.
- The `#`-comment-line skip in the CSV reader has no counterpart in the real provider, which would fail on such a line.

### Suggested direction

Have each validator delegate rather than reimplement: `prom.Validate` via `Accesses.DataSource` for Prometheus, the CSV provider's own `csvutil` header + decode for the pricing file, and an exported form of `getRegisteredPlugins` for plugins. That collapses most of the 266 new lines and — more importantly — guarantees the diagnostic and the runtime agree by construction, rather than by two implementations happening to stay in sync.

---

Issue #3940, opened 2026-07-18
feat(cloud): Automatic Cache Refresh Worker for Cloud Provider Pricing Data


Rucha0901:
**Is your feature request related to a problem? Please describe.**
Currently, OpenCost downloads cloud provider pricing data once on startup (via `a.CloudProvider.DownloadPricingData()`). There is no mechanism to automatically or periodically refresh the cached pricing data while the service is running. If cloud pricing changes or if new node instances/PV configurations are registered in the cluster that require pricing details, OpenCost may rely on stale or base pricing values unless the pod is restarted or the pricing data is retrieved synchronously on-demand, which can introduce latency and rate-limiting issues.

**Describe the solution you'd like**
Introduce an automatic background cache refresh worker for cloud provider pricing data. The worker should run as a background goroutine and periodically refresh the cloud provider pricing cache by invoking `DownloadPricingData()`.
Specifically:
1. Define a new environment variable `PRICING_REFRESH_RATE_HOURS` to specify the interval for cache refreshes (defaulting to 24 hours).
2. Set the background worker to be disabled if `PRICING_REFRESH_RATE_HOURS` is set to 0 or a negative number.
3. Start the background worker gracefully from the main entrypoint when Kubernetes is enabled.
4. Stop the background worker cleanly when a shutdown signal (Interrupt/SIGTERM) is received.

**Describe alternatives you've considered**
- Restarting OpenCost periodically via cron job or Kubernetes CronJob, which creates unnecessary service interruption.
- Alternatively, fetching pricing data on every node/PV query, which causes excessive CPU/network overhead and API rate limits.

**Additional context**
None.

PR #3988, opened 2026-08-04
aws: cache the pricing offer file and revalidate it with a conditional request

### LINKED ISSUE #15: aws pricing URL honors ETag
Since that file is currently over 600MB, it can be a very real win to use the `If-None-Match:` (or its `If-Modified-Since:` friend) to check if fetching those bytes could possibly result in any new information

Currently, since that URL is being served off of S3, it comes with all the expected headers, including `ETag:` containing the md5 of its contents:
```
	< Content-Length: 685820220
	< Last-Modified: Fri, 29 Mar 2019 04:08:55 GMT
	< ETag: "678d4f456aad44cb7a03708d7fd4511c"
```

If one sends that header back up, S3 is bright enough to indicate that downloading it will do you no good:
```
	> If-None-Match: "678d4f456aad44cb7a03708d7fd4511c"
	< HTTP/1.1 304 Not Modified
```

tormodmacleod:
## Description

`DownloadPricingData` re-fetches the region's entire AWS offer file every time it runs. The file is several hundred MB and changes roughly weekly, so almost every one of those transfers moves bytes that are already known.

It runs far more often than once a week. `NodePricing` calls it whenever a key misses `aws.Pricing` but hits `aws.ValidPricingKeys`, and those two maps are populated at different scopes:

- `populatePricing` **replaces** `aws.Pricing` on every download and refills it only for instance types present in the cluster at that moment — `if inputkeys[key] || inputkeys[spotKey]`, where `inputkeys` comes from the `GetAllNodes()` snapshot taken at the top of `DownloadPricingData`.
- `ValidPricingKeys[key]` / `[spotKey]` are written **outside** that guard, so they cover every on-demand instance type in the offer file.

So when an instance type leaves the node set its pricing is evicted by the next download, and when a node of that type returns the lookup misses, the `ValidPricingKeys` check passes, and the whole file is pulled again. Under autoscaler churn that repeats indefinitely.

### Impact observed

On a production EKS cluster in `eu-west-1` the exporter downloaded the offer file — **435,867,095 bytes** — **108 times a day**, about **53 GB/day** of NAT gateway egress, roughly **$72/month**. `starting download of` appears 1–9 times an hour, around the clock, with no diurnal shape.

Supporting detail:

- The cluster ran 8 distinct instance types across 11 nodes, 6 of them singletons, so a single consolidation evicts a type outright. Node launch rate (~2.5–4/hour) tracks the download rate.
- Downloads succeed — no fetch or decode errors — so nothing is failing; the same work is simply repeated.
- `GET /allNodePricing` returned exactly the keys for the running node types (6 instance types × 2 for the `,preemptible` twin, 15 EBS volume types × 2, plus `LoadBalancerUsage`) against hundreds of types in the file, which shows the scoping directly.
- A quieter cluster still did ~0.7 downloads/hour.

### The change

The endpoint serves an `ETag` and honours `If-None-Match` with a **304 and no body** — verified against the live endpoint. So: write the offer file to disk beside its ETag, and revalidate instead of re-downloading. An unchanged file then costs one conditional request, and repeated refreshes are cheap no matter what triggers them. This is what #15 asked for.

**Caching is opt-in, off by default** (`AWS_PRICING_CACHE_ENABLED`). It needs local storage roughly the size of the offer file, which on Kubernetes counts against the pod's `ephemeral-storage`, so it is not something to start consuming on upgrade without being asked. `AWS_PRICING_CACHE_DIR` points it at a volume where the writable layer is the wrong place.

**The all-region offer file is never cached, enabled or not.** It is ~9 GB against a region's ~436 MB, and `getPricingListURL` requests it whenever nodes span regions, whenever a node is missing its region label, and at startup before the cluster cache is populated. `pricingListURL` now reports whether it resolved a region and the cache is skipped when it did not, so an operator enabling the flag cannot be surprised by it.

Deliberately conservative about failure modes:

- The cache destination is opened **before** the request is issued, so an unwritable or missing directory is detected without having spent a download, and simply falls back to streaming.
- A cache write that fails midway latches caching off for an hour, so the fallback costs one extra download rather than one per refresh, and a transient full disk does not cost the cache for the life of the process.
- The body is renamed into place before its validator is written, so an interruption leaves a cache that is merely unvalidated, never mismatched.
- A validator with a missing body re-requests unconditionally rather than failing the refresh, and a validator that `net/http` would reject is deleted rather than retried — otherwise a single truncated `.etag` file fails every subsequent refresh and pricing never updates again.
- Cache paths are keyed by a hash of the URL, so a region change or an `AWS_PRICING_URL` override cannot read another URL's body.
- The cache lives in a private `0700` directory rather than the shared temp root, and cached files are opened `O_NOFOLLOW`. Both are only reachable by something that already shares the temp namespace — a sidecar on an emptyDir, a hostPath, an ephemeral debug container — but closing them is cheap.
- Partial downloads left by a crash or OOM-kill mid-copy are swept on first use, rather than accumulating one per crash at up to the full file size.

This also closes the pricing response body, which was previously never closed — a leaked connection or file handle on every refresh.

### What this does not do

It makes the repeated downloads cheap rather than removing them. The remaining cost is re-parsing the file on each refresh, which the conditional request does not help with. The narrower fix for that is to stop discarding pricing that is still relevant — merge into `Pricing` rather than replacing it, or widen `inputkeys` to recently-seen types — so a returning instance type is a cache hit and triggers no refresh at all. Following up separately.

I have not touched the `ValidPricingKeys` / `Pricing` scope asymmetry. Narrowing `ValidPricingKeys` to match `Pricing` would make the retry branch unreachable and remove the self-heal for types appearing between downloads, which looks intentional.

## Related Issues

- Closes #15 — conditional request on the offer file, opened in 2019 and still applicable; the header behaviour described there is unchanged today.
- Relates to #3940 — a periodic refresh worker addresses staleness, this makes each refresh cheap. Complementary, and this change makes a scheduled refresh much less costly to adopt.

## User Impact

No API change, not breaking, and nothing changes for anyone who does not set `AWS_PRICING_CACHE_ENABLED`. Where the cache is enabled, pricing content is unchanged: a revalidated cache is byte-identical to what would have been downloaded, and where the cache cannot be used behaviour falls back to exactly what happens today.

Two new environment variables, both documented in `AGENTS.md`:

| Variable | Default | Description |
|----------|---------|-------------|
| `AWS_PRICING_CACHE_ENABLED` | `false` | Cache the offer file and revalidate it with a conditional request |
| `AWS_PRICING_CACHE_DIR` | private `0700` dir under `os.TempDir()` | Where to keep it |

## Testing

`TestPricingOfferFileCache`, using an `httptest.Server` that implements `ETag` / `If-None-Match`:

- caching does nothing and writes nothing while disabled;
- a second refresh revalidates instead of downloading (one full GET and one 304 across two refreshes);
- a changed ETag produces a fresh download with the new body;
- a missing cached body with the validator still present falls back to a full download;
- a malformed validator — unquoted, unterminated, containing a control character, or oversized — is discarded rather than sent, and the next refresh revalidates normally instead of failing identically;
- a read-only cache directory, and a cache directory that cannot be created, both still serve pricing and latch the cache off;
- the latch expires rather than lasting the process.

`TestPricingListURLRegionScoping` pins which node sets produce the all-region index, since getting that flag wrong is what puts 9 GB on disk. `TestSweepPricingCachePartials` covers the orphan sweep.

`go test ./pkg/cloud/aws/ ./pkg/env/ -count=1` passes, `go test ./pkg/cloud/aws/ -run 'TestPricing|TestSweepPricing' -race -count=2` passes, `gofmt` and `go vet` are clean.

Not yet verified in-cluster — I have not deployed a build of this to the affected cluster. The measurements above are from the unpatched exporter.


peatey (changes requested):
Checked out the branch, verified the premise against the live AWS endpoint, and wrote throwaway probes for the failure modes. The diagnosis is correct and the approach is right. Blocking on one issue the PR's own reasoning misses, plus some `/tmp` hardening I confirmed by exploit.

## Validation — what holds up

- The analysis is accurate, not hand-waved. `populatePricing` does replace `aws.Pricing` wholesale and gate refills on `inputkeys` (`provider.go:1333-1337`) while writing `ValidPricingKeys` unconditionally outside that guard (`:1338-1339`), and `NodePricing` re-triggers a full `DownloadPricingData` on exactly that gap (`:1905-1908`). The churn loop is real.
- The endpoint behaves as claimed. Confirmed live: the `eu-west-1` offer file is **435,868,309 bytes**, serves an `ETag`, and honours `If-None-Match` with `304` / `size_download=0`.
- The `defer resp.Body.Close()` at `:1202` is a genuine pre-existing leak fix — `populatePricing` never closed it.
- Build, `go vet` and `gofmt` clean; the full `./pkg/cloud/aws/` suite passes; the new test passes under `-race -count=2`.

## Blocker: the multiregion fallback writes ~9 GB to `os.TempDir()`

`getPricingListURL` (`:819`) drops the region segment and requests the **all-region** index whenever nodes span regions, whenever any single node lacks a region label, **or whenever the node list is empty** — the startup case before the cluster cache is warm, since `region == "" && !multiregion` fails the `region != ""` test.

Measured just now:

```
regional  eu-west-1 index.json:    435,868,309 bytes
all-region          index.json:  8,953,850,152 bytes   <- ~9 GB
```

Today those 9 GB stream through the decoder and are discarded. With this change they land on disk, and since the temp file is created beside an existing cache, peak usage is ~2x — roughly **18 GB** of container writable layer. Kubernetes counts that toward `ephemeral-storage`, so the pod gets evicted, or the node hits imagefs pressure and evicts its neighbours. There is no `Content-Length` check and the `io.Copy` at `:895` has no ceiling.

Suggest capping before storing — check `resp.ContentLength`, with `io.CopyN` as the hard stop since Content-Length can be absent or wrong — or gating the cache behind the explicit opt-in and configurable path you offered in the description. Given this, I'd lean towards the explicit switch.

## Security findings

Each of these reproduced against the branch.

**Cache poisoning, no integrity binding.** The path is fully predictable — `/tmp/opencost-aws-pricing-<sha256(url)[:8]>.json` (`:863`) — and the ETag is public, one HEAD away. Planting a body plus the current ETag makes the exporter serve that content verbatim off the 304 path. Nothing verifies the cached bytes are the bytes AWS sent: no digest, no size check, no ownership check. The log line reads `unchanged, parsing the cached copy`, so there is no signal either. Impact is falsified cost/chargeback data.

**Symlink following.** `os.Open(bodyPath)` at `:872` follows symlinks, so a link planted at the cache path causes the exporter to read and parse any file readable by uid 1001. The write side is fine — `os.Rename` replaces the link rather than following it — so this is specifically the 304 read path.

Both require write access to the same `/tmp` namespace, so neither is reachable in the default single-container pod. They are in play with a sidecar sharing an emptyDir at `/tmp`, a hostPath `/tmp`, an ephemeral debug container, or a non-container deployment. Both close at once if the cache lives in a private `0700` directory rather than the shared temp root, plus `O_NOFOLLOW` on the read.

**A malformed `.etag` file wedges pricing permanently.** This is the one most likely to bite with no attacker involved. The file contents go into `req.Header.Set("If-None-Match", ...)` at `:973` after only `TrimSpace`, and Go's transport rejects invalid header values *before dialling*:

```
attempt 0 err=Get "http://...": net/http: invalid header field value for "If-None-Match"
attempt 1 err=...   attempt 2 err=...
```

The error propagates out of `DownloadPricingData`, and nothing deletes the offending file — so every subsequent refresh fails identically and pricing never updates again until someone clears `/tmp`. A truncated write gets you there as easily as a hostile one. Worth validating the value (`^(W/)?"[^"]*"$`), bounding the read, and unlinking the file when it doesn't parse.

Nothing else raised a flag: the diff is confined to two files, no new dependencies, no credential handling touched, no new network destinations (the `AWS_PRICING_URL` override pre-dates this PR), and the cached content is public data written `0600`.

## Lower severity

- **No cleanup, ever** (`:930`). Cache files are never removed — not at shutdown, not when the URL flips region — and orphaned `.part-` files from a crash or OOM-kill mid-copy accumulate one per crash, each up to the full file size. A startup sweep of `opencost-aws-pricing-*` would cover it.
- **The unusable-latch is permanent and process-wide.** One transient ENOSPC and caching is off until the pod restarts, straight back to 53 GB/day behind a single `Warnf`. A timed reset, or at least `Errorf`, would be better.
- **Retry keeps the stale validator** (`:1017`). `req.Clone` carries `If-None-Match` into the post-failure retry. It can't 304 today since that branch is only reachable from a 200, but if it ever did, the caller would parse a zero-byte body as pricing. A `Header.Del` on the clone makes it obviously safe.
- **`cachedPricingResponse` sets `ContentLength: 0`**, which in Go means "0 bytes" rather than "unknown"; `-1` is the correct sentinel. Nothing reads it today.
- **The test leaves the global latched** — the last subtest exits with `pricingCacheUnusable == true` for the rest of the package run. And the "unwritable cache directory" case actually exercises a *non-existent* directory; the realistic `readOnlyRootFilesystem: true` case (directory exists, EROFS) isn't covered.

The PR is upfront that it doesn't stop the churn — each 304 still re-parses 436 MB of JSON, so the CPU cost of 108 refreshes/day remains. That's reasonable as a follow-up.

Requesting changes on the multiregion path and the ETag wedge, both of which are reachable without an adversary. The `/tmp` hardening is cheap and closes the poisoning and symlink findings together. Everything else is polish.

tormodmacleod:
Pushed 9677716 and 2d6ab3d. All of it addressed; where I did something other than what you suggested it is called out below.

## Blocker: the multiregion 9 GB

Both belt and braces.

Caching is now behind `AWS_PRICING_CACHE_ENABLED`, default false, with `AWS_PRICING_CACHE_DIR` for the path — the explicit switch you leaned towards.

I also stopped the all-region index being cached at all, enabled or not. The switch alone still lets an operator turn the cache on and get the 9 GB, and the empty-node-list case means that happens at startup on the first refresh, before the cluster cache is warm — so it is not really an operator choice. `getPricingListURL` is now a wrapper over `pricingListURL`, which also reports whether it resolved a region; `getRegionPricing` streams as before when it did not. `TestPricingListURLRegionScoping` pins all four node-set shapes.

I did not add the `Content-Length` / `io.CopyN` ceiling. With the all-region file excluded by construction the cap has nothing left to catch, and a byte limit invites a wrong guess about how large a regional file is allowed to get. Say the word if you would rather have it as a backstop.

## The ETag wedge

Fixed, and your `^(W/)?"[^"]*"$` was too permissive — I wrote the test before the regex and it failed on `"with\nnewline"`. `[^"]` admits control characters, which is exactly what the transport rejects, so that pattern still wedges. It is now `^(?:W/)?"[\x21\x23-\x7e]*"$` — RFC 9110 etagc minus obs-text. Stricter than the transport, which only costs a full download on an exotic-but-valid ETag.

The read is bounded at 4 KiB, and anything that does not parse is unlinked rather than kept. The test covers unquoted, unterminated, control-character and oversized values, and asserts the *next* refresh revalidates normally rather than repeating the failure.

## Cache poisoning and symlinks

Both closed by moving out of the shared temp root, as you suggested: the cache lives in a private `0700` directory, verified after `MkdirAll` with `Lstat` — which also rejects a symlinked cache directory — and chmod'd back if it exists too wide. Reads go through `openPricingCacheFile`, which is `O_NOFOLLOW` on unix and an `Lstat` regular-file check elsewhere.

To be clear about what I did not do: there is still no digest binding the cached bytes to what AWS sent. The private directory means anything able to plant a body can already write the directory, so a digest stored beside it buys nothing. If you want integrity independent of the filesystem it needs to come from the endpoint, which is a different change.

## Lower severity

- **No cleanup** — `sweepPricingCachePartials` runs once per process behind a `sync.Once`, removing `opencost-aws-pricing-*.part-*`. `TestSweepPricingCachePartials` covers what it must not delete.
- **Permanent latch** — now a timed latch, one hour, and `Errorf` rather than `Warnf`.
- **Retry keeps the stale validator** — `Header.Del("If-None-Match")` on the clone. You are right that it is unreachable today; it is now obviously safe rather than argued safe.
- **`ContentLength: 0`** — now `-1`.
- **Test leaves the global latched** — reset via `t.Cleanup` as well as on entry. The unwritable case is now a real `0500` directory (skipped as root, which ignores the mode), and the non-existent-directory case is kept as a separate subtest.

## Still outstanding

The re-parse cost is unchanged — each 304 still parses 436 MB of JSON, so the CPU cost of 108 refreshes/day remains. Following up separately with the `populatePricing` scoping fix, which removes most of those refreshes rather than making them cheap.

tormodmacleod:
@peatey nudge on this one — everything from your review was addressed in 9677716 and 2d6ab3d back on 1 September, with a note on the two places I did something other than what you suggested. The branch was 21 commits behind, so I have just merged `develop` in cleanly (f556985) and re-verified: `gofmt` and `go vet` clean, `./pkg/cloud/aws/` and `./pkg/env/` pass, and the pricing tests pass under `-race -count=2`.

The one thing still open is a question for you rather than a change: I did **not** add the `Content-Length` / `io.CopyN` ceiling. With the all-region index excluded by construction the cap has nothing left to catch, and a byte limit invites a wrong guess about how large a regional offer file is allowed to grow. Happy to add it as a belt-and-braces backstop if you would rather have it — just say which way you want it and I will push.

No rush, but a re-review when you have a window would be appreciated.

Issue #2247, opened 2023-10-27
Move from deprecated Azure Rate Card API to Pricesheet API


lostcities:
**Is your feature request related to a problem? Please describe.**

We have discounted offerIDs for our Azure subscriptions, and weren't seeing the discounts applied via the Rate Card API.  MSFT support confirmed that it is deprecated (so this will not be fixed) and pointed us to the Consumption Pricesheet API: https://learn.microsoft.com/en-us/rest/api/consumption/price-sheet/get?tabs=HTTP .  This API does appear to be discovering/applying discounting based on the subscription ID.

**Describe the solution you'd like**

We have addressed this in a fork of OpenCost because it was urgent for us. (We also forked for workload identity support, and because we deploy a handful of Windows node types).   But this will help anyone else out there with discounted offer IDs (those other than MS-AZR-0003P).  And, we'd love not to use/maintain an opencost fork forever ...

**Describe alternatives you've considered**

We had a very long support exchange with MSFT.  There don't appear to be other options for discounted pricing.  RateCardAPI is still fine for those with default pricing.

**Additional context**
Add any other context, documentation links, or screenshots about the feature request here.

We found the SDK https://github.com/hashicorp/go-azure-sdk/tree/main/resource-manager/consumption/2023-05-01/pricesheet and hoped it would do most of the work for us, but unfortunately it does not support paginated responses, and the Consumption Pricesheet API's returning 224 pages.   (It also doesn't support workload identity.). But we did find its JSON model helpful for unmarshalling the data. 

I gather from the conversation in our other issue, there is some refactoring planned that should come first.  But if/when things are ready, we should be able to contribute this work back (pending organizational approval) -- especially after we've tested it for a while in clusters across multiple subscriptions and regions.

mattray:
@lostcities Thanks for the detailed report! I'll raise this with Kubecost Engineering in our next call, it sounds like something we should address sooner rather than later. Do you have any associated PRs?

lostcities:
We are working in a private Azure DevOps repo, so there is no publicly visible PR.  Most of the additions for the pricesheet API are in one new file.  I can dump out our forked version into Github, so that you can see the nature of the changes.  I'm still waiting for my internal PR review so it makes sense to get that feedback first.

We forked from 1.103, so the merge back is going to be involved, and would also include the workload identity support, which changes the auth flow.  

[...]

PR #3992, opened 2026-08-10
Azure Price Sheet: support keyless auth via DefaultAzureCredential

abhisheks-devspace:
## What does this PR change?

The Azure Price Sheet downloader hardcodes `azidentity.NewClientSecretCredential`, so negotiated-pricing retrieval requires a service principal **client secret** even on clusters that authenticate with Azure Workload Identity or managed identity. The CloudCost integration already supports keyless auth via its `AzureDefaultCredential` authorizer (`pkg/cloud/azure/authorizer.go`); this PR brings the Price Sheet path to parity.

- When `clientID`/`clientSecret`/`tenantID` are configured, the existing `ClientSecretCredential` path is used **unchanged**.
- When they are absent, `DefaultAzureCredential` resolves workload-identity federated tokens, managed identity, and Azure CLI credentials via the standard chain.

## Why

Secretless auth (Workload Identity) is the recommended posture on AKS — many clusters (ours included, a healthcare platform where long-lived secrets are a compliance finding) cannot mount SP secrets. Today that means no negotiated pricing at all on such clusters.

## Testing

- `go build ./pkg/cloud/azure/` and `go test ./pkg/cloud/azure/` pass.
- Behavior note: a *partially* configured SP (e.g. clientID set, secret missing) previously failed at credential construction; it now falls through to the DefaultAzureCredential chain, which is strictly more useful and surfaces a clear auth error if nothing in the chain applies.

Related: #2247


pkg/cloud/azure/pricesheetdownloader.go
```diff
@@ -52,7 +53,16 @@ func (d *PriceSheetDownloader) GetPricing(ctx context.Context) (map[string]*Azur
...
+	var cred azcore.TokenCredential
+	var err error
+	if d.ClientID != "" && d.ClientSecret != "" && d.TenantID != "" {
+		cred, err = azidentity.NewClientSecretCredential(d.TenantID, d.ClientID, d.ClientSecret, nil)
+	} else {
+		// Keyless: resolves workload identity, managed identity, and CLI
+		// credentials via the standard chain, matching the CloudCost
+		// integration's AzureDefaultCredential authorizer.
+		cred, err = azidentity.NewDefaultAzureCredential(nil)
+	}
```
peatey, line 65: **Suggestion: log the chosen auth path, pass the configured tenant through, and make the selection testable.**

Three small things, folded into one change because they touch the same lines:

1. **Log which path was taken.** `NewDefaultAzureCredential` never fails at construction — unavailable credentials become deferred error reporters, so a bad ambient setup surfaces only much later, wrapped as `beginning pricesheet download: ...`. Without a log line an operator can't tell whether OpenCost used the service principal or the ambient chain. `log` is already imported, and `GetPricing` logs at Info just above.

2. **Pass `d.TenantID` to the default chain.** In the `else` branch the tenant may well be set — tenant configured but no secret is precisely the keyless case. `DefaultAzureCredentialOptions.TenantID` sets the default tenant for the workload identity and CLI credentials. An empty string reproduces today's behaviour, so this is a strict improvement. This is the one spot where the downloader legitimately differs from `DefaultAzureCredentialHolder.GetCredential()`, which passes `nil` only because it has no tenant to hand.

3. **Extracting `credential()`** makes the selection unit-testable with no network — see the test in the next comment.

```suggestion
// credential resolves the token credential used to request the price sheet.
// A fully configured service principal takes precedence; otherwise we fall
// back to the default chain, which resolves workload identity, managed
// identity, and CLI credentials. This mirrors both the gate the rate card
// path already uses in provider.go and the CloudCost integration's
// AzureDefaultCredential authorizer.
func (d *PriceSheetDownloader) credential() (azcore.TokenCredential, error) {
	if d.ClientID != "" && d.ClientSecret != "" && d.TenantID != "" {
		log.Infof("pricesheet: authenticating with the configured service principal client secret")
		return azidentity.NewClientSecretCredential(d.TenantID, d.ClientID, d.ClientSecret, nil)
	}
	log.Infof("pricesheet: no complete service principal configured, using the default Azure credential chain")
	// TenantID is empty unless it was configured, in which case it sets the
	// default tenant for the workload identity and CLI credentials.
	return azidentity.NewDefaultAzureCredential(&azidentity.DefaultAzureCredentialOptions{
		TenantID: d.TenantID,
	})
}

func (d *PriceSheetDownloader) getDownloadURL(ctx context.Context) (string, error) {
	cred, err := d.credential()
```

Verified against the PR head: builds, `go vet` clean, `gofmt` clean, existing tests pass.

pkg/cloud/azure/pricesheetdownloader.go
```diff
@@ -52,7 +53,16 @@ func (d *PriceSheetDownloader) GetPricing(ctx context.Context) (map[string]*Azur
...
 	if err != nil {
 		return "", fmt.Errorf("creating credential: %w", err)
 	}
```
peatey, line 68: **Test coverage for the credential selection.**

The PR currently adds no tests. With the `credential()` extraction from the previous comment, the selection becomes testable offline — I confirmed `NewClientSecretCredential` constructs fine from dummy values, and `NewDefaultAzureCredential` returns a credential with no error and makes no network calls at construction, so this is safe in CI.

Posting as a code block rather than a suggestion since `pricesheetdownloader_test.go` isn't part of this diff:

```go
func TestCredential(t *testing.T) {
	base := PriceSheetDownloader{
		TenantID:     "test-tenant-id",
		ClientID:     "test-client-id",
		ClientSecret: "test-client-secret",
	}

	t.Run("complete service principal uses client secret", func(t *testing.T) {
		cred, err := base.credential()
		require.NoError(t, err)
		require.IsType(t, &azidentity.ClientSecretCredential{}, cred)
	})

	// Any missing piece of the service principal falls back to the default
	// chain, matching the rate card gate in provider.go.
	for field, clear := range map[string]func(*PriceSheetDownloader){
		"tenant ID":     func(d *PriceSheetDownloader) { d.TenantID = "" },
		"client ID":     func(d *PriceSheetDownloader) { d.ClientID = "" },
		"client secret": func(d *PriceSheetDownloader) { d.ClientSecret = "" },
	} {
		t.Run("missing "+field+" falls back to default chain", func(t *testing.T) {
			d := base
			clear(&d)
			cred, err := d.credential()
			require.NoError(t, err)
			require.IsType(t, &azidentity.DefaultAzureCredential{}, cred)
		})
	}
}
```

Needs `github.com/Azure/azure-sdk-for-go/sdk/azidentity` added to the test file's imports; `require` is already there. I ran this against the refactor above and all four subtests pass.

peatey (commented):
## Review: keyless auth for the Azure price sheet

Well-scoped and I believe correct — 11 lines, one file, no new dependencies, closing a real gap. Nothing below is blocking.

### The premise checks out

I confirmed the bug isn't theoretical. `provider.go:1002` gates the price-sheet download only on `AzureBillingAccount != ""` — nothing there requires a client secret — yet `getDownloadURL` hardcoded `NewClientSecretCredential`, which fails at construction on a secretless cluster (`secret can't be empty string`). So negotiated pricing was genuinely unreachable under workload identity.

### The partial-SP caveat is smaller than the description suggests

The PR lists the partially-configured service principal behaviour change as a caveat. It's actually the pre-existing semantic in the very same function: the new condition is character-for-character the gate already at `provider.go:915`, and the rate-card path immediately below it (lines 925-936) already falls back to ambient auth when the triple is incomplete. This brings the price sheet in line with the rate card rather than inventing new behaviour — worth stating plainly, since it removes the main reason a reviewer might hesitate.

Two more points in the change's favour: the `var cred azcore.TokenCredential` + if/else shape matches existing precedent at `core/pkg/storage/azurestorage.go:553-563`, and `azcore` is already a direct dependency (`go.mod:17`, also imported by `authorizer.go` in this package), so there's no go.mod churn.

### Verified locally

On the PR head (`e40f021`): `go build`, `go vet`, `go test ./pkg/cloud/azure/`, and `gofmt` all clean; DCO passes. The branch has since been updated with `develop`, so the earlier `behind` state is resolved.

I also checked the SDK behaviour the change leans on: `NewDefaultAzureCredential(nil)` does not fail at construction and makes no network calls there — unavailable credentials become deferred error reporters, so failures surface at `GetToken` time. That confirms the description's claim that a clear auth error still surfaces when nothing in the chain applies, and it's what motivates the logging suggestion inline.

### One question, not a blocker

The price-sheet block only runs if the rate-card query at `provider.go:964` already succeeded. When the SP triple is incomplete, the rate card uses the legacy autorest `auth.NewAuthorizerFromEnvironment()` (lines 925-936), which handles env client-credentials and IMDS managed identity but *not* workload identity's federated token file. On a pure workload-identity cluster, `DownloadPricingData` could therefore return at line 968 before ever reaching the price sheet. You report this working in production, which suggests your cluster also has a managed identity the rate card can use — can you confirm? If so, bringing the rate-card path to parity looks like a natural follow-up rather than something to fold into this PR.

### Inline

Two suggestions below: one consolidating change that adds an auth-path log line, passes the configured tenant through, and makes the selection unit-testable; and the test to go with it. Both verified to build and pass.

abhisheks-devspace:
Thanks for the thorough look — the `authorizer.go` precedent citation and the `azcore` dependency check saved me from having to point that out myself.

You're right to push on this, and I should be more precise than the PR description was: **we haven't actually confirmed the price sheet downloads successfully end-to-end in our own production.** Re-reading my own wording, "ours included" was describing the problem (no negotiated pricing today), not a confirmed fix — sorry if that read as a stronger claim than intended.

To your actual question — does our cluster have a managed identity the rate card can fall back on — **no**, and I can be specific about it: our own `/pricingSourceStatus` shows the rate card call failing with an empty `AzureSubscriptionID` and IMDS returning `"error":"invalid_request","error_description":"Identity not found"`. No managed identity there, just workload identity. So the gap you're describing isn't hypothetical — it's exactly what's happening on our cluster right now, and this PR alone won't get us negotiated pricing until the rate-card path also handles WI.

There's actually a second, independent blocker stacked on top: our `AzureBillingAccount` isn't even set, so the price-sheet block never gets reached regardless of the rate-card outcome. Two separate reasons the code path is currently dead for us — not proud of that, but better to say it plainly.

Given that, I think you've got the scope call right: this PR is still worth taking as-is (it's correct, and it unblocks the credential-construction failure, which is a real bug independent of whether the rate card also succeeds), and bringing the legacy auth at 925-936 up to WI parity is a genuinely separate change — bigger, touches a path with its own history, deserves its own review rather than riding in here. I'll open that as a follow-up once I've got something concrete rather than hand-wave it.

Appreciate you actually checking the SDK behavior instead of taking the description at face value — that's exactly the kind of review this needed.

PR #4000, opened 2026-08-13
azure: use the Cost Management price sheet API for EA billing accounts

### LINKED ISSUE #3993: Azure Price Sheet API is still using retired Microsoft.Consumption endpoint
**Describe the bug**
OpenCost fails to download the Azure Enterprise Agreement price sheet because it calls the legacy Microsoft.Consumption/pricesheets/download API.

Reference: https://learn.microsoft.com/en-us/rest/api/cost-management/price-sheet/download-by-billing-account?view=rest-cost-management-2025-03-01&utm_source=chatgpt.com&tabs=HTTP
`All versions of the Microsoft.Consumption Azure Price Sheet - Download by Billing Account (including 2022-06-01, 2021-10-01, 2020-01-01-preview, 2019-10-01, 2019-05-01) are scheduled to be retired on 01 June 2026 and will no longer be supported after this date.`

The request returns:

AuthorizationFailed:
does not have authorization to perform action
'Microsoft.Consumption/pricesheets/action'

The service principal has the required EnrollmentReader billing role.

**To Reproduce**
Steps to reproduce the behaviour: 

Follow instruction and grand Enrollment Reder role to OpenCost SPN https://opencost.io/docs/configuration/azure#customer-specific-pricing

The old API call returns an error
`curl -i -X POST   "https://management.azure.com/providers/Microsoft.Billing/billingAccounts/${BILLING_ACCOUNT_ID}/billingPeriods/202608/providers/Microsoft.Consumption/pricesheets/download?api-version=2022-06-01"   -H "Authorization: Bearer ${TOKEN}"   -H "Content-Type: application/json"   --data-raw ''`

{"error":{"code":"AuthorizationFailed","message":"The client 'xxxxxxx' with object id 'xxxxxx' does not have authorization to perform action 'Microsoft.Consumption/pricesheets/action' over scope '/providers/Microsoft.Billing/billingAccounts/xxxx/billingPeriods/202608/providers/Microsoft.Consumption/pricesheets/download' or the scope is invalid. If access was recently granted, please refresh your credentials."}}

The new API call returns the data
`curl -i -X POST   "https://management.azure.com/providers/Microsoft.Billing/billingAccounts/${BILLING_ACCOUNT_ID}/billingPeriods/202608/providers/Microsoft.CostManagement/pricesheets/default/download?api-version=2025-03-01"   -H "Authorization: Bearer ${TOKEN}"   -H "Content-Type: application/json"   --data-raw ''`
 

**Expected behavior**
OpenCost should use the current Microsoft.CostManagement Price Sheet API for EA billing accounts.

[opencost/opencost-ui:1.120.4](opencost:1.120.4)

dakshshrivastav:
## What

Moves the Azure EA price sheet download off the retired
`Microsoft.Consumption/pricesheets/download` API and onto
`Microsoft.CostManagement/pricesheets/default/download`, and teaches the parser
to read what that API actually returns.

Fixes #3993

## Why

`Microsoft.Consumption/pricesheets/download` retired on 1 June 2026, but it had
already stopped working before that. The endpoint authorises against
`Microsoft.Consumption/pricesheets/action`, which the `EnrollmentReader` billing
role no longer carries, so a service principal configured exactly as
[our docs describe](https://opencost.io/docs/configuration/azure#customer-specific-pricing)
gets:

```
AuthorizationFailed: The client '...' does not have authorization to perform action
'Microsoft.Consumption/pricesheets/action' over scope
'/providers/Microsoft.Billing/billingAccounts/.../providers/Microsoft.Consumption/pricesheets/download'
```

The effect is that EA customers silently fall back to Rate Card list pricing
instead of their negotiated rates.

## What changed

This is not just a URL swap, because the replacement API changes the payload as
well as the route:

| | Legacy (`Microsoft.Consumption`) | Current (`Microsoft.CostManagement`) |
|---|---|---|
| Result field | `properties.downloadUrl` | `properties.reportUrl` |
| Artifact | one `.csv`, two preamble lines | **`.zip` of multiple `.csv` parts**, 75MB each |
| Schema | 14 fixed-position columns | 21 renamed/reordered columns, no `Unit`, no `OfferID`, **`ReservedInstance` rows included** |

**1. Client.** Deleted the hand-rolled ARM client. Its comment said it only
existed because the SDK didn't support the API yet; `armcostmanagement/v3` now
does, so the URL, api-version (`2025-03-01`) and LRO polling are maintained
upstream. This bumps `azcore` to v1.22.0 and `msal` to v1.7.0. No other module
needed changing.

**2. Zip support.** The download is sniffed for the zip magic number, then every
`.csv` member is parsed and merged. One malformed part is logged and skipped
rather than discarding the whole sheet. A bare CSV still works.

**3. Columns by name.** Fixed positions can't survive a reordered schema, so
columns are now resolved by normalised name (case, spaces and dashes stripped).
One code path reads both layouts, and `"Meter sub-category"` matches
`"MeterSubCategory"`. The offer ID filter only applies when the sheet actually
has that column — the schema reference and the price sheet terms page disagree
about whether it still exists, so both cases are handled.

**4. Price type filter (behaviour change worth reviewing).** The old code only
skipped `"Savings Plan"`. The new sheet also carries `ReservedInstance` rows
whose `UnitPrice` is the *total cost of a one or three year commitment*. Those
sort after the consumption row for the same meter, so left in they would
overwrite hourly on-demand prices with a number thousands of times too large.
Only consumption rows are kept now. There's a regression test for exactly this.

**5. Diagnosability.** `no matching pricing from price sheet` now explains
itself. A mismatched `AZURE_OFFER_ID` is the most common cause and there was
previously nothing in the error or the logs pointing at it.

## Testing

- `go test ./...` on the main module, plus `go build ./...` on `core/`,
  `modules/prometheus-source` and `modules/collector-source` against the bumped
  `azcore`.
- `go fmt ./...` / `go vet ./...` clean.
- New tests: `httptest`-backed coverage of the full 202 + `Location` -> poll ->
  200 long running operation; zip merging, partial failure and no-CSV cases;
  both schemas producing identical prices; the `ReservedInstance` overwrite
  regression; BOM and CRLF handling; column mapping.
- Both commits build and pass tests independently.

## One thing I'd like a second opinion on

The 2025-03-01 spec (and therefore the generated SDK model) says the result
field is `reportUrl`. Microsoft's
[migration guide for this same endpoint](https://learn.microsoft.com/en-us/azure/cost-management-billing/automate/migrate-ea-price-sheet-api)
says `downloadUrl`. The reference page's `OperationStatus` definition is also
visibly copy-pasted from the reservation report API — it types the URL field as
the `ReservationReportSchema` *enum*.

I couldn't settle this without a live EA enrollment, so the code accepts either:
it prefers the typed SDK value and falls back to reading the raw response body
through a small pipeline policy (`runtime.Payload`, which rewinds, so the SDK
still parses the response normally). Both paths are tested.

**@yalmaty** — if you're able to share the response body from the curl
in the issue (with the SAS token redacted), I can drop whichever branch turns
out to be dead.

## Follow-ups, not in this PR

- The docs at opencost.io still describe the old permission model; the
  `EnrollmentReader` guidance is still correct but the API name should be
  updated. That lives in a different repo.
- We only request the current billing period. If Azure hasn't generated it yet
  we fall back to Rate Card, same as before. Retrying against the previous
  period would be a reasonable improvement.
- `conversions` was built from the legacy `Unit` vocabulary. If `UnitOfMeasure`
  spells anything differently those prices pass through un-normalised; the
  existing `all units in pricesheet:` log line surfaces that and is kept.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	cloud.google.com/go/compute/metadata v0.9.0
 	cloud.google.com/go/storage v1.60.0
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.0
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.22.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.4
 	github.com/Azure/go-autorest/autorest v0.11.30
@@ -69,6 +69,7 @@ require (
 )
 
 require (
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/costmanagement/armcostmanagement/v3 v3.0.0
 	github.com/google/go-cmp v0.7.0
 	github.com/opencost/bingen v0.2.0
 	github.com/oracle/oci-go-sdk/v65 v65.117.1
@@ -81,7 +82,7 @@ require (
 	cloud.google.com/go/auth/oauth2adapt v0.2.8 // indirect
 	cloud.google.com/go/iam v1.5.3 // indirect
 	cloud.google.com/go/monitoring v1.24.3 // indirect
-	github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.2 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/internal v1.12.0 // indirect
 	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
 	github.com/Azure/go-autorest/autorest/adal v0.9.22 // indirect
 	github.com/Azure/go-autorest/autorest/azure/cli v0.4.6 // indirect
@@ -90,7 +91,7 @@ require (
 	github.com/Azure/go-autorest/autorest/validation v0.3.2 // indirect
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
-	github.com/AzureAD/microsoft-authentication-library-for-go v1.6.0 // indirect
+	github.com/AzureAD/microsoft-authentication-library-for-go v1.7.0 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp v1.31.0 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.55.0 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.55.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -29,12 +29,18 @@ github.com/Azure/azure-sdk-for-go v68.0.0+incompatible h1:fcYLmCpyNYRnvJbPerq7U0
 github.com/Azure/azure-sdk-for-go v68.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
 github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.0 h1:fou+2+WFTib47nS+nz/ozhEBnvU96bKHy6LjRsY4E28=
 github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.0/go.mod h1:t76Ruy8AHvUAC8GfMWJMa0ElSbuIcO03NLpynfbgsPA=
+github.com/Azure/azure-sdk-for-go/sdk/azcore v1.22.0 h1:aokoqcHvaGjiM3VpjKDfMMnF/8epJ+Q1HLJ7CudztqE=
+github.com/Azure/azure-sdk-for-go/sdk/azcore v1.22.0/go.mod h1:/WYEx9pcM9Y+Dd/APJaNlSvVSvzl54rrMdZT5+Oi2LM=
 github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1 h1:Hk5QBxZQC1jb2Fwj6mpzme37xbCDdNTxU7O9eb5+LB4=
 github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1/go.mod h1:IYus9qsFobWIc2YVwe/WPjcnyCkPKtnHAqUYeebc8z0=
 github.com/Azure/azure-sdk-for-go/sdk/azidentity/cache v0.3.2 h1:yz1bePFlP5Vws5+8ez6T3HWXPmwOK7Yvq8QxDBD3SKY=
 github.com/Azure/azure-sdk-for-go/sdk/azidentity/cache v0.3.2/go.mod h1:Pa9ZNPuoNu/GztvBSKk9J1cDJW6vk/n0zLtV4mgd8N8=
 github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.2 h1:9iefClla7iYpfYWdzPCRDozdmndjTm8DXdpCzPajMgA=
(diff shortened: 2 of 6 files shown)


dakshshrivastav:
### Note on the failing Snyk check

The `security/snyk` failure here is pre-existing and not introduced by this PR.

`govulncheck` reports exactly one vulnerability, and it is identical on `develop`
and on this branch:

```
Vulnerability #1: GO-2026-6061
  Module:   google.golang.org/grpc
  Found in: google.golang.org/grpc@v1.79.3
  Fixed in: google.golang.org/grpc@v1.82.1
```

This PR does not touch gRPC. The complete module graph change is four Azure
entries, three of which are upgrades forced by the new SDK's own `go.mod`:

```
- github.com/Azure/azure-sdk-for-go/sdk/azcore                     v1.21.0  -> v1.22.0
- github.com/Azure/azure-sdk-for-go/sdk/internal                   v1.11.2  -> v1.12.0
- github.com/AzureAD/microsoft-authentication-library-for-go       v1.6.0   -> v1.7.0
+ github.com/Azure/.../costmanagement/armcostmanagement/v3                     v3.0.0
```

Nothing is downgraded and no new vulnerable package is pulled in.

The behaviour also matches the other open PRs: every currently open PR that
touches a `go.mod` fails this check, and the number of failed tests tracks the
number of changed manifests rather than anything in the diff.

| PR | manifests changed | Snyk |
|---|---|---|
| #3976 | `go.mod`, `core/go.mod` | fail — "2 tests have failed" |
| #3983 | `go.mod`, `modules/huawei-obs-plugin/go.mod` | fail — "1 test has failed" |
| this one | `go.mod` | fail — "1 test has failed" |
| #3974 | `go.mod`, `modules/collector-source/go.mod` | pass — merged before GO-2026-6061 was published |

PRs that don't touch a manifest pass trivially, which is why this only shows up
on dependency-touching changes.

Bumping gRPC to v1.82.1 would clear it, but that's unrelated to this fix and
belongs in its own PR — happy to open one if that's useful.

Also worth noting: `go.sum` here is additive only (`+8/-0`). I deliberately
avoided a bare `go mod tidy`, which would have pruned a dozen unrelated stale
entries and added noise to the diff.

dakshshrivastav:
Thanks — pushed 49749731. Three of the four were fair; I pushed back on one.

**Wrong issue number in the test comment** — fixed, should have been #3993. Leftover
from an earlier mistake on my side; I'd corrected the commit message and missed the
code comment. Grepped the tree, no other stragglers.

**`buildZip` claims to sort and doesn't** — fixed by adding the `sort.Strings(names)`
that the comment was describing, rather than deleting the comment. A deterministic
archive is genuinely worth having in a test helper.

**Line counter off by one** — fixed, but not the way suggested. `lines := 0` would
make the number a data-row ordinal rather than a file line, and it would still be
wrong by two on legacy sheets, whose preamble is a title line plus a blank line
before the header. Since the message says "line", it should mean the line you'd
land on opening the CSV in an editor.

`readPricesheetHeader` now returns how many lines it consumed and the counter
continues from there, so a warning about line 4 means line 4 on a legacy sheet and
line 2 on a current one. That also resolves the "line vs row" wording note, because
it's now genuinely a line number. Pinned by two tests, including the preamble case
that the `lines := 0` version would have got wrong.

While there: row warnings now name the CSV part they came from. A zipped sheet is
split across parts that each restart at line 1, so a bare line number wasn't
actually enough to locate a bad row — my oversight when adding multi-part support.

**`pollFrequency` 30s vs the 60s in the comment** — comment fixed, value left alone,
because the suggested change would be a no-op at best.

`Frequency` is documented in azcore as "the time to wait between polling intervals
**in absence of a Retry-After header**", and the poller implements exactly that
([`runtime/poller.go`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/poller.go)):

```go
d := cp.Frequency
if retryAfter := shared.RetryAfter(resp); retryAfter > 0 {
    d = retryAfter          // Retry-After wins outright
}
```

So when Cost Management sends `Retry-After: 60` we already wait 60s; raising the
constant to 60 wouldn't reduce polling at all, and it would double the wait in the
one case it does apply (header missing). 30s is also azcore's own default. The real
defect was my comment implying the two competed, so I reworded it to state the
actual semantics.

`go build`, `go vet`, `gofmt` and the full `go test ./...` are clean; the azure
package is at 55 passing tests.

Issue #3942, opened 2026-07-20
format-check failing on develop: configmapsource_test.go not gofmt-clean


vtushar06:
`just fmt-check` is currently failing on `develop` because `core/pkg/external/configmapsource_test.go` isn't gofmt-clean. Looks like it came in with the KCM-5392 configmap change (86698e97) — just some misaligned map literals in a couple of test funcs.

Since format-check is a required check, this now shows up red on every PR branched off develop, not only the one that added it.

Repro:

```
git checkout develop
gofmt -l core/pkg/external/configmapsource_test.go
# -> prints the file
```

Putting up a small PR that just runs gofmt on it.

cc @ameijer

PR #4002, opened 2026-08-14
fix(clustercache): include native sidecar containers in pod cost calculations

### LINKED ISSUE #3956: OpenCost omits Native Sidecar container resources from Pod cost calculations
**Describe the bug**
As described in the [Sidecar Containers  Kubernetes Documentation](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/), these sidecars are defined as init Containers. 
As far as I understand the implementation, opencost is querying the kube state metric `kube_pod_container_resource_requests` . Bu the kubernetes sidecar resources are been reported in the metric `kube_pod_init_container_resource_requests`. 
If containers sidecars are been used, the reported costs by opencost are not entirely correct.

Native Sidecars are been used in example by the istio project (istio proxy on the pod level).

**To Reproduce**
1. run a normal installation of opencost and a a standard deployment with a sidecar container (take a look on the documentation linked above) and request some resources for this sidecar
2. calculating the cost for the pod / deployment by your own and compare it against the reported value of opencost.

**Expected behavior**
I would expect that kubernetes sidecars are been included in the cost calculation (pod, deployment, waste, ...)

**Screenshots**

1.120.2 but as far as I understand, the problem is still present in the current version

**Additional context**
Add any other context about the problem here. Kubernetes versions and which public clouds you are working with are especially important.
* kubernetes 1.34

** Additional observation:**
I do not have validated it but kubernetes allows in the meantime to request resources on the pod level as well. From my perspective this is missing as well.

somaz94:
## Description

`TransformPodSpec` only copies `spec.containers` into the cluster cache, so native sidecars (init containers with `restartPolicy: Always`, such as an injected `istio-proxy`) never reach the cost model and are costed at zero. `InitContainer` does not appear anywhere in the codebase today.

This adds init containers with `restartPolicy: Always` to the cached container list, and deliberately leaves ordinary init containers out.

That split is the one Kubernetes itself makes. In `k8s.io/component-helpers/resource.PodRequests`, a restartable init container's requests are added to the pod's cumulative requests exactly like a regular container, while an ordinary init container only reaches the total through a `max()`. See [KEP-753](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/753-sidecar-containers#exposing-pod-resource-requirements). Including every init container would over-charge any pod that runs a plain one for the pod's whole lifetime; including none under-charges every pod with a sidecar.

One change at the cache level covers all three readers of `pod.Spec.Containers`:

- `costmodel.ComputeCostData`, the request figures behind allocation
- `costmodel.NewContainerMetricsFromPod`, the per-container allocation keys
- `metrics.KubePodCollector`, the `kube_pod_container_resource_requests` series OpenCost emits itself. That emitter is on by default (`EmitKsmV1Metrics` defaults to `true`), so the Prometheus-sourced path picks the sidecars up too.

Sidecars are appended after the regular containers so that index 0 stays a regular container, which `ComputeCostData` relies on when it assigns PV claims to the first container only.

Not covered here: deployments that switch OpenCost's own emitter off and scrape a real kube-state-metrics instead. There a sidecar arrives as `kube_pod_init_container_resource_requests`, which carries no `restartPolicy` label, so it cannot be separated from an ordinary init container at the metric level. That needs its own design, so I left it out rather than guess.

## Related Issues

Fixes #3956

## User Impact

Pods that run native sidecars will report higher, and now correct, costs. The size of the change depends on the sidecar: for a pod whose sidecar requests are comparable to its app container, the reported pod cost roughly doubles.

Anyone who has been reading OpenCost numbers for a mesh-injected workload has been seeing the app container only, so this will look like a cost increase in dashboards even though nothing about the workload changed. Pods without native sidecars are unaffected, and pods with ordinary init containers are unaffected.

No configuration change, no API change, no new dependency.

## Testing

Added `core/pkg/clustercache/clustercache_test.go`:

- `TestTransformPodSpecIncludesNativeSidecars`, table driven over five cases: regular containers only, ordinary init container excluded, native sidecar included and ordered after the regular containers, `restartPolicy: OnFailure` not treated as a sidecar, and a mix of ordinary init containers and two sidecars.
- `TestTransformPodSpecKeepsSidecarRequests`, so a sidecar cannot be carried into the cache and then still costed at zero.

Both were confirmed to fail without the change (removing the `restartPolicy` branch makes them fail, restoring it makes them pass).

`just` is not installed here, so I ran the justfile recipes by hand as CONTRIBUTING suggests: `go test ./...` plus `go vet ./...` in `core`, `modules/prometheus-source`, `modules/collector-source` and the root module. 88 packages pass, vet is clean everywhere. `golangci-lint run ./pkg/clustercache/...` reports 0 issues.

One note: `format-check` currently fails on `develop` for `modules/pricing/public/httpclient/httpclient.go`, which is unrelated to this PR and already unformatted upstream (same area as #3942). `gofmt -l` is clean for both files touched here, and I left that file alone.

### DIFF
--- a/core/pkg/clustercache/clustercache.go
+++ b/core/pkg/clustercache/clustercache.go
@@ -254,10 +254,28 @@ func TransformPodStatus(input v1.PodStatus) PodStatus {
 }
 
 func TransformPodSpec(input v1.PodSpec) PodSpec {
-	containers := make([]Container, len(input.Containers))
-	for i, container := range input.Containers {
-		containers[i] = TransformPodContainer(container)
+	containers := make([]Container, 0, len(input.Containers)+len(input.InitContainers))
+	for _, container := range input.Containers {
+		containers = append(containers, TransformPodContainer(container))
 	}
+
+	// Native sidecars are init containers with restartPolicy: Always. They run for the
+	// whole life of the pod, and Kubernetes sums their requests into the pod's requests
+	// exactly like a regular container, so they have to be costed like one. Ordinary init
+	// containers are deliberately left out: they are transient, and they reach the pod's
+	// requests through a max() rather than a sum.
+	//
+	// This mirrors k8s.io/component-helpers/resource.PodRequests. See KEP-753:
+	// https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/753-sidecar-containers#exposing-pod-resource-requirements
+	//
+	// Sidecars are appended after the regular containers so that index 0 stays a regular
+	// container, which ComputeCostData relies on when it assigns PV claims.
+	for _, container := range input.InitContainers {
+		if container.RestartPolicy != nil && *container.RestartPolicy == v1.ContainerRestartPolicyAlways {
+			containers = append(containers, TransformPodContainer(container))
+		}
+	}
+
 	return PodSpec{
 		NodeName:      input.NodeName,
 		Containers:    containers,
--- a/core/pkg/clustercache/clustercache_test.go
+++ b/core/pkg/clustercache/clustercache_test.go
@@ -1,0 +1,121 @@
+package clustercache
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func containerWithCPU(name, cpu string, restartPolicy *v1.ContainerRestartPolicy) v1.Container {
+	return v1.Container{
+		Name:          name,
+		RestartPolicy: restartPolicy,
+		Resources: v1.ResourceRequirements{
+			Requests: v1.ResourceList{
+				v1.ResourceCPU: resource.MustParse(cpu),
+			},
+		},
+	}
+}
+
+// Native sidecars are init containers with restartPolicy: Always. Kubernetes sums their
+// requests into the pod's requests like a regular container, so they must be carried into
+// the cache and costed. Ordinary init containers must not be, or every pod that runs one
+// would be over-charged for the whole life of the pod.
+func TestTransformPodSpecIncludesNativeSidecars(t *testing.T) {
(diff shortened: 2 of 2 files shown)


subpathdev:
> Not covered here: deployments that switch OpenCost's own emitter off and scrape a real kube-state-metrics instead. There a sidecar arrives as kube_pod_init_container_resource_requests, which carries no restartPolicy label, so it cannot be separated from an ordinary init container at the metric level. That needs its own design, so I left it out rather than guess.

You make a great point. If you look at a standard kube-state-metrics deployment, you are right that the restartPolicy is not included in the labels for kube_pod_init_container_resource_requests. Instead, it is exposed on the kube_pod_init_container_info metric. This is documented in the [kube-state-metrics pod workload metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md).

Because of this, we can actually separate sidecars from ordinary init containers at the metric level by joining these two metrics. The PromQL query would look something like this:

```promql
sum by (namespace, pod, uid, resource) (
  kube_pod_init_container_resource_requests
  * on (namespace, pod, uid, container) group_left()
  kube_pod_init_container_info{restart_policy="Always"}
)
```

somaz94:
Thanks @subpathdev, that's a useful correction. I only checked the labels on `kube_pod_init_container_resource_requests` itself and missed that `kube_pod_init_container_info` carries `restart_policy`, so the join does make sidecars separable at the metric level. My scope note was wrong to call that undecidable.

I'd still keep it out of this PR. This one is `core/pkg/clustercache` only, and the KSM path lives in `modules/prometheus-source`, which today doesn't query `kube_pod_init_container_*` at all. So it's a new query plus the plumbing to feed it into the allocation pipeline, rather than an extension of what's here.

Happy to open a follow-up with your query as the starting point if that suits.

PR #4004, opened 2026-08-14
fix(prom): treat null 'result' field as empty query result

markusheiden:
## Description
Google Managed Prometheus returns "result": null instead of an empty array for empty matrix results, e.g. for subqueries over ranges without samples:

    {"status":"success","data":{"resultType":"matrix","result":null}}

NewQueryResults rejected this with "'result' field improperly formatted ... Response: '<nil>'", which made CostModel.ComputeAllocation fail with "failed to build pod map" (after three retries per window) whenever a queried window contained no samples, even though an empty result is a perfectly valid answer.

Treat a null result field like an empty array instead.

## Related Issues
None.

## User Impact
No.

## Testing
Added a test.

## AI
This PR was created using Claude Code because I don't know Go. So I hope it meets your standards.

### DIFF
--- a/modules/prometheus-source/pkg/prom/result.go
+++ b/modules/prometheus-source/pkg/prom/result.go
@@ -102,8 +102,13 @@ func NewQueryResults(query string, queryResult interface{}, resultKeys *source.R
 	}
 	resultsData, ok := resultData.([]interface{})
 	if !ok {
-		qrs.Error = ResultFieldFormatErr(query, resultData)
-		return qrs
+		// Google Managed Prometheus returns "result": null instead of an empty array
+		// for empty matrix results, e.g. for subqueries over ranges without samples.
+		// Treat it like an empty result instead of failing.
+		if resultData != nil {
+			qrs.Error = ResultFieldFormatErr(query, resultData)
+			return qrs
+		}
 	}
 
 	// Result vectors from the query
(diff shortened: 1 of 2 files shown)


cliffcolvin (approved):
Problem gate: **real and reproducible**. The description includes the observed Google Managed Prometheus payload and the resulting OpenCost failure path. I decoded that literal JSON through `encoding/json` on current `develop`; `NewQueryResults` rejects it with the reported `result field improperly formatted ... <nil>` error. Empty query results are already a supported success case when encoded as `[]`, so treating this provider's explicit `null` representation identically is compatibility handling, not a new product concept.

The change is narrow and fails closed where it should: a missing `result` field still gets `ResultFieldDoesNotExistErr`, and every non-null value that is not an array still gets `ResultFieldFormatErr`. A decoded JSON `null` is exactly an untyped nil here, and the resulting nil slice is safe for `len` and iteration. I found no security or existing-backend compatibility regression.

- literal-JSON regression probe on `develop` — fails as described
- same probe on this branch — pass
- `go test ./pkg/prom/... -count=1` — pass
- `go test -race ./pkg/prom/... -count=1` — pass
- `go test ./... -count=1` for `modules/prometheus-source` — pass
- `go vet ./...` for `modules/prometheus-source` — pass
- `git diff --check` — pass

Non-blocking PR-description request: please replace `Testing: Added a test` with the exact command(s) you ran and their result(s), as required by the contribution guidelines. Also, `User Impact: No` contradicts the description: this prevents allocation computation from failing and retrying on empty Google Managed Prometheus windows, so please state that user-visible impact explicitly.

PR #4005, opened 2026-08-14
Correct inference counter deltas per replica before aggregating

cliffcolvin:
## What

`queryCounterDelta` in `modules/prometheus-source/pkg/prom/inference_queries.go` summed replicas at each window endpoint and only then subtracted. Summing before subtracting erases replica identity, and the counter-reset guard then applies a per-series heuristic to an aggregate.

Both endpoints now group by `pod` as well as `model_name` and `namespace`, the reset correction is applied per replica, and the per-replica deltas are folded back into the `map[string]float64` keyed `"model_name:namespace"` that callers and result types already expect. **No caller or result-type changes.**

## Why

Reported by @dev404ai in https://github.com/opencost/opencost/pull/3915#issuecomment-5256436968, with a reproduction that holds exactly.

One model, one namespace, two replicas over the same window:

| replica | start | end | reset-aware delta |
|---|---|---|---|
| A | 900 | 100 (reset) | 100 |
| B | 100 | 1,000 | 900 |
| **model total** | | | **1,000** |

Summed at each endpoint, Prometheus reports 1,000 to 1,100, so the delta is **100**. The reset guard never fires, because the aggregate delta is positive. A 10x undercount, silently.

The same erasure also **overcounts**. With A 900 to 100 and B 100 to 150, the aggregate reads 1,000 to 250, the delta is negative, the guard *does* fire, and it returns the aggregate end value of 250 against a true 150. The error is unbounded in sign.

This affects all five vLLM cost counters, since they share the helper: `vllm:prompt_tokens_total`, `vllm:generation_tokens_total`, `vllm:request_prefill_time_seconds_sum`, `vllm:request_time_per_output_token_seconds_sum`, `vllm:prefix_cache_hits_total`. These are the token and time denominators for inference cost, so the error moves billing-visible totals and can skew derived cost per million tokens in either direction.

## When `pod` is not available

`pod` is never self-reported by the serving engine. It comes from Kubernetes service discovery relabeling, so it is legitimately absent under:

- non-Kubernetes service discovery (`static_configs`, DNS, EC2, Consul), where `namespace` can still be a static target label
- `write_relabel_configs` or `metric_relabel_configs` `labeldrop` that strip it for cardinality
- federation of pre-aggregated recording rules such as `sum by (model_name, namespace) (...)`

Those series collapse into a single per-`(model_name, namespace)` bucket, which **reproduces the previous behaviour exactly** rather than dropping the data, and a warning names the metric so operators know they are on the less accurate path. Deployments on the standard Kubernetes scrape config get the corrected number with no configuration change.

## Testing

`modules/prometheus-source/pkg/prom/inference_counter_delta_test.go`, driving the real helper through a scripted Prometheus client:

- `TestQueryCounterDelta_ReplicaResetIsNotMaskedByAnotherReplica` (the reported undercount)
- `TestQueryCounterDelta_ReplicaResetIsNotOvercounted` (the overcount direction)
- `TestQueryCounterDelta_SeriesWithoutPodLabelStillCount` (the degraded path, pinned to the exact number every deployment returned before this change)
- `TestQueryCounterDelta_GroupsByPod` (query shape, since the correction is only possible if both endpoints carry `pod`)

All three value tests fail against the previous implementation. `go build ./...` and `go test ./...` pass on `prometheus-source`.

## Scope

Split out of #3915 deliberately. The bug is in the merged AI inference cost path from #3845 and affects `develop` users today, independent of the saturation work, so it deserves its own review and is cherry-pickable on its own.

Two adjacent items intentionally **not** included here, to keep this reviewable:

- **The start-lookback asymmetry.** The end query uses a full-window lookback while the start query uses a fixed `[2m]`. It is a no-op under normal 30s scraping, but a replica whose scrape gapped for more than two minutes around the window start contributes zero at start and its full cumulative value at end. Happy to fold in if reviewers prefer.
- **The collector source.** `newInferenceCostCounterCollector` groups the same five counters by `(model_name, namespace)` and `increaseAggregator` sums across replicas before comparing cycle totals. Bounded per scrape cycle rather than per window, so a smaller blast radius, but the same class of bug. Worth a follow-up.

### DIFF
--- a/modules/prometheus-source/pkg/prom/inference_counter_delta_test.go
+++ b/modules/prometheus-source/pkg/prom/inference_counter_delta_test.go
@@ -1,0 +1,185 @@
+package prom
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"testing"
+	"time"
+)
+
+// counterDeltaClient dispatches canned vector responses keyed by a substring of
+// the PromQL query. The end-of-window query carries the full window as its
+// lookback and the start-of-window query a fixed 2m one, so "[2m] @" selects
+// the start endpoint and "[60m] @" the end endpoint.
+type counterDeltaClient struct {
+	responses map[string]string
+	queries   []string
+}
+
+func (c *counterDeltaClient) URL(ep string, args map[string]string) *url.URL {
+	u, _ := url.Parse("http://fake-prometheus:9090" + ep)
+	return u
+}
+
+func (c *counterDeltaClient) Do(_ context.Context, req *http.Request) (*http.Response, []byte, error) {
+	query := req.URL.Query().Get("query")
+	c.queries = append(c.queries, query)
+
+	result := "[]"
+	for substr, res := range c.responses {
+		if strings.Contains(query, substr) {
+			result = res
+			break
+		}
+	}
+	body := fmt.Sprintf(`{"status":"success","data":{"resultType":"vector","result":%s}}`, result)
+	return &http.Response{StatusCode: http.StatusOK}, []byte(body), nil
+}
+
+func podSample(modelName, namespace, pod string, value float64) string {
+	return fmt.Sprintf(`{"metric":{"model_name":"%s","namespace":"%s","pod":"%s"},"value":[1712000000,"%g"]}`,
+		modelName, namespace, pod, value)
+}
+
+func podlessSample(modelName, namespace string, value float64) string {
+	return fmt.Sprintf(`{"metric":{"model_name":"%s","namespace":"%s"},"value":[1712000000,"%g"]}`,
+		modelName, namespace, value)
+}
+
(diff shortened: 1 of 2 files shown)


peatey:
just a heads up on this cluster_id is missing from the group-by, so the same bug survives one level up. 

inference_queries.go:246,255

Every other query in this module threads cfg.ClusterLabel / cfg.ClusterFilter (see metricsquerier.go); the inference queries thread neither — pre-existing from #3845, but it now matters more. Against a multi-cluster Prometheus or Thanos, sum by (model_name, namespace, pod) merges vllm-0 in llm-d on cluster A with vllm-0 in llm-d on cluster B into one "replica" bucket. If A resets while B grows, you get back exactly the masking this PR fixes. Adding ctx.config.ClusterLabel to both by-clauses and to the internal …|pod key leaves the output key — and the caller contract — untouched.

PR #4007, opened 2026-08-14
fix(storage): normalize memory storage paths on write

cliffcolvin:
## What

`MemoryStorage` cannot read back a file it just wrote unless the caller's path is already in cleaned host form. On Windows that is every path, because storage paths are built with `path.Join`.

## Why

`MemoryStorage` keys files twice: in the directory tree, and in the `directPaths` map. Every reader (`Read`, `ReadStream`, `Stat`, `Exists`, `Remove`, `ReadToLocalFile`) normalized its key with `filepath.Clean`, but `Write` and `WriteStream` stored the caller's path verbatim.

```go
// Write
ms.directPaths[path] = f        // "test/test/collector/20260813120000.bingen"

// Read
path = filepath.Clean(path)     // "test\test\collector\20260813120000.bingen" on Windows
_, ok := ms.directPaths[path]   // miss -> DoesNotExistError
```

The tree half of the store normalizes on both sides, so `List` kept working while every read failed. On a clean checkout of `develop` on Windows, the collector WAL restore fails:

```
--- FAIL: TestWalinator_restore (0.00s)
    failed to load file contents for 'test/test/collector/20260813120000.bingen': file does not exist
    walinator_test.go:188: failed to get collector from repo2: failed to find MetricCollector for interval '1d' for time '2026-08-14T00:00:00Z'
```

`Walinator.restore` lists its three WAL files, fails to read all three, and the restored repository ends up empty. `TestWalinator_clean` passes only because `Remove` works through the tree, which also leaves the stale `directPaths` entry behind.

This is not Windows only. Since `Write` never normalized at all, `Write("./a/b.json")` followed by `Read("./a/b.json")` fails on any platform. CI runs on `ubuntu-latest`, where canonical paths hide the common case.

## What changed

Memory storage paths are logical keys, like bucket object keys, so they are now normalized to cleaned, slash separated form on every platform, on the write side as well as the read side.

- `memfile`: added `Normalize`, `Join` and `Separator`, and moved `Split`/`SplitPaths` from `filepath` to `path` semantics. This is a no-op on Linux, where `filepath.ToSlash` is the identity and `path.Clean` matches `filepath.Clean`.
- `MemoryStorage`: `Write` and `WriteStream` now share a `write` helper that normalizes before storing, readers use `memfile.Normalize`, and `ListDirectories` returns slash joined names instead of `opencost\storage\dir0/`.
- `core/pkg/pricing`: an embedded fixture was read through `filepath.Join`. `embed.FS` paths are always slash separated, so that read never resolved on Windows. Same class of bug, one line.

## Testing

`TestMemoryStorage_PathNormalization` covers the equivalence class of a path (canonical, `./` prefixed, doubled separator, `..` segment). Every form must read, stat and exist after a write in any other form, and `Remove` must clear both indexes. It fails on `develop` (4 of 4 subtests on Windows, 3 of 4 on Linux) and passes with this change.

On Windows, `modules/collector-source` is now fully green, `core/pkg/pricing` passes, and `core/pkg/storage` has only the pre-existing `FileStorage` failures described below.

## Known remaining Windows gaps, not addressed here

- `core/pkg/util/fileutil/locks_windows.go` stubs every locked read and write with "not implemented on Windows. Please open an issue.". `FileStorage` is built on it, so that single gap accounts for the `core/pkg/util/fileutil` failures, the `FileStorage` failures in `core/pkg/storage`, and the CSV pricing tests in `pkg/cloud/provider`. Closing it means real `LockFileEx` support, which is a feature rather than a fix.
- `TestFileStorageListDirectoriesSymlink` needs symlink privilege (elevation or developer mode).
- `TestCacheGroupMaxRollOff` in `core/pkg/util/cache` is timing sensitive and fails independently of these changes.

A `windows-latest` job on the unit test workflow would catch this class of bug, but it cannot go green until the fileutil locking gap is closed.

### DIFF
--- a/core/pkg/pricing/mock.go
+++ b/core/pkg/pricing/mock.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -113,8 +114,8 @@ func (mpm *MockPricingModule) Checksum(ctx context.Context) (string, error) {
 var pricingTestFS embed.FS
 
 func (mpm *MockPricingModule) loadTestFile(filename string) error {
-	path := filepath.Join("test", filename)
-	bs, err := pricingTestFS.ReadFile(path)
+	// embedded file paths are always slash separated, so they cannot be built with filepath
+	bs, err := pricingTestFS.ReadFile(path.Join("test", filename))
 	if err != nil {
 		return fmt.Errorf("failed to read embedded pricing file: %w", err)
 	}
--- a/core/pkg/storage/memfile/util.go
+++ b/core/pkg/storage/memfile/util.go
@@ -2,26 +2,43 @@ package memfile
 
 import (
 	"fmt"
+	"path"
 	"path/filepath"
 	"strings"
 )
 
+// Separator is the separator used within memory storage paths. Memory storage paths are logical
+// keys, like the object keys used by the bucket storages, so they are separated by '/' on every
+// platform rather than by the host file system separator.
+const Separator = "/"
+
+// Normalize returns the cleaned, slash separated form of the provided path. It is the canonical
+// form used to key a file within memory storage, so every path entering memory storage must pass
+// through it.
+func Normalize(p string) string {
+	return path.Clean(filepath.ToSlash(p))
+}
+
+// Join combines the provided path elements into a single memory storage path.
+func Join(elem ...string) string {
+	return path.Join(elem...)
+}
+
 // SplitPaths splits the directory path into a slice of directory names.
-func SplitPaths(path string) []string {
-	path = filepath.Clean(path)
-	if path[len(path)-1] == filepath.Separator {
-		path = path[:len(path)-1]
-	}
-	return strings.Split(path, string(filepath.Separator))
+func SplitPaths(p string) []string {
+	p = Normalize(p)
+	p = strings.TrimSuffix(p, Separator)
+
+	return strings.Split(p, Separator)
 }
 
 // Split splits the path into a slice of directory names and the file name.
-func Split(path string) ([]string, string) {
-	path = filepath.Clean(path)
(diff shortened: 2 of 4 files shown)


thomasvn (approved):
Approving. Existing memorystorage unit tests passed.

cliffcolvin (commented):
Problem gate: **real and reproducible**. Although there is no linked issue, this is not a hypothetical portability cleanup: on current `develop` I independently reproduced `Write("./a/b.json")` followed immediately by `Read("./a/b.json")` returning `file does not exist`. The Windows failure follows directly from mixing slash-built logical storage keys with host-`filepath.Clean` lookup keys, and the described collector WAL impact exercises a real production consumer.

The implementation matches the abstraction: `storage.DirDelim` and shared storage tests already define `/` object-key semantics, callers build keys with `path.Join`, and `embed.FS` likewise requires slash paths. Normalizing before both tree insertion and direct-map insertion removes the split-brain behavior; applying the same normalization to every lookup/removal preserves equivalent-path behavior and prevents stale `directPaths` entries. I did not find a security, data-loss, or compatibility blocker in the changed paths.

- pre-fix regression probe on `develop` — fails as described
- `go test ./pkg/storage/... ./pkg/pricing -count=1` — pass
- same packages with `-race` — pass
- `go vet ./pkg/storage/... ./pkg/pricing` — pass
- collector `TestWalinator_restore`/`TestWalinator_clean` — pass
- same collector tests with `-race`, plus vet — pass
- Windows/amd64 cross-compilation of affected core and collector packages — pass
- `git diff --check` — pass

I also audited the red checks: format-check names only untouched `modules/pricing/public/httpclient/httpclient.go`; integration failed on an HTTP connection reset and live metric/cost drift, neither of which can be caused by this in-memory path patch. No blockers found. I cannot submit an approval because the authenticated GitHub account is the PR author; the existing maintainer approval is consistent with this review.

PR #4009, opened 2026-08-18
fix: use actual usage in MCP efficiency recommendations

### LINKED ISSUE #3984: MCP `get_efficiency` reports allocated resources as usage, so efficiency is always 100 % and recommendations are always +20 %
**Version:** OpenCost 1.120.3 (`opencost:1.120.3`), MCP server enabled (`MCP_SERVER_ENABLED=true`, `MCP_HTTP_PORT=8081`), Prometheus-compatible backend (VictoriaMetrics).

## Summary

The MCP tool `get_efficiency` populates its `cpuCoresUsed` / `ramBytesUsed` fields with **allocated** quantities rather than measured usage. Because OpenCost allocates `max(request, usage)`, those values are identical to the request for any workload whose request exceeds its usage — i.e. nearly all of them. The reported efficiency is therefore always exactly `1.0`, and the derived recommendation is always `request × bufferMultiplier` (default `1.2`), i.e. a 20 % **increase** presented as a right-sizing suggestion with `costSavingsPercent: -20`.

The REST `/allocation` endpoint on the same process, over the same window, returns the correct usage — so this is not a metrics-pipeline or configuration problem.

## Reproduction

Same pod, same 24 h window, namespace `apps`:

`tools/call get_efficiency {"window":"24h","aggregate":"namespace","filter":"namespace:\"apps\""}`

```json
{
  "name": "apps",
  "cpuEfficiency": 1.0000000000000095,
  "memoryEfficiency": 1.0000000000000009,
  "cpuCoresRequested": 3.0432631495818496,
  "cpuCoresUsed":      3.0432631495818785,
  "ramBytesRequested": 11074386021.743427,
  "ramBytesUsed":      11074386021.743437,
  "recommendedCpuRequest": 3.651915779498254,
  "costSavingsPercent": -20.00000000000085
}
```

`GET /allocation?window=24h&aggregate=namespace&accumulate=true` — same namespace, same process:

```
cpuCoreUsageAverage   0.6722      cpuCoreRequestAverage   3.0426     ->  22.1 %
ramByteUsageAverage   9.43e9      ramByteRequestAverage   1.107e10   ->  85.2 %
```

Cross-check straight against the metrics backend agrees with `/allocation`: 0.6853 cores used vs 3.1449 requested (21.8 %).

Note `ramBytesUsed` from the tool (`11074386021`) equals `ramByteRequestAverage` from `/allocation` exactly — the used field carries the request value, it is not a zero/absent fallback.

## Cause

`pkg/mcp/server.go`, `computeEfficiencyMetric`:

```go
// Get current usage (average over the period)
cpuCoresUsed := alloc.CPUCoreHours / hours
ramBytesUsed := alloc.RAMByteHours / hours
```

`CPUCoreHours` / `RAMByteHours` are allocated resource-hours. `Allocation.CPUCores()` in `core/pkg/opencost/allocation.go` is this same expression and is documented as converting `CPUCoreHours` into average cores — allocation, not usage.

The model already exposes the measured values and the correct ratio:

```go
// Allocation.CPUEfficiency()
return a.CPUCoreUsageAverage / a.CPUCoreRequestAverage
```

## Suggested fix

Use the usage fields (or delegate to the existing helpers):

```go
cpuCoresUsed := alloc.CPUCoreUsageAverage
ramBytesUsed := alloc.RAMBytesUsageAverage
```

which also makes `cpuEfficiency` agree with `Allocation.CPUEfficiency()` and turns the recommendation back into `usage × buffer`.

## Impact

Any agent using this tool for right-sizing is told to increase every request by the buffer percentage, with numbers that look computed. In our estate the correct direction was the opposite: the namespace above runs at ~22 % CPU, so it is a candidate for a cut. We have banned the tool internally and kept right-sizing on measured usage.

`get_efficiency` also does not appear in the MCP integration docs (which list `get_allocation_costs`, `get_asset_costs`, `get_cloud_costs`) — if it is not considered ready, saying so in the docs would help; it is exposed by default when the MCP server is enabled.

AAH20:
## Summary
- use `CPUCoreUsageAverage` and `RAMBytesUsageAverage` instead of deriving "usage" from allocated resource-hours
- switch current efficiency calculations to OpenCost's canonical `CPUEfficiency()` and `RAMEfficiency()` helpers
- add regression coverage showing over-provisioned workloads no longer appear ~100% efficient

## Why
`get_efficiency` was using `CPUCoreHours / hours` and `RAMByteHours / hours` as if they were usage metrics. Those fields represent allocated resources (`max(request, usage)` over time), so heavily over-provisioned workloads could look fully efficient and receive bad right-sizing guidance.

This change makes MCP efficiency recommendations reflect real utilization, so AI agents can recommend cost reductions instead of accidental spend increases.

Closes #3984.

## Test plan
- `go test ./pkg/mcp/...`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/mcp/server.go
+++ b/pkg/mcp/server.go
@@ -1254,23 +1254,25 @@ func computeEfficiencyMetric(alloc *opencost.Allocation, bufferMultiplier float6
 		return nil
 	}
 
-	// Calculate time duration in hours
-	hours := alloc.Minutes() / 60.0
-	if hours <= 0 {
+	if alloc.Minutes() <= 0 {
 		return nil
 	}
 
-	// Get current usage (average over the period)
-	cpuCoresUsed := alloc.CPUCoreHours / hours
-	ramBytesUsed := alloc.RAMByteHours / hours
+	hours := alloc.Minutes() / 60.0
 
-	// Get requested amounts
+	// Use the actual usage averages, NOT CPUCoreHours/hours which represents
+	// allocated resources (max of request and usage) and makes efficiency
+	// appear ~100% even for grossly over-provisioned workloads.
+	cpuCoresUsed := alloc.CPUCoreUsageAverage
+	ramBytesUsed := alloc.RAMBytesUsageAverage
+
 	cpuCoresRequested := alloc.CPUCoreRequestAverage
 	ramBytesRequested := alloc.RAMBytesRequestAverage
 
-	// Calculate current efficiency (will be 0 if no requests are set)
-	cpuEfficiency := safeDiv(cpuCoresUsed, cpuCoresRequested)
-	memoryEfficiency := safeDiv(ramBytesUsed, ramBytesRequested)
+	// Use the canonical efficiency helpers which handle the no-request edge
+	// cases (usage-without-request → 100%, zero-usage-zero-cost → 0%).
+	cpuEfficiency := alloc.CPUEfficiency()
+	memoryEfficiency := alloc.RAMEfficiency()
 
 	// Calculate recommendations with buffer for headroom
 	recommendedCPU := cpuCoresUsed * bufferMultiplier
--- a/pkg/mcp/server_test.go
+++ b/pkg/mcp/server_test.go
@@ -1057,11 +1057,13 @@ func TestComputeEfficiencyMetric_ValidAllocation(t *testing.T) {
 		Name:  "test-pod",
 		Start: now.Add(-24 * time.Hour),
 		End:   now,
-		// 24 hours = 1440 minutes
-		CPUCoreHours:           24.0,   // 1 core for 24 hours
-		RAMByteHours:           24.0e9, // ~1GB for 24 hours
-		CPUCoreRequestAverage:  2.0,    // Requested 2 cores
-		RAMBytesRequestAverage: 2.0e9,  // Requested 2GB
+		// Usage averages (the actual resource consumption, NOT allocated)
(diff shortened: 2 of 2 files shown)


cliffcolvin (approved):
Problem gate: **real and independently demonstrated**. Issue #3984 was filed by a different user against OpenCost 1.120.3 with a concrete MCP response, the matching `/allocation` response, a raw-metrics cross-check, and operational impact. The current implementation does indeed label allocated `max(request, usage)` resource-hours as usage, which makes over-provisioned workloads appear ~100% efficient and can turn a right-sizing response into a request increase.

The fix is at the correct layer and follows existing OpenCost semantics: `CPUCoreUsageAverage`/`RAMBytesUsageAverage` are the measured, time-weighted fields already returned by allocations, and `CPUEfficiency()`/`RAMEfficiency()` are the canonical request edge-case implementations. Aggregated allocations preserve these averages by duration weighting. The regression fixture deliberately makes allocated hours differ from measured usage, so it would fail under the old implementation rather than merely restating the new code. I found no security or compatibility regression in the changed path.

- `go test ./pkg/mcp/... -count=1` — pass
- `go test -race ./pkg/mcp/... -count=1` — pass
- `go vet ./pkg/mcp/...` — pass
- `git diff --check` — pass

The PR description names the production failure, links the independent report, explains the field semantics, and supplies the relevant test command.

PR #4012, opened 2026-08-18
fix(azure): changed storage cleanup to delete files older than two days

TrevorKWalker:
## Description
Changed function `deleteFilesOlderThan7d` in opencost/pkg/cloud/azure/storageconnection.go to be ~~`deleteFilesOlderThan2d`~~ `deleteFilesOlderThanRetention` and updated external references to the function ( only one in [storagebillingparser.go](https://github.com/opencost/opencost/blob/30608b82c2aadf797bff39820ef8aa8bee632d7a/pkg/cloud/azure/storagebillingparser.go#L120)) to reflect the updated function name. 

Reduce the duration that we keep unneeded azure cloud billing files in the PVC to reduce storage size requirements. 

In environments with large billing integrations the usage will slowly creep up throughout the month and due to keeping 7 days of files, at the end of the month the PVC size requirement is effectively 7 * the size of the a full month billing export even though only the most recent file is utilized. Reducing this to 2 will reduce total PVC requirement by 3.5x. 

Edit: 
Changed the retention length to be configured by a EnvVar `CLOUD_COST_PV_RETENTION` with a default of 2 days.

## Related Issues
Relates to Kubecost issues and customer Feedback.

## User Impact
Users will see a decrease in the cloud costs PVC usage if using Azure billing integrations and are downloading files instead of streaming.

## Testing

Unit tests "go test ./..."


thomasvn (approved):

peatey:
just on this y'all - this is hard coded and is changing behavior -
1. I would like an env var setting for this
2. The actual garbage is superseded files: Azure writes a new month-to-date blob (new GUID name) each day, and getMostRecentBlobs only ever uses the newest per month. Old daily files are dead the moment a newer one lands, yet the age heuristic keeps them for 2 more days so the PVC still needs ~2-3× a full-month export at month end

TrevorKWalker:
@peatey , I will work on set it up as an Env Var. 
As for the second point you made, would an Env Var be sufficient to address this root cause or would we instead want to rewrite the function (or write a new function) that deletes the oldest files until there is only the newest one left?
If we had the Env Var and set it to 1d of retention it would effectively do the same in my mind in terms of peak PVC usage. It seems to me even if we wanted to keep only one file we would need room for two because we wouldn't want to delete the current file before the new file was successfully ingested meaning we would need capacity to fit two of the files in the PVC.

pkg/cloud/azure/storageconnection.go
```diff
@@ -133,13 +134,13 @@ func (sc *StorageConnection) DownloadBlobToFile(localFilePath string, blob conta
 	return nil
 }
 
-// deleteFilesOlderThan7d recursively walks the directory specified and deletes
-// files which have not been modified in the last 7 days. Returns a list of
+// deleteFilesOlderThan2d recursively walks the directory specified and deletes
+// files which have not been modified in the last 2 days. Returns a list of
 // files deleted.
-func (sc *StorageConnection) deleteFilesOlderThan7d(localPath string) ([]string, error) {
+func (sc *StorageConnection) deleteFilesOlderThan2d(localPath string) ([]string, error) {
```
thomasvn: We should probably rename this function so that it no longer explicitly calls out `2d`? Now that this is determined via env var?

pkg/cloud/azure/storageconnection.go
```diff
@@ -133,13 +134,13 @@ func (sc *StorageConnection) DownloadBlobToFile(localFilePath string, blob conta
...
 	sc.lock.Lock()
 	defer sc.lock.Unlock()
-	duration := 7 * 24 * time.Hour
+	duration := time.Duration(env.GetCloudCostPvcRetention()) * 24 * time.Hour
```
thomasvn: More accurately, it should be `env.GetCloudCostPvcRetention()`. This is the retention of the billing files on the PersistentVolume, not the PersistentVolumeClaim

pkg/cloud/azure/storageconnection.go
TrevorKWalker: Agreed. Will get to this in the morning.

pkg/cloud/azure/storageconnection.go
TrevorKWalker: I changed it to `env.GetCloudCostPvRetention()`, I believe this is what you meant it should be going off the second sentence since it already was `env.GetCloudCostPvcRetention()`.

pkg/cloud/azure/storageconnection.go
TrevorKWalker: updated to deleteFilesOlderThanRetention

pkg/cloud/azure/storagebillingparser.go
```diff
@@ -117,7 +117,7 @@ func (asbp *AzureStorageBillingParser) ParseBillingData(start, end time.Time, re
 	if env.IsAzureDownloadBillingDataToDisk() {
 		// clean up old files that have been saved to disk before downloading new ones
 		localPath := env.GetAzureDownloadBillingDataPath()
-		if _, err := asbp.deleteFilesOlderThan7d(localPath); err != nil {
+		if _, err := asbp.deleteFilesOlderThan2d(localPath); err != nil {
```
thomasvn: I think there are a few locations where you need to update to use the new function name? `deleteFilesOlderThanRetention`

TrevorKWalker: I didnt save the file in VS code so it wasn't included in the commit. Fixed now!

Sean-Holcomb (approved):
LGTM

PR #4013, opened 2026-08-18
Pricing Module Updates (dedup) & Update of USD/CNY data

alexsouthard:
## Description
Update AWS/Azure to dedup data. Previously it wasn't considering how multiple line items could show up w/ similar attributes based on stuff like OS or Spot pricing, and this can be seen in the data as identical lines that have different prices for them, which is obviously wrong. So it updates that and also reruns to fix nodes for AWS/Azure USD/CNY (GCP is unaffected by this)

## Related Issues
No related issue, the whole pricing premise is that it should be updated regularly and this should kick of the 1.0.0 action in the repo

## User Impact
Not a breaking change and doesn't affect anything right now since kubemodel isn't out

## Testing
Not really testable

### DIFF
--- a/modules/pricing/public/azure/azurepricingsource.go
+++ b/modules/pricing/public/azure/azurepricingsource.go
@@ -51,6 +51,7 @@ func (a *AzurePricingSource) GetPricing() (*pricing.PricingSet, error) {
 	// Fetch VM pricing
 	url := a.buildVMURL()
 	pageCount := 0
+	seenNodes := make(map[nodeKey]struct{})
 
 	for url != "" {
 		resp, err := azureHTTPClient.Get(url)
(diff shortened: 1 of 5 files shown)


modules/pricing/public/azure/azurepricingsource.go
```diff
@@ -232,8 +239,13 @@ func (a *AzurePricingSource) includeItem(item AzurePricingAttributes) bool {
 		return false
 	}
 
+	// The Azure API appends an exact suffix to SkuName for non-on-demand rows.
+	// We only want on-demand Linux pricing, so reject Spot and Low Priority.
```
alexsouthard, line 243: Leaving this as true for now (in that it doesn't do spot pricing). The point of this PR is more to update/fix the pricing release than add spot support

Sean-Holcomb (approved):

PR #4014, opened 2026-08-20
Update device and include in bingen

Sean-Holcomb:
## Description
  Redesigns and completes the DCGM/GPU data path in `kubemodel`:

  - **GPU struct redesign**: replaces the nested `DCGMDevice → DCGMPod → DCGMContainer` model with a flat `Device` (identity: UUID/model/uptime) plus `DeviceUsage` living
  directly on `Container.DeviceUsages` (keyed by GPU UUID) — mirroring how CPU/RAM usage already lives on `Container`. Wires the corresponding `computeDevices` step into
  the active kubemodel compute pipeline (it was previously implemented but never enabled).
  - **`DaemonSet.Arguments`**: adds a generic, vendor-agnostic field capturing `--key=value` container arguments on a DaemonSet, backed by a new `daemonset_arguments`
  metric (one row per parsed arg, values kept as label *values* rather than dynamic label names to avoid Prometheus label-name sanitization issues). Emitted by both the
  OpenCost-native Prometheus-facing collector (`pkg/metrics/kubemodel.go`) and the collector-source scraper, so it's populated under either data source.
  - **Collector-source DCGM `pod_uid` fix**: found and fixed a real bug where DCGM scrapes under collector-source never carried a `pod_uid` label (only namespace/pod by 
  name), causing `computeDevices` to silently drop all container-level GPU usage. Adds a small, reusable `UpdateEnricher` hook on `TargetScraper` and uses it to backfill 
  `pod_uid` on DCGM scrapes from a clustercache-derived pod index.

  ## Related Issues
  

  ## User Impact
  - No breaking changes to existing API responses. The kubemodel binary (bingen) format gains new fields (`Device`, `Container.DeviceUsages`, `DaemonSet.Arguments`); 
  serialization is additive/versioned so it remains backward compatible.
  - Collector-source deployments scraping DCGM will now correctly attribute GPU usage to containers — previously this was silently dropped due to the missing `pod_uid`.
  - Deployments using either Prometheus or collector-source now get a `daemonset_arguments` metric alongside the existing `daemonset_info`/`labels`/`annotations` metrics.

  ## Testing
  - Added/updated unit tests across `core/pkg/model/kubemodel`, `core/pkg/compute/kubemodel`, `core/pkg/source`, `modules/collector-source` (scrape + collector), and 
  `modules/prometheus-source`, covering the new `Device`/`DeviceUsage` model, `DaemonSet.Arguments` end-to-end (metric emission → query → compute), and the new 
  `TargetScraper` enrichment hook (generic case + DCGM-specific `pod_uid` resolution).
  - `go build ./...`, `go vet ./...`, and `go test ./...` pass clean in `core`, `modules/collector-source`, `modules/prometheus-source`, and the root module.
  - Regenerated bingen codecs (`go generate ./core/pkg/model/kubemodel/...`) and confirmed clean compilation plus round-trip equality via the existing `KubeModelSetEquals` 
  test helpers.


core/pkg/model/kubemodel/container.go
```diff
@@ -56,3 +62,11 @@ func (kms *KubeModelSet) RegisterContainer(container *Container) error {
 
 	return nil
 }
+
+func (c *Container) GetKey() string {
+	return ContainerKey(c.PodUID, c.Name)
+}
+
+func ContainerKey(podUID, conatinerName string) string {
```
bbylovas: typos

modules/collector-source/pkg/scrape/dcgm.go
```diff
@@ -16,18 +17,40 @@ var dcgmRegex = regexp.MustCompile("(?i)(.*dcgm-exporter.*)")
...
+		namespace, pod := update.Labels[source.NamespaceLabel], update.Labels[source.PodLabel]
+		if namespace == "" || pod == "" {
+			return update
+		}
+
+		if uid, ok := index[podKey{namespace: namespace, name: pod}]; ok {
+			update.Labels[source.PodUIDLabel] = string(uid)
+		}
+		return update
+	}
```
mbolt35, line 54: any issue with statefulset pods here? I suppose since it's a singular scrape and we update the uid per cycle, it's probably fine yeah?

modules/collector-source/pkg/scrape/index.go
```diff
@@ -46,3 +46,18 @@ func buildPVIndex(pvs []*clustercache.PersistentVolume) map[string]types.UID {
...
+}
+
+// buildPodIndex returns a map from (namespace, name) to Pod UID.
+func buildPodIndex(pods []*clustercache.Pod) map[podKey]types.UID {
+	m := make(map[podKey]types.UID, len(pods))
+	for _, pod := range pods {
+		m[podKey{namespace: pod.Namespace, name: pod.Name}] = pod.UID
+	}
+	return m
+}
```
mbolt35, line 63: statefulset pods creating issues here?

modules/collector-source/pkg/scrape/targetscraper.go
```diff
@@ -29,6 +34,7 @@ func newTargetScrapper(name string, provider target.TargetProvider, metricNames
 		targetProvider: provider,
 		metricNames:    metricSet,
 		includeMetrics: includeMetrics,
+		enrich:         enrich,
```
mbolt35, line 37: the enrich stuff is super clever. definitely a cool addition.

pkg/metrics/kubemodel.go
```diff
@@ -378,6 +382,18 @@ func (c KubeModelCollector) scrapeDaemonSets(
...
+			for _, arg := range slices.Sorted(maps.Keys(daemonSetArguments)) {
+				out = append(out, newInfoMetric("daemonset_arguments", map[string]string{
+					"uid":           string(ds.UID),
+					"namespace_uid": string(nsUID),
+					"daemonset":     ds.Name,
+					"arg":           arg,
+					"value":         daemonSetArguments[arg],
+				}))
+			}
+		}
```
mbolt35, line 396: So, we do have something like this in cluster cache where we scrape args for GPU Allocation. Not sure if you want to try to reuse the `coreutils` helper here: 

https://github.com/opencost/opencost/blob/e977d842ed2ae977726f29637061df0bac352ac6/modules/collector-source/pkg/scrape/clustercache.go#L1469-L1492

Up to you

modules/collector-source/pkg/scrape/index.go
mbolt35: See above - I think because the namespace/pod name mapping updates each cycle, it always has the correct UID.

pkg/metrics/kubemodel.go
Sean-Holcomb: I copied this function over to `pkg/db/internal/kubemodel/helper.go` in https://github.com/kubecost/kubecost-cost-model/pull/4421. The helper will move back to opencost when we finish building out the kubemodel pipeline there.

The core of the choice to move this over is that for me it crosses a level of business logic that I don't want to be a part of KubeModel, which I view as a recording data structure. Since we will need to support device plugins in general, I don't know how likely it will be that we need additional information out of the daemonsets for other vendors, but I do want to leave the door open to it as much as possible.

modules/collector-source/pkg/scrape/dcgm.go
Sean-Holcomb: Exactly, The best case scenario would be do have the Pod_uid just be attached. but because this a single point in time and the index gets rebuilt each scrape, it should not be possible for there to be collisions

mbolt35 (approved):
Looks good to me!

modules/collector-source/pkg/scrape/dcgm.go
```diff
@@ -16,18 +17,40 @@ var dcgmRegex = regexp.MustCompile("(?i)(.*dcgm-exporter.*)")
...
-		true)
+		true,
+		enrich)
+}
+
+// podUIDEnricher backfills pod_uid on a DCGM update using its own namespace/pod name labels,
+// resolved against a freshly built index of the cluster's current pods. Left unset if pod_uid
+// is already present, or if namespace/pod can't be resolved to a known pod.
+func podUIDEnricher(clusterCache clustercache.ClusterCache) UpdateEnricher {
+	index := buildPodIndex(clusterCache.GetAllPods())
```
thomasvn: I agree with this finding!

## 1. HIGH — Pod index is built once at process startup and never refreshed

```go
func podUIDEnricher(clusterCache clustercache.ClusterCache) UpdateEnricher {
	index := buildPodIndex(clusterCache.GetAllPods())   // evaluated once
	return func(update metric.Update) metric.Update { ... }
}
```

`podUIDEnricher` is called from `newDCGMScrapper` → `NewScrapeController` (`scrapecontroller.go:106`), which runs exactly once from `modules/collector-source/pkg/collector/datasource.go:99`. The closure captures that snapshot for the lifetime of the process, while `Scrape()` runs every scrape interval forever.

Consequences:

- Any GPU pod created after startup never gets `pod_uid` backfilled, so `computeContainers` drops its DCGM usage at `container.go:169` (`res.PodUID == ""`) and the container shows zero device usage permanently.
- A StatefulSet pod recreated under the same namespace/name resolves to the **dead** pod's UID, attributing GPU usage to a container that no longer exists.
- If the cluster-cache informers haven't synced when `NewScrapeController` runs, the index is empty forever.

The doc comment directly above says "resolved against a **freshly built** index of the cluster's current pods," so per-scrape was clearly the intent. Every other index in this package (`buildNamespaceIndex`, `buildPVIndex`, `buildPVCIndex`) is rebuilt inside `ClusterCacheScraper.Scrape()` (`clustercache.go:58-61`).

**Fix:** move `buildPodIndex(clusterCache.GetAllPods())` inside the returned closure, or rebuild it once per `Scrape()` call.

Sean-Holcomb: resolved by changing to a bulk enrich and building the index at the start of that. Triggered once per scrape

PR #4015, opened 2026-08-21
Add Spot to Public Pricing + Pipeline Error Fixes

alexsouthard:
## Description
Adds spot pricing to public pricing! This is quite a simple change for Azure/GCP as it just changes the filter of what it was sorting out.

However, for AWS it adds back in the spot API caller which runs on all regions found in the original, single API hit for AWS. Includes a TODO on gov pricing because the AWS account used for the credentials doesn't have gov built in

This PR also includes a formatting error that popped up because of go 1.27.x
Also has a fix for some stack overflow issue that appears for some reason

## Related Issues
https://apptio.atlassian.net/browse/CLDYCON-7761

## User Impact
No user impact because it's part of the kubemodel pricing updates

## Testing
Tested locally by running w/ an AWS account, should require credentials to an AWS account that has access to all opt-in regions

### DIFF
--- a/modules/pricing/public/aws/awspricingsource.go
+++ b/modules/pricing/public/aws/awspricingsource.go
@@ -243,6 +243,7 @@ func (p *AWSPricingSource) GetPricing() (*pricing.PricingSet, error) {
 
 		resultCh := make(chan regionResult, len(regions))
 		var wg sync.WaitGroup
+		// TODO: Add separate credential path for aws gov regions. Current AWS account cannot hit it
 		for r := range regions {
 			wg.Add(1)
 			go func(r string) {
--- a/modules/pricing/public/aws/awspricingsource.go
+++ b/modules/pricing/public/aws/awspricingsource.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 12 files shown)


modules/pricing/public/cny/nodes.jsonl
Sean-Holcomb, line 1: This seems like a pretty big change in this file. AWS doesn't have spot CNY but Azure and GCP do I guess?

modules/pricing/public/aws/awspricingsource.go
Sean-Holcomb, line 1: Am I understanding this correctly that you build up a list of regions from the other api and use it to do individual queries on this other api.

Sean-Holcomb (approved):
In general looks good. For AWS I believe that the existing pricing api that we are using has a parallel structure to "OnDemand" in its response that might contain some spot date. It might even have something for CNY, but I don't know.

modules/pricing/public/aws/awspricingsource.go
alexsouthard: Yep, it just pulls the regions from the on-demand API and then cycles through it since each spot API request requires region!

modules/pricing/public/cny/nodes.jsonl
alexsouthard: Yeah there's no AWS spot pricing for CNY but the Azure and GCP spot pricing is just baked into the normal pricing API so it's just taking the regular values and converting them to the yuan.

Mostly this is just a consequence of the difference between AWS and Azure/GCP. Like the AWS CNY has special logic scoped for the region whereas Azure/GCP doesn't have any special logic going (and therefore still includes all regions), and so it's pretty verbose for Azure/GCP

Boils down to: AWS CNY requires chinese region whereas Azure/GCP support CNY currency for non-chinese regions

Issue #3195, opened 2025-06-12
Update ClusterCache to use New K8s Client Informers API  + Provider Thread Safety


mbolt35:
# Goals

### Cluster Cache 
* Manages Unstructured Kubernetes Resources - allows omitting fields at the watcher.
    * Allows omitting fields at the watcher level. 
    * Allows use of the Kubernetes resource Data Structures without the full memory footprint (omitted fields). 
* More efficient Kubernetes provided API (avoids awkward faux Store implementation)
* Fix/Address concurrent modification issues happening in existing opencost implementation. https://github.com/opencost/opencost/issues/2910

### Concurrency Issues in Providers
* Fix issue sharing pointers across thread boundaries and concurrent access panics: https://github.com/opencost/opencost/issues/2821

PR #4023, opened 2026-09-01
fix(clustercache): isolate node labels from cache consumers

### LINKED ISSUE #3388: alloy-opencost - Container restarted
Alert: alloy-opencost in alloy-opencost-689b4fbdd7-ql675 has been restarted

Can you take a look at the investigation and see if thet is something that can be fixed from your side, in order to not suffer from container restarts?

Looking to be a Bug in OpenCost Go code — unsafe concurrent access to a map.

Container restarted because OpenCost hit a Go runtime panic:

2025-09-28 01:13:34.192	fatal error: concurrent map iteration and map write
2025-09-28 01:13:34.201	/home/runner/work/opencost/opencost/core/pkg/errors/panic.go:76 +0x6c

Cause of panic:

One goroutine was iterating over a Go map.
Another goroutine was writing to the same map at the same time.
Maps in Go are not thread-safe, so this combination immediately triggers a runtime panic.

Relevant function:

2025-09-28 01:13:34.196github.com/opencost/opencost/core/pkg/util/promutil.SanitizeLabels(0xc0047c4690)
SanitizeLabels was actively executing at the moment of the crash.

This is the point where the unsafe map access occurred.

In summary

The crash and container restart are caused by concurrent map iteration and modification in OpenCost and not Memory related issue, specifically during execution of SanitizeLabels

Best Regards
Amelco SD

riccardomenegazzo:
## Description

Prevent ClusterCache V2 consumers from sharing the cached node label map.

`KubernetesClusterCacheV2.GetAllNodes()` currently returns the node pointers
stored by `GenericStore` directly. That means callers also receive the exact
cached `Labels` map.

`CostModel.GetNodeCost()` enriches those labels with `providerID`, while
`KubeNodeCollector.Collect()` can concurrently iterate the same map through
`promutil.SanitizeLabels()` during a `/metrics` scrape.

That creates a concrete read/write race on a Go map and matches the fatal
`concurrent map iteration and map write` failure reported in #3388. Because
this is a Go runtime fatal error, it terminates the OpenCost process rather
than returning a recoverable request error.

This change makes `GetAllNodes()` return a shallow node copy with an
independently cloned labels map. Callers keep the same node data and can
continue deriving/enriching labels, but those writes no longer mutate the
map held by ClusterCache V2 or race with other consumers.

A focused regression test mutates labels returned by one `GetAllNodes()` call
and verifies a subsequent read is unchanged.

## Related Issues

Fixes #3388
Relates to #3195
Relates to #2910

## User Impact

No API, configuration, or pricing behavior changes are intended.

The change prevents one class of process-level crashes when node metadata is
consumed concurrently, including the `/metrics` scrape path reported in
#3388.

The tradeoff is one shallow node copy and one label-map clone per node
returned from the V2 cache.

## Testing

- Added `TestKubernetesClusterCacheV2GetAllNodesIsolatesLabels`.
- The test verifies that mutating labels returned by one cache read cannot
  affect a later read.
- The regression test is deterministic and does not rely on race timing.
- Repository CI should run the project format, vet, and test checks.

### DIFF
--- a/pkg/clustercache/clustercache2.go
+++ b/pkg/clustercache/clustercache2.go
@@ -1,6 +1,7 @@
 package clustercache
 
 import (
+	"maps"
 	"sync"
 
 	cc "github.com/opencost/opencost/core/pkg/clustercache"
@@ -93,7 +94,18 @@ func (kcc *KubernetesClusterCacheV2) GetAllNamespaces() []*cc.Namespace {
 }
 
 func (kcc *KubernetesClusterCacheV2) GetAllNodes() []*cc.Node {
-	return kcc.nodeStore.GetAll()
+	nodes := kcc.nodeStore.GetAll()
+	isolated := make([]*cc.Node, len(nodes))
+	for i, node := range nodes {
+		if node == nil {
+			continue
+		}
+
+		clone := *node
+		clone.Labels = maps.Clone(node.Labels)
+		isolated[i] = &clone
+	}
+	return isolated
 }
 
 func (kcc *KubernetesClusterCacheV2) GetAllPods() []*cc.Pod {
--- a/pkg/clustercache/clustercache2_test.go
+++ b/pkg/clustercache/clustercache2_test.go
@@ -1,0 +1,37 @@
+package clustercache
+
+import (
+	"testing"
+
+	cc "github.com/opencost/opencost/core/pkg/clustercache"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestKubernetesClusterCacheV2GetAllNodesIsolatesLabels(t *testing.T) {
+	nodeStore := NewGenericStore(cc.TransformNode)
+	node := &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			UID:  types.UID("node-1"),
+			Name: "node-1",
+			Labels: map[string]string{
+				"kubernetes.io/arch": "amd64",
+			},
+		},
+	}
+	require.NoError(t, nodeStore.Add(node))
+
+	cache := &KubernetesClusterCacheV2{nodeStore: nodeStore}
+
+	firstRead := cache.GetAllNodes()
+	require.Len(t, firstRead, 1)
+	firstRead[0].Labels["providerID"] = "aws:///us-east-1a/i-1234567890"
+
+	secondRead := cache.GetAllNodes()
+	require.Len(t, secondRead, 1)
+	require.NotSame(t, firstRead[0], secondRead[0])
+	require.Equal(t, "amd64", secondRead[0].Labels["kubernetes.io/arch"])
+	require.NotContains(t, secondRead[0].Labels, "providerID")
+}


pkg/clustercache/clustercache2.go
```diff
@@ -93,7 +94,18 @@ func (kcc *KubernetesClusterCacheV2) GetAllNamespaces() []*cc.Namespace {
...
-	return kcc.nodeStore.GetAll()
+	nodes := kcc.nodeStore.GetAll()
+	isolated := make([]*cc.Node, len(nodes))
+	for i, node := range nodes {
+		if node == nil {
+			continue
+		}
+
+		clone := *node
+		clone.Labels = maps.Clone(node.Labels)
```
cliffcolvin, line 105: [P1] Protect the supported cache-v1 path too. `USE_CACHE_V1=true` still returns nodes through `TransformNode`, which assigns `Labels: input.Labels`; `GetNodeCost` then writes `providerID` into that shared map, so the reported fatal race remains possible there. Clone at the writer in `GetNodeCost` (as its pod-label path already does), or apply an equivalent contract to both cache implementations; the writer-side fix also avoids cloning every node label map for read-only consumers.

cliffcolvin (changes requested):
Problem gate: pass. #3388 reports this exact process-fatal map iteration/write crash, and #2910 contains repeated large-cluster reports after earlier partial fixes. The description identifies a concrete writer (`GetNodeCost`) and reader (`KubeNodeCollector`), and the deterministic isolation test is a good way to avoid a timing-dependent regression. I found no security issue. Locally, the focused test passed 10 times under `-race`, the full `pkg/clustercache` package passed, and `go vet ./pkg/clustercache` was clean.

The blocker below is scope: this only changes cache V2, while the supported `USE_CACHE_V1=true` path still returns `TransformNode` values whose `Labels` map aliases the watcher object. The same `GetNodeCost` mutation therefore remains capable of racing metrics consumers for those users. The narrowest cross-cache fix is to clone `n.Labels` at the mutation site (matching the existing pod-label pattern), or otherwise give both implementations the same isolation guarantee.

Please also replace the prospective testing statement (`CI should run`) with the exact commands and results you actually ran. This PR's integration workflow took the no-op path, so it currently provides no author/CI execution evidence beyond the added test source.

mbolt35:
This is a good find, but cloning all labels for all nodes is incredibly expensive, which is exactly why we return the map pointer. The memory savings from avoiding label map cloning has been well documented over the years, and we'd like to continue returning the pointer. In order to fix this, we need to update the call site to clone the labels: 
https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/costmodel/costmodel.go#L955

I'd consider this a P0 bug, since this could very easily create a concurrent modification panic.

mbolt35:
The specific case here has been resolved in: https://github.com/opencost/opencost/pull/4050 

We will continue to address these at the modification site, but generally, we want to consider a more elegant broader fix as part of this task: https://github.com/opencost/opencost/issues/3195

peatey:
> In order to fix this, we need to update the call site to clone the labels

@mbolt35 agreed on fixing at the call site. I did a sweep of the consumers of ClusterCache results that mutate the returned maps, and `GetNodeCost` isn't the only writer. For the call-site approach to fully close this class of bug, these need the same treatment:

1. **[`pkg/metrics/podlabelmetrics.go#L86-L92`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/metrics/podlabelmetrics.go#L86-L92)**: `podLabels := pod.Labels` followed by `delete(podLabels, lname)` when `UseLabelsWhitelist` is enabled. This deletes non-whitelisted labels from the *cached* pod on every scrape, so labels silently disappear for `ComputeCostData` and other `GetAllPods()` consumers. Overlapping scrapes can also hit the same fatal concurrent map error.
2. **[`pkg/costmodel/costmodel.go#L292`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/costmodel/costmodel.go#L292)**: `podAnnotations := pod.Annotations`, then namespace annotations are merged into it. Pod labels a few lines above are already cloned, but annotations aren't.
3. **[`pkg/costmodel/costmodel.go#L824-L833`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/costmodel/costmodel.go#L824-L833)** (`addPVData`): writes `annotationStorageCost` into the cached StorageClass `Parameters` map (`TransformStorageClass` aliases `input.Parameters`).
4. **[`pkg/cloud/aws/provider.go#L947-L950`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/cloud/aws/provider.go#L947-L950)** (`DownloadPricing`): writes `provisioner` into the same cached StorageClass `Parameters` map on every pricing refresh.
5. **[`pkg/cloud/digitalocean/provider.go#L241-L257`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/cloud/digitalocean/provider.go#L241-L257)** (`DOKS.GetKey`): writes `providerID` and `node.opencost.io/cpu`/`ram` into the labels it's passed. `GetNodeCost` calls `GetKey` four times with `nodeLabels`, so the clone needs to happen once, before the first call, and be reused.

For `GetNodeCost` itself, the minimal fix (`maps` is already imported) covers both V1 and V2 as well as the DOKS writes:

```go
nodeLabels := maps.Clone(n.Labels)
if nodeLabels == nil {
	log.Warnf("GetNodeCost: Found node '%s' with no labels", name)
	nodeLabels = make(map[string]string)
}
nodeLabels["providerID"] = n.SpecProviderID
```

The remaining sites need the same `maps.Clone` before mutation. For the StorageClass params, keep the existing `params != nil` guard, since `maps.Clone(nil)` returns nil.

mbolt35:
> > In order to fix this, we need to update the call site to clone the labels
> 
> @mbolt35 agreed on fixing at the call site. I did a sweep of the consumers of ClusterCache results that mutate the returned maps, and `GetNodeCost` isn't the only writer. For the call-site approach to fully close this class of bug, these need the same treatment:
> 
> 1. **[`pkg/metrics/podlabelmetrics.go#L86-L92`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/metrics/podlabelmetrics.go#L86-L92)**: `podLabels := pod.Labels` followed by `delete(podLabels, lname)` when `UseLabelsWhitelist` is enabled. This deletes non-whitelisted labels from the _cached_ pod on every scrape, so labels silently disappear for `ComputeCostData` and other `GetAllPods()` consumers. Overlapping scrapes can also hit the same fatal concurrent map error.
> 2. **[`pkg/costmodel/costmodel.go#L292`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/costmodel/costmodel.go#L292)**: `podAnnotations := pod.Annotations`, then namespace annotations are merged into it. Pod labels a few lines above are already cloned, but annotations aren't.
> 3. **[`pkg/costmodel/costmodel.go#L824-L833`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/costmodel/costmodel.go#L824-L833)** (`addPVData`): writes `annotationStorageCost` into the cached StorageClass `Parameters` map (`TransformStorageClass` aliases `input.Parameters`).
> 4. **[`pkg/cloud/aws/provider.go#L947-L950`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/cloud/aws/provider.go#L947-L950)** (`DownloadPricing`): writes `provisioner` into the same cached StorageClass `Parameters` map on every pricing refresh.
> 5. **[`pkg/cloud/digitalocean/provider.go#L241-L257`](https://github.com/opencost/opencost/blob/45243e33438b8f51092bb3931784696f105eeba7/pkg/cloud/digitalocean/provider.go#L241-L257)** (`DOKS.GetKey`): writes `providerID` and `node.opencost.io/cpu`/`ram` into the labels it's passed. `GetNodeCost` calls `GetKey` four times with `nodeLabels`, so the clone needs to happen once, before the first call, and be reused.
> 
> For `GetNodeCost` itself, the minimal fix (`maps` is already imported) covers both V1 and V2 as well as the DOKS writes:
> 
> ```go
> nodeLabels := maps.Clone(n.Labels)
> if nodeLabels == nil {
> 	log.Warnf("GetNodeCost: Found node '%s' with no labels", name)
> 	nodeLabels = make(map[string]string)
> }
> nodeLabels["providerID"] = n.SpecProviderID
> ```
> 
> The remaining sites need the same `maps.Clone` before mutation. For the StorageClass params, keep the existing `params != nil` guard, since `maps.Clone(nil)` returns nil.

Thank you for getting those tracked down - will update 4050 with specific updates.

PR #4024, opened 2026-09-01
Follow the allocation pattern for the network insight account field

kevinguy-ibm:
## Description

Follow-up to #4025, which added `account` to the network insight grammar.

`networkInsightFieldMap` has no case for `account`. An in-memory filter on it falls through to the generic error:

```
Failed to find string identifier on Network Insight: account
```

and `Matches` returns `false`. So the filter matches nothing, and the error makes it look like someone forgot to add the field.

Per review on #4025, `Account` should not be a member of `NetworkInsight`. Storage joins the account in after ingestion, so the SQL compilers resolve the field and this path cannot. `allocationFieldMap` already hits the same situation and returns `"account property not implemented"`, while `AllocationProperties` carries no Account field either. This does the same, so both filters behave alike.

## Related Issues

Relates to #4025.

## User Impact

None for existing callers. No type changes, no signature changes, no new JSON fields.

An in-memory filter on `account` returns `false` before and after. Only the error changes. It now says the field is not implemented instead of looking like a lookup bug.

## Testing

There was no test for `NetworkInsightMatcher`, so this adds one.

- `TestNetworkInsightMatcher_Fields` parses a filter, compiles it, and matches it against a fixture. It covers the fields that do resolve. If the grammar and the field map drift apart, the test fails instead of the runtime.
- `TestNetworkInsightMatcher_AccountIsNotResolvable` locks the gap in on purpose, so nobody makes `account` match in memory later without the test saying so.

`go build ./...` and `go test ./...` pass on both modules.

### DIFF
--- a/core/pkg/opencost/networkinsightmatcher.go
+++ b/core/pkg/opencost/networkinsightmatcher.go
@@ -52,6 +52,9 @@ func networkInsightFieldMap(ni *NetworkInsight, identifier ast.Identifier) (stri
 		return ni.Namespace, nil
 	case nfilter.FieldPod:
 		return ni.Pod, nil
+	case nfilter.FieldAccount:
+		// Resolved against storage, not in memory. Same as allocationFieldMap.
+		return "", fmt.Errorf("account property not implemented")
 	}
 
 	return "", fmt.Errorf("Failed to find string identifier on Network Insight: %s", identifier.Field.Name)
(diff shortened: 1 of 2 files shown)


kevinguy-ibm:
Closing this.

After the review on #4025 it reduced to returning a not-implemented error on a path that returns false either way, so it changes no behavior, only a log message. Not worth the review cycles.

Matt's point on #4025 covers it: the account belongs in the storage join, not on the NetworkInsight struct, the same way allocations already work. That is what we are doing on our side, so nothing is blocked by closing this.

Thanks both for looking at it.

PR #4025, opened 2026-09-01
Add Account filter to networkinsights

daniel-spray:
## Description
Adds `account` as a supported filter field for the Network Insight filter package. This includes:

- `FieldAccount` constant in `core/pkg/filter/networkinsight/fields.go`, using `fieldstrings.FieldAccount` consistent with all other filter packages
- `FieldAccount` wired into `networkInsightFilterFields` in `parser.go` so `NewNetworkInsightFilterParser()` recognises `account` as a valid filter term
- Unit tests added in `core/pkg/filter/networkinsight/parser_test.go` covering parse, negation, multi-value, compound expressions, unknown field rejection, and `DefaultFieldByName`

## User Impact
Non-breaking additive change. Existing filter expressions are unaffected.

## Testing
- Added unit tests in `core/pkg/filter/networkinsight/parser_test.go`:
  - `TestParse` — valid filter expressions for all fields including `account`, negation, multi-value, and compound `+`/`|` expressions
  - `TestFailingParses` — confirms unknown fields and invalid operators are rejected
  - `TestDefaultFieldByName` — confirms all four `FieldXxx` constants resolve correctly via `DefaultFieldByName`
- All existing tests in `core/pkg/opencost` and `core/pkg/filter/networkinsight` pass

### DIFF
--- a/core/pkg/filter/networkinsight/fields.go
+++ b/core/pkg/filter/networkinsight/fields.go
@@ -1,14 +1,16 @@
 package networkinsight
 
+import "github.com/opencost/opencost/core/pkg/filter/fieldstrings"
+
 type NetworkInsightField string
 type NetworkInsightDetailField string
 
-// Only support namespace, pod and cluster on day 1
 // Field used for Network Insight filtering
 const (
-	FieldClusterID NetworkInsightField = "cluster"
-	FieldNamespace NetworkInsightField = "namespace"
-	FieldPod       NetworkInsightField = "pod"
+	FieldClusterID NetworkInsightField = NetworkInsightField(fieldstrings.FieldClusterID)
+	FieldNamespace NetworkInsightField = NetworkInsightField(fieldstrings.FieldNamespace)
+	FieldPod       NetworkInsightField = NetworkInsightField(fieldstrings.FieldPod)
+	FieldAccount   NetworkInsightField = NetworkInsightField(fieldstrings.FieldAccount)
 )
 
 // Field used for Network Insight Details filtering
--- a/core/pkg/filter/networkinsight/parser.go
+++ b/core/pkg/filter/networkinsight/parser.go
@@ -10,6 +10,7 @@ var networkInsightFilterFields []*ast.Field = []*ast.Field{
 	ast.NewField(FieldClusterID),
 	ast.NewField(FieldNamespace),
 	ast.NewField(FieldPod),
+	ast.NewField(FieldAccount),
 }
 
 var fieldMap map[NetworkInsightField]*ast.Field
--- a/core/pkg/filter/networkinsight/parser_test.go
+++ b/core/pkg/filter/networkinsight/parser_test.go
@@ -1,0 +1,104 @@
+package networkinsight
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/opencost/opencost/core/pkg/filter/ast"
+)
+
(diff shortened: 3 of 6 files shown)


mbolt35:
There are a few major problems with this change set:

1. You have added a field to one of the exported opencost structs. See: [bingen.go](https://github.com/opencost/opencost/blob/develop/core/pkg/opencost/bingen.go)  
  - Run: `go install github.com/opencost/bingen/cmd/bingen@latest` on your command line to install bingen. 
  - You'll need to advance the version of the network insights version set from `1` to `2`
  - You'll need to annotate your new `Account` field with a `//@bingen:field[version=2]`
  - You'll need to regenerate the binary codecs after making the two above changes. This will allow the account to be encoded and decoded and maintain backwards compatibility. 
  - There are examples of when new fields were added to the `Allocation` struct here: https://github.com/opencost/opencost/blob/2841b19cdc265d837e625f01bb577cf0379960e6/core/pkg/opencost/allocation.go#L55-L112 -- note the `//@bingen:field` 

2. You never populate the `Account` field - If you look at https://github.com/opencost/opencost/blob/2841b19cdc265d837e625f01bb577cf0379960e6/pkg/costmodel/networkinsight.go#L13 you'll see we execute a handful of queries against the metric querier for data pertaining to scraped egress/ingress byte accounting data. The general structure for the query result is: 
```
type NetworkGiBResult struct {
	UID       string
	Cluster   string
	Namespace string
	Pod       string
	Service   string

	Data []*util.Vector
}
```
Note that this data does *not* include an `Account`, and attempting to join in an account here feels a bit odd, but it might be possible (just depends on what is expected there).
  - Account data would need to come from the provider (?) -- or should it be joined later in the pipeline. *IF* we want this to get joined later in the pipeline (similar to Allocation's -- which doesn't have an actual Account field, but does have an Account filter).
  - TLDR; There's no way to inherently derive the account information from the metric query alone, and it's _unclear_ how exactly that data should be joined. So, you may want to hold off on adding this change until we can strategize about how we want to populate this data first. 
  - I believe that the newer Kubemodel solves this problem because of the parent-child structure, where the consumer would have the account context before the network insights were even read.

mbolt35:
@daniel-spray Ok, so had a few discussions, and I think the _actual_ change we want here is a slight reduction in what you already have (which is, of course, MILES from what I was suggesting we do).

Basically, the filters are template-like structures for allowing you to define "filterable" keywords and their types. You've done this already for NetworkInsights. The filters don't necessarily have to perfectly map to the actual data structure representation as long as the mechanism applying the filter has an understanding of what each field maps to. 

In other words, if there was a database representation of network insights that had an `account` column, then your filter compiler could leverage that for account filtering. 

When it comes to the way Kubecost processes allocations, we perform a join of account with the allocation data in our database pipeline AFTER consuming the exported allocation data. This is why the allocation filter parser allows the "account" field despite the `Allocation` data structure lacking an actual `Account` field. While I _do_ think that this is a bit misleading for opencost users,  I don't believe the costs associated with this detail are worth making a fuss over. 

In short, I think you would want to follow the allocation implementation here -- keep the current filter changes, but omit adding an Account field to the NetworkInsight. Then, in the Kubecost implementation, add a derivation step that joins the account field into the network_insight table.

mbolt35 (approved):
Looks great! Thank you sir

PR #4027, opened 2026-09-01
aws: stop evicting pricing for instance types that have merely scaled to zero

tormodmacleod:
## Description

`DownloadPricingData` re-fetches the region's entire AWS offer file — several hundred MB — far more often than the file changes, because pricing for an instance type is discarded as soon as no node of that type is running.

Two maps are populated at different scopes:

- `populatePricing` **replaces** `aws.Pricing` on every download and refills it only for instance types present in the cluster at that moment — `if inputkeys[key] || inputkeys[spotKey]`, where `inputkeys` comes from the `GetAllNodes()` snapshot taken at the top of `DownloadPricingData`.
- `ValidPricingKeys[key]` / `[spotKey]` are written **outside** that guard, so they cover every on-demand instance type in the offer file.

`NodePricing` re-triggers a full `DownloadPricingData` on exactly that gap: a key that misses `aws.Pricing` but hits `aws.ValidPricingKeys`. So when an instance type scales to zero its pricing is evicted by the next download, and when a node of that type comes back the lookup misses, the `ValidPricingKeys` check passes, and the whole file is pulled again. Under autoscaler churn that repeats for as long as the cluster keeps cycling through the same handful of types.

### Impact observed

On a production EKS cluster in `eu-west-1` the exporter downloaded the offer file — **435,867,095 bytes** — **108 times a day**, about **53 GB/day** of NAT gateway egress, roughly **$72/month**. `starting download of` appears 1–9 times an hour, around the clock, with no diurnal shape.

The cluster ran 8 distinct instance types across 11 nodes, 6 of them singletons, so a single consolidation evicts a type outright. Node launch rate (~2.5–4/hour) tracks the download rate. Downloads succeed — no fetch or decode errors — so nothing is failing; the same work is simply repeated. `GET /allNodePricing` returned exactly the keys for the running node types against hundreds of types in the file, which shows the scoping directly.

### The change

`DownloadPricingData` populates pricing for instance types seen within the last 24 hours, not only those running at that instant. A returning node is then a hit in `aws.Pricing` and triggers nothing.

`NodePricing` also notes the key it is about to ask about before triggering the download. Without that, a node whose type the cluster cache has not yet picked up produces a download that does not populate the key that caused it, and the next call makes the same request again.

A type genuinely new to the process still costs one download. That is the self-heal the retry branch exists for, and it is unchanged — what goes away is repeating it for types the cluster has already run.

`pricingKeyMemory` is a 24 hour constant rather than a setting: it only has to outlast a scale-down, and a day covers a cluster that scales to zero overnight. The memory is bounded by the number of distinct instance types the cluster runs, and expired entries are dropped from the map rather than merely ignored.

### Why not the alternatives

- **Merging into `Pricing` instead of replacing it** keeps the evicted type's pricing, but at its old price — nothing re-parses it until a node of that type returns and puts it back in `inputkeys`, and there is no periodic refresh to correct it in the meantime. Widening the filter keeps every remembered type re-parsed from the current offer file on each download, so pricing stays as fresh as it is today.
- **Narrowing `ValidPricingKeys` to match `Pricing`** would make the retry branch unreachable and remove the self-heal for types appearing between downloads, which looks intentional. Left alone.

## Related Issues

- Follows #3988, which makes each refresh cheap with a conditional request. This one stops most of the refreshes happening at all. They are independent — either helps on its own, and together the conditional request covers the downloads this cannot remove (a genuinely new instance type, a restart).
- Relates to #3940 — with the churn removed, a periodic refresh worker becomes the main thing driving downloads, which is the intended behaviour rather than an accident of eviction.

## User Impact

No configuration or API change, and not breaking. Pricing content is unchanged: remembered types are re-parsed from the current offer file on every download exactly as running types are, so no price is served from a staler source than it is today.

The `Pricing` map holds entries for types seen in the last 24 hours rather than only those currently running — a few dozen extra entries on a churning cluster.

## Testing

- `TestPricingKeyMemory` covers the memory itself: a running type is populated, a type whose nodes went away is still populated, a key noted by `NodePricing` is populated, and a type gone longer than `pricingKeyMemory` is both dropped from the returned set and removed from the map.
- `TestPricingSurvivesScaleToZero` is the regression, driven through `populatePricing` with the existing `testdata/pricing-us-east-2.json` fixture. It first pins the behaviour being fixed — handed only the currently-running keys, `populatePricing` drops the type while `ValidPricingKeys` keeps it, which is the loop — and then asserts the type survives a refresh once the key memory is in play.

`go test ./pkg/cloud/aws/ -count=1` passes, `go test ./pkg/cloud/aws/ -run 'TestPricingKeyMemory|TestPricingSurvives' -race -count=2` passes, `gofmt` and `go vet ./pkg/cloud/aws/` are clean.

Not yet verified in-cluster — I have not deployed a build of this to the affected cluster. The measurements above are from the unpatched exporter.


pkg/cloud/aws/provider.go
```diff
@@ -1738,6 +1799,10 @@ func (aws *AWS) NodePricing(k models.Key) (*models.Node, models.PricingMetadata,
 	if ok {
 		return aws.createNode(terms, usageType, k)
 	} else if _, ok := aws.ValidPricingKeys[key]; ok {
+		// Note the key before downloading: the cluster cache may not have caught up with
+		// the node yet, and a download that does not populate the key being asked for
+		// leaves the next call to make the same request again.
+		aws.notePricingKey(key)
```
cliffcolvin, line 1812: [P1] Refresh this timestamp on normal pricing hits too. `recentPricingKeys` is otherwise updated only during a download or this miss: if a type runs for more than 24 hours with no download, scales to zero, and another key then triggers a refresh, its old timestamp expires immediately even though it disappeared moments ago; its return causes the same full download this PR is meant to prevent. Record successful `NodePricing` lookups (or continuously observed node keys) and cover that quiet-period sequence in the regression test.

cliffcolvin (changes requested):
Problem gate: pass. This is a measured production problem, not an invented one: the description ties 108 successful 436 MB downloads/day to the existing `Pricing`/`ValidPricingKeys` scope mismatch, explains user cost, documents alternatives, and supplies focused test commands. The change is narrowly scoped, uses a dedicated mutex, and introduces no security concern. I independently ran the focused tests with `-race -count=2`, the full `pkg/cloud/aws` test package, and `go vet`; all passed.

There is one blocker below: the 24-hour timestamp currently measures the last pricing download/miss, not the last time the type was actually present or priced. That leaves the core loop reproducible after a quiet period. Please make normal successful sightings refresh the timestamp and add the long-running-then-scale-to-zero case to the regression test.

The integration check is also red. Its log includes a target-branch `pathspec` failure plus broad allocation/network comparison failures that do not look related to this AWS-only patch, but it should be rerun or explicitly waived on an up-to-date branch before merge.

tormodmacleod:
@cliffcolvin thanks — the blocker was right, and it is worse than it reads. Pushed 6f3705f plus a `develop` merge.

## The timestamp

It was stamped only when a download ran or when `NodePricing` missed, so it measured time since the last *download*. Those two diverge exactly when this change works: downloads stop, so a type that has been running continuously throughout still ages out of the map, and is then evicted the moment it scales down. The original loop, reproduced after any quiet period longer than `pricingKeyMemory` — and the more effective the change, the more likely it is.

`NodePricing` now calls `notePricingKey` on the hit path as well, so the memory measures 24 hours from the last time a node of that type was actually priced rather than from the last download.

## The regression case

New subtest in `TestPricingSurvivesScaleToZero`: populate, age the memory past `pricingKeyMemory` with no download in between, price a node of that type through the real `NodePricing`, then scale to zero. Pricing must survive.

I checked it bites rather than merely passing — with the `notePricingKey` call reverted it fails with `pricing for "us-east-2,p3.8xlarge,windows" was evicted after a quiet period`.

It uses `p3.8xlarge` rather than `m5.large` because that is the only compute instance in `testdata/pricing-us-east-2.json` whose on-demand terms carry `HourlyRateCode`. With `m5.large` the fixture's rate code is `2TG2D8R56U`, so `createNode` returns `cannot find the following key in OnDemand pricing data` and the sighting would not be a real one.

## The integration check

The branch was 24 commits behind; `develop` is now merged in, cleanly. Upstream's only changes to `pkg/cloud/aws/provider.go` in that range were a `maps.Clone` fix and a logging tidy, so nothing overlaps the pricing paths. The rerun should be a fair test — if the allocation/network comparison failures repeat on the up-to-date branch I will dig into them rather than assume they are unrelated.

`gofmt` and `go vet ./pkg/cloud/aws/` clean, the full `./pkg/cloud/aws/` package passes, and `go test ./pkg/cloud/aws/ -run 'TestPricingKeyMemory|TestPricingSurvives' -race -count=2` passes.

PR #4028, opened 2026-09-03
fix(aws): include marketplace fees

### LINKED ISSUE #4022: AWS Athena/CUR integration excludes Marketplace subscription fees (line_item_line_item_type = 'Fee')
### Summary

The AWS Athena/CUR cloud cost integration filters `line_item_line_item_type` down to a fixed whitelist that does not include `Fee`, so AWS Marketplace third-party SaaS subscription costs (and other recurring fee line items) are silently excluded from OpenCost's reported AWS costs.

### Where

`pkg/cloud/aws/athenaintegration.go`, in the `AthenaWhereUsage` constant used by `getCloudCost()`:

​```go
const AthenaWhereUsage = "(line_item_line_item_type = 'Usage' OR line_item_line_item_type = 'DiscountedUsage' OR line_item_line_item_type = 'SavingsPlanCoveredUsage' OR line_item_line_item_type = 'EdpDiscount' OR line_item_line_item_type = 'PrivateRateDiscount')"
​```

This is AND-ed into the final WHERE clause alongside the date/partition filters, so any CUR row whose `line_item_line_item_type` is not one of those five values is dropped before aggregation.

### Impact

In the AWS Cost and Usage Report schema, AWS Marketplace subscriptions (and other flat-rate recurring charges) are recorded with `line_item_line_item_type = 'Fee'`. Since `'Fee'` is not in the whitelist above, these costs never appear in OpenCost's AWS totals when using the Athena/CUR integration — with no warning or documentation that this category is omitted.

### Expected behavior

Either:
1. `Fee` line items (or at minimum, AWS Marketplace fees specifically, e.g. filtered via `product_product_name = 'AWSMarketplace'`) should be included in the query so Marketplace SaaS subscription costs are reflected in totals, or
2. If this exclusion is intentional (e.g. to avoid double-counting or because Marketplace costs are handled elsewhere), it should be clearly documented as a known limitation of the Athena/CUR integration.

### Actual behavior

AWS Marketplace subscription/SaaS costs billed through the CUR are completely absent from OpenCost's AWS cost data pulled via Athena, with no indication to the user that this category is missing.

### Suggested fix

Add `'Fee'` (or a more targeted `product_product_name = 'AWSMarketplace'` condition) to `AthenaWhereUsage`, or expose this as a configurable option, and update the Athena/CUR docs to note which `line_item_line_item_type` values are included/excluded by default.

### Environment

- OpenCost version: 1.20.3

Kush172005:
## Summary

Fixes #4022 by including AWS Marketplace **Fee** charges in the Athena CUR integration.

Currently, the Athena query filters "**line_item_line_item_type**" to a fixed set of values, so **Marketplace fee charges are left out of OpenCost costs**.

## Changes

* Include **`Fee`** line items when **`bill_billing_entity = 'AWS Marketplace'**
* Keep **non-Marketplace Fee** line items excluded
* Add a **regression test** for the new filter
* Update the related comments

This keeps the change limited to **Marketplace fees** instead of including all AWS `Fee` charges.

## Testing

* `go test ./pkg/cloud/...`
* `go vet ./...`
* `gofmt`
* `git diff --check`

**Fixes #4022**

### DIFF
--- a/pkg/cloud/aws/athenaintegration.go
+++ b/pkg/cloud/aws/athenaintegration.go
@@ -54,8 +54,22 @@ var AthenaNetSPPricingCoalesce = fmt.Sprintf("COALESCE(%s, %s, 0)", AthenaNetSPP
 const AthenaDateColumn = "line_item_usage_start_date"
 const AthenaDateTruncColumn = "DATE_TRUNC('day'," + AthenaDateColumn + ") as usage_date"
 
+// AthenaBillingEntityColumn distinguishes standard AWS charges ('AWS') from AWS
+// Marketplace charges ('AWS Marketplace') on a CUR line item.
+const AthenaBillingEntityColumn = "bill_billing_entity"
+const AthenaMarketplaceBillingEntity = "AWS Marketplace"
+
 const AthenaWhereDateFmt = `line_item_usage_start_date >= date '%s' AND line_item_usage_start_date < date '%s'`
-const AthenaWhereUsage = "(line_item_line_item_type = 'Usage' OR line_item_line_item_type = 'DiscountedUsage' OR line_item_line_item_type = 'SavingsPlanCoveredUsage' OR line_item_line_item_type = 'EdpDiscount' OR line_item_line_item_type = 'PrivateRateDiscount')"
+
+// AthenaWhereUsage includes usage-driving line item types plus AWS Marketplace 'Fee'
+// line items (flat-rate/subscription charges for third-party SaaS products). Marketplace
+// is scoped to bill_billing_entity = 'AWS Marketplace' so this does not also pull in
+// non-Marketplace 'Fee' rows, such as Reserved Instance upfront purchases,
+// which are outside the scope of this Marketplace-specific fix.
+var AthenaWhereUsage = fmt.Sprintf(
+	"(line_item_line_item_type = 'Usage' OR line_item_line_item_type = 'DiscountedUsage' OR line_item_line_item_type = 'SavingsPlanCoveredUsage' OR line_item_line_item_type = 'EdpDiscount' OR line_item_line_item_type = 'PrivateRateDiscount' OR (line_item_line_item_type = 'Fee' AND %s = '%s'))",
+	AthenaBillingEntityColumn, AthenaMarketplaceBillingEntity,
+)
 
 // AthenaQueryIndexes is a struct for holding the context of a query
 type AthenaQueryIndexes struct {
@@ -183,9 +197,9 @@ func (ai *AthenaIntegration) getCloudCost(start, end time.Time, limit int) (*ope
 	whereDate := fmt.Sprintf(AthenaWhereDateFmt, start.Format("2006-01-02"), end.Format("2006-01-02"))
 	wherePartitions := ai.GetPartitionWhere(start, end, isCUR20(allColumns))
 
-	// Query for all line items with a resource_id or from AWS Marketplace, which did not end before
-	// the range or start after it. This captures all costs with any amount of
-	// overlap with the range, for which we will only extract the relevant costs
+	// Query for all line items whose usage start date falls within the given range and
+	// partition, restricted to usage-driving line item types and AWS Marketplace fees
+	// (see AthenaWhereUsage).
 	whereConjuncts := []string{
 		wherePartitions,
 		whereDate,
--- a/pkg/cloud/aws/athenaintegration_coverage_test.go
+++ b/pkg/cloud/aws/athenaintegration_coverage_test.go
@@ -7,6 +7,29 @@ import (
 	"github.com/opencost/opencost/pkg/cloud"
 )
 
+func TestAthenaWhereUsage(t *testing.T) {
+	// Regression test for https://github.com/opencost/opencost/issues/4022 -
+	// AWS Marketplace subscription/fee charges (line_item_line_item_type = 'Fee')
+	// must be included, but only when billed through AWS Marketplace, so that
+	// non-Marketplace 'Fee' rows (e.g. Reserved Instance upfront purchases,
+	// already captured via 'DiscountedUsage' amortization) are not swept in and
+	// double counted.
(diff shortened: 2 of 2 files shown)


Kush172005:
@peatey, @Sean-Holcomb have a look when you get a chance. Thanks!

pkg/cloud/aws/athenaintegration.go
```diff
@@ -57,8 +57,22 @@ var AthenaNetSPPricingCoalesce = fmt.Sprintf("COALESCE(%s, %s, 0)", AthenaNetSPP
...
+
 const AthenaWhereDateFmt = `line_item_usage_start_date >= date '%s' AND line_item_usage_start_date < date '%s'`
-const AthenaWhereUsage = "(line_item_line_item_type = 'Usage' OR line_item_line_item_type = 'DiscountedUsage' OR line_item_line_item_type = 'SavingsPlanCoveredUsage' OR line_item_line_item_type = 'EdpDiscount' OR line_item_line_item_type = 'PrivateRateDiscount')"
+
+// AthenaWhereUsage includes usage-driving line item types plus AWS Marketplace 'Fee'
+// line items (flat-rate/subscription charges for third-party SaaS products). Marketplace
+// is scoped to bill_billing_entity = 'AWS Marketplace' so this does not also pull in
+// non-Marketplace 'Fee' rows, such as Reserved Instance upfront purchases,
+// which are outside the scope of this Marketplace-specific fix.
+var AthenaWhereUsage = fmt.Sprintf(
```
peatey, line 80: `AthenaWhereUsage` now always references `bill_billing_entity`, but unlike the other optional columns in this file (`resource_tags`, `line_item_usage_account_name`, `bill_payer_account_name`, net/amortized pricing), its presence is never checked against `allColumns`. The file itself notes that CUR 2.0 can disable any column.

If a CUR 2.0 data export omits `bill_billing_entity`, every Athena query will fail with `COLUMN_NOT_FOUND`. `GetCloudCost` would then error on every run, dropping **all** cloud cost data for that integration (not just Marketplace fees), and `RefreshStatus` would mark it `FailedConnection`. Those same setups work today.

Suggestion: build the usage filter inside `getCloudCost` and only append the `OR (line_item_line_item_type = 'Fee' AND bill_billing_entity = 'AWS Marketplace')` clause when `allColumns[AthenaBillingEntityColumn]` is true, falling back to the previous filter otherwise.

pkg/cloud/aws/athenaintegration.go
```diff
@@ -57,8 +57,22 @@ var AthenaNetSPPricingCoalesce = fmt.Sprintf("COALESCE(%s, %s, 0)", AthenaNetSPP
...
+	"(line_item_line_item_type = 'Usage' OR line_item_line_item_type = 'DiscountedUsage' OR line_item_line_item_type = 'SavingsPlanCoveredUsage' OR line_item_line_item_type = 'EdpDiscount' OR line_item_line_item_type = 'PrivateRateDiscount' OR (line_item_line_item_type = 'Fee' AND %s = '%s'))",
+	AthenaBillingEntityColumn, AthenaMarketplaceBillingEntity,
+)
```
peatey: Suggested fix for the missing-column issue above (1/2): keep the original filter as `AthenaWhereUsageBase` so there is a fallback when `bill_billing_entity` is absent. `AthenaWhereUsage` is unchanged, so `TestAthenaWhereUsage` still passes.

```suggestion
// AthenaWhereUsageBase filters to usage-driving line item types only.
const AthenaWhereUsageBase = "(line_item_line_item_type = 'Usage' OR line_item_line_item_type = 'DiscountedUsage' OR line_item_line_item_type = 'SavingsPlanCoveredUsage' OR line_item_line_item_type = 'EdpDiscount' OR line_item_line_item_type = 'PrivateRateDiscount')"

// AthenaWhereUsage extends AthenaWhereUsageBase with AWS Marketplace 'Fee'
// line items (flat-rate/subscription charges for third-party SaaS products). Marketplace
// is scoped to bill_billing_entity = 'AWS Marketplace' so this does not also pull in
// non-Marketplace 'Fee' rows, such as Reserved Instance upfront purchases,
// which are outside the scope of this Marketplace-specific fix. CUR 2.0 exports can
// omit bill_billing_entity, so only use this filter when that column is present.
var AthenaWhereUsage = fmt.Sprintf(
	"(line_item_line_item_type = 'Usage' OR line_item_line_item_type = 'DiscountedUsage' OR line_item_line_item_type = 'SavingsPlanCoveredUsage' OR line_item_line_item_type = 'EdpDiscount' OR line_item_line_item_type = 'PrivateRateDiscount' OR (line_item_line_item_type = 'Fee' AND %s = '%s'))",
	AthenaBillingEntityColumn, AthenaMarketplaceBillingEntity,
)
```

pkg/cloud/aws/athenaintegration.go
```diff
@@ -186,9 +200,9 @@ func (ai *AthenaIntegration) getCloudCost(start, end time.Time, limit int) (*ope
...
 
-	// Query for all line items with a resource_id or from AWS Marketplace, which did not end before
-	// the range or start after it. This captures all costs with any amount of
-	// overlap with the range, for which we will only extract the relevant costs
+	// Query for all line items whose usage start date falls within the given range and
+	// partition, restricted to usage-driving line item types and AWS Marketplace fees
+	// (see AthenaWhereUsage).
 	whereConjuncts := []string{
 		wherePartitions,
 		whereDate,
```
peatey, line 216: Suggested fix (2/2): only use the Marketplace-inclusive filter when `bill_billing_entity` exists in `allColumns`.

**Note:** the next line (`AthenaWhereUsage,`) is outside the diff, so it can't be part of this suggestion. After applying, change it to `whereUsage,` by hand.

```suggestion
	// Query for all line items whose usage start date falls within the given range and
	// partition, restricted to usage-driving line item types and, when the
	// bill_billing_entity column exists, AWS Marketplace fees (see AthenaWhereUsage).
	whereUsage := AthenaWhereUsage
	if !allColumns[AthenaBillingEntityColumn] {
		whereUsage = AthenaWhereUsageBase
	}
	whereConjuncts := []string{
		wherePartitions,
		whereDate,
```

Kush172005:
Hey @peatey, addressed both. Thanks for catching this!

- Added base + Marketplace-aware filters with a safe fallback.
- GetWhereUsage now applies the Marketplace clause only when bill_billing_entity exists.
- Added coverage for both branches

peatey (approved):

PR #4037, opened 2026-09-11
fix(costmodel): include K8s native sidecar init containers in pod metrics (#3956)

Tyagiquamar:
## Problem
OpenCost pod allocation calculations omitted Kubernetes 1.28+ Native Sidecar containers defined in \pod.Spec.InitContainers\ with \
estartPolicy: Always\, causing under-reported pod resource usage and cost attribution.

## Root Cause
\NewContainerMetricsFromPod\ in \pkg/costmodel/containerkeys.go\ iterated only over \pod.Spec.Containers\, ignoring native sidecars configured under \pod.Spec.InitContainers\.

## Fix
Updated \NewContainerMetricsFromPod\ to inspect \pod.Spec.InitContainers\ and include any init container where \
estartPolicy: Always\.

## Verification
Ran \go test ./pkg/costmodel/...\ green.


peatey:
Thanks for picking this up, @Tyagiquamar! Closing this one for now, for a few reasons:

1. **It doesn't compile.** `clustercache.PodSpec` has no `InitContainers` field and `clustercache.Container` has no `RestartPolicy`. `go build ./pkg/costmodel/` fails with `pod.Spec.InitContainers undefined`. Init containers would first need to be added to the clustercache types and transforms.
2. **It would drop sidecar usage.** Even with those fields in place, adding sidecar keys to `currentContainers` sends them down the branch in `ComputeCostData` that only builds `CostData` for `pod.Spec.Containers`. Today they fall through to the `else` branch and still emit usage-only data, so this change would remove sidecar CPU/RAM usage from `/costDataModel` and the metrics emitter.
3. **It doesn't address the core of #3956.** The missing sidecar requests come from `kube_pod_init_container_resource_requests`, which the allocation queries don't read. This PR only touches the legacy `ComputeCostData` key path, so `/allocation` would be unchanged.

We also have a new version of kubemodel on the way, and native sidecar support is best handled there (container discovery and requests, including init containers with `restartPolicy: Always`). Let's keep tracking the requirement in #3956. Contributions against the new implementation once it lands would be very welcome.

Tyagiquamar:
Thanks for the detailed explanation, really appreciate it. I’m here mainly to learn and contribute where I can, so this feedback is genuinely helpful for understanding the codebase better.

I’ll keep an eye on #3956 and the new kubemodel implementation, and I’d be happy to contribute once the relevant changes land. Thanks again for taking the time to review this!

PR #4039, opened 2026-09-11
Fix PV ProviderID

Sean-Holcomb:
## Description
Update logic on PV Info metric so that it emits the proper ProviderID, Change name Of CSI Handle property on KubeModel PV to avoid bingen version increment. 

ProviderID for PVs is not explicitly part of the spec like it is for nodes however it can be found in various places depending on the provider and is meant to hold the identifier given to the PV by the provider. CSI Volume Handle is one place this can be but the pv spec struct has multiple (deprecated) vender specific locations that it can be.


modules/collector-source/pkg/scrape/clustercache.go
```diff
@@ -1339,7 +1359,7 @@ func getPersistentVolumeClaimClass(claim *clustercache.PersistentVolumeClaim) st
 // toResourceUnitValue accepts a resource name and quantity and returns the sanitized resource, the unit, and the value in the units.
 // Returns an empty string for resource and unit if there was a failure.
 func toResourceUnitValue(resourceName v1.ResourceName, quantity resource.Quantity) (resource string, unit string, value float64) {
-	resource = promutil.SanitizeLabelName(string(resourceName))
+	resource = resourceName.String()
```
Sean-Holcomb, line 1340: This is a separate issue, but I am fixing it here. Resource name is recorded as a value not a label in the metric and therefore does not need to be sanitized.

mbolt35 (approved):
This looks good, just making sure you re-ran bingen (instead of just updated by hand) just to be sure that ordering did not change. But this change should be totally fine!

PR #4040, opened 2026-09-11
fix pod metric UID and phase lookup across namespaces

### LINKED ISSUE #4041: Allocation metrics attach pod UIDs from other namespaces
OpenCost v1.121.1 keys its pod UID and phase maps by pod name, omitting namespace. For identically named pods in different namespaces, the last pod returned by `GetAllPods()` supplies both entries. Allocation metrics retain their namespace labels but receive the same UID, breaking joins on `(namespace, pod, uid)`.

 As of September 11, 2026, The defect remains in [`develop` at a1b6790e](https://github.com/opencost/opencost/blob/a1b6790e879971225b3ee20aaed89b02b0b8c0e1/pkg/costmodel/metrics.go#L454)

```go
podStatus[pod.Name] = pod.Status.Phase
podUIDs[pod.Name] = string(pod.UID)
```

Direct exporter output confirmed the mismatch in `container_cpu_allocation` and `container_memory_allocation_bytes`. The emitted UID switched repeatedly while Kubernetes reported stable, distinct UIDs for the two pods.

Source shows the same lookup supplies UID labels for `container_gpu_allocation` and `pod_pvc_allocation`. The phase map controls `containerSeen` and metric cleanup, so an identically named pod in another namespace can also affect retention. 

## Reproduce

1. Run `database-0` in namespaces `team-a` and `team-b`, with nonzero CPU and RAM requests. Confirm both are Running and record their Kubernetes UIDs.
2. Inspect their CPU/RAM allocation series directly at OpenCost's `/metrics` endpoint.
3. Compare each series' `(namespace, pod, uid)` labels with Kubernetes. Both pods receive the UID of the last same-named pod enumerated by the exporter. Node placement does not affect the lookup.

Example with placeholder UIDs when `team-b` is enumerated last:

| Pod | Kubernetes UID | Exported UID |
| --- | --- | --- |
| team-a/database-0 | uid-a | uid-b |
| team-b/database-0 | uid-b | uid-b |

## Fix

Key UID and phase by `(namespace, pod)`. Allocation labels must use that pod's UID; lifecycle bookkeeping must use its phase. Missing metadata must not fall back to an identically named pod in another namespace.

https://github.com/opencost/opencost/pull/4040 implements the fix and tests namespace isolation with distinct UIDs/phases, reversed enumeration order, and missing metadata.

brianr2600:
Key pod metric metadata by both name and namespace, otherwise we may attach a single UID to multiple pods with  identical names but different namespaces. Given the 'U' in UID, the value will be wrong for all but one of them. Fixes: #4041


peatey:
can we get a desc pls

cliffcolvin (commented):
Problem gate: this fixes a real defect, not an invented Kubernetes edge case. Pod names are namespace-scoped, but the current emitter collapses UID and phase by pod name alone; common same-named workloads in two namespaces can therefore export the other pod's UID and can retain/delete a series using the other pod's phase. `types.NamespacedName` is the repository-idiomatic key and the implementation is narrowly scoped. I found no security concern or code-level blocker.

The PR still has no description, linked report, reproduction, or author-provided test results, and its integration workflow took the no-op path. Please update the description with the concrete failure/user impact and exact commands/results used to validate it. In particular, say whether the observable `container_*_allocation` metrics were checked for two namespaces with the same pod name; the added unit test proves the helper map in both input orders, but does not exercise the emitter wiring, label values, or stale-series handling.

Reviewer verification: `go test ./pkg/costmodel -run '^TestBuildPodMetricMetadataNamespaceIsolation$' -count=1`, `go test ./pkg/costmodel -count=1`, `go vet ./pkg/costmodel`, and `gofmt -d` all passed.

brianr2600:
We ran into this problem in a cluster which deploys ephemeral environments into various different namespaces. Since the environments contain different versions of the same application stack, they may have identically named pods (especially those belonging to `StatefulSet` and cnpg `Cluster` objects). The UID duplication and churn creates chaos for alerting/recording rules and dashboard queries which join on the UID. Issue #4041 has steps to reproduce.

PR #4042, opened 2026-09-11
support label aliases for allocation autocomplete

aligon:
## Description
Support the label aliases in autocomplete

## User Impact
Autocomplete API should suggest label alias values


pkg/allocation/autocompletequeryservice.go
```diff
@@ -88,10 +89,50 @@ func allocationAutocompleteValues(props *opencost.AllocationProperties, field st
...
+// aliasLabelValues looks up each comma-separated key from configuredKey in labels
+// and returns all found values.
+func aliasLabelValues(labels map[string]string, configuredKey string) []string {
+	if configuredKey == "" || len(labels) == 0 {
+		return nil
+	}
+	var results []string
+	for _, key := range strings.Split(configuredKey, ",") {
+		key = promutil.SanitizeLabelName(strings.TrimSpace(key))
+		if v, ok := labels[key]; ok && v != "" {
```
peatey, line 129: **Blocking:** this only consults `props.Labels`, but every other alias resolver falls back to `props.Annotations` with the same sanitized key: `GenerateKey` in `core/pkg/opencost/allocationprops.go` (the `AllocationDepartmentProp` … `AllocationTeamProp` cases) and the alias filter pass in `core/pkg/opencost/allocationmatcher.go` (`convertAliasFilterToLabelAnnotationFilter`).

Concretely: an allocation with `Annotations{"team":"foo"}` and no `team` label aggregates under `team=foo` and matches the filter `team:"foo"`, but `field=team` autocomplete returns `[]` for it. The UI can then never suggest a value that the API accepts and aggregates on.

Please mirror `GenerateKey`: label lookup first, then annotation fallback, per key. Once that is in, the annotation case also needs an end-to-end test (see the test file comment).

---

pkg/allocation/autocompletequeryservice.go
```diff
@@ -88,10 +89,50 @@ func allocationAutocompleteValues(props *opencost.AllocationProperties, field st
...
+
+// aliasLabelValues looks up each comma-separated key from configuredKey in labels
+// and returns all found values.
+func aliasLabelValues(labels map[string]string, configuredKey string) []string {
+	if configuredKey == "" || len(labels) == 0 {
+		return nil
+	}
+	var results []string
+	for _, key := range strings.Split(configuredKey, ",") {
+		key = promutil.SanitizeLabelName(strings.TrimSpace(key))
```
peatey, line 128: Two things on this line.

1. `promutil.SanitizeLabelName(strings.TrimSpace(key))` is exactly `(*opencost.LabelConfig).Sanitize`, which is what `GenerateKey` uses. Please call `labelConfig.Sanitize(key)` and drop the new `promutil` import, so the normalization has a single owner and can't drift from aggregation.

2. This runs inside the per-allocation loop in `QueryAllocationAutocompleteFromSetRange`, so the configured key string is re-split, re-trimmed, and regexp-sanitized once per allocation even though it is constant for the request. For a large `AllocationSetRange` on an interactive endpoint that adds up. Resolve the sanitized `[]string` once after `NormalizeRequest` and pass it in.

---

pkg/allocation/autocompletequeryservice.go
```diff
@@ -88,10 +89,50 @@ func allocationAutocompleteValues(props *opencost.AllocationProperties, field st
 		if v, ok := autocomplete.MapValueFold(props.NamespaceLabels, label); ok {
 			return []string{v}
 		}
+	// Alias fields resolve to user-configured label keys via LabelConfig.
+	// Each configured key string may be comma-separated; all keys are checked.
+	case field == "department", field == "environment", field == "owner", field == "product", field == "team":
```
peatey, line 94: The five alias names are now string literals in three places: `validate.go`, this `case`, and the `switch` in `aliasLabelKey`. `opencost.AllocationDepartmentProp` / `EnvironmentProp` / `OwnerProp` / `ProductProp` / `TeamProp` already exist, and `(*AllocationProperty).IsAliasedLabel()` already answers "is this an alias". Please use those. If `validate.go` and this switch ever drift, `ValidateField` accepts a field that this function silently returns `nil` for, and the endpoint returns an empty 200 instead of a 400.

Also worth noting: alias lookup here is exact-case (`labels[key]`), while the adjacent `label:<key>` path uses `autocomplete.MapValueFold`. With the existing fixture `Labels{"Team":"platform"}`, `field=label:team` returns the value and `field=team` returns nothing. Exact-case does match `GenerateKey`, so this may be deliberate, but please add a comment saying so.

---

pkg/allocation/autocompletequeryservice.go
```diff
@@ -88,10 +89,50 @@ func allocationAutocompleteValues(props *opencost.AllocationProperties, field st
...
+		return aliasLabelValues(props.Labels, aliasLabelKey(field, labelConfig))
 	}
 	return nil
 }
 
+// aliasLabelKey returns the configured label key string for the given alias field.
+func aliasLabelKey(alias string, labelConfig *opencost.LabelConfig) string {
+	if labelConfig == nil {
```
peatey, line 102: This `nil` fallback is unreachable on the production path: `QueryAllocationAutocompleteFromSetRange` calls `NormalizeRequest` with `EnsureLabelConfig: true`, which guarantees `req.LabelConfig` is non-nil before this is called. `TestAliasLabelKey_DefaultConfig` is the only thing exercising it. Please drop the branch and rely on `NormalizeRequest` as the single place defaults are applied.

---

pkg/allocation/autocompletequeryservice_test.go
```diff
@@ -97,3 +97,114 @@ func TestQueryAllocationAutocompleteFromSetRange(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		got := aliasLabelKey(tt.alias, nil)
+		if got != tt.want {
+			t.Errorf("aliasLabelKey(%q, nil) = %q, want %q", tt.alias, got, tt.want)
+		}
+	}
+}
+
+func TestAliasLabelValues(t *testing.T) {
```
peatey, line 148: These tests only cover the private helpers. Nothing drives an alias field through `QueryAllocationAutocompleteFromSetRange`, so removing the alias `case` from `allocationAutocompleteValues`, or dropping the `req.LabelConfig` argument, keeps every new test green while the endpoint returns `[]` for every alias field.

Please add cases to `TestQueryAllocationAutocompleteFromSetRange` for `Field: "team"` with the value in `Labels`, and one with the value only in `Annotations` (which will fail until the annotation fallback is added).

---

core/pkg/autocomplete/allocation/validate.go
```diff
@@ -17,6 +17,11 @@ func ValidateField(field string) (string, error) {
 	switch f {
 	case "account", "cluster", "namespace", "node", "controllerkind", "controllername", "pod", "container", "label", "namespacelabel":
 		return f, nil
+	// Alias fields map to user-configured label keys via LabelConfig (e.g. "department" →
+	// LabelConfig.DepartmentLabel). Resolution to the actual label key happens in
+	// allocationAutocompleteValues using the LabelConfig on the request.
+	case "department", "environment", "owner", "product", "team":
```
peatey, line 23: `ValidateField` is in the shared `core` module and now accepts alias fields, but the sibling `RouteField` in `route.go` still returns `RouteDefault` for them. Nothing in this repo switches on `RouteField` for allocations today, but a downstream storage-backed implementation would now accept `team`, get `RouteDefault`, and fall through to plain-property handling (an empty 200, or a query against a non-existent column) where it previously returned a 400. Please add a `RouteAlias` route so the core validator and router stay in agreement.

Also, the comment says "user-configured label keys", but the only call site in this repo (`pkg/costmodel/autocomplete.go`) hardcodes `opencost.NewLabelConfig()`, so in OpenCost proper the aliases always resolve to the defaults. That matches how aggregation behaves here, so not a bug, but the wording overstates what this PR delivers for OpenCost itself.

---

peatey (changes requested):
Thanks for this. The direction is right and CI is green, but there is one correctness gap that needs fixing before merge, plus a few cleanups.

**Blocking**
- Alias autocomplete only reads `Labels`. `GenerateKey` and the alias filter pass both fall back to `Annotations` with the same key, so annotation-based aliasing aggregates and filters but never autocompletes. Details inline.

**Should fix**
- Use `LabelConfig.Sanitize` instead of re-implementing it; resolve the sanitized key list once per request rather than per allocation.
- Use the existing `Allocation*Prop` constants / `IsAliasedLabel()` instead of three hardcoded literal lists.
- Drop the unreachable `labelConfig == nil` fallback.
- Add end-to-end cases for an alias field (label-backed and annotation-backed) to `TestQueryAllocationAutocompleteFromSetRange`.
- Add a `RouteAlias` route in `core` so `RouteField` agrees with the widened `ValidateField`.

**Note**
- In this repo the only call site hardcodes `NewLabelConfig()`, so aliases resolve to defaults only. Consistent with aggregation, just worth being accurate about in the description.

---

pkg/allocation/autocompletequeryservice.go
```diff
@@ -40,7 +41,7 @@ func QueryAllocationAutocompleteFromSetRange(asr *opencost.AllocationSetRange, r
 				continue
 			}
 
-			values := allocationAutocompleteValues(alloc.Properties, field)
+			values := allocationAutocompleteValues(alloc.Properties, field, req.LabelConfig)
```
biancaburtoiu, line 44: One issue here - this fix pertains to the opencost autocomplete implementation, this code path is not touched in KCM (the db-backed implementation). We would need to handle that side separately from this.

core/pkg/autocomplete/allocation/validate.go
```diff
@@ -17,6 +17,11 @@ func ValidateField(field string) (string, error) {
...
+		return f, nil
```
biancaburtoiu, line 24: One of the solutions I proposed in the ticket was to introduce the resolution to the actual label key at this level, i.e. convert from e.g. `environment` to the corresponding label alias `label:env`

pkg/allocation/autocompletequeryservice.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/opencost/opencost/core/pkg/autocomplete"
 	coreallocation "github.com/opencost/opencost/core/pkg/autocomplete/allocation"
 	"github.com/opencost/opencost/core/pkg/opencost"
+	"github.com/opencost/opencost/core/pkg/util/promutil"
```
peatey, line 10: No longer needed once `LabelConfig.Sanitize` is used below.

```suggestion
```

---

pkg/allocation/autocompletequeryservice.go
```diff
@@ -56,7 +57,7 @@ func QueryAllocationAutocompleteFromSetRange(asr *opencost.AllocationSetRange, r
 	return &autocomplete.Response{Data: autocomplete.UniqueSortedLimited(results, req.Limit)}, nil
 }
 
-func allocationAutocompleteValues(props *opencost.AllocationProperties, field string) []string {
+func allocationAutocompleteValues(props *opencost.AllocationProperties, field string, labelConfig *opencost.LabelConfig) []string {
 	switch {
```
peatey, line 61: Typed property so the alias case below can use `IsAliasedLabel()` instead of a literal list.

```suggestion
func allocationAutocompleteValues(props *opencost.AllocationProperties, field string, labelConfig *opencost.LabelConfig) []string {
	prop := opencost.AllocationProperty(field)
	switch {
```

---

pkg/allocation/autocompletequeryservice.go
```diff
@@ -88,10 +89,50 @@ func allocationAutocompleteValues(props *opencost.AllocationProperties, field st
...
+	}
+	var results []string
+	for _, key := range strings.Split(configuredKey, ",") {
+		key = promutil.SanitizeLabelName(strings.TrimSpace(key))
+		if v, ok := labels[key]; ok && v != "" {
+			results = append(results, v)
+		}
+	}
+	return results
+}
```
peatey, line 134: This is the substantive fix. It adds the annotation fallback (mirroring `GenerateKey`), uses `LabelConfig.Sanitize`, keys off the `Allocation*Prop` constants, and drops the unreachable nil fallback. Accept together with the two suggestions above (line 10 and lines 60-61) and the test suggestion below; the four compile and pass as a set.

```suggestion
	case prop.IsAliasedLabel():
		return aliasLabelValues(props, aliasLabelKeys(field, labelConfig))
	}
	return nil
}

// aliasLabelKeys returns the sanitized label keys configured for an alias field
// (e.g. "team" -> LabelConfig.TeamLabel), or nil if field is not an alias. The
// configured value may be comma-separated. Keys are sanitized with
// LabelConfig.Sanitize, exactly as GenerateKey does, so autocomplete agrees with
// aggregation.
func aliasLabelKeys(field string, labelConfig *opencost.LabelConfig) []string {
	var configured string
	switch opencost.AllocationProperty(field) {
	case opencost.AllocationDepartmentProp:
		configured = labelConfig.DepartmentLabel
	case opencost.AllocationEnvironmentProp:
		configured = labelConfig.EnvironmentLabel
	case opencost.AllocationOwnerProp:
		configured = labelConfig.OwnerLabel
	case opencost.AllocationProductProp:
		configured = labelConfig.ProductLabel
	case opencost.AllocationTeamProp:
		configured = labelConfig.TeamLabel
	default:
		return nil
	}

	var keys []string
	for _, key := range strings.Split(configured, ",") {
		if key = labelConfig.Sanitize(key); key != "" {
			keys = append(keys, key)
		}
	}
	return keys
}

// aliasLabelValues returns the value of each alias key from the allocation's
// labels, falling back to its annotations when the label is absent. This
// mirrors GenerateKey and the alias filter pass, so every value returned here
// is one that aggregation and filtering will accept. Lookups are exact-case,
// as in GenerateKey; this intentionally differs from the case-folded
// label:<key> path.
func aliasLabelValues(props *opencost.AllocationProperties, keys []string) []string {
	var results []string
	for _, key := range keys {
		if v, ok := props.Labels[key]; ok {
			results = append(results, v)
			continue
		}
		if v, ok := props.Annotations[key]; ok {
			results = append(results, v)
		}
	}
	return results
}
```

Optional follow-up that GitHub won't let me suggest because the lines aren't in the diff: hoist the key resolution out of the per-allocation loop. After the `NormalizeRequest` error check add

```go
	// Alias fields resolve to configured label keys once per request rather
	// than once per allocation; nil for non-alias fields.
	aliasKeys := aliasLabelKeys(field, req.LabelConfig)
```

then pass `aliasKeys` instead of `req.LabelConfig` at the call site, change the signature to `aliasKeys []string`, and make this case `return aliasLabelValues(props, aliasKeys)`. That version is on `peatey/modest-clarke-0ac0z6` (commit `670d244`) if you'd rather cherry-pick the whole thing.

---

pkg/allocation/autocompletequeryservice_test.go
```diff
@@ -97,3 +97,114 @@ func TestQueryAllocationAutocompleteFromSetRange(t *testing.T) {
...
+			t.Errorf("case %q aliasLabelValues(%q): got %v, want %v", tt.name, tt.configuredKey, got, tt.want)
+			continue
+		}
+		for i := range got {
+			if got[i] != tt.want[i] {
+				t.Errorf("case %q aliasLabelValues(%q)[%d]: got %q, want %q", tt.name, tt.configuredKey, i, got[i], tt.want[i])
+			}
+		}
+	}
+}
```
peatey, line 210: Replaces the helper-only tests with the new helper signatures plus an end-to-end alias test through `QueryAllocationAutocompleteFromSetRange`: label-backed, annotation-backed, label-over-annotation precedence, comma-separated configured keys with sanitization, search, and a configured key nothing carries.

```suggestion
func TestQueryAllocationAutocompleteFromSetRange_Alias(t *testing.T) {
	start := time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC)
	as := opencost.NewAllocationSet(start, start.Add(24*time.Hour))
	// label-backed alias value
	as.Set(opencost.NewMockUnitAllocation("a1", start, 24*time.Hour, &opencost.AllocationProperties{
		Namespace: "ns-a",
		Labels:    map[string]string{"team": "platform"},
	}))
	// annotation-backed alias value: GenerateKey and the alias filter pass
	// both fall back to annotations, so autocomplete must too
	as.Set(opencost.NewMockUnitAllocation("a2", start, 24*time.Hour, &opencost.AllocationProperties{
		Namespace:   "ns-b",
		Annotations: map[string]string{"team": "data"},
	}))
	// label wins over annotation when both are present
	as.Set(opencost.NewMockUnitAllocation("a3", start, 24*time.Hour, &opencost.AllocationProperties{
		Namespace:   "ns-c",
		Labels:      map[string]string{"team": "sre"},
		Annotations: map[string]string{"team": "ignored"},
	}))
	// second configured key, needing sanitization (cost-center -> cost_center)
	as.Set(opencost.NewMockUnitAllocation("a4", start, 24*time.Hour, &opencost.AllocationProperties{
		Namespace: "ns-d",
		Labels:    map[string]string{"cost_center": "eng-123"},
	}))

	asr := opencost.NewAllocationSetRange(as)
	window := opencost.NewClosedWindow(start, start.Add(24*time.Hour))

	// default LabelConfig: team -> "team"
	resp, err := QueryAllocationAutocompleteFromSetRange(asr, autocomplete.Request{
		Field:  "team",
		Window: window,
	})
	if err != nil {
		t.Fatalf("unexpected error: %v", err)
	}
	want := []string{"data", "platform", "sre"}
	if !equalStrings(resp.Data, want) {
		t.Fatalf("team autocomplete = %v, want %v", resp.Data, want)
	}

	// search applies to alias values
	resp, err = QueryAllocationAutocompleteFromSetRange(asr, autocomplete.Request{
		Field:  "Team",
		Search: "dat",
		Window: window,
	})
	if err != nil {
		t.Fatalf("unexpected error: %v", err)
	}
	if !equalStrings(resp.Data, []string{"data"}) {
		t.Fatalf("team autocomplete with search = %v, want [data]", resp.Data)
	}

	// custom LabelConfig with comma-separated keys
	cfg := opencost.NewLabelConfig()
	cfg.DepartmentLabel = "squad, cost-center"
	resp, err = QueryAllocationAutocompleteFromSetRange(asr, autocomplete.Request{
		Field:       "department",
		Window:      window,
		LabelConfig: cfg,
	})
	if err != nil {
		t.Fatalf("unexpected error: %v", err)
	}
	if !equalStrings(resp.Data, []string{"eng-123"}) {
		t.Fatalf("department autocomplete = %v, want [eng-123]", resp.Data)
	}

	// alias configured to a key no allocation carries
	cfg.OwnerLabel = "missing"
	resp, err = QueryAllocationAutocompleteFromSetRange(asr, autocomplete.Request{
		Field:       "owner",
		Window:      window,
		LabelConfig: cfg,
	})
	if err != nil {
		t.Fatalf("unexpected error: %v", err)
	}
	if len(resp.Data) != 0 {
		t.Fatalf("owner autocomplete = %v, want empty", resp.Data)
	}
}

func TestAliasLabelKeys(t *testing.T) {
	cfg := &opencost.LabelConfig{
		DepartmentLabel:  "cost-center",
		EnvironmentLabel: "stage, tier",
		OwnerLabel:       "owner",
		ProductLabel:     "",
		TeamLabel:        "squad",
	}

	tests := []struct {
		field string
		want  []string
	}{
		{"department", []string{"cost_center"}},
		{"environment", []string{"stage", "tier"}},
		{"owner", []string{"owner"}},
		{"product", nil},
		{"team", []string{"squad"}},
		{"namespace", nil},
		{"label:team", nil},
	}
	for _, tt := range tests {
		got := aliasLabelKeys(tt.field, cfg)
		if !equalStrings(got, tt.want) {
			t.Errorf("aliasLabelKeys(%q) = %v, want %v", tt.field, got, tt.want)
		}
	}

	defaults := opencost.NewLabelConfig()
	if got := aliasLabelKeys("product", defaults); !equalStrings(got, []string{"app"}) {
		t.Errorf("aliasLabelKeys(product, defaults) = %v, want [app]", got)
	}
}

func TestAliasLabelValues(t *testing.T) {
	tests := []struct {
		name  string
		props *opencost.AllocationProperties
		keys  []string
		want  []string
	}{
		{
			name:  "label match",
			props: &opencost.AllocationProperties{Labels: map[string]string{"team": "platform"}},
			keys:  []string{"team"},
			want:  []string{"platform"},
		},
		{
			name:  "annotation fallback",
			props: &opencost.AllocationProperties{Annotations: map[string]string{"team": "platform"}},
			keys:  []string{"team"},
			want:  []string{"platform"},
		},
		{
			name: "label wins over annotation",
			props: &opencost.AllocationProperties{
				Labels:      map[string]string{"team": "from-label"},
				Annotations: map[string]string{"team": "from-annotation"},
			},
			keys: []string{"team"},
			want: []string{"from-label"},
		},
		{
			name: "multiple keys, mixed sources",
			props: &opencost.AllocationProperties{
				Labels:      map[string]string{"cost_center": "eng-123"},
				Annotations: map[string]string{"squad": "core"},
			},
			keys: []string{"squad", "cost_center", "missing"},
			want: []string{"core", "eng-123"},
		},
		{
			name:  "exact-case only, matching GenerateKey",
			props: &opencost.AllocationProperties{Labels: map[string]string{"Team": "platform"}},
			keys:  []string{"team"},
			want:  nil,
		},
		{
			name:  "no keys",
			props: &opencost.AllocationProperties{Labels: map[string]string{"team": "platform"}},
			keys:  nil,
			want:  nil,
		},
		{
			name:  "nil maps",
			props: &opencost.AllocationProperties{},
			keys:  []string{"team"},
			want:  nil,
		},
	}

	for _, tt := range tests {
		got := aliasLabelValues(tt.props, tt.keys)
		if !equalStrings(got, tt.want) {
			t.Errorf("case %q: aliasLabelValues = %v, want %v", tt.name, got, tt.want)
		}
	}
}

func equalStrings(a, b []string) bool {
	if len(a) != len(b) {
		return false
	}
	for i := range a {
		if a[i] != b[i] {
			return false
		}
	}
	return true
}
```

---

core/pkg/autocomplete/allocation/validate.go
```diff
@@ -17,6 +17,11 @@ func ValidateField(field string) (string, error) {
...
 	}
```
peatey, line 25: Not a plain suggestion because it also needs an import that is outside the diff. Replace these lines with:

```go
	}

	// Alias fields (department, environment, owner, product, team) resolve to
	// label keys via the LabelConfig on the request. Implementations route
	// them with RouteAlias.
	if prop := opencost.AllocationProperty(f); prop.IsAliasedLabel() {
		return f, nil
	}
```

and add `"github.com/opencost/opencost/core/pkg/opencost"` to the import block. No cycle: `core/pkg/opencost` does not import `core/pkg/autocomplete`.

For `route.go` (not in the PR, so no inline comment possible), add a `RouteAlias` constant after `RouteNamespaceLabelValue` and, at the end of `RouteField` before the default return:

```go
	if prop := opencost.AllocationProperty(field); prop.IsAliasedLabel() {
		return RouteAlias, field, nil
	}
```

plus the five alias rows in `TestRouteField`. Both files, done and tested, are in commit `670d244` on `peatey/modest-clarke-0ac0z6`.

---

peatey (commented):
Follow-up to my earlier review with the code for each point, as accept-able suggestions where GitHub allows it.

The four suggestions in `pkg/allocation/` (import removal, function signature, helper bodies, tests) are a set: accepted together they compile and pass `go test ./pkg/allocation/`. The `core` changes (`validate.go`, `route.go`) touch lines outside the diff so they're spelled out in comments instead.

Everything, including the optional once-per-request key hoist, is also in one signed commit on `peatey/modest-clarke-0ac0z6` (`670d244`, based on this PR's head) if cherry-picking is easier: https://github.com/opencost/opencost/commit/670d2447ef9f9f6c1f4a6bd7b6d3ba0cd4b9e3ab

Verified locally: `go vet` and `go test` on `pkg/allocation`, `pkg/costmodel`, and `core/pkg/autocomplete/...`, and `go build ./...` in both modules.

---

peatey:
Correction: the commit link in my last review summary has a mangled hash. The correct one is https://github.com/opencost/opencost/commit/670d244c99908ad5a0841976c417155427c9927d (short form `670d244` was right).

---

core/pkg/autocomplete/allocation/validate.go
biancaburtoiu: This review is solid and the suggestion in https://github.com/opencost/opencost/commit/670d244c99908ad5a0841976c417155427c9927d looks good to me (except one question). This will need an associated KCM PR, though, to support the new `RouteAlias` option in `kubecost-cost-model/pkg/db/allocation/autocompletequeryservice.go`

biancaburtoiu:
Would like to close in favor of https://github.com/opencost/opencost/pull/4045

aligon:
Closing in favor of https://github.com/opencost/opencost/pull/4045

PR #4043, opened 2026-09-13
Include namespace in controller aggregation key

### LINKED ISSUE #1964: Include namespace in ui/metrics

I am using the opencost ui and I do a breakdown based on this:
controller,service,pod
Unfortunately  the listing does not contain the namespace.

For example instead of the current listing like:
deployment:traefik
I would rather like to see
deployment:traefik:trafik

because different teams could deploy the same app in their own namepsace like
deployment:team-a:traefik
deployment:team-b:traefik

I have not found any option to include the namespace.

**Additional context**
Add any other context, documentation links, or screenshots about the feature request here.

Retr0-XD:
When aggregating allocations by controller, include the namespace in the generated key so controllers with the same kind and name in different namespaces are not collapsed into a single entry.

If namespace is already part of the aggregation list, it is not duplicated in the controller key.

Closes #1964


core/pkg/opencost/allocationprops.go
```diff
@@ -399,7 +409,16 @@ func (p *AllocationProperties) GenerateKey(aggregateBy []string, labelConfig *La
 				// Indicate that allocation has no controller
 				controller = UnallocatedSuffix
 			} else if p.ControllerKind != "" {
-				controller = fmt.Sprintf("%s:%s", p.ControllerKind, controller)
+				if p.Namespace != "" && !namespaceInAgg {
+					// Include the namespace so that controllers with the same
+					// kind/name in different namespaces are not collapsed into
+					// a single entry
+					controller = fmt.Sprintf("%s:%s:%s", p.ControllerKind, p.Namespace, controller)
```
cliffcolvin, line 416: This rewrites the public aggregate=controller key from kind:name to kind:namespace:name and changes grouping cardinality for every existing caller. The linked issue confirms aggregate=namespace,controller already supplies namespace and identifies the remaining problem as UI presentation. Preserve controller semantics; fix the UI or add a separately named, opt-in dimension with compatibility and migration documentation.

cliffcolvin (changes requested):
Problem validation: this does not establish a backend defect that warrants changing existing API semantics. In #1964, the first maintainer response explains that multi-aggregation is already supported, the reporter replies that it works, and a later project contributor demonstrates aggregate=namespace,controller and recommends moving the remaining work to opencost-ui. The issue is still labeled enhancement and needs-triage.

This implementation silently changes the key format and cardinality of the long-standing controller aggregation for every API, MCP, and library consumer. Users that intentionally aggregate a same-named application across namespaces would also get different totals. That is a breaking change with no opt-in, migration path, or release documentation. Please re-scope this to the UI gap, or propose a new opt-in aggregation dimension and get explicit maintainer agreement on that API before implementation.

The PR description also needs a Testing section with the exact commands run and their results. For review, I locally ran the focused aggregation tests, the full core/pkg/opencost package, and go vet; they pass, but currently only prove the proposed new behavior—not backward compatibility.

Retr0-XD:
@cliffcolvin thanks for the clarification. I see now that the intended way to distinguish same-named controllers across namespaces is to use multi-aggregation (`aggregate=namespace,controller`), and changing the single `controller` key format would break existing callers. I am closing this PR.

PR #4044, opened 2026-09-13
Add IBM Cloud Usage Reports cloud cost integration

cliffcolvin:
## Summary
- Adds IBM Cloud billing ingestion via the [Usage Reports API](https://cloud.ibm.com/docs/apis/metering-reporting) (`GetResourceUsageAccount`), modeled on the Oracle/STACKIT cloud-cost path so Kubecost can consume it through OpenCost’s existing pipeline.
- Auth uses an IAM API key authorizer (`authorizerType: IBMAPIKey`) nested in `cloud-integration.json`; secrets are redacted in status/export.
- Monthly instance usage is prorated across full-month or month-to-date covered days and mapped into standard `CloudCost` properties (account, region, service, category, provider ID, tags).

### Example `cloud-integration.json`
```json
{
  "ibm": {
    "usageApi": [{
      "accountID": "<ibm-account-id>",
      "authorizer": {
        "authorizerType": "IBMAPIKey",
        "apiKey": "<ibm-cloud-api-key>"
      }
    }]
  }
}
```

The API key (user or service ID) needs `billing.usage-report.read`. Enable with `CLOUD_COST_ENABLED=true`.

## Test plan
- [x] `go test ./pkg/cloud/ibm/... ./pkg/cloud/config/ ./pkg/cloudcost/`
- [x] HTTP-mocked Usage Reports tests cover account/month options, pagination, response mapping, empty data, and API failures
- [x] `go vet` on touched packages
- [ ] Configure a real IBM account + API key in `cloud-integration.json` and verify `/cloudCost` and `/cloudCost/status`
- [ ] Include an enterprise child account invoiced to its parent and confirm Usage Reports returns billable usage that is ingested
- [ ] Confirm Kubecost cloud-cost process ingests `billing_integrations/...` bingen after bumping the OpenCost dependency

Made with [Cursor](https://cursor.com)


cliffcolvin:
Addressed the Copilot review findings in cc155eb8:

1. **Non-billable rows** — skip `InstanceUsage` when `Billable == false` before projecting costs.
2. **CurrencyRate** — multiply `cost` / `rated_cost` by `CurrencyRate` when present and > 0 (else treat as 1.0 / already USD).
3. **Tags** — parse `key:value` string tags; merge `Tags` + `ServiceTags`.
4. **rated_cost == 0** — keep list/amortized at zero (no fallback to net).
5. **MTD proration** — completed months ÷ calendar days; current month ÷ day-of-month (`asOf`).
6. **Connection status** — `SuccessfulConnection` when the API returned items, even if all were filtered/zero-cost.
7. **ParseProvider aliases** — removed `ibm`/`iks` aliases for this billing-only change; `IBMProvider` constant remains for CloudCost `Provider`.
8. **DCO** — squashed history so Author and Signed-off-by both use `clifford.colvin@ibm.com`.

Left for a follow-up: injectable/mocked HTTP client covering the full `GetCloudCost` pager path (helpers + SDK projection are covered in unit tests).

cliffcolvin:
Reviewing this against a second producer that emits the same `CloudCost` shape from IBM billing-export data rather than the Usage Reports API. The two are mutually exclusive per deployment, but they have to be interchangeable, because cloud cost aggregation keys on:

```
ProviderID | InvoiceEntityID | AccountID | Service | UsageType | Category | RegionID | AvailabilityZone
```

Any field the two producers disagree on splits one resource into two rows. Three findings, one blocking.

### 1. `Service` uses the display name — this one will split rows

```go
options.SetNames(true)
...
service := item.ResourceName
if service == "" {
    service = item.ResourceID
}
```

With `names=true`, `ResourceName` is essentially always populated, so `Service` becomes the human label. Your own test fixture shows the gap: `ResourceID: "containers-kubernetes"` produces `Service: "IBM Cloud Kubernetes Service"`.

A billing-export producer only has the service identifier, so it emits `containers-kubernetes`. Same resource, two `Service` values, two rows.

Suggest `Service: item.ResourceID` unconditionally, and dropping `SetNames(true)` unless it is needed elsewhere. Beyond cross-producer agreement, the identifier is the better key on its own merits: it is stable, machine-readable, and not subject to marketing renames or localization — the same reason the AWS integration keys on `ProductCode` (`AmazonEKS`) rather than "Amazon Elastic Kubernetes Service". `ResourceName` is still valuable, just as a label rather than as an identity field.

Note the header comment already documents the current behavior as `Service = ResourceName when names=true, else ResourceID`, so this is a deliberate choice worth revisiting rather than an oversight.

### 2. `selectIBMCategory` — the prefix arms are broader than they look

The exact-match members line up with the billing-export side. The `strings.HasPrefix` arms do not:

```go
strings.HasPrefix(id, "is.vpn"),
strings.HasPrefix(id, "transit"),
strings.HasPrefix(id, "internet-svcs"),
strings.HasPrefix(id, "codeengine"),
```

Each is redundant with the `==` case directly above it *unless* a real service id extends that prefix — `is.vpn-server` and `transit-gateway` being the obvious candidates. So these arms are either dead code or a silent widening, depending on IBM's actual service-id list.

Worth making explicit either way: drop them if the exact matches are complete, or keep them and document the intended prefix families so a second implementation can reproduce the same categories. I will mirror whichever this lands on. Category strings surface as report filter values, so divergence is visible to users rather than just internally inconsistent.

### 3. `AccountID` format — thanks, this resolves an open question

```go
//	AccountID = account_id from Usage Reports (published samples are bare 32-hex, no "a/" prefix)
```

That is the answer needed on the other side. Billing exports carry the account in CRN-adjacent form (`a/<32-hex>`), so that producer will strip the `a/` prefix to match this. Flagging it here so the normalization direction is recorded on both sides rather than rediscovered.

One request: if a real account is ever observed returning a prefixed value, it is worth normalizing defensively here too, since the field feeds object storage paths where a `/` changes the path shape.

### Non-blocking, just confirmations

- `AvailabilityZone` left unset — matches. IBM billing data has no AZ concept, so both sides leave it empty.
- `InvoiceEntityID = AccountID` — matches, and forced rather than chosen: there is no payer or enterprise column in row data on either side.
- The five cost metrics (rated → `ListCost`/`AmortizedCost`, post-discount → `NetCost`/`AmortizedNetCost`/`InvoicedCost`) — matched exactly on the other side, including `AmortizedCost` from rated cost.
- No changes to `CloudCost`, `CloudCostProperties`, or the codecs, so producers pinned to different core versions stay wire-compatible. Worth preserving.
- Even proration across days is consistent with how billing exports arrive, since IBM's source granularity is monthly in both cases. Neither path produces true daily cost, which is worth stating plainly somewhere so a flat daily series is not read as real.

Happy to align the other implementation to whatever you decide on (1) and (2) — the key is that one of us moves rather than both shipping.

cliffcolvin:
Aligned with the CAC / billing-export contract review:

1. **Service = ResourceID** (blocking) — display name moves to label `ibm_resource_name`; `SetNames(true)` kept only for that label.
2. **selectIBMCategory** — exact matches only for the locked table; sole prefix family is documented `databases-for-*`. Dropped the silent wideners (`is.vpn*`, `transit*`, `codeengine*`, `internet-svcs*`).
3. **AccountID** — defensive `normalizeAccountID` strips leading `a/` on API rows, config Key, and Usage Reports requests.

Please mirror Service=`ResourceID` and the same category table on the billing-export side.

cliffcolvin:
## CI triage notes

**Integration (`run-tests` / `run-comparison-tests` / `check-success`)** — failures look infra/flake-related (demo Prometheus 500s, spot/UID coverage, Prom vs collector drift), not caused by the IBM Usage Reports cloud-cost package. Failed jobs were re-triggered on this run.

**Snyk** — still red (`1 test has failed`). OSV reports no known vulns for the direct new deps (`IBM/platform-services-go-sdk@v0.103.2`, `IBM/go-sdk-core/v5@v5.23.4`). Need a maintainer with `ameijer` Snyk org access to open the PR-check dashboard and confirm whether this is a real CVE, transitive finding, or policy/license flag.

pkg/cloud/ibm/usageintegration.go
```diff
@@ -0,0 +1,364 @@
...
+				itemsSeen++
+				record, ok := instanceUsageFromSDK(item)
+				if !ok {
+					continue
+				}
+				for _, cc := range cloudCostsFromInstance(record, start, end, asOf) {
+					ccsr.LoadCloudCost(cc)
+				}
+			}
+		}
```
peatey, line 90: This is the only path in the package that touches the wire, and nothing exercises it: pagination, the empty-result → `MissingData` branch, and the pager error → `FailedConnection` transition are all unverified. I'd like this closed before merge rather than in a follow-up.

It's cheap with the SDK: `usagereportsv4.NewUsageReportsV4(&UsageReportsV4Options{URL: srv.URL, Authenticator: &core.NoAuthAuthenticator{}})` against an `httptest.Server` that returns two pages (second via `next_url`) plus one 500 covers all three branches. The only production change needed is making the client injectable — either an optional service URL on `UsageConfiguration` (which is also useful for private endpoints) or a client-factory field on `UsageIntegration` that defaults to `GetUsageReportsClient`.

---

pkg/cloud/ibm/usageintegration.go
```diff
@@ -0,0 +1,364 @@
...
+//	InvoiceEntityID   = AccountID (no payer/enterprise column in row data)
+//	Service           = ResourceID (stable service id — not the display name)
+//	Category          = selectIBMCategory(ResourceID) only — pure function of service id
+//	UsageType         = N/A on CloudCostProperties in this tree (and must remain unset if added later)
+//	ListCost / AmortizedCost           = sum(rated_cost) converted to USD, prorated
+//	NetCost / AmortizedNetCost / InvoicedCost = sum(cost) converted to USD, prorated
+//
+// ResourceName (when names=true) is stored as label "ibm_resource_name", not Service.
+// Daily values are synthetic: report totals ÷ covered days (full month, or MTD day-of-month),
+// emitted for days overlapping the query window. Non-billable instances are skipped.
```
peatey: Worth stating the ingestor interaction here explicitly, since it is not obvious from "synthetic". With defaults (`CLOUD_COST_REFRESH_RATE_HOURS=6`, MTD run every 6th run, `CLOUD_COST_RUN_WINDOW_DAYS=3`), ordinary runs rewrite only the trailing 3 days using today's day-of-month divisor and today's MTD total, while earlier days keep values from the last MTD run. Between MTD runs the sum of the month's daily rows does not equal IBM's MTD figure; it reconciles every 36h. The previous month gets its final full-month proration only because the 72h run window exceeds the 36h MTD cadence — if an operator sets the run window shorter than that, the prior month can be left with a stale divisor permanently.

Not a bug in this PR, but a doc line here (and in the eventual opencost.io page) will save a support thread.

---

pkg/cloud/ibm/usageintegration.go
```diff
@@ -0,0 +1,364 @@
...
+	RatedCost          float64
+	Tags               []any
+}
+
+// instanceUsageFromSDK projects an SDK row. ok is false when the row should be skipped
+// (explicitly non-billable).
+func instanceUsageFromSDK(item usagereportsv4.InstanceUsage) (instanceUsageRecord, bool) {
+	if item.Billable != nil && !*item.Billable {
+		return instanceUsageRecord{}, false
+	}
```
peatey, line 140: Correct for Lite plans and free tiers. I could not confirm from the SDK what IBM sets `billable` to on an enterprise child account whose usage is invoiced to the parent. If that comes back `false`, enterprise deployments emit nothing. Please include an enterprise child account in the real-account verification in the test plan — no code change needed unless it turns out to be the case.

---

peatey (changes requested):
Implementation looks sound; one request before merge (the pager test, inline), two non-blocking notes inline.

**Verified locally on `8cdda4c`**
- Build, `go vet`, and tests for `pkg/cloud/ibm`, `pkg/cloud/config`, `pkg/cloudcost` pass. `go mod tidy` is a no-op, so the transitive bumps in go.mod are pulled in by the IBM SDK rather than hand-edited.
- Wiring covers every site where STACKIT is registered (configurations, statuses, controller parse, integration factory); skipping `provider.go` for a billing-only change is fine.
- Currency direction is right: the SDK documents `CurrencyRate` as the account currency's value in USD, so cost × rate is USD.
- `databases-for-*` → Storage matches the AWS mapping (RDS/DynamoDB land in Storage).
- The earlier review rounds were applied as described: `Service` keys on `ResourceID`, prefix wideners removed, `Billable == false` and `NonChargeable` skipped, rated-cost zero preserved, account ID normalized.

**CI**
- The integration failures are not this PR's. The re-run of `run-tests / test-stack` polled `pr-4044-oc-8cdda4c.infra.opencost.io` and got 404 for 80s because the stack had been torn down ~21h after the original build. Develop's own scheduled integration run on the same base commit (run 34852824430) is also red.
- Snyk: I don't have dashboard access either; that needs someone with the org access you named.

---

pkg/cloud/ibm/usageintegration.go
cliffcolvin: Addressed in 421d8249. I added a private SDK-client factory and HTTP-backed tests using the real IBM SDK. The success case asserts the normalized account/month path, _limit/_names/_tags options, next.href pagination via _start across two pages, mapped costs, and SuccessfulConnection. Separate cases cover an empty response -> MissingData and HTTP 500 -> FailedConnection. The affected package suite, go vet, and the IBM race test all pass.

pkg/cloud/ibm/usageintegration.go
cliffcolvin: Addressed across 884d9d51 and 421d8249. The integration now widens each touched range to whole months and rewrites every covered day together, so a refresh cannot leave one month with mixed MTD divisors. I also updated the package comment and PR summary to state clearly that the daily values are synthetic rather than true daily usage.

pkg/cloud/ibm/usageintegration.go
cliffcolvin: Added this explicitly to the unchecked real-account test plan: verify an enterprise child account invoiced to its parent still returns billable usage and is ingested. I left production filtering unchanged until that behavior can be observed against a real enterprise account, as requested.

peatey (approved):

PR #4045, opened 2026-09-14
Allocation autocomplete: Support label aliases

biancaburtoiu:
## Description
Introduce support for label aliases `Department / Owner / Environment / Team / Product` in allocation autocomplete.

## User Impact
Not a breaking change. The five keys above are now accepted as valid label alias keys.

## Testing
- Covered in unit tests


thomasvn:
Looks good. Do we also need to update the file "[pkg/allocation/autocompletequeryservice.go](https://github.com/opencost/opencost/blob/develop/pkg/allocation/autocompletequeryservice.go)" to support label aliases?

<details>
<summary>Alias fields now pass validation, but the handler has no way to answer them</summary>

- Where it breaks: QueryAllocationAutocompleteFromSetRange in pkg/allocation/autocompletequeryservice.go never calls RouteField or ResolveAliasLabelKey. Its allocationAutocompleteValues switch has no case for department, environment, owner, product or team, so these fields return nil.
- Result: GET /allocation/autocomplete?field=team&window=1d used to return 400. It now returns 200 with empty data, even when pods have a team label.
- Fix: in the service, turn the alias into its label key and look that key up in props.Labels. Also look in props.Annotations, since aggregation checks both. Right now RouteAlias and ResolveAliasLabelKey are only called from tests.

</details>

biancaburtoiu:
> Looks good. Do we also need to update the file "[pkg/allocation/autocompletequeryservice.go](https://github.com/opencost/opencost/blob/develop/pkg/allocation/autocompletequeryservice.go)" to support label aliases?
> 
> Alias fields now pass validation, but the handler has no way to answer them

Thanks for calling this out @thomasvn! I've updated the allocation autocomplete API handler accordingly.

thomasvn (approved):

cpetersen5 (approved):
LGTM :shipit:

PR #4049, opened 2026-09-16
prom: probe range semantics for non-semver backends

### LINKED ISSUE #4046: Prometheus 3 range offset is keyed on the buildinfo version, so Grafana Cloud / Mimir backends lose the first step of every query window
**Describe the bug**

OpenCost decides whether to compensate for Prometheus 3's left-open range selectors by parsing the version string from `/api/v1/status/buildinfo` as semver. Grafana Cloud (Mimir) returns a version string that is not semver(based on Mimir's weekly release), so the compensation is disabled. 

**To Reproduce**

Confirmed live against Grafana Cloud, and against Prometheus 3.11.3 for the control:

1. Point OpenCost 1.120.0 at a Grafana Cloud hosted Prometheus endpoint.
2. Check the version string OpenCost is trying to parse:
```
   curl -u <instance-id>:<token> \
     https://<stack>.grafana.net/api/prom/api/v1/status/buildinfo
```
   Returns `"version":"r409-4d1cd395"`.
3. Confirm the backend actually uses left-open ranges. At a minute-aligned timestamp:
```
   curl -u <instance-id>:<token> \
     'https://<stack>.grafana.net/api/prom/api/v1/query?query=count_over_time(vector(1)[1m:1m])&time=<unix time on a minute boundary>'
```
   Returns `1` on Grafana Cloud and on Prometheus 3.11.3; returns `2` on Prometheus 2.x.
4. OpenCost logs `Failed to parse prometheus version r409-4d1cd395` at startup and leaves the offset off.

**Expected behavior**

The resolution offset should be applied whenever the backend evaluates range selectors as left-open, regardless of whether its version string happens to parse as semver.

OpenCost 1.120.0, Helm chart 2.5.14

git-jxj:
## Description

Prometheus 3 changed range selectors to exclude the left boundary, so OpenCost offsets range resolution for Prometheus 3 and newer. The current detection relies entirely on parsing buildinfo.version as semver. Grafana Cloud and Mimir return versions such as r409-4d1cd395, leaving the offset disabled even though they use the Prometheus 3 behavior.

When the build version is not semver, this change runs a constant PromQL subquery at a minute-aligned timestamp. A left-open backend returns one sample and enables the offset; a legacy inclusive backend returns two and keeps it disabled. This measures the behavior OpenCost needs instead of treating every non-semver backend as Prometheus 3.

The regression test exercises the real Prometheus HTTP client path for both semantics.

## Related Issues

Fixes #4046

## User Impact

Grafana Cloud and Mimir users with non-semver build versions no longer lose the first sample in OpenCost query windows. Non-semver Prometheus-compatible backends that retain Prometheus 2 inclusive semantics remain unchanged. No configuration or API changes are introduced.

## Testing

- go test ./pkg/prom -run TestPrometheusDataSourceDetectsRangeSemanticsForNonSemverVersions -count=1
- go test ./pkg/prom -count=1
- go test -race ./pkg/prom -run TestPrometheusDataSourceDetectsRangeSemanticsForNonSemverVersions\|TestQueryLogs -count=1
- go test ./...
- go vet ./pkg/prom
- git diff --check


modules/prometheus-source/pkg/prom/datasource_test.go
```diff
@@ -0,0 +1,74 @@
...
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				switch r.URL.Path {
+				case "/api/v1/query":
+					if err := r.ParseForm(); err != nil {
+						http.Error(w, err.Error(), http.StatusBadRequest)
+						return
+					}
+					value := "1"
+					if r.Form.Get("query") == rangeSemanticsProbe {
```
cliffcolvin, line 37: 🟡 risk: this test supplies the desired probe value but never validates the time parameter, so it still passes if Truncate(time.Minute) is removed—even though alignment is what makes the 1-vs-2 result meaningful. Assert that the received evaluation timestamp is exactly minute-aligned.

modules/prometheus-source/pkg/prom/datasource.go
```diff
@@ -15,8 +15,33 @@ import (
 
 	prometheus "github.com/prometheus/client_golang/api"
 	prometheusAPI "github.com/prometheus/client_golang/api/prometheus/v1"
+	prometheusModel "github.com/prometheus/common/model"
```
cliffcolvin, line 18: 🔵 maintenance: this makes github.com/prometheus/common a direct dependency, but modules/prometheus-source/go.mod still marks it indirect. go mod tidy -diff reports that exact move; commit the resulting go.mod change.

cliffcolvin (changes requested):
Problem validation: this fixes a real OpenCost failure rather than inventing one. #4046 was filed independently by a Grafana Cloud user and includes live results showing the non-semver build version, the incorrect disabled offset, and Prometheus 2/3 controls. The implementation is appropriately narrow. I locally ran the focused regression test, all pkg/prom tests, the focused race tests, and go vet ./pkg/prom; all passed. The integration workflow is red, but its log shows broad shared-environment/API failures rather than a failure isolated to this two-file change.

Two changes are needed before merge: make the regression test protect the probe's minute-alignment invariant, and commit the module metadata update required by the new direct import.

git-jxj:
Addressed both review points in `14daebff`: the regression server now verifies that the probe timestamp is an exact minute boundary, and `github.com/prometheus/common` is recorded as a direct dependency in the module metadata. Validation passed: focused regression, all `./pkg/prom` tests, focused race test, `go vet ./pkg/prom`, module-wide `go fmt ./...`, module-wide `go vet ./...`, and `go mod tidy -diff` with no remaining diff.

mbolt35 (approved):
Looks good! Thanks for providing the better fallback options!

git-jxj:
Rebased onto the latest `develop` in `3cae5b31`; the reviewed change remains unchanged and the branch is no longer behind. @cliffcolvin, the two requested updates are in the current head and their threads are resolved—could you re-review when convenient?

git-jxj:
The current-head integration run failed while waiting for the API and in allocation comparisons. A develop-branch run from about an hour earlier also failed API readiness and comparison tests, which suggests a shared CI/environment issue, though the logs do not prove every failure has the same cause: https://github.com/opencost/opencost/actions/runs/35903690707 and https://github.com/opencost/opencost/actions/runs/35897581696. I do not have permission to rerun the upstream workflow or view the Snyk ERROR details. Could a maintainer rerun the integration workflow and inspect the Snyk result? The two earlier review requests remain addressed in the current head.

